PR #15, opened 2024-01-31
[1/] Improvement: Add utility functions for CLI improvements

zeyadkhaled:
Part of an effort to improve the osdk cli. 
The work to get the improvements across will be done over the followings PRs:
[1/] [Add new utility functions](https://github.com/palantir/osdk-ts/pull/15)  --> this
[2/] [ Introduce new command structure ](https://github.com/palantir/osdk-ts/pull/16)
[3/] [Use loadedToken for fetching data](https://github.com/palantir/osdk-ts/pull/17)

## This PR
Introduces 3 utility functions that will assist with the CLI improvements.

- configFileUtils -> Project config file loading and validating
- versionUtils -> For autoVersioning capabilities 
- loadTokenUtils -> Loading token file containing JWT tokens used for Auth

## Considerations

- Adding tests for each of the utility tools. Should that be part of a local tests sub-package within the CLI package?
- Licensing information is set to 2023 instead of 2024


packages/cli/src/utils/loadTokenUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+    checkIsValidToken(loadedToken);
+    consola.debug(`Using token from --tokenFile argument`);
+    return loadedToken;
+  }
+
+  for (const envVar of TOKEN_ENV_VARS) {
+    const environmentToken = process.env[envVar];
+    if (environmentToken) {
+      checkIsValidToken(environmentToken);
+      consola.debug(`Using token from ${envVar} environment variable`);
```
zeyadkhaled, line 48: If we use `FOUNDRY_SDK_AUTH_TOKEN` show a depreciation warning

packages/cli/src/utils/versionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    const version = gitVersion.trim().replace(prefix, "");
+
+    if (!isSemVerCompliant(version)) {
+      throw new Error(`The version string ${version} is not SemVer compliant.`); // Will fall back to package.json version
+    }
+
+    return version;
+  } catch {
+    // If git describe fails, find the nearest package.json and use its version
+    const packageJsonPath = await findUp("package.json");
```
zeyadkhaled, line 43: How does the versioning work on other providers? Is it expected that you will always use a package.json?

packages/cli/src/utils/versionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+import { execSync } from "node:child_process";
+import { promises as fsPromises } from "node:fs";
+
+/**
+ * Gets the version string using git describe or falls back to package.json version.
+ * @param prefix The prefix to use for matching against tags. Defaults to an empty string.
+ * @returns A promise that resolves to the version string. Throws an error if the version string is not SemVer compliant or if the version cannot be determined.
+ * @throws An error if the version string is not SemVer compliant or if the version cannot be determined.
+ */
+export async function getAutoVersion(prefix: string = ""): Promise<string> {
```
zeyadkhaled, line 26: Provide the prefix through a key in the config file

packages/cli/src/utils/token/loadTokenUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+import { consola } from "consola";
+import * as fs from "node:fs";
+import * as path from "node:path";
+
+const TOKEN_ENV_VARS: string[] = ["FOUNDRY_TOKEN", "FOUNDRY_SDK_AUTH_TOKEN"];
+
+/**
+ * Loads a JWT Auth Token from an argument, a file, or environment variable.
```
tzyl, line 24:
```suggestion
 * Loads a JWT from an argument, a file, or environment variable returning the first found.
```

packages/cli/src/utils/token/loadTokenUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (token) {
+    checkIsValidToken(token);
+    consola.debug(`Using token from --token argument`);
+    return token;
+  }
+
+  if (tokenFile) {
+    const loadedToken = loadTokenFile(tokenFile);
+    checkIsValidToken(loadedToken);
+    consola.debug(`Using token from --tokenFile argument`);
```
tzyl, line 40: Are we going to log the file path used as well? I know this one is in the CLI args but may be still helpful to have the paths used always in the debug stdout for verbose mode

packages/cli/src/utils/token/loadTokenUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { consola } from "consola";
+import * as fs from "node:fs";
+import * as path from "node:path";
+
+const TOKEN_ENV_VARS: string[] = ["FOUNDRY_TOKEN", "FOUNDRY_SDK_AUTH_TOKEN"];
```
tzyl, line 21: suggestion: Narrow the types so the deprecated check to FOUNDRY_SDK_AUTH_TOKEN later will be type safe

```suggestion
const TOKEN_ENV_VARS = ["FOUNDRY_TOKEN", "FOUNDRY_SDK_AUTH_TOKEN"] as const;
```

packages/cli/src/utils/version/versionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    );
+  }
+}
+
+/**
+ * Checks if a given version string is SemVer compliant.
+ * @param version The version string to check.
+ * @returns true if the version string is SemVer compliant, false otherwise.
+ */
+function isSemVerCompliant(version: string): boolean {
```
tzyl, line 66: We can reuse [semver util](https://github.com/palantir/osdk-ts/blob/main/packages/cli/src/util/isValidSemver.ts)

packages/cli/src/utils/token/tests/loadTokenUtils.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    warn: vi.fn(),
+  },
+}));
+
+vi.mock("node:fs", () => ({
+  readFileSync: vi.fn(),
+}));
+
+const originalEnv = process.env;
+vi.stubGlobal("process", { ...process, env: { ...originalEnv } });
```
tzyl, line 33: Should this use the dedicated stubEnv and unstubEnv?
https://vitest.dev/api/vi#vi-stubenv-0-26-0

packages/cli/src/utils/token/tests/loadTokenUtils.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    expect(() => tokenModule.loadToken()).toThrow("No token found.");
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+    vi.restoreAllMocks();
+  });
+});
+
+describe("loadTokenFile", () => {
```
tzyl, line 73: It looks like these tests are testing `loadToken` instead of `loadTokenFile` / `checkIsValidToken`

zeyadkhaled: Because these are internal functions, I wanted to test them indirectly and clearly label it

packages/cli/src/utils/version/versionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+import { execSync } from "node:child_process";
+import { promises as fsPromises } from "node:fs";
+
+/**
+ * Gets the version string using git describe or falls back to package.json version.
+ * @param prefix The prefix to use for matching against tags. Defaults to an empty string.
+ * @returns A promise that resolves to the version string. Throws an error if the version string is not SemVer compliant or if the version cannot be determined.
+ * @throws An error if the version string is not SemVer compliant or if the version cannot be determined.
+ */
+export async function getAutoVersion(prefix: string = ""): Promise<string> {
```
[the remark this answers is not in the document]
zeyadkhaled: The idea was to do it through the config ( or else user could directly just pass a version). Yeah that doesn't sound bad. Will think about it.

packages/cli/src/utils/version/versionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+  const { findUp } = await import("find-up");
+
+  try {
+    const gitVersion = execSync(
+      `git describe --tags --always --first-parent --match="${prefix}*"`,
```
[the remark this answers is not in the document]
zeyadkhaled: I think this depends in the first place if we want to expose this implementation details to the user. I would say, this is more of a best-effort, if it fails just prompt the user to supply a --version

packages/cli/src/utils/version/versionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    );
+    const version = gitVersion.trim().replace(prefix, "");
+
+    if (!isSemVerCompliant(version)) {
+      throw new Error(`The version string ${version} is not SemVer compliant.`); // Will fall back to package.json version
+    }
+
+    return version;
+  } catch {
+    // If git describe fails, find the nearest package.json and use its version
```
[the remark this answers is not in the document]
zeyadkhaled: No, as it deemed a bit flaky and unreliable when dev-ing locally

packages/cli/src/util/config.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export interface FoundryConfig {
+  foundryUrl: string;
+  site: SiteConfig;
+}
+
+export interface LoadedFoundryConfig {
+  foundryConfig: FoundryConfig;
+  configFilePath: string;
+}
```
zeyadkhaled, line 36: Added this to be able to call `consola.debug` with the fileName when we first load this file

packages/cli/src/util/config.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Asynchronously searches for a configuration file in the current directory and its parents.
+ * @returns A promise that resolves to the path of the configuration file, or undefined if not found.
+ */
+async function loadConfigFile(): Promise<string | undefined> {
+  const { findUp } = await import("find-up");
+
+  const matcher = async (directory: string) => {
+    const files = await fsPromises.readdir(directory);
```
tzyl, line 107: Do we need to read the directory within the matcher? I was imagining this would just be a simple pattern string that would be applied to the files within the traversal itself which is actually done by passing the name array rather than the matcher function

packages/cli/src/util/config.ts
```diff
@@ -0,0 +1,125 @@
...
+      required: ["application", "directory"],
+    },
+  },
+  required: ["foundryUrl", "site"],
+  additionalProperties: false,
+};
+
+const Ajv = ajvModule.default; // https://github.com/ajv-validator/ajv/issues/2132
+const ajv = new Ajv({ allErrors: true });
+const validate = ajv.compile(configFileSchema);
```
tzyl, line 62: Given that `loadFoundryConfig()` will only be called once per lifetime of the command it may be better to construct this when needed lazily with dynamic import rather than immediately once the script is loaded

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,45 @@
...
+import { execSync } from "node:child_process";
+import { isValidSemver } from "./isValidSemver.js";
+
+/**
+ * Gets the version string using git describe or falls back to package.json version.
+ * @param prefix The prefix to use for matching against tags. Defaults to an empty string.
+ * @returns A promise that resolves to the version string.
+ * @throws An error if the version string is not SemVer compliant or if the version cannot be determined.
+ */
+export async function getAutoVersion(prefix: string = ""): Promise<string> {
```
tzyl, line 26:
```suggestion
export async function autoVersion(prefix: string = ""): Promise<string> {
```

packages/cli/src/util/autoVersion.test.ts
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { execSync } from "node:child_process";
+import { describe, expect, it, vi } from "vitest";
+import { getAutoVersion } from "./autoVersion.js";
+
+vi.mock("node:child_process", () => ({
```
tzyl, line 21: Have you tried removing the factory argument (just use `vi.mock("node:child_process")`)? I think vitest should automatically mock out the object with defaults

packages/cli/src/utils/token/tests/loadTokenUtils.test.ts
tzyl: I recommend not testing them indirectly and test them explicitly

packages/cli/src/util/token.ts
```diff
@@ -0,0 +1,94 @@
...
+  } catch (error) {
+    throw new Error(`Unable to read token file "${filePath}": ${error}`);
+  }
+
+  return token;
+}
+
+function validate(token: string): void {
+  if (!isJWT(token)) {
+    throw new Error(`Token "${token}" does not appear to be a JWT`);
```
tzyl, line 86: Avoid logging the token in stdout

```suggestion
    throw new Error(`Token does not appear to be a JWT`);
```

packages/cli/src/util/token.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Loads a JWT Auth Token from an argument, a file, or environment variable returning the first found.
+ * @param token The token as a string.
+ * @param tokenFile The path to the token file.
+ * @returns The token as a string.
+ * @throws An error if no token is found.
+ */
+export function loadToken(token?: string, tokenFile?: string): string {
+  if (token) {
+    validate(token);
+    consola.debug(`Using token from --token argument`);
```
tzyl, line 33: Let's say we're using the `--token` argument as soon as we enter this branch so that if validate fails this debug information still shows up

packages/cli/src/util/token.ts
```diff
@@ -0,0 +1,94 @@
...
+    return token;
+  }
+
+  if (tokenFile) {
+    const loadedToken = loadTokenFile(tokenFile);
+    validate(loadedToken);
+    consola.debug(`Using token from --tokenFile=${tokenFile} argument`);
```
tzyl, line 40: Same comment here about logging we're using the token from file as soon as we know it. Also we may want to log the full resolved path for debugging as I guess this may be a relative path in the input

packages/cli/src/util/token.ts
```diff
@@ -0,0 +1,94 @@
...
+    return loadedToken;
+  }
+
+  for (const envVar of TOKEN_ENV_VARS) {
+    const environmentToken = process.env[envVar];
+    if (environmentToken) {
+      validate(environmentToken);
+      consola.debug(`Using token from ${envVar} environment variable`);
```
tzyl, line 48: same

packages/cli/src/util/token.ts
```diff
@@ -0,0 +1,94 @@
...
+const TOKEN_ENV_VARS = ["FOUNDRY_TOKEN", "FOUNDRY_SDK_AUTH_TOKEN"] as const;
+
+/**
+ * Loads a JWT Auth Token from an argument, a file, or environment variable returning the first found.
+ * @param token The token as a string.
+ * @param tokenFile The path to the token file.
+ * @returns The token as a string.
+ * @throws An error if no token is found.
+ */
+export function loadToken(token?: string, tokenFile?: string): string {
```
tzyl, line 30: Should we make all of these async rather than sync for consistency with config loading?

packages/cli/src/util/token.ts
zeyadkhaled: validate throws, so the CLI exists before printing the debug message.

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,43 @@
...
+import { execSync } from "node:child_process";
+import { isValidSemver } from "./isValidSemver.js";
+
+/**
+ * Gets the version string using git describe.
+ * @param prefix The prefix to use for matching against tags. Defaults to an empty string.
+ * @returns A promise that resolves to the version string.
+ * @throws An error if the version string is not SemVer compliant or if the version cannot be determined.
+ */
+export async function autoVersion(prefix: string = ""): Promise<string> {
```
tzyl, line 26:
```suggestion
export async function autoVersion(tagPrefix: string = ""): Promise<string> {
```

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,43 @@
...
+      `git describe --tags --always --first-parent --match="${prefix}*"`,
+      { encoding: "utf8" },
+    );
+    const version = gitVersion.trim().replace(new RegExp(`^${prefix}`), "");
+    if (!isValidSemver(version)) {
+      throw new Error(`The version string ${version} is not SemVer compliant.`);
+    }
+
+    return version;
+  } catch (error) {
```
tzyl, line 38: I do think it would be worth checking what exactly this will show when git is not installed, not in git repo etc. and if it would be helpful to have more specific messages as a follow up

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Gets the version string using git describe.
+ * @param prefix The prefix to use for matching against tags. Defaults to an empty string.
+ * @returns A promise that resolves to the version string.
+ * @throws An error if the version string is not SemVer compliant or if the version cannot be determined.
+ */
+export async function autoVersion(prefix: string = ""): Promise<string> {
+  try {
+    const gitVersion = execSync(
+      `git describe --tags --always --first-parent --match="${prefix}*"`,
```
tzyl, line 29: - `--always` will mean that we get a commit object as a fallback and then probably immediately fail on the semver check is it preferable to do this or throw the git error on no match
- Do you think it's valuable to support `--dirty` to add a suffix on unclean state or should the version still be the same as the last commit?

packages/cli/src/util/config.ts
```diff
@@ -0,0 +1,113 @@
...
+
+import type { JSONSchemaType } from "ajv";
+import { promises as fsPromises } from "node:fs";
+
+type AutoVersionType = "git-describe";
+
+interface AutoVersion {
+  type: AutoVersionType;
+  tagPrefix?: string;
+}
```
tzyl, line 25: suggestion: Start this off as a type union as tagPrefix may only make sense for git-describe

```suggestion
interface GitDescribeAutoVersionConfig {
  type: "git-describe";
  tagPrefix?: string;
}

type AutoVersionConfig = GitDescribeAutoVersionConfig;
type AutoVersionType = AutoVersionConfig["type"];
```

packages/cli/src/util/config.ts
```diff
@@ -0,0 +1,113 @@
...
+    foundryUrl: { type: "string" },
+    site: {
+      type: "object",
+      properties: {
+        application: { type: "string" },
+        directory: { type: "string" },
+        autoVersion: {
+          type: "object",
+          nullable: true,
+          properties: {
```
tzyl, line 59: Can we express this more correctly as the actual type union? Maybe oneOf?

packages/cli/src/util/token.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(token).toBe(validToken);
+  });
+
+  it("should load a valid token from an environment variable", async () => {
+    vi.stubEnv("FOUNDRY_TOKEN", validToken);
+
+    const token = await loadToken();
+
+    expect(token).toBe(validToken);
+    delete process.env.FOUNDRY_TOKEN;
```
tzyl, line 48: Should we be using `vi.unstubAllEnvs()` rather than delete here?

packages/cli/src/util/token.ts
tzyl: I meant that we will want to see the debug message if validate throws otherwise it will just say token is not valid but it would be helpful to know which path we entered during debugging to know which token it's referring to

zeyadkhaled: Gotcha 👍  That makes sense

packages/cli/src/util/autoVersion.ts
zeyadkhaled: I think omitting always might be better because the error message will probably be more descriptive and helpful to the user. Given we chatted about this workflow being more important for development, then supporting the unclean state should be considered 👍

packages/cli/src/util/autoVersion.ts
zeyadkhaled: Yeah, I just played around with having the tagPrefix set to `v` and it looks good even in the cases where we have the string not starting with a `v`. Add a test case to cover this. Incase it becomes evident that we want to support multiple defaults say (v, version,ver), I think the implementation could allow for that easily.

packages/cli/src/util/token.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(token).toBe(validToken);
+  });
+
+  it("should load a valid token from an environment variable", async () => {
+    vi.stubEnv("FOUNDRY_TOKEN", validToken);
+
+    const token = await loadToken();
+    expect(token).toBe(validToken);
+
+    vi.unstubAllEnvs();
```
tzyl, line 48: Maybe move to `afterEach` to make sure things are cleaned up

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { execSync } from "node:child_process";
+import { isValidSemver } from "./isValidSemver.js";
+
+/**
+ * Gets the version string using git describe.
```
tzyl, line 21: Worth a note about the behaviour when tagPrefix is empty string? (all tags, expects version as tag possibly with v prefix) We'll have to document this on the other layers on the CLI and user docs as well later obviously

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param tagPrefix The prefix to use for matching against tags. Defaults to an empty string.
+ * @returns A promise that resolves to the version string.
+ * @throws An error if the version string is not SemVer compliant or if the version cannot be determined.
+ */
+export async function autoVersion(tagPrefix: string = ""): Promise<string> {
+  const matchRegExp = new RegExp(tagPrefix == "" ? "v?" : `^${tagPrefix}`);
```
tzyl, line 27:
```suggestion
  const matchRegExp = new RegExp(tagPrefix == "" ? "^v?" : `^${tagPrefix}`);
```

packages/cli/src/util/autoVersion.ts
```diff
@@ -0,0 +1,46 @@
...
+  const matchClause = tagPrefix != "" ? ` --match="${matchRegExp}*"` : ""; // Support filtering when tagPrefix is passed
```
tzyl, line 28: Can probably do without the comment and just read the code

```suggestion
  const matchClause = tagPrefix != "" ? ` --match="${matchRegExp}*"` : "";
```

tzyl (approved):

PR #16, opened 2024-01-31
[2/] Improvement: Introduce new command structure 

zeyadkhaled:
Part of an effort to improve the osdk cli. 
The work to get the improvements across will be done over the followings PRs:
[1/] [Add new utility functions](https://github.com/palantir/osdk-ts/pull/15) 
[2/][ Introduce new command structure ](https://github.com/palantir/osdk-ts/pull/16) --> this
[3/] [Use loadedToken for fetching data](https://github.com/palantir/osdk-ts/pull/17)

## This PR

Deleted the following subfolders under site:
- versions
- upload
- delete

Updated the following subfolder under site:
- deploy

Added the following subfolders under site:
- version
    - list
    - set
    - unset
    - delete
    - get

The cli now depends on a the config file (if found) to set default values to arguments and decide on behavior of the commands conflicting.

## New Command Structure

If a project config file(foundry.config.json) with required values is provided, most of these commands without additional arguments.

- osdk unstable site version list              List application versions
- osdk unstable site version set <version>     Set the live site version
- osdk unstable site version unset             Clear the currently set live site ve
- osdk unstable site version delete <version>  Delete the specified site version.
- osdk unstable site version get               Get the currently set live site version
- osdk unstable site deploy                    Deploy an uploaded version            

## Considerations

All requests are still being made using `FOUNDRY_SDK_AUTH_TOKEN` environment variable, the token and tokenFile arguments aren't respected yet. This will be handled in part 3.


packages/cli/src/commands/site/deploy/handleSiteDeploy.mts
```diff
@@ -17,28 +17,76 @@
...
+  }
 
-  const ctx = createConjureContext(baseUrl, "/artifacts/api");
-  if (version) {
+  consola.start("Zippping site files");
+
+  const archive = archiver("zip").directory(directory, false);
+
+  await Promise.all([
+    artifacts.SiteAssetArtifactsService.uploadZippedSiteAsset(
```
tzyl, line 68: I notice when on errors such as version already exists we also print out the CLI help text which makes it hard to see the error at the bottom do you know where that's coming from?

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -14,54 +14,118 @@
...
+          "Common Arguments",
+        )
+        .check((argv) => {
+          // This is required because we can't use demandOption with conflicts. conflicts protects us against the case where both are provided.
+          // So this case is for when nothing is provided.
+          if (
+            autoVersion == null && argv.autoVersion == null
+            && argv.version == null
+          ) {
+            throw new Error(
```
tzyl, line 94: What's the difference in error handling when throwing ExitProcessError vs Error here?

packages/cli/src/commands/site/version/delete/versionDeleteCommand.mts
```diff
@@ -15,17 +15,19 @@
...
-import { artifacts } from "../../../net/index.mjs";
-import type { DeleteArgs } from "./DeleteArgs.js";
+import { artifacts } from "../../../../net/index.mjs";
+import type { SiteVersionArgs } from "../siteVersionArgs.js";
 
-export default async function siteDeleteCommand(args: DeleteArgs) {
+export default async function versionDeleteCommand(
+  { version, application, foundryUrl }: SiteVersionArgs,
+) {
   await artifacts.SiteAssetArtifactsService.deleteSiteVersion(
```
tzyl, line 24: It looks like some of these handwritten clients use default consola and always print out the success status code (204) on debug level. We can probably get rid of this and just rely on the shared consola with this showing up when verbose is enabled.

tzyl: Additionally, the delete here is idempotent which means if you put a non-existent version in it will always say successful not sure if this will be slightly misleading

packages/cli/src/commands/site/deploy/handleSiteDeploy.mts
```diff
@@ -17,28 +17,76 @@
...
+    throw new ExitProcessError(
+      2,
+      "Either version or autoVersion must be specified",
+    );
+  }
+
+  const siteVersion = !version ? await findAutoVersion(gitTagPrefix) : version;
+  if (!version) {
+    consola.info(
+      `No version was specified, and autoVersion is enabled. Inferred version: ${siteVersion}`,
```
tzyl, line 53:
```suggestion
      `Auto version inferred next version to be: ${siteVersion}`,
```

packages/cli/src/commands/site/deploy/handleSiteDeploy.mts
```diff
@@ -17,28 +17,76 @@
...
+    );
+  }
```
tzyl, line 55: We are assuming only one of these are defined at this point is that right? If that's true then I think it might be clearer to understand if we flip the check to be `if autoVersion` rather than `if !version`

packages/cli/src/commands/site/version/get/versionGetCommand.mts
```diff
@@ -0,0 +1,41 @@
...
+  createConjureContext,
+  thirdPartyApplicationService,
+} from "#net";
+import { consola } from "consola";
+import type { CommonSiteArgs } from "../../CommonSiteArgs.js";
+
+export default async function versionGetCommand(
+  { foundryUrl, application }: CommonSiteArgs,
+) {
+  consola.start("Getting deployed version");
```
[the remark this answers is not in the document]
zeyadkhaled: For the most part it does, so I will just add it to the ones that don't exist.

packages/cli/src/cli.ts
```diff
@@ -44,11 +47,22 @@ export async function cli(args: string[] = process.argv) {
...
-          .demandCommand();
+          .demandCommand()
+          .check((_args) => {
+            if (config != null) {
+              consola.debug(
+                `Using configuration from file: "${config.configFilePath}"`,
+              );
+            }
+            return true;
+          });
```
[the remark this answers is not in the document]
zeyadkhaled: This is the one location that I found where the logging level had already been set. I will try to work on this with the lazy loading of the config file.

packages/cli/src/cli.ts
```diff
@@ -44,11 +47,22 @@ export async function cli(args: string[] = process.argv) {
       command: "unstable",
       aliases: ["experimental"],
       builder: async (argv) => {
+        const config: LoadedFoundryConfig | undefined =
+          await loadFoundryConfig();
```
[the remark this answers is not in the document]
zeyadkhaled: Actually this could be moved outside the scope of the builder to a higher level because the logging now works as part of the `.check`

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -14,54 +14,118 @@
...
+              }
+              : {},
+          },
+        }).group(
+          ["autoVersion", "gitTagPrefix"],
+          "Autoversion Arguments",
+        )
+        .group(
+          ["version", "directory", "uploadOnly"],
+          "Common Arguments",
```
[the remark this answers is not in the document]
zeyadkhaled: Are you mentioning that it would be better to move these arguments on each of the subcommands? So that when you write a non-complete command like `osdk unstable site version` you won't get these "common" arguments filed?

packages/cli/src/cli.ts
```diff
@@ -19,12 +19,14 @@ import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
 import type { CliCommonArgs } from "./CliCommonArgs.js";
 import auth from "./commands/auth/index.js";
-import site from "./commands/site/index.js";
+import siteHandler from "./commands/site/index.js";
 import typescript from "./commands/typescript/index.js";
 import { ExitProcessError } from "./ExitProcessError.js";
+import { logConfigFileMiddleware } from "./util/configLoader.js";
```
tzyl, line 25: Let's put this together with existing middleware

```suggestion
import { logConfigFileMiddleware } from "./yargs/configLoader.js";
```

packages/cli/src/cli.ts
```diff
@@ -39,6 +41,7 @@ export async function cli(args: string[] = process.argv) {
     )
     .demandCommand()
     .middleware(logVersionMiddleware, true)
+    .middleware(logConfigFileMiddleware, true)
```
tzyl, line 44: Is it desirable to run before validation with `true` for this middleware? If we wanted exactly the same behaviour as check I believe it runs with false

packages/cli/src/util/configLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+    const loadedConfig = await loadFoundryConfig();
+    config = loadedConfig?.foundryConfig;
+    configFilePath = loadedConfig?.configFilePath;
+  }
+  return config;
+}
+
+// This middleware will be run after parsing is complete so we know
+// we have loaded the config prior during construction of the commands
+export async function logConfigFileMiddleware(): Promise<void> {
```
tzyl, line 34: Let's move this out to a separate middleware file alongside the verbose log level setting middleware

packages/cli/src/util/configLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (config === null) {
+    const loadedConfig = await loadFoundryConfig();
+    config = loadedConfig?.foundryConfig;
+    configFilePath = loadedConfig?.configFilePath;
+  }
+  return config;
+}
+
+// This middleware will be run after parsing is complete so we know
+// we have loaded the config prior during construction of the commands
```
tzyl, line 33: Maybe can get rid of this comment now that it's clear enough where this is being applied in the yargs builder

packages/cli/src/util/configLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { FoundryConfig } from "./config.js";
+import { loadFoundryConfig } from "./config.js";
+
+let config: FoundryConfig | undefined | null = null;
+let configFilePath: string | undefined | null = null;
```
tzyl, line 21: Can we group both together in the loaded foundry config interface? Some people also prefer not to rely on different meanings for undefined vs null (so use double equals for nullish) but if we use the promise approach then it doesn't matter:

```suggestion
let configPromise: Promise<LoadedFoundryConfig | undefined> | undefined = undefined;
```

packages/cli/src/util/configLoader.ts
```diff
@@ -0,0 +1,44 @@
...
+
+async function getConfig(): Promise<FoundryConfig | undefined> {
+  if (config === null) {
```
tzyl, line 24: It would technically be possible for us to load the foundry config multiple times if this is called before the internal promise resolves. To prevent that we can check and reuse the promise between multiple calls e.g.

```
function getConfig(): Promise<LoadedFoundryConfig | undefined> {
  if (configPromise == null) {
    configPromise = loadFoundryConfig();
  }
  return configPromise;
}
```

packages/cli/src/cli.ts
tzyl: I'm noticing this is run multiple times is this because it runs for each sub command as well? We may want to follow the existing version middleware which only logs on first run

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -14,54 +14,126 @@
...
+              `Overriding "autoVersion" from config file with ${argv.autoVersion}`,
+            );
+            if (argv.autoVersion != "git-describe") {
+              throw new Error(
+                `Only 'git-describe' is supported for autoVersion`,
+              );
+            }
+          }
+
+          if (directory != null && argv.directory !== directory) {
```
tzyl, line 118: As a follow up I think these logging side effects also don't make sense to be part of "check" as they aren't actually checking for any violations

packages/cli/src/commands/site/version/delete/versionDeleteCommand.mts
```diff
@@ -15,17 +15,18 @@
...
-import { artifacts } from "../../../net/index.mjs";
-import type { DeleteArgs } from "./DeleteArgs.js";
+import { artifacts } from "../../../../net/index.mjs";
+import type { SiteVersionArgs } from "../SiteVersionArgs.js";
 
-export default async function siteDeleteCommand(args: DeleteArgs) {
+export default async function versionDeleteCommand(
+  { version, application, foundryUrl }: SiteVersionArgs,
+) {
   await artifacts.SiteAssetArtifactsService.deleteSiteVersion(
```
tzyl, line 24: Let's add a start message for this one as well

packages/cli/src/commands/site/version/get/versionGetCommand.mts
```diff
@@ -0,0 +1,45 @@
...
+
+  const deployedVersion = await ArtifactsSitesAdminV2Service.getDeployedVersion(
+    ctx,
+    repositoryRid,
+  );
+
+  if (!deployedVersion) {
+    consola.info("No live site version set");
+  } else {
+    consola.success(`Current deployed version is ${deployedVersion.version}`);
```
tzyl, line 43:
```suggestion
    consola.success(`Live site version is ${deployedVersion.version}`);
```

packages/cli/src/commands/site/deploy/siteDeployCommand.mts
```diff
@@ -0,0 +1,94 @@
...
+  thirdPartyApplicationService,
+} from "#net";
+import archiver from "archiver";
+import * as fs from "node:fs";
+import { Readable } from "node:stream";
+import { ExitProcessError } from "../../../ExitProcessError.js";
+import { autoVersion as findAutoVersion } from "../../../util/autoVersion.js";
+import type { SiteDeployArgs } from "./SiteDeployArgs.js";
+
+export default async function handleSiteDeploy(
```
tzyl, line 32:
```suggestion
export default async function siteDeployCommand(
```

packages/cli/src/commands/site/version/index.ts
```diff
@@ -14,9 +14,31 @@
...
+    return argv
+      .command(list)
+      .command(set)
+      .command(unset)
+      .command(deleteCmd)
+      .command(get)
+      .demandCommand();
+  },
+  handler: async (args) => {
+  },
```
tzyl, line 41:
```suggestion
  handler: async (args) => {},
```

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -14,54 +14,126 @@
...
-          type: "boolean",
-          conflicts: "siteVersion",
-          // group: "Deploy Version",
-          // implies: { "siteVersion": "" },
-        },
-      }).group(
-        ["siteVersion", "clearVersion"],
-        "Version To Deploy (requires one of)",
-      );
+async function deployHandler(): Promise<
```
tzyl, line 24: Is it possible to keep these as constants like they were before and move the async config loading into the builder? For example `.builder()` accepts returning a `PromiseLike`

packages/cli/src/commands/site/index.ts
```diff
@@ -14,40 +14,83 @@
...
 import type { CommonSiteArgs } from "./CommonSiteArgs.js";
-import siteDelete from "./delete/index.js";
-import siteDeploy from "./deploy/index.js";
-import upload from "./upload/index.js";
-import versions from "./versions/index.js";
+import deploy from "./deploy/index.js";
+import version from "./version/index.js";
 
-const site: yargs.CommandModule<CliCommonArgs, CommonSiteArgs> = {
+export const command: CommandModule<CliCommonArgs, CommonSiteArgs> = {
```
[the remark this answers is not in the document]
zeyadkhaled: I agree, I was just trying to keep it consistent. But I think its better to make it clean

PR #19, opened 2024-01-31
Add @osdk/create-app package

tzyl:
Initial code for package for CLI to bootstrap OSDK apps with guided prompts and multiple tempaltes rather than cloning a starter repo and following instructions to update multiple files


packages/create-osdk-app/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "create-osdk-app",
+  "private": true,
```
tzyl, line 3: Will turn on publishing in a separate step

packages/create-osdk-app/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
+    "dev:transpile": "tsup --watch",
+    "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
+    "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "transpile": "tsup",
+    "transpileWatch": "tsup --watch",
+    "typecheck": "tsc-absolute --build",
+    "update-snapshots": "UPDATE_SNAPSHOTS=true vitest run"
```
tzyl, line 28: This is the only custom script in the package all others are the same as monorepolint

packages/create-osdk-app/src/cli.ts
```diff
@@ -0,0 +1,408 @@
...
+      consola.info(`Reusing existing project directory`);
+    }
+  } else {
+    consola.info(`Creating project directory`);
+    fs.mkdirSync(root, { recursive: true });
+  }
+
+  consola.info(`Copying files into project directory`);
+
+  const templatesDir = findUpSync("templates", {
```
tzyl, line 124: Originally this had a hardcoded relative path but this is slightly different now running from test or bundle:

When run in vitest this looks for `create-osdk-app/src/cli.ts` -> `create-osdk-app/templates`
When run in tsup bundle this looks for `create-osdk-app/build/js/index.mjs` -> `create-osdk-app/templates`

packages/create-osdk-app/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "create-osdk-app",
```
styu, line 2: should this have the `@osdk/` prefix? or is that handled somewhere else?

styu (commented):
setup largely lgtm. i can't comment on the file directly, but it does feel like we should find a different .svg instead of the resource icon to use?

mfedderly (commented):
I'm good with the changes outside of packages/create-osdk-app. I didn't look at the code inside of that package though.

packages/create-osdk-app/package.json
mfedderly: (I think that it has to be named this in order for us to match the other `npx create-foo-app` syntax floating around the JS ecosystem)

One thing we want to do though is make sure the NPM token has permissions to publish to this package as well.

PR #20, opened 2024-01-31
Send consistent UserAgent strings

mfedderly:
We seem to have a pretty broad need for getting the current package's version for UA strings, logging, etc. This give it to us at compile time, no package.json finding or async file reads at runtime.

This changes the various network requests to include the relevant entries from the list below:

```
osdk-client/version - requests coming from  the 2.0 client
osdk-legacy-client/version - requests coming from the 1.1 client
typescript-sdk/version - the version of the user-generated ontology, version only
typescript-sdk-generator/version
  - the version of the @osdk/foundry-sdk-generator used to generate the ontology
osdk-cli/version - a handful of requests originate from @osdk/cli directly
```


.lintstagedrc.mjs
```diff
@@ -23,7 +23,7 @@ export default {
   "*": (files) => {
     const mrlFiles = micromatch(files, [
       "package.json",
-      "**/package.json",
+      // "**/package.json",
```
mfedderly, line 26: Temporarily disabling this until we fix the issues with dprint and autorelease arguing about whitespace.

packages/cli/src/commands/typescript/generate/generate.ts
```diff
@@ -67,12 +67,17 @@ export const command: CommandModule<
             default: "commonjs",
             choices: ["commonjs", "module"],
           },
+          version: {
+            type: "string",
+            description: "Version of the generated code, or 'dev'",
+            demandOption: true,
+          },
```
mfedderly, line 74: Do we want to restrict this to only valid semver strings or something?

packages/generator/src/v2.0/generateMetadata.ts
```diff
@@ -31,7 +32,7 @@ export async function generateOntologyMetadataFile(
       export const OntologyMetadata = {
         ontologyRid: "${ontology.ontology.rid}",
         ontologyApiName: "${ontology.ontology.apiName}",
-        userAgent: "foundry-typescript-osdk/2.0.0",
+        userAgent: "${userAgent}",
```
mfedderly, line 35: this tripped me up on self-review, the whole thing is actually inside backticks so this is valid syntax.

packages/cli/src/commands/typescript/generate/generate.ts
tzyl: Hmm, do you know what foundry-sdk-generator CLI does? In Dev Console the server side SDK versions are validated before the CLI gets called I'm not sure if the CLI itself also validates. 

How valuable do you think this is for the intended usage of this CLI? I'm not too familiar with the plans on the 2.0 CLI which only generates the files but not the full NPM package with a name/version

tzyl (approved):

mfedderly:
foundry-sdk-generator checks the command line version string against semver before accepting it. We're currently using the cli to just do codegen directly into a project and piggy back its bundling process. Given that, I'm not even sure that we can provide meaningful versions in their user agents because the same generated code is potentially used across a large range of shipped versions.

I guess I'll just check that it's dev or semver just in case and then just tell them to use `dev` for now.

PR #21, opened 2024-01-31
Remove v2 client Attachment class and use typed strings instead

mfedderly:


packages/client/src/util/toDataValue.ts
```diff
@@ -36,8 +35,11 @@ export function toDataValue(value: any): DataValue {
     return Array.from(value, toDataValue);
   }
 
-  // attachments just send the rid directly
-  if (isAttachment(value)) {
+  // if we have an object that only has a rid property, assume its an attachment and send the rid directly
+  // it is possible for the user to pass in an AttachmentRid directly, which will just get passed through at the end
+  if (
+    typeof value === "object" && value.rid && Object.keys(value).length === 1
+  ) {
```
mfedderly, line 42: ^ the whole PR hinges on our ability to do this. Its a little sketchy because any `{rid: string}` will get caught in this path, and we support structs in a later codepath.

ericanderson: I read this a few times and I think this is relatively okay. I am trying to think of a scenario when you have exactly a `{rid: string}` given your keys length check and the wire can't handle just getting the rid string.

I guess the question is why would they have this `{rid: string}`. If its because we created one of these for them, we could always add a marker when we do with a symbol or something like that. Doesn't enumerate. Maybe? Or leverage a parent prototype.

If they are passing a rid string directly this problem doesnt exist of course.

mfedderly:
Sad that this didn't work out quite well enough, going to close.

We also considered some kind of javascript guards like adding a non-enumerable symbol to the attachment object, but at that point why not just have a class so we can hang methods off of it?

PR #22, opened 2024-02-01
Update docs about changelog/release process

tzyl:
Include changelog folders in published package
Rename old CHANGELOG.md files to CHANGELOG_OLD.md
Note about changesets usage


mfedderly (approved):
I'm not actually sure why we're shipping our changelogs in the final packages, but we're being consistent here at least. Thanks for the docs update!

PR #23, opened 2024-02-01
action params type fix

ssanjay1:


packages/legacy-client/src/client/actions/actions.ts
```diff
@@ -45,21 +45,37 @@ export interface ValidLegacyActionParameterTypes {
 export type ActionArgs<
   O extends OntologyDefinition<any>,
   A extends keyof O["actions"],
-> =
-  & {
+> = NonNullableKeys<O["actions"][A]["parameters"]> extends never ? {
```
ssanjay1, line 48: Basically, the problem was that our types as they were allowed you to pass in arrays to action args (based on our tests atleast). It was because if `NonNullableKeys` or `NullableKeys` were `never`, the type would look something like `{nullableKeys} & {}`, and for some reason having that `& {}` just allowed users to put in anything. Attempted to fix that here, but there may be a cleaner solution for this

mfedderly: Specifically `{[P in never]: any}` accepts `[]`

packages/legacy-client/src/client/actions/actions.test.ts
```diff
@@ -81,6 +80,10 @@ describe("Actions", () => {
         ]
       >();
 
+      expectTypeOf<typeof actions.createTask>()
```
ssanjay1, line 83: Not really sure if this is the best thing to do here. Using `toBeCallableWith` was the only way I could get the arrays to pass through when we had the **_incorrect_** types, unfortunately, you can't use `.not` to check that we don't regress this behavior, so I put in expect-error, which should flag an error if it actually is callable with the array.

packages/legacy-client/src/client/actions/actions.test.ts
```diff
@@ -32,14 +32,13 @@ import {
...
-import {
-  type ActionError,
-  ActionExecutionMode,
-  type ActionExecutionOptions,
-  type ActionResponseFromOptions,
-  type Edits,
-  type Result,
-  ReturnEditsMode,
+import { ActionExecutionMode, ReturnEditsMode } from "../..";
+import type {
```
ssanjay1, line 36: This happened automatically on save, not sure if this is what we want

mfedderly: 🤷 unless you want to propose changes to dprint's config its fine by me

packages/legacy-client/src/client/actions/actions.test.ts
mfedderly: Yeah I think this as good as we can get for now. Once you add bulk actions support this test is going to just get deleted I guess anyhow.

mfedderly (approved):
🥳

PR #31, opened 2024-02-05
Refactor create-app dotfile generation

tzyl:
- Match repo structure with test file next to source file
- Extract `TemplateContext` type to make it clearer what variables are available during templating
- Refactor .npmrc / .env file generation with unit test coverage

The .npmrc / .env files are shared across all templates but could technically be in the individual template folders. For now I've kept the structure the same as before


packages/create-app/src/generate/generateNpmRc.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+export function generateNpmRc({
+  osdkPackage,
+  osdkRegistryUrl,
+}: {
+  osdkPackage: string;
+  osdkRegistryUrl: string;
+}): string {
+  const withoutProtocol = osdkRegistryUrl.replace(/^https:\/\//, "");
```
zeyadkhaled, line 24: I guess this is safe given `promptOsdkRegistryUrl` already tests it against a regex. If we plan to use this somewhere that isn't part of this code path, should we consider passing the protocol but have it with a default value?

zeyadkhaled (approved):

packages/create-app/src/generate/generateNpmRc.ts
tzyl: I think the only other protocol that would be relevant would be `http` but should be safe to only allow `https` for this url

PR #32, opened 2024-02-05
Refactor create-app prompts with tests

tzyl:
Move prompts to separate files with unit tests to cover the different branches that can happen with interactive prompts. The CLI snapshot tests only use the args to skip all interactive prompts so some of this is not covered automatically right now.


packages/create-app/src/cli.ts
```diff
@@ -90,7 +97,7 @@ export async function cli(args: string[] = process.argv) {
 
   const parsed: CliArgs = base.parseSync();
   const project: string = await promptProject(parsed);
-  const overwrite: boolean = await promptOverwrite(parsed, project);
+  const overwrite: boolean = await promptOverwrite({ ...parsed, project });
```
zeyadkhaled, line 100: I think its more clear to pass the actual expected object here?
```suggestion
  const overwrite: boolean = await promptOverwrite({ overwrite: parsed.overwrite, project });
```

zeyadkhaled (approved):
Looks good with just a small comment

packages/create-app/src/cli.ts
tzyl: Hmm I actually feel it's more consistent with the other prompts which is saying you can have access to all of the parsed defaults but the project is replaced with the final name from the previous prompt

PR #36, opened 2024-02-05
Add bulk actions

ssanjay1:


packages/generator/src/v1.1/generateActions.ts
```diff
@@ -43,49 +43,71 @@ export async function generateActions(
...
-        Promise<Result<ActionResponseFromOptions<O, Edits<${
-        addedObjects.length > 0
-          ? addedObjects.join(" | ")
-          : "void"
-      }, ${
-        modifiedObjects.length > 0
-          ? modifiedObjects.join(" | ")
-          : "void"
-      }>>, ActionError>>;`,
+      ${action.apiName}<P extends ${interiorParameters} | ${interiorParameters}[],O extends P extends ${interiorParameters}[]? BulkActionExecutionOptions: ActionExecutionOptions>(
```
ssanjay1, line 72: So, I made the O generic something like `O extends BulkActionExecutionOptions | ActionExecutionOptions`, which it seemed to like, but I wasn't able to test if that would still enforce our internal argument constraints because I got blocked on another type issue below

packages/generator/src/v1.1/generateActions.ts
```diff
@@ -43,49 +43,71 @@ export async function generateActions(
...
+        params:P ,options?: O): 
+        Promise<Result<P extends ${interiorParameters}[] ? BulkActionResponseFromOptions<O,Edits<${
```
ssanjay1, line 74: This also gave an issue because of the conditional with `P extends`, it couldn't tell what options to use

packages/legacy-client/src/client/actions/actions.ts
```diff
@@ -153,8 +181,19 @@ export type Actions<
       ? <Op extends ActionExecutionOptions>(
         options?: Op,
       ) => WrappedActionReturnType<O, A, Op>
-      : <Op extends ActionExecutionOptions>(
-        params: ActionArgs<O, A>,
-        options?: Op,
-      ) => WrappedActionReturnType<O, A, Op>;
+      :
+        & (<
```
ssanjay1, line 185: The &s seemed to work here

ssanjay1:
Decided to take a different approach and put bulkActions at top level instead: https://github.com/palantir/osdk-ts/pull/53

PR #38, opened 2024-02-06
Update api-gateway definitions

mfedderly:
Additionally this adds the SLS dependency information to @osdk/gateway package.json


packages/generator/src/shared/wireActionTypeV2ToSdkActionDefinition.ts
```diff
@@ -69,7 +70,7 @@ function wireActionParameterV2ToSdkParameterDefinition(
     case "array":
       return {
         multiplicity: true,
-        type: actionPropertyToSdkPropertyDefinition(value.dataType),
+        type: actionPropertyToSdkPropertyDefinition(value.dataType.subType),
```
mfedderly, line 73: Pretty sure this was just wrong before

packages/gateway-generator/src/cli/updateSls.ts
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2023 Palantir Technologies, Inc. All rights reserved.
```
styu, line 2:
```suggestion
 * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
```

styu: @ericanderson do we have eslint set up for the copyright header?

packages/gateway/package.json
```diff
@@ -52,5 +52,13 @@
...
   "main": "./build/js/index.js",
   "module": "./build/js/index.mjs",
-  "types": "./build/types/index.d.ts"
+  "types": "./build/types/index.d.ts",
+  "sls": {
+    "dependencies": {
+      "com.palantir.foundry.api:api-gateway": {
+        "minVersion": "1.763.0",
+        "maxVersion": "1.x.x"
+      }
```
styu, line 61: just so i understand, is this so this will propagate to the asset that the generator is included in? apologies i hadn't been following ultra closely

packages/gateway/src/generated/openapi/errors/ObjectTypesNotSynced.ts
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectTypeApiName } from "../components/ObjectTypeApiName";
+
+/**
+ * One or more of the requested object types are not synced into the ontology. Please reach out to your Ontology
+ * Administrator to re-index the object type(s) in Ontology Management Application.
```
styu, line 21: in theory this should only happen the very first time the object type is not synced

```suggestion
 * Administrator to index the object type(s) in Ontology Management Application.
```

mfedderly: This is just autogenerated from the upstream docs, we don't control it.

packages/gateway/package.json
mfedderly: This is halfway to that. Right now we're just adding it to @osdk/gateway so that anyone that depends on that will get it in their package tree so our internal code should pick this up. It doesn't wind up on the package built by foundry-sdk-generator OR the create-osdk-app assets.

styu: ok great, thanks for explaining! that's what I was hoping for

packages/gateway/src/generated/openapi/errors/ObjectTypesNotSynced.ts
styu: i can clearly read the file path, my bad

styu (approved):

packages/gateway-generator/src/cli/updateSls.ts
mfedderly: Its in `monorepo/eslint-config-sane/library.cjs`. We should probably split that into its own PR through.

styu: yupyup no worries

PR #39, opened 2024-02-07
Make sure to run all tests

tzyl:
- `@osdk/cli` has `test:watch` but no `test` script so turbo is not running tests under test -> check
- `@osdk/generator` has tests but looks like it was copied in and converted to vitest but also not running properly. I updated this to match the repo style with the test file next to the source file and fixed the escaping within backticks
- Some random clean up on existing tests to make them properly await promise rejections and stub out local token envs always to not conflict when running locally

### DIFF
--- a/packages/cli/src/util/config.ts
+++ b/packages/cli/src/util/config.ts
@@ -58,7 +58,7 @@ const CONFIG_FILE_SCHEMA: JSONSchemaType<FoundryConfig> = {
           oneOf: [
             {
               properties: {
-                "type": { enum: ["git-describe"], type: "string" },
+                "type": { const: "git-describe", type: "string" },
                 "tagPrefix": { type: "string", nullable: true },
               },
             },
--- a/packages/cli/src/util/token.test.ts
+++ b/packages/cli/src/util/token.test.ts
@@ -35,19 +35,25 @@ describe("loadToken", () => {
       validToken,
     );
     const token = await loadToken(undefined, "valid-token.txt");
-
     expect(token).toBe(validToken);
   });
 
-  it("should load a valid token from an environment variable", async () => {
+  it("should load a valid token from the FOUNDRY_TOKEN environment variable", async () => {
     vi.stubEnv("FOUNDRY_TOKEN", validToken);
+    vi.stubEnv("FOUNDRY_SDK_AUTH_TOKEN", "");
+    expect(await loadToken()).toBe(validToken);
+  });
 
-    const token = await loadToken();
-    expect(token).toBe(validToken);
+  it("should load a valid token from the deprecated FOUNDRY_SDK_AUTH_TOKEN environment variable", async () => {
+    vi.stubEnv("FOUNDRY_TOKEN", "");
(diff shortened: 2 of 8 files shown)


packages/cli/src/util/token.test.ts
```diff
@@ -35,19 +35,25 @@ describe("loadToken", () => {
...
+  it("should load a valid token from the deprecated FOUNDRY_SDK_AUTH_TOKEN environment variable", async () => {
+    vi.stubEnv("FOUNDRY_TOKEN", "");
+    vi.stubEnv("FOUNDRY_SDK_AUTH_TOKEN", validToken);
+    expect(await loadToken()).toBe(validToken);
   });
 
   it("should throw an error if no token is found", async () => {
-    expect(() => loadToken()).rejects.toThrow("No token found.");
+    vi.stubEnv("FOUNDRY_TOKEN", "");
+    vi.stubEnv("FOUNDRY_SDK_AUTH_TOKEN", "");
```
zeyadkhaled, line 55: Isn't that equivalent to them not being set? And is it necessary set them to empty above as well?

packages/cli/src/util/config.ts
```diff
@@ -58,7 +58,7 @@ const CONFIG_FILE_SCHEMA: JSONSchemaType<FoundryConfig> = {
           oneOf: [
             {
               properties: {
-                "type": { enum: ["git-describe"], type: "string" },
+                "type": { const: "git-describe", type: "string" },
```
zeyadkhaled, line 61: Is this because we are using oneOf so if we introduce a new "type" we will just add a new object?

zeyadkhaled (approved):
Approved on the cli changes with minor questions

packages/cli/src/util/token.test.ts
tzyl: This is for running the tests locally where you may already have your own env set with these variables so the test can be affected by your local env. We make sure to stub out all the input env variables for the test to get a predictable result

packages/cli/src/util/config.ts
tzyl: Yes, I believe so otherwise we wouldn't use oneOf at all if we wanted to express everything under one object but then you wouldn't get the stricter validation on which properties belong to which type in the union and everything would have to be nullable

packages/gateway-generator/src/generate/error.test.ts
```diff
@@ -47,8 +47,8 @@ describe("Errors", () => {
     const sourceFile = project.getSourceFile("errors/MyError.ts");
     expect(sourceFile?.getFullText()).toMatchInlineSnapshot(`
             "export interface MyError {
-                errorCode: \\"ERROR_TYPE\\";
-                errorName: \\"MyError\\";
+                errorCode: "ERROR_TYPE";
+                errorName: "MyError";
```
mfedderly, line 51: ^ the snapshots changed when I upgraded vitest a while back

mfedderly (approved):
How close are we to just enforcing `test` and `test:watch` scripts in MRL? I'm not sure there's any packages where these shouldn't exist at this point.

PR #41, opened 2024-02-07
Display errors more consistently in @osdk/cli

zeyadkhaled:
- Throw ExitProcessError from all clients and command handlers
- Throw normal Errors in yargs validation and check paths
- Use yargs [`fail`](https://yargs.js.org/docs/#api-reference-failfn-boolean) to handle failures manually and display `ExitProcessError` as consola.error without the cli help text

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -59,12 +59,27 @@ export async function cli(args: string[] = process.argv) {
       if (err instanceof ExitProcessError) {
         Consola.consola.error(err);
       } else {
-        yargs.showHelp();
-        // eslint-disable-next-line no-console
-        console.error(msg);
+        // if yargs.check returns a string it populates the err variable with a string
+        if (err && typeof err !== "string") {
+          throw err;
+        } else {
+          yargs.showHelp();
+          // eslint-disable-next-line no-console
+          console.log(""); // intentional new line
+          // eslint-disable-next-line no-console
+          console.error(msg);
+        }
       }
       process.exit(1);
     });
 
-  return base.parseAsync();
+  // Special handling where failures happen before yargs does its error handling within .fail
+  try {
+    return await base.parseAsync();
+  } catch (err) {
+    if (err instanceof ExitProcessError) {
+      const Consola = await import("consola");
+      Consola.consola.error(err);
+    }
+  }
 }
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -53,14 +53,18 @@ export async function cli(args: string[] = process.argv) {
           .demandCommand();
       },
       handler: (_args) => {},
+    })
+    .fail(async (msg, err, yargs) => {
+      const Consola = await import("consola");
+      if (err && err instanceof ExitProcessError) {
+        Consola.consola.error(err);
+      } else {
+        yargs.showHelp();
+        // eslint-disable-next-line no-console
+        console.error(msg);
+      }
+      process.exit(1);
     });
 
-  try {
-    return base.parseAsync();
-  } catch (e) {
-    if (e instanceof ExitProcessError) {
-      const Consola = await import("consola");
-      Consola.consola.error(e.message);
-    }
-  }
+  return base.parseAsync();
 }
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -55,15 +55,23 @@ export async function cli(args: string[] = process.argv) {
       },
       handler: (_args) => {},
     })
-    .fail(async (msg, err, argv) => {
+    .fail(async (msg, err, yargsContext) => {
       const Consola = await import("consola");
+      const isVerbose = args.some(arg =>
+        ["-v", "--v", "--verbose"].includes(arg)
+      );
+
       if (err instanceof ExitProcessError) {
-        Consola.consola.error(err);
+        if (isVerbose) {
+          Consola.consola.error(err);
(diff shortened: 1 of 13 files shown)


packages/cli/src/cli.ts
```diff
@@ -53,14 +53,18 @@ export async function cli(args: string[] = process.argv) {
...
+    .fail(async (msg, err, yargs) => {
+      const Consola = await import("consola");
+      if (err && err instanceof ExitProcessError) {
+        Consola.consola.error(err);
+      } else {
+        yargs.showHelp();
+        // eslint-disable-next-line no-console
+        console.error(msg);
+      }
+      process.exit(1);
```
tzyl, line 66: - If err is defined but not ExitProcessError should we follow the yargs example and rethrow the error with stacktrace? These should be uncaught runtime exceptions given the ExitProcessError special casing is msg defined in these cases?
- What are the cases in which err is not defined is that for built in validation errors from yargs? Is it desirable to show help before or after the message? The yargs examples seem to default to printing help after the message

zeyadkhaled: > What are the cases in which err is not defined is that for built in validation errors from yargs? Is it desirable to show help before or after the message? The yargs examples seem to default to printing help after the message

Yup it seems to be for built in validation errors. The default out of the box case is to actually showHelp then print 
the message.  I prefer the error message to be shown last as that is more or less the expectation when faced with an error.

> If err is defined but not ExitProcessError should we follow the yargs example and rethrow the error with stacktrace? These should be uncaught runtime exceptions given the ExitProcessError special casing is msg defined in these cases?

I am not sure about this one, because the Errors thrown during the `check` will be in that category and for those I think we don't want to show the stacktrace because it doesn't add value to the user. So if we did want to show stacktrace to runtime exceptions we can throw a different kind of Error inside of `check` perhaps define a `CheckError`.

This is what this would look like for passing an invalid foundryUrl:

image

tzyl: What happens if you return string rather than an error in `.check()` out of interest? Does yargs throw that in a wrapped error or does it populate msg?
https://yargs.js.org/docs/#api-reference-checkfn-globaltrue

zeyadkhaled: Sweet, I played around and yargs returns the string in place of `err` object. So adding to the else part of fail this:

- String returned from check (Show help + returned string) OR validation error:
image
image

- ExitProcessorError (no help + Error and stacktrace with consola )
image

- runtime exception
image

zeyadkhaled: There is a 4th kind where any Error kind is thrown before yargs does it's parsing for example I changed the ConfigFile loading to throw ExitProcessError but that is handled outside of the `.fail` because yargs hasn't reached that part yet.

image


But then if I re-add the original try/catch around parseAsync then I can special handle ExitProcessError again 👍 
image

zeyadkhaled: That way I think we covered all possible handled/unhandled exceptions/checks within and outside yargs

packages/cli/src/cli.ts
```diff
@@ -53,14 +53,33 @@ export async function cli(args: string[] = process.argv) {
           .demandCommand();
       },
       handler: (_args) => {},
+    })
+    .fail(async (msg, err, yargs) => {
+      const Consola = await import("consola");
+      if (err instanceof ExitProcessError) {
```
tzyl, line 59: There are some places that use ExitProcessError with `errorCode` which we don't log anywhere but note that it's not an exit code but the error code from the http requests. This means we're currently logging the stacktrace by using the whole `err` object. I could see an option where we just always include the stacktrace in verbose mode as well

packages/cli/src/cli.ts
```diff
@@ -53,14 +53,33 @@ export async function cli(args: string[] = process.argv) {
...
+        Consola.consola.error(err);
+      } else {
+        // if yargs.check returns a string it populates the err variable with a string
+        if (err && typeof err !== "string") {
+          throw err;
+        } else {
+          yargs.showHelp();
+          // eslint-disable-next-line no-console
+          console.log(""); // intentional new line
```
tzyl, line 68: Would `consola.log("")` work? Same in the other existing places that use this eslint disable

packages/cli/src/cli.ts
```diff
@@ -53,14 +53,33 @@ export async function cli(args: string[] = process.argv) {
           .demandCommand();
       },
       handler: (_args) => {},
+    })
+    .fail(async (msg, err, yargs) => {
```
tzyl, line 57: In this case I think we should be careful with `yargs` as we're shadowing the import which could cause confusion maybe `argv` would be better

packages/cli/src/cli.ts
```diff
@@ -53,14 +53,33 @@ export async function cli(args: string[] = process.argv) {
...
+      const Consola = await import("consola");
+      if (err instanceof ExitProcessError) {
+        Consola.consola.error(err);
+      } else {
+        // if yargs.check returns a string it populates the err variable with a string
+        if (err && typeof err !== "string") {
```
tzyl, line 63: This is maybe not so obvious to be more explicit we could define a similar custom error class for custom check/validation errors?

zeyadkhaled: Yeah I suggested that above. I agree that would be the cleanest will do it

tzyl (approved):
Approving to unblock let's track the follow ups

PR #42, opened 2024-02-07
Move argv override logging to middleware

zeyadkhaled:
Addresses <https://github.com/palantir/osdk-ts/pull/16#discussion_r1481347100>


packages/cli/src/commands/site/index.ts
```diff
@@ -68,25 +68,15 @@ const command: CommandModule<CliCommonArgs, CommonSiteArgs> = {
...
-          );
-        }
-
-        if (foundryUrl != null && argv.foundryUrl !== foundryUrl) {
-          consola.debug(
-            `Overriding "foundryUrl" from config file with ${argv.foundryUrl}`,
-          );
-        }
-
+        argv.foundryUrl = argv.foundryUrl.replace(/\/$/, "");
```
tzyl, line 71: This also seems like a weird side effect to put in a check method and not obvious when this would run in the yargs lifecycle. Can we do this in a more explicit yargs method for example coerce or something else?

packages/cli/src/yargs/logCommandDebugMiddlewares.ts
```diff
@@ -0,0 +1,69 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Arguments } from "yargs";
+import type { CommonSiteArgs } from "../commands/site/CommonSiteArgs.js";
+import type { SiteDeployArgs } from "../commands/site/deploy/SiteDeployArgs.js";
+import type { GitDescribeAutoVersionConfig } from "../util/config.js";
+
+export async function logSiteCommandDebugMiddleware(
```
tzyl, line 22: Hmm, these feel extremely specific to the single command they live in so I think it would be fine and easier to follow to just define them in the same file as the command

tzyl, line 22: suggestion: not obvious what "debug middleware" will do we can name it more intentionally for what it logs and if we need other debug middleware we can always chain multiple small specific middlewares together rather than putting everything in a single middleware
```suggestion
async function logConfigFileOverrides(argv, config)
```

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -101,30 +101,24 @@ const command: CommandModule<
...
-            );
-          }
-        }
-
-        if (directory != null && argv.directory !== directory) {
-          consola.debug(
-            `Overriding "directory" from config file with ${argv.directory}`,
+        if (
+          autoVersion != null && argv.autoVersion !== autoVersion.type
+          && argv.autoVersion !== "git-describe"
```
tzyl, line 106: Minor note I feel this block and the one above are going to be difficult to maintain if we add another auto version type. The checks also seem slightly loose for example the above one will throw with `--autoVersion git-describe --tagPrefix abc` if there is no config file but that would be a valid state

packages/cli/src/yargs/logCommandDebugMiddlewares.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+
+  if (directory != null && argv.directory !== directory) {
+    consola.debug(
+      `Overriding "directory" from config file with ${argv.directory}`,
+    );
+  }
+
+  if (gitTagPrefix != null && argv.gitTagPrefix !== gitTagPrefix) {
```
tzyl, line 64: Let's also make these checks tighter currently this is going to say misleading things like: `Overriding "gitTagPrefix" from config file with undefined`

packages/cli/src/commands/site/deploy/index.ts
zeyadkhaled: Yeah I agree, but it is required I think to not get the user in a state that doesn't make sense.

As for the part that didn't work, that was just because I forgot to check if autoVersion was defined.

Open to ideas here honestly to make this easier to handle.

packages/cli/src/yargs/logCommandDebugMiddlewares.ts
zeyadkhaled: Which command did you run to get to that state?

tzyl: This is because I have `"tagPrefix": ""` in my config file which I think would be good to illustrate that the key exists if someone discovers it and wants to edit it. We probably have some loose truthy checks on strings which is why some people use `===` everywhere

packages/cli/src/commands/site/index.ts
```diff
@@ -61,32 +61,24 @@ const command: CommandModule<CliCommonArgs, CommonSiteArgs> = {
           description: "Path to a file containing your Foundry API Token",
         },
       })
+      .coerce({
+        foundryUrl: (foundryUrl) => foundryUrl.replace(/\/$/, ""),
+      })
```
tzyl, line 66: Can we make this match the existing coerce where we do it directly in the option definition it relates to? That way I think we don't need to specify the key again and it's more tightly grouped with the definition

packages/cli/src/commands/site/deploy/index.ts
tzyl: I would recommend breaking up the check a bit for example:

```
const autoVersionType = argv.autoVersion ?? autoVersion;
if (autoVersionType !== "git-describe") { throw }
```

To make it extensible in the future we may be able to do something with the AutoVersionType type and visit against each of the cases in the union and if we don't match any then we would throw

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -92,7 +92,7 @@ const command: CommandModule<
         }
 
         if (
-          (autoVersion?.type !== "git-describe"
+          ((autoVersion != null && autoVersion.type !== "git-describe")
```
tzyl, line 95: maybe:


```
const autoVersionType = argv.autoVersion ?? autoVersion;
if (argv.gitTagPrefix != null && autoVersionType !== "git-describe") { ... }
```

tzyl (approved):

PR #43, opened 2024-02-07
Fix intellisense for aggregations and fill in more number types

ericanderson:


packages/client/src/objectSet/ObjectSet.ts
```diff
@@ -69,7 +73,9 @@ export interface BaseObjectSet<
   //   OsdkObjectFrom<K, O, PropertyKeysFrom<O, K>>
   // >;
 
-  aggregateOrThrow: <const AO extends AggregateOpts<O, K, any>>(
+  aggregateOrThrow: <
+    AO extends AggregateOpts<O, K, AggregationClause<O, K>>,
```
ericanderson, line 77: This change made it make suggestions again

packages/client/src/query/aggregations/AggregationsClause.ts
```diff
@@ -38,7 +39,8 @@ export type AggregationClause<
     K
   >[P]["type"] extends "string"
     ? StringAggregateOption | StringAggregateOption[]
-    : ObjectTypePropertyDefinitionsFrom<O, K>[P]["type"] extends "double"
+    : ObjectTypePropertyDefinitionsFrom<O, K>[P]["type"] extends
+      "double" | "integer" | "float" | "decimal" | "byte" | "long" | "short"
```
ericanderson, line 43: Just make more types work that I suspect will work

packages/client/src/query/aggregations/AggregationsClause.ts
```diff
@@ -38,7 +39,8 @@ export type AggregationClause<
...
       ? NumericAggregateOption | NumericAggregateOption[]
-    : ObjectTypePropertyDefinitionsFrom<O, K>[P]["type"];
+    : never;
```
ericanderson, line 45: This old one was wrong because you cant have an aggregation of `{ geohash: "geopoint" }`. Switching to never to be more accurate.

mfedderly (approved):

PR #47, opened 2024-02-09
Increase CLI max width to 150 columns

zeyadkhaled:
## wrap set to min(150 columns, windowWidth)
image

### DIFF
--- a/packages/cli/changelog/@unreleased/pr-47.v2.yml
+++ b/packages/cli/changelog/@unreleased/pr-47.v2.yml
@@ -1,5 +1,5 @@
 type: improvement
 improvement:
-  description: Set yargs wrap property to null
+  description: Increase CLI max width to 150 columns
   links:
   - https://github.com/palantir/osdk-ts/pull/47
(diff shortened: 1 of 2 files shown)


packages/cli/src/cli.ts
```diff
@@ -28,6 +28,7 @@ import { YargsCheckError } from "./YargsCheckError.js";
 
 export async function cli(args: string[] = process.argv) {
   const base: Argv<CliCommonArgs> = yargs(hideBin(args))
+    .wrap(Math.min(150, process.stdout.columns))
```
tzyl, line 31: Should we use yargs built in method for this?

```suggestion
    .wrap(Math.min(150, yargs().terminalWidth()))
```

packages/cli/changelog/@unreleased/pr-47.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Set yargs wrap property to null
```
tzyl, line 3: Remember to update this:
```
Increase CLI max width to 150 columns
```

tzyl (approved):

PR #48, opened 2024-02-09
Prep for client(Foo) syntax

mfedderly:
Converts many of the types in `@osdk/client` to work on a `ObjectOrInterfaceDefinition` instead of on a pair of `Ontology, K extends keyof Ontology["objects"]`.

Updates v2 generation to include a `type` field on each definition.

There is still remaining work to convert actions.


packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
```diff
@@ -60,9 +60,10 @@ describe("OsdkObjectSet", () => {
 
   it("exposes descriptions", () => {
     const os = createBaseTodoObjectSet(client);
+    console.log(os);
```
mfedderly, line 63:
```suggestion
```

packages/legacy-client/src/client/OsdkLegacyObject.test.ts
```diff
@@ -22,9 +22,12 @@ import type { OsdkLegacyObjectFrom } from "./OsdkLegacyObject";
 describe("OsdkLegacyObject", () => {
   it("compiles", async () => {
     const x: Todo = {} as Todo;
-    const y: OsdkLegacyObjectFrom<typeof MockOntology, "Todo"> = x;
+    const y: OsdkLegacyObjectFrom<MockOntology, "Todo"> = x;
```
mfedderly, line 25: I don't think that this is actually an example of a breaking change, but better to flag it than miss it.

packages/api/src/index.ts
```diff
@@ -43,9 +45,11 @@ export type {
   ObjectTypeLinkDefinition,
   ObjectTypeLinkDefinitionFrom,
   ObjectTypeLinkKeysFrom,
+  ObjectTypeLinkKeysFrom2,
   ObjectTypeLinkTargetTypeFrom,
   ObjectTypePropertyDefinition,
   ObjectTypePropertyDefinitionFrom,
+  ObjectTypePropertyDefinitionFrom2,
```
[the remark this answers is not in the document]
ericanderson: Can't overwrite, that would be breaking changes.

packages/legacy-client/src/client/OsdkLegacyObject.test.ts
ericanderson: that mock is not generated, its old. this shouldnt be an issue for those using generated.

mfedderly:
👍

mfedderly:
LGTM but can't approve my own PR

PR #50, opened 2024-02-12
Harmonize usage of options `foundryUrl` & `clientId` across all cli commands

zeyadkhaled:
## Changes

Replaces the option "stack" in generate command and option "baseUrl" in login command with "foundryUrl" to be consistent with the site command.

Similarly, replace the option "applicationId" to "clientId"

Left both old options as aliases of the new option to maintain backwards compatibility. If that shouldn't be the case will remove the aliases.

## Open Questions

- Considered changing applicationId to clientId in the login command for consistency with the typescript command, but wasn't sure if application could be anything but the clientId.
- `outDir` reflects the output directory unlike in site command `directory` actually maps to input directory, so didn't think there is much use of changing that unless we want to have consistency of using dir vs directory and in/out?

### DIFF
--- a/packages/cli/changelog/@unreleased/pr-50.v2.yml
+++ b/packages/cli/changelog/@unreleased/pr-50.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Harmonize usage of optional `foundryUrl` across all cli commands
+  links:
+  - https://github.com/palantir/osdk-ts/pull/50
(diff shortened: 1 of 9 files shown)


tzyl (approved):
> Considered changing applicationId to clientId in the login command for consistency with the typescript command, but wasn't sure if application could be anything but the clientId.

I vote for going ahead and renaming this from applicationId -> clientId as this is the actual terminology used in the UI and is what is used in create-app

PR #51, opened 2024-02-12
Clean up yargs and consola imports

tzyl:
- Consistently name `argv` as the `yargs.Argv` builder type and `args` as the parsed `yargs.Arguments` type
- Make all consola imports static imports at the moment there's a big mix of both and consola is always going to be needed immediately in the CLI
- Add description for top level unstable command so that it actually shows up in the default generated help text
- Simplify custom fail logging verbose logic to just log stack in debug level
- Print CLI errors as consola error for consistency and more visibility at end of CLI

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -79,7 +79,6 @@ export async function cli(args: string[] = process.argv) {
   } catch (err) {
     if (err instanceof ExitProcessError) {
       consola.error(err);
-      consola.debug(err.stack);
     }
   }
 }
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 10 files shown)


packages/cli/src/cli.ts
```diff
@@ -85,8 +78,8 @@ export async function cli(args: string[] = process.argv) {
     return await base.parseAsync();
   } catch (err) {
     if (err instanceof ExitProcessError) {
-      const Consola = await import("consola");
-      Consola.consola.error(err);
+      consola.error(err);
+      consola.debug(err.stack);
```
zeyadkhaled, line 82: This is logging the stack as well so need for the second statement. Also the second statement won't be logged as consola level will not be set because this try/catch is for handling errors prior to parsing.
```suggestion
      consola.error(err);
```

zeyadkhaled (approved):
I like the simplification of logging the stack for `ExitProcessError` 👍

PR #52, opened 2024-02-12
Handle version/autoVersion selection in the parsing layer

zeyadkhaled:
The handling layer needed to know about the internals of the parsing layer where version and autoVersion could be defined at the same time due to how default is being used in the autoVersion yargs option. This introduces a new type just for the handler arguments with a unionType `selectedVersion` that helps remove any confusion about which key from the parsed args to use.


packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -117,7 +117,13 @@ const command: CommandModule<
   },
   handler: async (args) => {
     const command = await import("./siteDeployCommand.mjs");
-    await command.default(args);
+    const selectedVersion = args.version != null
+      ? args.version
+      : {
+        type: args.autoVersion as "git-describe",
```
tzyl, line 123: Can we similarly make this type more specific and/or use coerce similar to third party app rid?

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -117,7 +117,13 @@ const command: CommandModule<
...
+        tagPrefix: args.gitTagPrefix,
+      };
+    await command.default({ ...args, selectedVersion });
```
tzyl, line 126: May make debugging slightly easier down the line if no extra information is passed in:

```suggestion
    const { version, autoVersion, gitTagPrefix, ...restArgs } = args;
    await command.default({ ...restArgs, selectedVersion });
```

tzyl (approved):

PR #53, opened 2024-02-12
Adding bulk actions to 1.1

ssanjay1:
We added bulk actions to the top level client ontology due to difficulties wrangling with the types that would conditionally enforce options and return types based on what parameters were passed in (array vs single input)

### DIFF
--- a/packages/generator/src/v1.1/generateBulkActions.ts
+++ b/packages/generator/src/v1.1/generateBulkActions.ts
@@ -61,14 +61,14 @@ export async function generateBulkActions(
           `* @param {${typeScriptType}} params.${parameterName}`,
         );
       }
-      parameterBlock += "} ";
+      parameterBlock += "}[], ";
     }
 
     jsDocBlock.push(`*/`);
     actionSignatures.push(
       `
       ${jsDocBlock.join("\n")}
-      ${action.apiName}<O extends BulkActionExecutionOptions>(${parameterBlock}[],options?: O): 
+      ${action.apiName}<O extends BulkActionExecutionOptions>(${parameterBlock}options?: O): 
         Promise<Result<BulkActionResponseFromOptions<O, Edits<${
         addedObjects.length > 0
           ? addedObjects.join(" | ")
--- a/packages/generator/src/v1.1/generateBulkActions.ts
+++ b/packages/generator/src/v1.1/generateBulkActions.ts
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-import type { ActionParameterType } from "@osdk/gateway/types";
 import path from "node:path";
 import type { MinimalFs } from "../MinimalFs";
 import { getModifiedEntityTypes } from "../shared/getEditedEntities";
 import { formatTs } from "../util/test/formatTs";
 import type { WireOntologyDefinition } from "../WireOntologyDefinition";
+import { getTypeScriptTypeFromDataType } from "./generateActions";
 
 export async function generateBulkActions(
   ontology: WireOntologyDefinition,
@@ -97,43 +97,4 @@ export async function generateBulkActions(
     }
   `),
   );
-}
-
-function getTypeScriptTypeFromDataType(
-  actionParameter: ActionParameterType,
-  importedObjects: Set<string>,
-): string {
-  switch (actionParameter.type) {
-    case "objectSet": {
-      const objectType = actionParameter.objectTypeApiName!;
-      importedObjects.add(objectType);
-      return `ObjectSet<${objectType}>`;
-    }
-    case "object": {
-      const objectType = actionParameter.objectTypeApiName!;
-      importedObjects.add(objectType);
-      return `${objectType} | ${objectType}["__primaryKey"]`;
-    }
-    case "array":
-      return `Array<${
-        getTypeScriptTypeFromDataType(actionParameter.subType, importedObjects)
-      }>`;
-    case "string":
-      return `string`;
-    case "boolean":
-      return `boolean`;
-    case "attachment":
-      return `Attachment`;
-    case "date":
-      return `LocalDate`;
-    case "double":
-    case "integer":
-    case "long":
-      return `number`;
-    case "timestamp":
-      return `Timestamp`;
-    default:
-      const _: never = actionParameter;
-      throw new Error(`Unsupported action parameter type: ${actionParameter}`);
-  }
 }
--- a/packages/generator/src/v1.1/generateBulkActions.ts
+++ b/packages/generator/src/v1.1/generateBulkActions.ts
@@ -63,7 +63,7 @@ export async function generateBulkActions(
       }
       parameterBlock += "}[], ";
     } else {
-      parameterBlock = `params: {}[], `;
+      parameterBlock = `params: Record<string,never>[], `;
     }
 
     jsDocBlock.push(`*/`);
--- a/packages/generator/src/v1.1/generateBulkActions.ts
+++ b/packages/generator/src/v1.1/generateBulkActions.ts
@@ -62,6 +62,8 @@ export async function generateBulkActions(
         );
       }
       parameterBlock += "}[], ";
+    } else {
+      parameterBlock = `params: {}[], `;
     }
 
     jsDocBlock.push(`*/`);
--- a/packages/generator/src/v1.1/generateBulkActions.ts
+++ b/packages/generator/src/v1.1/generateBulkActions.ts
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2023 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ActionParameterType } from "@osdk/gateway/types";
+import path from "node:path";
+import type { MinimalFs } from "../MinimalFs";
+import { getModifiedEntityTypes } from "../shared/getEditedEntities";
+import { formatTs } from "../util/test/formatTs";
+import type { WireOntologyDefinition } from "../WireOntologyDefinition";
+
+export async function generateBulkActions(
+  ontology: WireOntologyDefinition,
+  fs: MinimalFs,
+  outDir: string,
+  importExt: string = "",
+) {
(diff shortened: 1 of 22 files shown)


packages/legacy-client/src/client/baseTypes/ActionType.ts
```diff
@@ -141,8 +163,8 @@ export const ActionResponse = {
       ],
     };
     if (response.edits?.type === "edits") {
-      const added = [];
-      const modified = [];
+      const added: any[] = [];
```
ssanjay1, line 166: Didn't touch this, but for some reason had to add this type, or else checks error out with something along the lines of :
`Argument of type '{ apiName: string; primaryKey: any; get: () => Promise<Result<OntologyObject<string, NonNullable<ParameterValue>>, GetObjectError>>; }' is not assignable to parameter of type 'never'.ts(2345)`

packages/legacy-client/src/client/ontology.ts
```diff
@@ -38,6 +41,10 @@ export class Ontology<O extends OntologyDefinition<any>> {
     return createActionProxy(this.#client);
   }
 
+  get bulkActions(): BulkActions<O> {
```
ssanjay1, line 44: I only added this here, and not in the actual OntologyDefinition since this seems like we're only adding this for 1.1

packages/legacy-client/src/client/actions/actions.ts
```diff
@@ -158,3 +182,15 @@ export type Actions<
         options?: Op,
       ) => WrappedActionReturnType<O, A, Op>;
 };
+
+export type BulkActions<O extends OntologyDefinition<any>> = {
+  [A in keyof O["actions"]]:
+    IsEmptyRecord<O["actions"][A]["parameters"]> extends true
+      ? <Op extends BulkActionExecutionOptions>(
+        options?: Op,
+      ) => WrappedBulkActionReturnType<O, A, Op>
```
ssanjay1, line 191: Right now I'm just following pattern of how we've handled other actions with no parameters, but the way its setup bulk actions don't really work for actions without parameters. This _could_ be modified so it accepts an array of `{}` potentially? And we just use those to tell us how many empty requests we need for the batch endpoint?

packages/legacy-client/src/client/baseTypes/ActionType.ts
```diff
@@ -30,6 +33,11 @@ export type ActionExecutionOptions = {
   returnEdits?: ReturnEditsMode;
 };
 
+export type BulkActionExecutionOptions = {
+  mode?: never;
```
ssanjay1, line 37: Don't want users trying to pass in any sort of execution mode as that isn't supported by the endpoint

ssanjay1 (commented):
Left some comments as well, also general question, where is the best place to expose some sort of docs to indicate that the bulk actions endpoint only supports up to 20 requests at a time?

packages/generator/src/v1.1/generateBulkActions.ts
```diff
@@ -0,0 +1,100 @@
...
+  outDir: string,
+  importExt: string = "",
+) {
+  const importedObjects = new Set<string>();
+  let actionSignatures: any[] = [];
+  for (const action of Object.values(ontology.actionTypes)) {
+    const entries = Object.entries(action.parameters);
+
+    const modifiedEntityTypes = getModifiedEntityTypes(action);
+    const addedObjects = Array.from(modifiedEntityTypes.addedObjects);
```
ericanderson, line 36: I think `.addedObjects`/`.modifiedObjects` is a `Set<string>`. So we are converting it to an array so we can `forEach` add it to a different set. At the very least you could just `forEach` on the returned sets instead of the memory copy.

packages/legacy-client/src/client/actions/actions.ts
ericanderson: Yeah, I think `{}[]` or `undefined[]` is better than nothing, otherwise you arent really calling bulk.

packages/legacy-client/src/client/actions/createActionProxy.ts
```diff
@@ -41,11 +48,18 @@ export function createActionProxy<
             };
           }
 
-          return async function<Op extends ActionExecutionOptions>(
+          return async function<
+            Op extends ActionExecutionOptions,
+          >(
```
ericanderson, line 53: nit: unnecessary change

packages/legacy-client/src/client/actions/createActionProxy.ts
```diff
@@ -41,11 +48,18 @@ export function createActionProxy<
...
             params: ActionArgs<O, typeof p>,
             options?: Op,
           ): Promise<WrappedActionReturnType<O, typeof p, Op>> {
-            return executeAction<O, typeof p, Op>(client, p, params, options);
+            return executeAction<O, typeof p, Op>(
```
ericanderson, line 57: nit: unnecessary change

packages/legacy-client/src/client/baseTypes/ActionType.ts
ericanderson: The problem with this though is it is suggested in intellisense even though it can only be undefined.

It can just be omitted right? https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBCBXANgawIIDGMAlhAHYCiIwBGJ5A8mPWQM5wC8cA3gFBxwAthAAmwAPwAuOGWAA3YFADc-OFGAwMUSiOIxWUmRkEAjRXF4BfXmIJYAhurgAzDGSKkyLsgAoIYaUxcQhYqGjpPJhZWAEppOQhiERVeZ18+AQEAeiy4YjIYYCwsYlZgNlRWAAsIAHd2AANhMQarGN5eUEhYBGRUIPwPcjDaFijPVgAmTh5VdU1tCl19QzJjMygLa1sHJ1d3Fm9JvwD0bEHQ6lHI5gnJuLgEpJU044zM5uBpAFYlOByZBB4FhgABzexYOBgRz2QR5ZxwRAQDBwABWGFY8Aggj0bSAA

packages/legacy-client/src/client/baseTypes/ActionType.ts
ericanderson: Hmm. I am a little worried this means we broke something somewhere. If we need to keep a change like this we could at least type it to the real shape we are pushing to prevent future errors.

packages/legacy-client/src/client/baseTypes/ActionType.ts
```diff
@@ -179,3 +201,52 @@ export const ActionResponse = {
...
+      string,
+      NonNullable<ParameterValue>
+    >,
+  >(
+    client: ClientContext<OntologyDefinition<any>>,
+    response: BatchApplyActionResponseV2,
+  ): BulkActionResponse<Edits<TAddedObjects, TModifiedObjects> | undefined> => {
+    if (response.edits?.type === "edits") {
+      const added: any[] = [];
+      const modified: any[] = [];
```
ericanderson, line 218: same here

packages/legacy-client/src/client/baseTypes/ActionType.ts
```diff
@@ -179,3 +201,52 @@ export const ActionResponse = {
...
+      for (const edit of response.edits.edits) {
```
ericanderson, line 219: This seems like a copy/paste of the above.  Maybe extract function for getting the `edits` shape (`{type:"edits", added,modified}`)?

packages/legacy-client/src/client/net/executeAction.ts
```diff
@@ -71,6 +77,44 @@ export async function executeAction<
...
+  return wrapResult(
+    async () => {
+      const response = await applyActionBatchV2(
+        createOpenApiRequest(client.stack, client.fetch),
+        client.ontology.metadata.ontologyApiName,
+        actionApiName as string,
+        {
+          requests: params
+            ? remapBulkActionParams<O, A>(params)
+            : [],
```
ericanderson, line 99: Given empty array is used here, does this mean that any actions that take zero arguments can never be called in bulk? If thats the intention, should we just exclude them in the type mapping for creating bulk actions?

styu (commented):
defer to @ericanderson on the meat of the changes

packages/legacy-client/src/client/ontology.ts
ericanderson: That would be correct, since there is no custom definition for actions vs bulk actions. Only the runtime `client.ontology.` should be showing `bulkActions`.

ericanderson (commented):
Some notes to fix up before merging.

packages/legacy-client/src/client/baseTypes/ActionType.ts
ssanjay1: Ah yea you're right, I think I needed this before when we tried the original syntax and I had some conditional types for the options, but yea I think I can just omit now, on it

packages/legacy-client/src/client/actions/actions.ts
ssanjay1: Would `undefined[]` be preferable here? Because with `{}[]`won't that just let you pass in any sort of object https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBCBXANgawIIDGMAlhAHYCiIwBGJ5A8mPWQM5wC8cA3gFBxwAthAAmwAPwAuOGQyCARsCgBufnCjAYGKJRHEYrKTLmKocXgF9eYglgCGGuADMMZIqTLOyACghhpmLiELFQ0dB5MLKwAlNIAbhDEIqq8Tj7c5gICAPTZcMRkMMBYWMSswGyorAAWEADu7AAGwmKNltG8vKCQsAjIqIH47uShtCyRHqwATJw8ahpaOhR6BkayCkrmVjb2ji5uLF5TADwACnCgRWQi7IPBHqPhjMyTUwB8vv5wp7FwCUmqNJTbx8LJCUTAaQAVlyMgg8CwwAA5nYsHAwA47IJ8k44IgIBg4AArDCseAQQT6dqdIHHO7DSjUMYRF7kVgfUFZFqQuAwvJkeFwREotEYqBYnF4gnE0nkykwaldcDQeBIFBweTYHDAQTMRCzTkFapKSHcCwAbQAuqorLwCGyEZoyVBEJJNbgdXqDUaTZJzdwAJIEOzXerVOxFOJKSQARgsABoeBZLRYgA

ssanjay1: But also feel like it could be weird synatx to do something like `client.ontology.bulkActions.noParamaction([undefined, undefined,...], {options})`?

ssanjay1: Oh wait could we use `Record<string,never>[]`?

packages/generator/src/v1.1/generateBulkActions.ts
ssanjay1: Ah yea I just pulled this from the `generateActions`, it looks like we converted to Array so we could just do a simple join on line 74 when generating the edits return type

ericanderson (approved):

PR #55, opened 2024-02-12
Address minor issues

zeyadkhaled:


packages/cli/src/commands/site/deploy/siteDeployCommand.mts
```diff
@@ -59,7 +59,7 @@ export default async function siteDeployCommand(
   if (version != null) {
     siteVersion = version;
   } else {
-    siteVersion = await findAutoVersion(gitTagPrefix);
+    siteVersion = findAutoVersion(gitTagPrefix);
```
zeyadkhaled, line 62: `autoVersion` was labeled async but didn't do anything async so removed that and here for accuracy.

packages/cli/src/util/config.ts
```diff
@@ -91,10 +96,17 @@ export async function loadFoundryConfig(): Promise<
   const configFilePath = await findUp(CONFIG_FILE_NAMES);
 
   if (configFilePath) {
+    const extension = extname(configFilePath);
+    if (!SUPPORTED_EXTENSIONS.includes(extension)) {
```
tzyl, line 100: Hmm, I'm not really sure this is useful given that the extension is fixed in the config file name. I think we'd instead in the future when adding extra supported config file names run a switch over the extension and have different parsers or a map from extension to parser rather than an array of supported extensions. The unmatched case would then be "unsupported" rather than a check like this that we have to keep the array in sync with the actual parsing branches.

packages/cli/src/util/config.ts
```diff
@@ -116,3 +128,13 @@ export async function loadFoundryConfig(): Promise<
 
   return undefined;
 }
+
+function parseConfigFile(
```
tzyl, line 132: Per the other comment I think it would be much better to keep these in sync by having a JSON parser, etc. and either a switch on the extension to call the parser or a map from extension to parser

packages/cli/src/commands/site/version/list/versionListCommand.mts
```diff
@@ -55,7 +55,10 @@ export default async function versionListCommand(
   }
 
   consola.success("Found versions:");
-  for (const version of versions) {
+
+  const semver = await import("semver");
+  const sortedVersions = semver.rsort(versions.filter(v => semver.valid(v)));
```
tzyl, line 60: We can probably assume the versions are enforced semver given the timelines?

packages/cli/src/util/autoVersion.ts
```diff
@@ -25,30 +25,72 @@ import { isValidSemver } from "./isValidSemver.js";
...
+export function autoVersion(tagPrefix: string = ""): string {
   const [matchPrefix, prefixRegex] = tagPrefix !== ""
     ? [tagPrefix, new RegExp(`^${tagPrefix}`)]
     : [undefined, new RegExp(`^v?`)];
+
+  const gitVersion = gitDescribe(matchPrefix);
+  const version = gitVersion.trim().replace(prefixRegex, "");
+  if (!isValidSemver(version)) {
+    throw new ExitProcessError(
+      2,
```
tzyl, line 37: From earlier comments we don't actually do anything with this `errorCode` information on the error and it's inconsistent on what it means where some of the net code uses it as a HTTP status code and here it seems like we're using it as an exit code. What should we do with it?

packages/cli/src/util/autoVersion.ts
```diff
@@ -25,30 +25,72 @@ import { isValidSemver } from "./isValidSemver.js";
...
+      }
+
+      if (
+        errorMessage.includes(
+          "fatal: no names found, cannot describe anything.",
+        )
+      ) {
+        throw new ExitProcessError(
+          2,
+          `Unable to determine the version using git-describe as no tags were found.\nPlease tag a version or supply a --version option.`,
```
tzyl, line 85:
```suggestion
          `Unable to determine the version using git-describe as no matching tags were found.\nPlease tag a matching version or supply a --version option.`,
```

packages/cli/src/util/config.ts
zeyadkhaled: Yeah that makes sense. I updated the implementation to use switch and run it of the file extension directly. I think its much better now

packages/cli/src/commands/site/version/list/versionListCommand.mts
zeyadkhaled: I think its better to leave it and I could clarify with a comment. I personally have few broken apps that have LOOSE versions stored

packages/cli/src/util/autoVersion.ts
zeyadkhaled: I actually think it doesn't add any value as well. So I could remove the passing of errorCode in a FLUP. Unless you foresee a usecase where we could benefit from it consistently?

packages/cli/src/commands/site/version/list/versionListCommand.mts
tzyl: Yeah I guess the point is that after a certain point in time in the future we're not going to have the control to filter this and it will be on the service layer to decide how to handle this instead

packages/cli/src/util/autoVersion.ts
```diff
@@ -29,26 +30,71 @@ export async function autoVersion(tagPrefix: string = ""): Promise<string> {
...
+        );
+      }
     }
 
-    return version;
-  } catch (error) {
     throw new ExitProcessError(
       2,
-      `Unable to determine the version automatically. Please supply a --version argument. ${error}`,
+      `Unable to determine the version automatically: ${error}.\nPlease supply a --version argument.`,
```
tzyl, line 95:
```suggestion
      `Unable to determine the version automatically: ${error}.\nPlease supply a --version option.`,
```

packages/cli/src/util/config.ts
```diff
@@ -94,11 +95,12 @@ export async function loadFoundryConfig(): Promise<
     let foundryConfig: FoundryConfig;
     try {
       const fileContent = await fsPromises.readFile(configFilePath, "utf-8");
-      foundryConfig = JSON.parse(fileContent);
-    } catch {
+      const extension = extname(configFilePath);
+      foundryConfig = parseConfigFile(fileContent, extension);
```
tzyl, line 99: Given we don't use the extension anywhere outside maybe simpler to just move it into `parseConfigFile`

packages/cli/src/util/autoVersion.ts
```diff
@@ -29,26 +30,71 @@ export async function autoVersion(tagPrefix: string = ""): Promise<string> {
...
       { encoding: "utf8" },
     );
-    const version = gitVersion.trim().replace(prefixRegex, "");
-    if (!isValidSemver(version)) {
-      throw new ExitProcessError(
-        2,
-        `The version string ${version} is not SemVer compliant.`,
-      );
+    gitVersion = stdout;
+  } catch (error: any) {
```
tzyl, line 57: Hmm, have you tested for this new execAsync change that it throws? If it does not throw on non-zero exit code is the error message on stderr instead?

tzyl (approved):

packages/cli/src/util/autoVersion.ts
zeyadkhaled: Tested and throws 👍

PR #57, opened 2024-02-12
Display site link after deploy

tzyl:
Fetch registered domains (only one is exposed at the moment) and display link for live site or for preview version depending on `--uploadOnly`.

Note that the preview link sets a cookie and redirects to the root page so going back to the live link later will still show that preview version until clicking exit in the banner


zeyadkhaled (approved):
Looks really good, my only nit would be to refactor the consola.box repeated bit into a function that takes a title and body as arguments.

PR #59, opened 2024-02-13
Add $intersects for where clause

ssanjay1:


packages/client/src/query/WhereClause.ts
```diff
@@ -88,11 +89,23 @@ export type GeoFilter_Within = {
...
+    }
+    | Polygon;
+};
+
+export type GeoFilter = GeoFilter_Within | GeoFilter_Intersects;
 
 type FilterFor<PD extends ObjectTypePropertyDefinition> = PD["type"] extends
   "string" ? StringFilter
-  : PD["type"] extends "geopoint" ? GeoFilter
+  : PD["type"] extends "geopoint" | "geoshape" ? GeoFilter
```
ssanjay1, line 108: Where can we double check what filters each of these wire property types can actually accept? I'm not sure that a geopoint type can do an intersects filter? Atleast I remember in the 1.1 version that geopoint types were only allowed to do $within filters

ericanderson: unsure but this is worth getting right

ericanderson (approved):

packages/client/src/query/WhereClause.ts
ssanjay1: Looks like geopoint and geoshape do support that https://github.palantir.build/foundry/api-gateway/blob/a6d1c2ffb544592af830556d31a23fd4ca0f5a1e/adapters/ontology-adapters/src/main/java/com/palantir/foundry/api/adapters/ontology/filters/PropertyTypeValidatorVisitor.java#L380

PR #66, opened 2024-02-14
Update copyright to use new Date()

styu:
Since the years are now different, I updated the rule to use a pattern instead so we don't have to change the 2023 files


mfedderly (approved):
(or grab the current year using `Date`)

PR #68, opened 2024-02-15
Polish casing, option description, and YargsCheckErrors across CLIs and commands

zeyadkhaled:
- Update the create-app CLI to use camelCase for the options
- Make create-app CLI width consistent with other CLI
   - image
  
- Update several option description to be consistent across CLIs
- Update errors thrown inside yargs check
   - (After/Before) image
 - Throw YargsCheckError if version isn't valid strict semVer 
     - image

### DIFF
--- a/packages/cli/src/commands/site/deploy/index.ts
+++ b/packages/cli/src/commands/site/deploy/index.ts
@@ -101,7 +101,7 @@ const command: CommandModule<
 
         if (args.version != null && !isValidSemver(args.version)) {
           throw new YargsCheckError(
-          `--version "${args.version}" must be a valid SemVer string`,
+            `--version "${args.version}" must be a valid SemVer string`,
           );
         }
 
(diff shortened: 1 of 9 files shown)


tzyl (approved):

packages/cli/src/commands/site/deploy/index.ts
```diff
@@ -98,6 +99,12 @@ const command: CommandModule<
           );
         }
 
+        if (args.version != null && !isValidSemver(args.version)) {
+          throw new YargsCheckError(
+            "--version must be a valid SemVer string",
```
tzyl, line 104:
```suggestion
            `Version "${args.version}" must be a valid SemVer string",
```

zeyadkhaled: I feel leaving the `--version` could be helpful to clearly show this is coming directly from the user passing `--version` and not possibly due to autoVersion failing in some form?

```suggestion
          `--version "${args.version}" must be a valid SemVer string`,
```

tzyl: Looks like you made most of these existing error messages reference the option name so fine to keep it like that!

PR #69, opened 2024-02-15
Resurface `Version already exists` message

zeyadkhaled:
Due to the logic introduced [here](https://github.com/palantir/osdk-ts/pull/61) where fetch errors are being directly thrown, the logic of checking if the upload failed due to clashing versions wasn't reachable.

So we explicitly overwrite the error message if its of the specific Conjure error type we expect.

image

### DIFF
--- a/packages/cli/src/net/createFetch.mts
+++ b/packages/cli/src/net/createFetch.mts
@@ -61,11 +61,14 @@ function handleFetchError(e: unknown): Promise<Response> {
     tip = "Check your token is valid and has not expired or been disabled";
   } else if (e.statusCode === 403) {
     tip = "Check your token has the required scopes for this operation";
-  } else if (e.errorName === "Artifacts:ArtifactAlreadyExists") {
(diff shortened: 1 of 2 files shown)


packages/cli/src/net/createFetch.mts
```diff
@@ -61,6 +61,8 @@ function handleFetchError(e: unknown): Promise<Response> {
     tip = "Check your token is valid and has not expired or been disabled";
   } else if (e.statusCode === 403) {
     tip = "Check your token has the required scopes for this operation";
+  } else if (e.errorName === "Artifacts:ArtifactAlreadyExists") {
+    e.message = "Version already exists";
```
tzyl, line 65: Would prefer to not mutate the error (even though it's not passed on) and instead:

```
let message = e.message;
if (..) {
  ...
}
```

this means that we also keep the two branches for status code and error name separate so if it happens that they overlap they still both run on the message and tip

tzyl (approved):

PR #70, opened 2024-02-15
Log OSDK CLI version on start of CLI

zeyadkhaled:
Logging OSDK CLI version doesn't depend on verbosity level, so if we want it to be visible even when running `--help` we can move logging the version to top of the CLI.

image

Doesn't affect already existing behavior of when errors occur

image

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -29,6 +29,10 @@ import { logVersionMiddleware } from "./yargs/logVersionMiddleware.js";
 import { YargsCheckError } from "./YargsCheckError.js";
 
 export async function cli(args: string[] = process.argv) {
+  consola.info(
+    `Palantir OSDK CLI ${process.env.PACKAGE_VERSION}`,
+  );
+
   const base: Argv<CliCommonArgs> = yargs(hideBin(args))
     .wrap(Math.min(150, yargs().terminalWidth()))
     .env("OSDK")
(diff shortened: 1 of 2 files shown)


packages/cli/src/yargs/logVersionMiddleware.ts
```diff
@@ -21,10 +21,6 @@ let firstTime = true;
 export async function logVersionMiddleware(args: CliCommonArgs) {
```
tzyl, line 21: Can we rename this to `logLevelMiddleware` to be more specific now?

packages/cli/src/cli.ts
```diff
@@ -29,6 +29,10 @@ import { logVersionMiddleware } from "./yargs/logVersionMiddleware.js";
 import { YargsCheckError } from "./YargsCheckError.js";
 
 export async function cli(args: string[] = process.argv) {
+  consola.info(
+    `Palantir OSDK CLI ${process.env.PACKAGE_VERSION}`,
```
tzyl, line 33: Can we try to maintain the spacing to be consistent as well when you use `--help` or not? For example just add an extra new line
```suggestion
    `Palantir OSDK CLI ${process.env.PACKAGE_VERSION}\n`,
```

tzyl (approved):

packages/cli/src/cli.ts
zeyadkhaled: I don't think we can make the space consistent. In all cases other than `--help` we get an extra line from `logVersionMiddleware`. If we add a new line like above we will in all cases (except for `--help`) get the following:

(non verbose)
```
ℹ Palantir OSDK CLI 0.5.0              
.
.
other logs    
```                                                                                                                                                           

(verbose)
```
ℹ Palantir OSDK CLI 0.5.0              
.
⚙ Verbose logging enabled (debug)                                                                                                                                                                     
.
other logs    
```

So I would say in its current shape its consistent for almost all cases except for the `--help`.

PR #71, opened 2024-02-15
Check client env variables at runtime

tzyl:
Currently app renders fine but difficult to debug when client has not been initialized correctly. This will fail faster where the app will fail to render completely showing something is wrong with the console error showing the missing env variable.

### DIFF
--- a/packages/create-app/src/__snapshots__/expected-template-next-static-export/src/lib/client.ts
+++ b/packages/create-app/src/__snapshots__/expected-template-next-static-export/src/lib/client.ts
@@ -1,14 +1,30 @@
 import { FoundryClient, PublicClientAuth } from "@fake/sdk";
 
+const url = process.env.NEXT_PUBLIC_FOUNDRY_API_URL;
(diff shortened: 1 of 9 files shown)


packages/create-app/src/__snapshots__/expected-template-next-static-export/src/lib/client.ts
```diff
@@ -1,14 +1,30 @@
 import { FoundryClient, PublicClientAuth } from "@fake/sdk";
 
+const url = process.env.NEXT_PUBLIC_FOUNDRY_API_URL;
+const clientId = process.env.NEXT_PUBLIC_FOUNDRY_CLIENT_ID;
+const redirectUrl = process.env.NEXT_PUBLIC_FOUNDRY_REDIRECT_URL;
```
tzyl, line 5: Not doing a dynamic lookup based on the key as it may interfere with the env variable being bundled during build

https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

zeyadkhaled (approved):

PR #74, opened 2024-02-15
Dont fetch rid by default

ericanderson:
Per prior conversation with objects team.


packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -153,6 +155,38 @@ export function convertWireToOsdkObjects<
...
+    obj.$apiName = obj.__apiName;
+
+    // for now these are the same but when we start doing interface projections the $objectType will always be underlying and
+    // the $apiName will be for the current view (in current designs)
+    obj.$objectType = obj.__apiName;
+
+    // copying over for now as its always returned. In the future, this should just be inferred from underlying
+    obj.$primaryKey = obj.__primaryKey;
+
+    // After Feb 2024 (unless we have new apis):
```
ssanjay1, line 173: So what happens after this month, do we just set that const to true and we delete those fields?

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -153,6 +155,38 @@ export function convertWireToOsdkObjects<
...
+      // Hide these from things like `console.log` so that people
+      // don't think to use them.
+      Object.defineProperties(obj, {
+        "__apiName": {
+          enumerable: false,
+        },
+        "__primaryKey": {
+          enumerable: false,
+        },
+      });
```
mfedderly, line 187: So its a little weird that we're afraid about conflicts here given that the backend is sending us these over the wire, they would've had to avoid this conflict ahead of time.

Additionally, I think there's a bunch more references to __apiName and __primaryKey in client that need to get worked through. Check out `OsdkObject.ts` and `OsdkObjectFrom.ts` for the root of that typing

packages/client/src/object/convertWireToOsdkObjects.ts
ericanderson: Yeah, just trying to give some time for people to deal with breaks.

packages/client/src/object/convertWireToOsdkObjects.ts
ericanderson: Re conflicts, I am looking at consistency. In 2.0, the things that are "OSDK" on the objects have `$` names and the things that dont, dont. This is more about consistency than conflicts. Correctness is a benefit.

packages/client/src/OsdkObjectFrom.ts
```diff
@@ -32,44 +32,52 @@ type OsdkCommonFrom<
...
-     * as always optional to avoid breaking changes.
-     */
-    __rid?: string;
-  };
+    $primaryKey: Q extends ObjectTypeDefinition<any>
+      ? OsdkObjectPrimaryKeyType<Q>
+      : unknown;
+    // $uniqueId: string; // will be dynamic
+  }
+  & (R extends true ? { $rid: string } : {});
```
mfedderly, line 58: can you use `{}` safely here? or do you want `Record<string, never>`?

mfedderly (approved):

packages/client/src/OsdkObjectFrom.ts
ericanderson: It doesn't matter in this case because after they are joined it has the other types.

PR #78, opened 2024-02-16
Add orderBy syntax to fetchPageOrThrow

mfedderly:
Example syntax:

```typescript
    const { data: employees } = await client.objects.Employee.fetchPageOrThrow({
      orderBy: { "employeeId": "asc" },
    });
```

orderBy accepts any property apiNames as keys, even if there's a select list which does not include them


packages/client/src/object/fetchPageOrThrow.ts
```diff
@@ -28,16 +28,28 @@ import type { PageResult } from "../PageResult.js";
 import { convertWireToOsdkObjects } from "./convertWireToOsdkObjects.js";
 
 export interface SelectArg<
-  O extends ObjectOrInterfaceDefinition<any, any>,
-  L = ObjectOrInterfacePropertyKeysFrom2<O>,
+  Q extends ObjectOrInterfaceDefinition<any, any>,
```
ssanjay1, line 31: Why the rename from O to Q?

packages/client/src/object/fetchPageOrThrow.ts
```diff
@@ -28,16 +28,28 @@ import type { PageResult } from "../PageResult.js";
...
 > {
   select?: readonly L[];
 }
 
+export interface OrderByArg<
+  Q extends ObjectOrInterfaceDefinition<any, any>,
+  L = ObjectOrInterfacePropertyKeysFrom2<Q>,
+> {
+  orderBy?: {
+    [K in L & string]?: "asc" | "desc";
```
ssanjay1, line 42: Why do we need to & string here?

ssanjay1 (approved):

packages/client/src/object/fetchPageOrThrow.ts
mfedderly: Ah yeah the L typing is a little sloppy, can fix

PR #79, opened 2024-02-20
Minor aggregation syntax changes

ssanjay1:
Rename group to $group and remove values parameter from agg results

### DIFF
--- a/packages/client/src/object/aggregateOrThrow.test.ts
+++ b/packages/client/src/object/aggregateOrThrow.test.ts
@@ -173,8 +173,8 @@ describe("aggregateOrThrow", () => {
       },
     );
     expectType<Array<any>>(grouped);
-    expectType<string | undefined>(grouped[0].$group?.text);
-    expectType<number>(grouped[0].id?.approximateDistinct);
+    expectType<string | undefined>(grouped[0].$group.text);
+    expectType<number>(grouped[0].id.approximateDistinct);
   });
 
   it("works with where: todo", async () => {
--- a/packages/client/src/object/aggregateOrThrow.test.ts
+++ b/packages/client/src/object/aggregateOrThrow.test.ts
@@ -173,7 +173,7 @@ describe("aggregateOrThrow", () => {
       },
     );
     expectType<Array<any>>(grouped);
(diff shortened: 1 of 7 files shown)


packages/client/src/object/aggregateOrThrow.test.ts
```diff
@@ -165,16 +165,16 @@ describe("aggregateOrThrow", () => {
...
         },
         groupBy: {
           text: "exact",
         },
       },
     );
     expectType<Array<any>>(grouped);
-    expectType<string | undefined>(grouped[0].group.text);
-    expectType<number>(grouped[0].values.text.approximateDistinct);
+    expectType<string | undefined>(grouped[0].$group?.text);
```
ssanjay1, line 176: I had to optional chain here to get the tests to pass, but I have no idea why my changes did that

ssanjay1: For context ,the error tests gave was `cannot read properties of undefined` on group and id, which is why I optional chained it

ssanjay1: Actually, was able to fix this

packages/client/src/object/aggregateOrThrow.ts
```diff
@@ -85,8 +85,8 @@ export async function aggregateOrThrow<
   const ret: AggregationResultsWithGroups<Q, AO["select"], any> = result.data
     .map((entry) => {
       return {
-        group: entry.group as any,
-        values: legacyToModernSingleAggregationResult(entry),
+        $group: entry.group as any,
+        ...legacyToModernSingleAggregationResult(entry),
       };
     }) as any; // fixme
```
ssanjay1, line 91: Is there a good way to fix this, besides doing something like `as AggregationResultsWithGroups<Q, AO["select"], any>`?

mfedderly (approved):

packages/client/src/object/aggregateOrThrow.ts
ssanjay1: Ok I won't block merge on it, will think about it as I do other agg stuff potentially

PR #67, opened 2024-02-14
Finish future types

ericanderson:

PR #83, opened 2024-02-20
Add new client invocation syntax

ericanderson:
Blocks on https://github.com/palantir/osdk-ts/pull/67


ericanderson:
Admin merging since policy-bot has confused itsef and thinks this needs to pass some rules for another branch. Everything else is satisfied.

PR #84, opened 2024-02-20
Add count to aggregations

ssanjay1:
Syntax:
`client.objects.testObject.aggregateOrThrow({select:{$count:true}})`


examples/basic/cli/src/index.ts
```diff
@@ -173,6 +173,32 @@ async function runTests() {
...
+    // Should be 51 because it includes DC
+    console.log(
+      testAggregateCountNoGroup.$count,
+      testAggregateCountNoGroup.latitude.avg,
+      testAggregateCountNoGroup.latitude.max,
+      testAggregateCountNoGroup.latitude.min,
+    );
+    const testAggregateCountWithGroups = await client.objects.BoundariesUsState
+      .aggregateOrThrow({
+        select: { $count: true, latitude: ["min", "max", "avg"] },
```
ssanjay1, line 190: Ok so the types are working, but there's one weird thing happening. When I do something like:
`const testAggregateCountWithGroups = await client.objects.BoundariesUsState
      .aggregateOrThrow({
        select: { $count: true, misNameLatitiude: ["min", "max", "avg"] },
        groupBy: { usState: "exact" },
      });` the type checks let it through and it doesn't red line(the results are still fine though and don't pass through misNameLatitude)
     ```
 But if I do something like:
      const testAggregateCountWithGroups = await client.objects.BoundariesUsState
      .aggregateOrThrow({
        select: { $count: true, misNameLatitiude: ["min", "max", "avg"] },
        groupBy: { misNameUsState: "exact" },
      }); 
```
_then_ it recognizes that both the select and groupBy have a misnamed property. The groupby also correctly typechecks on its own.
**TLDR: the select statement only complains about misnamed or nonexistent param keys in its type if theres also a type issue with one of the other agg clauses like groupBy **

File: packages/client/src/query/aggregations/AggregateOpts.ts [hunk 2]
```diff
@@ -16,16 +16,17 @@
...
   AggregationClause,
   GroupByClause,
   WhereClause,
 } from "../../query/index.js";
 
 export type AggregateOpts<
-  Q extends ObjectOrInterfaceDefinition<any, any>,
-  AC extends AggregationClause<Q>,
+  Q extends ObjectOrInterfaceDefinition,
+  KK extends AggregatableKeys<Q> = AggregatableKeys<Q>,
```
ssanjay1, line 27: I don't really think I need this with the way I'm using it right now, I can probably just have AggregationClause use `AggregatableKeys<Q> `directly in its type definition, since I'm just piping that through anyways

ericanderson:
@ssanjay1 after you get this merged ill merge my type fixes

packages/client/src/object/aggregateOrThrow.test.ts
```diff
@@ -226,22 +230,22 @@ describe("aggregateOrThrow", () => {
...
     > = {
       select: {
         locationCity: "approximateDistinct",
+        text: "approximateDistinct",
       },
-    } as any;
+    };
 
-    expectType<"approximateDistinct">(f.select.locationCity);
+    // expectType<"approximateDistinct">(f.select.locationCity);
```
mfedderly, line 249: Flagging whether this needs to get re-enabled now (or after the next PR merges perhaps?)

mfedderly (approved):

packages/client/src/query/aggregations/AggregationsClause.ts
```diff
@@ -26,13 +26,18 @@ type NumericAggregateOption =
...
-    : never;
-};
+  Q extends ObjectOrInterfaceDefinition,
+  K extends AggregatableKeys<Q> = AggregatableKeys<Q>,
+> =
+  & {
+    [P in K]?: Q["properties"][P]["type"] extends "string"
+      ? StringAggregateOption | StringAggregateOption[]
+      : Q["properties"][P]["type"] extends
+        "double" | "integer" | "float" | "decimal" | "byte" | "long" | "short"
```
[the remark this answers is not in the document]
ssanjay1: Ah ok, I can FLUP to address that

packages/client/src/object/aggregateOrThrow.test.ts
ssanjay1: Yea I think this would be after next PR

PR #86, opened 2024-02-20
improve missing rid inference

ericanderson:
In case where its omitted or false, we use the short version of the type that relies on the default value for include rid (false)

Before:
* image
* image
* image
* image

After:
* image
* image
* image
* image


mfedderly (approved):
But you gotta fix `DefaultToFalse` first

PR #88, opened 2024-02-21
Add ObjectSet .get(pk) syntax for 2.0

mfedderly:
Add the ability to use `.get(primaryKey)` on the base object sets and also immediately after a pivotTo.

This also sets us up to replace the MultiLink with a regular BaseObjectSet and get more functionality with less code.


packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -143,3 +157,48 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
...
+          withPk,
+        );
+
+        if (data.length !== 1 || nextPageToken != null) {
+          throw new Error("Expected exactly one result");
+        }
+
+        return data[0];
+      }
+      : undefined as never) as BaseObjectSet<Q>["get"],
```
ssanjay1, line 202: So doing the `as never` will make it so that we can't use `get` on something that's not an objectTypeDef right?

packages/client/src/objectSet/ObjectSet.ts
```diff
@@ -97,11 +101,22 @@ export interface BaseObjectSet<Q extends ObjectOrInterfaceDefinition> {
     ...objectSets: ReadonlyArray<ObjectSet<Q>>
   ) => ObjectSet<Q>;
 
-  pivotTo: <L extends LinkNames<Q>>(type: L) => ObjectSet<LinkedType<Q, L>>;
+  pivotTo: <L extends LinkNames<Q>>(type: L) => BaseObjectSet<LinkedType<Q, L>>;
 
   subscribe: (listener: ObjectSetListener<Q>) => () => void;
 }
 
+export interface BaseObjectSet<
```
ssanjay1, line 106: Ah so you made a separate BaseObjectSet type so that we only allowed get calls when we came off like the base or a pivot, and not wheres or anything?

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -143,3 +157,48 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
...
+  objectSet: WireObjectSet = {
+    type: "base",
+    objectType: objectType["apiName"] as string,
+  },
+): BaseObjectSet<Q> {
+  return {
+    ...createObjectSet(objectType, clientCtx, objectSet),
+
+    get: (isObjectTypeDefinition(objectType)
+      ? async <A extends SelectArg<any>>(
```
ssanjay1, line 175: Why can't you give` SelectArg<Q>`?

ssanjay1: Or maybe why don't you need to give it Q

ssanjay1 (approved):
Approved, left some comments for my own understanding again

packages/client/src/objectSet/createObjectSet.ts
mfedderly: Yeah `Q` is probably more correct. It gets shadowed by the `BaseObjectSet` definition.

packages/client/src/objectSet/createObjectSet.ts
mfedderly: Yeah for interfaces you can't get by a PK (although you could consider it with the uniqueId instead). Until that gets implemented you just won't get this method for interfaces.

The `as never` is actually from an earlier attempt before I cast it to `BaseObjectSet<Q>['get']` so I can just drop it for now.

packages/client/src/objectSet/ObjectSet.ts
mfedderly: Correct. This is the same approach that legacy-client took. Next PR will give it to us at `object.$link.multiLinkType` as well.

PR #91, opened 2024-02-22
Convert multi-links to a full BaseObjectSet

mfedderly:
`MultiLinkAccessor` was really just a second more limited implementation of the BaseObjectSet now that it got a `.get(pk)`.

This drops out the duplicate link loading code and uses searchAround ObjectSets to support links. As a bonus we get to drop some of the link-only gateway endpoint usage.


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -291,7 +293,7 @@ export class ObjectSetListenerWebsocket<
               this.#client,
               this.#metadataContext,
               objects,
-            ),
+            ) as Array<Osdk<Q>>,
```
mfedderly, line 296: This was hitting the typescript infinite recursion error

packages/client/src/definitions/LinkDefinitions.ts
```diff
@@ -39,7 +35,7 @@ export type OsdkObjectLinksEntry<
   O extends ObjectTypeDefinition<any>,
   L extends ObjectTypeLinkKeysFrom2<O>,
 > = O["links"][L] extends ObjectTypeLinkDefinition<infer T, infer M> ? (
-    M extends false ? SingleLinkAccessor<T> : MultiLinkAccessor<T>
+    M extends false ? SingleLinkAccessor<T> : BaseObjectSet<T>
```
ericanderson, line 38: Should this just be `ObjectSet<T>`?

ericanderson (approved):
Pre-emptive approve on the question.

packages/client/src/definitions/LinkDefinitions.ts
mfedderly: BaseObjectSet also has the .get(pk) method.

PR #93, opened 2024-02-23
Add docs

ssanjay1:


examples/docs_example/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "private": true,
+  "version": "0.0.1",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/palantir/osdk-ts.git"
+  },
+  "scripts": {
+    "build": "tsc && vite build",
+    "codegen": "rm -rf src/generatedNoCheck/* src/generatedNoCheck2/* && osdk unstable typescript generate --outDir src/generatedNoCheck2 --beta --ontologyPath ontology.json --version dev",
```
mfedderly, line 12: You can probably just collapse this into src/generatedNoCheck (with --beta still)

examples/docs_example/src/osdkExample.tsx
```diff
@@ -0,0 +1,155 @@
...
+  // Within clauses take either a polygon or bounding box
+  const withinFilteredObjectSet = await client(Office).where({
+    $or: [
+      {
+        entrance: { $within: { distance: [1_000, "miles"], of: [0, 0] } },
+      },
+      { entrance: { $within: { bbox: [0, 1, 2, 3] } } },
+      { entrance: { $within: [0, 1, 2, 3] } },
+      {
+        entrance: { $within: { polygon: [[[0, 1], [0, 0], [1, 1], [1, 0]]] } },
```
mfedderly, line 53: Hrm is this the actual syntax? there should be a 5th entry here for `[0,1]` again.
https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6

(Same for the rest of the file everywhere)

examples/docs_example/src/osdkExample.tsx
```diff
@@ -0,0 +1,155 @@
...
+  // Simple where clause
+  // This defaults to AND together all the parameters
+  const simpleFilteredEmployeeObjectSet = await client(Employee).where({
+    fullName: { $contains: "Clooney" },
+    employeeId: { $eq: 12 },
+  });
+
+  // Where clause with ors, ands, nots
+  const complexFilteredEmployeeObjectSet = await client(Employee).where({
+    $or: [{ fullName: { $contains: "Clooney" }, startDate: { $gt: 10 } }, {
```
mfedderly, line 32: Shouldn't `startDate` be a date?

examples/docs_example/src/osdkExample.tsx
```diff
@@ -0,0 +1,155 @@
...
+    simpleFilteredEmployeeObjectSet,
+  );
+
+  // objectSet subtract
+  const subtractObjectSet = complexFilteredEmployeeObjectSet.subtract(
+    simpleFilteredEmployeeObjectSet,
+  );
+
+  // Pivots and links
+  const getPeeps = await client(Employee).pivotTo("peeps").fetchPageOrThrow({});
```
mfedderly, line 103:
```suggestion
  const getPeeps = await client(Employee).pivotTo("peeps").fetchPageOrThrow();
```

examples/docs_example/src/osdkExample.tsx
```diff
@@ -0,0 +1,155 @@
...
+  console.log(getPeeps.data);
```
mfedderly, line 104: I'm lukewarm on the `console.log`s in this file.

examples/docs_example/src/osdkExample.tsx
mfedderly: There's a few more of these around as well, I'm not sure why people are providing an empty object there when it isn't required?

mfedderly (approved):

examples/docs_example/src/osdkExample.tsx
ssanjay1: Ah wait you're right good flag, yea the points I gave would be invalid GeoJSON since I didn't close the loop

PR #96, opened 2024-02-26
Improve aggregation selection verification

ericanderson:
Mismatch IDE:
image

Hover over `mismatchedLatitude`:
image

Hover over `testAggregateCountWithGroups`:
image

### DIFF
--- a/packages/client/src/object/aggregateOrThrow.ts
+++ b/packages/client/src/object/aggregateOrThrow.ts
@@ -37,6 +37,23 @@ import type {
 } from "../query/index.js";
 import type { ArrayElement } from "../util/ArrayElement.js";
 
+export type AggregateOptsThatErrors<
+  Q extends ObjectOrInterfaceDefinition,
+  AO extends AggregateOpts<Q>,
+> =
+  & AO
+  & {
+    select:
+      & Pick<
+        AO["select"],
+        keyof AggregateOpts<Q>["select"] & keyof AO["select"]
+      >
+      & Record<
+        Exclude<keyof AO["select"], keyof AggregateOpts<Q>["select"]>,
+        never
+      >;
+  };
+
 export async function aggregateOrThrow<
   Q extends ObjectOrInterfaceDefinition,
   AO extends AggregateOpts<Q>,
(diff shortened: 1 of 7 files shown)


packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -63,20 +63,30 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
...
+      },
     ): Promise<AggregationsResults<Q, AO>> => {
       return aggregateOrThrow<Q, AO>(
         clientCtx,
         objectType,
         objectSet,
         req,
       );
-    },
+    }) as any,
```
ssanjay1, line 89: Why do we need the as any here?

packages/client/src/object/aggregateOrThrow.ts
```diff
@@ -111,3 +121,5 @@ function aggregationToCountResult(
     }
   }
 }
+
```
ssanjay1, line 124: Was this intentional?

ssanjay1 (approved):

packages/client/src/objectSet/createObjectSet.ts
ericanderson: I can try to take it out but the types got so complex here that it was erroring that two identical types didnt match.

ssanjay1: Ah probably fine to just keep it then, things work so yay

packages/client/src/query/aggregations/AggregationsResults.ts
```diff
@@ -25,12 +26,17 @@ import type {
...
   : Exclude<AO["groupBy"], undefined> extends never // groupBy is explicitly undefined
     ?
       & AggregationResultsWithoutGroups<Q, AO["select"]>
       & AggregationCountResult<Q, AO["select"]>
-  : AggregationResultsWithGroups<Q, AO["select"], AO["groupBy"]>;
+  : AggregationResultsWithGroups<Q, AO["select"], AO["groupBy"]>
+  : `Sorry, the following are not valid selectors for an aggregation: ${Exclude<
+    keyof AO["select"] & string,
+    AggregatableKeys<Q> | "$count"
+  >}`;
```
mfedderly, line 42: I kinda dig this way of surfacing more useful errors.

mfedderly (approved):

PR #101, opened 2024-02-28
Initial Interfaces Support

ericanderson:
This is one of many chained PRs.


packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,14 +115,53 @@ const protoConverterCache = createCache(
...
+  interfaceApiName: string,
 ) {
-  // Fix properties
+  fixObjectPropertiesInline(objects);
+
+  const localObjectCache = createAsyncCache((client, apiName: string) =>
+    client.ontology.provider.getObjectDefinition(apiName)
+  );
+
+  const uniqueApiNames = new Set<string>(objects.map(o => o.$apiName));
```
mfedderly, line 133: `objects` may be large enough that `for (const {$apiName} of objects) { uniqueApiNames.add($apiName); }` is notably faster by avoiding the .map step. you'd have to jsbench it to be sure.

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,14 +115,53 @@ const protoConverterCache = createCache(
...
+  const [interfaceType] = await Promise.all([
+    client.ontology.provider.getInterfaceDefinition(interfaceApiName),
+    ...Array.from(uniqueApiNames).map(n => localObjectCache.get(client, n)),
```
mfedderly, line 136: Do you want to do `Array.from(uniqueApiNames).map(n=>localObjectsCache.get(client, n))` above this await without blocking on it? That would let us proceed with conversions until we run into one of these that hasn't loaded.

Its probably not worth bucketing on apiName because you'd lose ordering.

packages/client/src/object/fetchPage.ts
```diff
@@ -83,6 +112,68 @@ export type FetchPageResult<
...
+
+  if (objectSet.type === "filter") {
+    return objectSetToSearchJsonV2(
+      objectSet.objectSet,
+      expectedApiName,
+      objectSet.where,
+    );
+  }
+
+  throw new Error(`Unsupported objectSet type: ${objectSet.type}`);
```
mfedderly, line 138: Does this work if you have like `client(iface).where().where().fetchPage()`? Not sure if the MinimalObjectSet removes the ability to make something more complicated.

packages/generator/src/shared/UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst.ts
```diff
@@ -14,49 +14,74 @@
...
+  // if we want links to work
+  const imports: string[] = [];
+
+  return `${imports.join("\n")}
+
+    ${v2 ? getV2Types() : ""}
+
+    export const ${definition.apiName}: ${objectDefIdentifier} = {
+      ${stringify(definition)}
+    };`;
```
mfedderly, line 86: I think we technically don't even need any of this if they're v1

packages/api/src/ontology/ObjectTypeDefinition.ts
```diff
@@ -65,6 +65,8 @@ export interface ObjectTypeDefinition<
   type: "object";
   primaryKeyApiName: keyof this["properties"];
   primaryKeyType: keyof WirePropertyTypes;
+
+  spts?: Record<string, any>;
```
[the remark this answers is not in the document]
ericanderson: Yup, I wanted to get the PRs smaller so kept it at chunks that will compile and then was going to work on followups that chain to this. I can probably just fix this one in place though.

packages/generator/src/shared/UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst.ts
ericanderson: This is what it was doing before. If we want to remove lets do in another PR.

packages/client/src/object/fetchPage.ts
ericanderson: Added an example to confirm it works

packages/client/src/object/convertWireToOsdkObjects.ts
ericanderson: Improved per suggestion

packages/client/src/object/Cache.ts
```diff
@@ -92,18 +105,23 @@ export function createAsyncCache<K, V extends {}>(
...
   }
 
   async function get(client: MinimalClient, key: K) {
-    return cache.get(client, key)
+    return cache.get(client, key) ?? inProgress.get(client, key)
       ?? set(client, key, fn(client, key));
   }
 
   async function set(client: MinimalClient, k: K, v: V | Promise<V>) {
-    return cache.set(client, k, await v);
```
mfedderly, line 106: Can you avoid needing `inProgress` by making this something like:

```
cache.set(client,k,v);
return await v;
```

I've had issues trying to make sure this sort of caching doesn't have subtle race conditions, so simpler is better.

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -153,10 +200,20 @@ export async function convertWireToOsdkObjectsInPlace(
...
+export async function convertWireToOsdkObjectsInPlace(
+  client: MinimalClient,
+  objs: OntologyObjectV2[],
+) {
+  fixObjectPropertiesInline(objs);
 
   // We dont want to refetch for each object type in this conversion
   const localObjectCache = createAsyncCache((client, apiName: string) =>
-    client.ontology.provider.getObjectOrInterfaceDefinition(apiName)
+    client.ontology.provider.getObjectDefinition(apiName)
```
mfedderly, line 216: Shouldn't `localObjectCache` have a longer lifetime than just a single call to `convertWireToOsdkObjectsInPlace`? Or do we really really need to have the newest definitions?

Or am I misunderstanding this, and its actually a second layer of caching (if so, then why?).

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -165,14 +222,15 @@ export async function convertWireToOsdkObjectsInPlace(
   );
 
   for (const obj of objs) {
-    const objectDef = await localObjectCache.get(client, obj.$apiName);
+    const objectDef: ObjectOrInterfaceDefinition = await localObjectCache.get(
```
mfedderly, line 225: This is still loading object definitions in serial as you traverse the objs array.

edit: oh it seems like you plan on only calling this with a single object type?

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,14 +118,55 @@ const protoConverterCache = createCache(
...
-  objs: OntologyObjectV2[],
+  objects: OntologyObjectV2[],
+  interfaceApiName: string,
 ) {
-  // Fix properties
+  fixObjectPropertiesInline(objects);
+
+  const localObjectCache = createAsyncCache((client, apiName: string) =>
+    client.ontology.provider.getObjectDefinition(apiName)
+  );
```
mfedderly, line 134: See comments on localObjectCache in the objects path, which are relevant here.

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,14 +118,55 @@ const protoConverterCache = createCache(
...
 const isAfterFeb2024OrNewApis = false;
 
-/**
- * @param objs the objects to be converted, the contents of this array will be mutated
+/*
+ * Currently loses all information about the interface. Future PRs will improve
+ * FIXME
  */
-export async function convertWireToOsdkObjectsInPlace(
+export async function convertWireToOsdkInterfaceInPlace(
```
mfedderly, line 125: This whole method feels pretty duplicative of `convertWireToOsdkObjectsInPlace` below. I'm kind of surprised that loading from an interface isn't a call to that, followed by an extra fixup step to do the `$as(Interface)` cast.

packages/client/src/object/fetchPage.ts
```diff
@@ -83,6 +112,85 @@ export type FetchPageResult<
...
+  );
+  await convertWireToOsdkInterfaceInPlace(
+    client,
+    result.data,
+    interfaceType.apiName,
+  );
+  return result as any;
+}
+
+export async function fetchPageInternal<
```
mfedderly, line 177: `/** @internal */` this too?

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -51,57 +68,33 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
...
+      objectType,
+      objectSet,
+    ) as ObjectSet<Q>["fetchPage"],
+
+    fetchPageOrThrow: fetchPageInternal.bind(
+      globalThis,
+      clientCtx,
+      objectType,
+      objectSet,
+    ) as ObjectSet<Q>["fetchPage"],
```
mfedderly, line 97: Super cool use of `fn.bind`

examples/basic/cli/src/index.ts
```diff
@@ -74,13 +80,35 @@ async function runTests() {
...
+      const r = true
+        ? await client(FooInterface)
+          .where({ name: { $ne: "Patti" } })
+          .where({ name: { $ne: "Roth" } })
+          .fetchPage({ pageSize: 5 })
+        : await fetchPage(clientCtx, FooInterface, {
+          pageSize: 5,
+        });
+
+      expectType<TypeOf<typeof r, PageResult<Osdk<FooInterface, "$all">>>>(
```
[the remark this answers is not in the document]
ericanderson: All of these type expectations leave things to be desired

packages/client/src/object/Cache.ts
ericanderson: wrote a ton of tests for you in the upcoming pr

packages/client/src/object/convertWireToOsdkObjects.ts
ericanderson: I forget if this PR or the followup did this but its okay at the end of the chain.

packages/client/src/object/convertWireToOsdkObjects.ts
ericanderson: This is actually what happened next in the chain when $as was implemented

packages/client/src/object/Cache.ts
ericanderson: I'm relying on the `getOrUndefined` to return instantly. I guess I don't have to but I like the semantics better.

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -41,7 +38,23 @@ function isObjectTypeDefinition(
   return def.type === "object";
 }
 
-const searchAroundPrefix = "searchAround_";
+export function isObjectSet(o: any): o is ObjectSet<any> {
+  return o != null && typeof o === "object"
+    && isWireObjectSet(objectSetDefinitions.get(o));
```
mfedderly, line 43: We can probably just shortcut this whole thing to `objectSetDefinitions.get(o) != null`

mfedderly (approved):

PR #106, opened 2024-03-04
Rework @osdk/client type mapping

mfedderly:
Using @osdk/api mapping interfaces is pretty flawed because of things like Attachments which cannot be represented there.

This changes @osdk/client to contain clearer mapping interfaces for Client<>Wire for both PropertyValue (object) and DataValue (action, query). The DataValue mapping interfaces exclude some of the more complicated types: ObjectSet, Object references, Sets, Structs, TwoDimensionalAggregation, ThreeDimensionalAggregation.

This also changes long and decimal data to be correctly marked as strings when we get them from the backend, but we will still accept numbers to send to the backend.

Are we happy with the naming here `DataValueWireToClient` `DataValueClientToWire` `PropertyValueWireToClient` `PropertyValueClientToWire`?

### DIFF
--- a/packages/client/src/mapping/DataValueMapping.ts
+++ b/packages/client/src/mapping/DataValueMapping.ts
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Attachment } from "../object/Attachment.js";
+
+/**
+ * Map from the DataValue type to the typescript type that we return
+ */
+export interface DataValueWireToClient {
+  attachment: Attachment;
+  boolean: boolean;
+  byte: number;
+  datetime: string;
+  decimal: string;
+  float: number;
+  double: number;
+  integer: number;
+  long: string;
+  marking: string;
+  null: null;
+  short: number;
+  string: string;
+  timestamp: string;
+}
+
+/**
+ * Map from the DataValue type to the typescript type that we accept
+ */
+export interface DataValueClientToWire {
+  attachment: Attachment;
+  boolean: boolean;
+  byte: number;
+  datetime: string;
+  decimal: string | number;
+  float: number;
+  double: number;
+  integer: number;
+  long: string | number;
+  marking: string;
+  null: null;
+  short: number;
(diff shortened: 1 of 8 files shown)


packages/client/src/mapping/PropertyValueMapping.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Map from the PropertyDefinition type to the typescript type that we return
+ */
+export interface PropertyValueWireToClient {
+  attachment: Attachment;
+  boolean: boolean;
+  byte: number;
+  datetime: string;
+  decimal: string;
```
mfedderly, line 27: This is actually an arbitrary precision number, so we must treat it as a string to safely represent it in javascript.

packages/client/src/mapping/PropertyValueMapping.ts
```diff
@@ -0,0 +1,65 @@
...
+  double: number;
+  float: number;
+  geopoint: GeoJSON.Point;
+  geoshape: GeoJSON.GeoJSON;
+  integer: number;
+  long: string;
```
mfedderly, line 33: Similarly, javascript's number types cannot represent the full range of values in a 64 bit integer, so we must treat longs as strings as well.

packages/client/src/mapping/PropertyValueMapping.ts
```diff
@@ -0,0 +1,65 @@
...
+  timestamp: string;
+
+  numericTimeseries: unknown;
+  stringTimeseries: unknown;
+}
+
+/**
+ * Map from the PropertyDefinition type to the typescript type that we accept
+ */
+export interface PropertyValueClientToWire {
```
mfedderly, line 46: As far as I can tell, there are ONLY used to map back to the primary key types, so some of these are not going to be used.

packages/client/src/mapping/DataValueMapping.ts
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Attachment } from "../object/Attachment.js";
+
+/**
+ * Map from the DataValue type to the typescript type that we return
+ */
+export interface DataValueWireToClient {
```
mfedderly, line 22: By specifying the direction that we intend to be mapping types, it lets us be specific about what we're returning, but more permissive with what we're accepting and sending to the backend. Some of the ClientToWire mappings we can rely on the backend to handle the transformations for us. For example, we can send a number for a long field, and it will be converted to the backend specific type as its parses the request.

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -195,16 +195,18 @@ export function createBaseObjectSet<
...
           type: "filter",
           objectSet: objectSet,
-          where: modernToLegacyWhereClause({
-            [objectType.primaryKeyApiName]: primaryKey,
-          }),
+          where: {
+            type: "eq",
+            field: objectType.primaryKeyApiName,
+            value: primaryKey,
+          },
```
ericanderson, line 209: Nice, thats better

ericanderson (approved):
Thank you for digging into these. I completely agree that we should just give strings if we cannot represent things. If the user wants the lossy results thats okay but we shouldnt do it for them

PR #108, opened 2024-03-05
Include create-app examples

tzyl:
This PR creates three rendered versions of the `@osdk/create-app` templates under the `examples/` folder where the existing examples were moved to `examples-extra/`.

- example-next-static-export
- example-react
- example-vue

These are almost exactly the raw output of the CLI apart from a few modifications:

- Swapped OSDK package to a locally generated package (reusing the 1.1 example)
- Removed .npmrc as we now use the locally generated package
- Replaced README explaining what the example actually is

With as few modifications as possible I'm hoping to next set up a script to automatically render these examples based on the template files and enforce that they're up to date so we know that the output of the templates successfully builds


packages/create-app/templates/template-next-static-export/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2020",
```
tzyl, line 3: This is to match the other templates based on vite which originally was needed when I had some of the generated code local to the example packages but might not matter anymore now that I reused an existing local package

packages/create-app/.gitignore
zeyadkhaled, line 1: Why was this deleted?

examples/example-next-static-export/README.md
```diff
@@ -0,0 +1,11 @@
+# example-next-static-export
```
zeyadkhaled, line 1: Do we want to add to the READMEs how to use/run these examples?

packages/create-app/.gitignore
tzyl: Every version of the next template has `lib/` folders under `src/` that need to be remembered to be unignored which was getting annoying to manage so I changed the root .gitignore instead to allow `src/lib/` instead

examples/example-next-static-export/README.md
tzyl: Hmm, the thing is that there's only so much that you can do with the example because it needs a real ontology and stack to login and start loading data. I agree it would be nice if there was something more to see do you have any ideas? Ideally it stays as close as possible to the stock rendered template so we can keep it in sync automatically and it serves as our end to end test that the template is buildable

zeyadkhaled (approved):

PR #111, opened 2024-03-05
Add startsWith for string filter

ssanjay1:
` fullName: { $startsWith: "Hemsworth" },`

### DIFF
--- a/examples/basic/sdk/ontology.json
+++ b/examples/basic/sdk/ontology.json
@@ -156,11 +156,6 @@
               "type": "string"
             }
           },
-          "name": {
-            "dataType": {
-              "type": "string"
-            }
-          },
           "latitude": {
             "dataType": {
               "type": "double"
--- a/examples/basic/sdk/ontology.json
+++ b/examples/basic/sdk/ontology.json
@@ -156,6 +156,11 @@
               "type": "string"
             }
           },
+          "name": {
+            "dataType": {
+              "type": "string"
+            }
+          },
           "latitude": {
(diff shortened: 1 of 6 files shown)


examples/basic/sdk/ontology.json
```diff
@@ -156,6 +156,11 @@
               "type": "string"
             }
           },
+          "name": {
+            "dataType": {
+              "type": "string"
+            }
+          },
```
mfedderly, line 163: is this logically a duplicate of `usState`?

packages/client/src/query/WhereClause.ts
```diff
@@ -32,7 +32,8 @@ export type PossibleWhereClauseFilters =
   | "$lt"
   | "$lte"
   | "$within"
-  | "$intersects";
+  | "$intersects"
+  | "$startsWith";
```
mfedderly, line 36: Its kind of wild that we didn't need to write any additional converter code for this to work.

mfedderly (approved):

examples/basic/sdk/ontology.json
ssanjay1: I guess yea logically, usState is the two letter abbreviation though and I wanted to make sure it worked with strings that had spaces in it. I can remove though

packages/client/src/query/WhereClause.ts
ssanjay1: Yea I was surprised too

PR #113, opened 2024-03-06
Fix long and decimal types to correctly be strings in legacy-client

mfedderly:
Update the types for Decimal and Long to correctly be `string` instead of `number` to match the actual type of the data that we return.

Queries and Actions should take `number | string` and not break back-compatability, Queries would return `string` now instead for longs. Object set `.get(primaryKey)` will also have to change from number to string for longs (and decimals? I forget if they can be used for PK).

### DIFF
--- a/packages/legacy-client/src/client/queries.ts
+++ b/packages/legacy-client/src/client/queries.ts
@@ -113,7 +113,7 @@ export interface ValidLegacyBaseQueryDataTypes {
   double: number;
   float: number;
   integer: number;
-  long: number;
+  long: number | string;
   boolean: boolean;
   string: string;
   date: LocalDate;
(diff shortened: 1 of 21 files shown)


packages/legacy-client/src/client/queries.ts
```diff
@@ -125,7 +125,11 @@ export type QueryDataTypeBase<
   O extends OntologyDefinition<any>,
   T extends QueryDataType<O, any, R>,
   R extends boolean,
-> = T extends PrimitiveDataType<infer X> ? ValidLegacyBaseQueryDataTypes[X]
+> = T extends PrimitiveDataType<infer X>
+  // returned longs are always strings, but we can accept numbers as input
+  ? (R extends true
```
ssanjay1, line 130: What is the R generic for?

ssanjay1: And also, so we don't need to worry about decimals here right?

mfedderly: R is for "is this a return type?"
`QueryDefinitions` don't allow Decimals 🤷

ssanjay1 (approved):

PR #119, opened 2024-03-07
Generate public examples from @osdk/example-generator

tzyl:
### Before this PR
The `examples/` packages were manually rendered from `@osdk/create-app` and a few patches applied to make it build against a locally generated SDK from `@osdk/examples.one.dot.one`.

### After this PR

The `@osdk/example-generator` package automates the generation of the examples based on the latest templates. A check command also runs which verifies if the templates were updated without updating the packages so that we ensure that the latest state of the templates can build successfully in the examples packages.

`pnpm generate` updates the example packages
`pnpm check` verifies if they are up to date

The old snapshot testing for `@osdk/create-app` can be removed in favour of actually building these example packages but still worth running the CLIs in the unit tests to verify no errors without saving the snapshot files.


tzyl:
Example check fail on examples out of date with templates:

https://github.com/palantir/osdk-ts/pull/119/commits/5dfd204fac987b3fdd96c9b01aa8b455723ad1dd
https://github.com/palantir/osdk-ts/actions/runs/8191673148/job/22401405744

```
  Error:  Found 1 differences in example-react please generate examples again.
  Error:  [ { path1: '/home/runner/work/osdk-ts/osdk-ts/examples/example-react/src',
      path2: '/tmp/tmp-5486-rGwkJZEPNWyc/example-react/src',
      relativePath: '/src',
      name1: 'Home.tsx',
      name2: 'Home.tsx',
      state: 'distinct',
      permissionDeniedState: 'access-ok',
      type1: 'file',
      type2: 'file',
      level: 1,
      size1: 1364,
      size2: 1345,
      date1: 2024-03-07T16:38:49.132Z,
      date2: 2024-03-07T16:39:48.708Z,
      reason: 'different-content' } ]
   ELIFECYCLE  Command failed with exit code 1.
```

packages/create-app/templates/template-next-static-export/.gitignore
```diff
@@ -10,8 +10,8 @@
 /coverage
 
 # next.js
-/.next/
-/out/
+/.next
+/out
```
tzyl, line 14: I changed this .gitignore because it wasn't being respected by the globby implementation in the sample code provided here they mention there are some bugs with it:

https://github.com/gliviu/dir-compare?tab=readme-ov-file#implement-gitignore-filter

packages/example-generator/src/run.ts
```diff
@@ -0,0 +1,220 @@
...
+    type: "modify",
+    newContent: readme(template),
+  }),
+};
+
+const MUTATORS: Mutator[] = [
+  DELETE_NPM_RC,
+  UPDATE_PACKAGE_JSON,
+  UPDATE_README,
+];
```
zeyadkhaled, line 195: Just a nit: can we move these under the interfaces as it makes reading the implementation for the first time easier without having to jump around few times to understand what is happening in the for loop

packages/example-generator/src/run.ts
```diff
@@ -0,0 +1,220 @@
...
+interface ModifyFile {
+  type: "modify";
+  newContent: string;
+}
+
+interface DeleteFile {
+  type: "delete";
+}
+
+export async function run({ outputDirectory, check }: RunArgs) {
```
zeyadkhaled, line 53: It took me few reads to understand the flow. Perhaps a small refactor of dividing into few chunks could make it better? Possibly have the run function look like this:

```
 const tmpDir = setupTemporaryDirectory();

  await generateAndMutateProjects(tmpDir.name);

  await formatProjectsUsingMonorepolint(tmpDir.name);

  if (check) {
    await checkGeneratedProjects(outputPath, tmpDir.name);
  } else {
    copy(overwrite?)GeneratedProjects(outputPath, tmpDir.name);
  }```

zeyadkhaled (approved):
Just minor nits otherwise looks great 👍

PR #120, opened 2024-03-07
Bump TS and make writing test network mocks easier

ericanderson:
## Changes
* Upgrade to latest typescript on packages that can
* Sever ts references for the legacy package because the TS 4 doesn't play nicely with modern syntax
* Implement a helper so you can write network mocks faster and more correctly.

## Why the helper is great
* I dont need to know the url / copy it from the generated code.
* I dont need to copy/paste the request response
* I am forced to specify all the path parameter names and cant do too few or too many
* I get typesafety on the path params instead of `Record<string, string | string[]>` 
* I don't have to cast the path params to `string` because we know they aren't `string[]`

## Example
```
  handleOpenApiCall(
    executeQueryV2,
    ["ontologyApiName", "queryApiName"],
    async (req, res, ctx) => {
...
});
```
* The number of path params is inferred from `executeQueryV2` to be 2. Giving more or fewer as the names to use will fail.
* `req` is inferred to be `RestRequest<ExecuteQueryRequest, Record<"ontologyApiName" | "queryApiName", string>>`
  * Note it declares the body of the request type AND the path params.
* `res` is inferred to be `ResponseComposition<BaseAPIError | ExecuteQueryResponse>`. Now I get errors if I try to return JSON that doesn't match the type.

### DIFF
--- a/.monorepolint.config.mjs
+++ b/.monorepolint.config.mjs
@@ -58,12 +58,16 @@ const formatedGeneratorHelper = (contents, ext) => async (context) => {
 
 /**
  * @param {string} baseTsconfigPath
- * @param {{customTsconfigExcludes?: string[]}} opts
+ * @param {{
+ *   customTsconfigExcludes?: string[]
+ *   skipTsconfigReferences?: boolean
+ * }} opts
  * @returns
  */
 function getTsconfigOptions(baseTsconfigPath, opts) {
   return {
     file: "tsconfig.json",
+    excludedReferences: opts.skipTsconfigReferences ? ["**/*"] : undefined,
     template: {
       extends: baseTsconfigPath,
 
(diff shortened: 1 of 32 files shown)


ssanjay1 (approved):

.monorepolint.config.mjs
```diff
@@ -108,7 +112,8 @@ function getTsconfigOptionsE2E(baseTsconfigPath) {
  *  packageDepth: number,
  *  type: "library" | "example",
  *  customTsconfigExcludes?: string[],
- *  tsVersion?: "^5.2.2"|"^4.9.0",
+ *  tsVersion?: "^5.4.2"|"^4.9.0",
```
mfedderly, line 115: Was curious so I looked it up: 4.9.5 is the last in the 4.9.x series

package.json
```diff
@@ -39,7 +39,7 @@
     "tsconfig": "workspace:*",
     "tsup": "^7.2.0",
     "turbo": "^1.10.13",
-    "typescript": "^5.2.2",
+    "typescript": "^5.3.2",
```
mfedderly, line 42: Should this be 5.4.2?

packages/api/package.json
```diff
@@ -40,12 +40,12 @@
...
   "devDependencies": {
     "ts-expect": "^1.3.0",
-    "typescript": "^5.2.2"
+    "typescript": "^5.4.2"
   },
   "publishConfig": {
     "access": "public"
   },
-  "keywords": [ ],
+  "keywords": [],
```
mfedderly, line 48: Reminder that we still don't have autoformatting enabled for package.json because of autorelease

package.json
ericanderson: Turns out PNPM already did that under the covers but explicit is better

PR #123, opened 2024-03-08
Add README to examples folder

tzyl:
Some helpful introduction if we link directly to this folder

### DIFF
--- a/examples/example-next-static-export/README.md
+++ b/examples/example-next-static-export/README.md
@@ -1,6 +1,6 @@
 # example-next-static-export
 
-This project was generated with `@osdk/create-app` from the `next-static-export` template. It is built against a locally generated SDK and a non-existent Foundry stack so is intended for reference purposes only.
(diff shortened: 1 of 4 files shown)


examples/example-next-static-export/README.md
```diff
@@ -1,6 +1,6 @@
 # example-next-static-export
 
-This project was generated with `@osdk/create-app` from the `next-static-export` template. It is built against a locally generated SDK and a non-existent Foundry stack so is intended for reference purposes only.
+This project was generated with [`@osdk/create-app`](https://www.npmjs.com/package/@osdk/create-app) from the `next-static-export` template. It is built against a locally generated SDK and a non-existent Foundry stack so is intended for reference purposes only.
```
zeyadkhaled, line 3: Just for better readability. Same applies for the other examples.

```suggestion
This project was generated with [`@osdk/create-app`](https://www.npmjs.com/package/@osdk/create-app) from the `next-static-export` template. It is built against a locally generated SDK and a non-existent Foundry stack, so it is intended for reference purposes only.
```

zeyadkhaled (approved):

examples/example-next-static-export/README.md
tzyl: Done

PR #92, opened 2024-02-23
Improvement: Add @osdk/cli README

zeyadkhaled:

PR #127, opened 2024-03-11
Update template READMEs with note about foundry.config.json

tzyl:
- Explain that a `foundry.config.json` file is created and can be deleted if not using Foundry website hosting
- Add instruction that `--corsProxy true` can be used if needed for local development
- Add detail that with Foundry website hosting CORS is automatically set up for production
- Link to NPM packages (will pick up https://github.com/palantir/osdk-ts/pull/92 when it's merged)

### DIFF
--- a/packages/create-app/templates/template-next-static-export/README.md.hbs
+++ b/packages/create-app/templates/template-next-static-export/README.md.hbs
@@ -1,6 +1,6 @@
 # {{project}}
 
-This project was generated with `@osdk/create-app` and demonstrates using the OSDK package `{{osdkPackage}}` with React on top of Next.js. Check out the [Next.js](https://nextjs.org/docs) docs for further configuration.
+This project was generated with [`@osdk/create-app`](https://www.npmjs.com/package/@osdk/create-app) and demonstrates using the Ontology SDK package `{{osdkPackage}}` with React on top of Next.js. Check out the [Next.js](https://nextjs.org/docs) docs for further configuration.
 
 ## Developing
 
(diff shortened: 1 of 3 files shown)


packages/create-app/templates/template-next-static-export/README.md.hbs
```diff
@@ -30,4 +30,6 @@ Production configuration is stored in `.env.production`.
 
 If you did not fill in the URL your production application will be hosted on you will need to fill in the `NEXT_PUBLIC_FOUNDRY_REDIRECT_URL` in `.env.production`.
 
-In order to make API requests to Foundry, CORS must be configured for the stack to allow the production origin to load resources. The configured OAuth client must also allow the production origin auth callback as a redirect URL.
+In order to make API requests to Foundry, CORS must be configured for the stack to allow the production origin to load resources. This will be automatically done for you if you are using Foundry website hosting. The configured OAuth client must also allow the production origin auth callback as a redirect URL.
+
+A `foundry.config.json` file is included in the root of this project to make deploying to Foundry website hosting with [`@osdk/cli`](https://www.npmjs.com/package/@osdk/cli) easier. If you are not using Foundry website hosting for your application you may delete this file.
```
zeyadkhaled, line 35: What do you think we make the generation of the file opt-out option with a default value of true. And here we can mention, If you aren't using foundry website hosting pass `--websiteHosting false`. I feel asking the user to delete here is a bit unclean. What do you think?

zeyadkhaled (approved):

PR #129, opened 2024-03-12
Fix array and string filters

ssanjay1:
Realized that the `contains` filter only works on array properties, so fixed that in the OSDK. Also added remaining string filters while I was there

### DIFF
--- a/packages/client/src/query/WhereClause.ts
+++ b/packages/client/src/query/WhereClause.ts
@@ -33,7 +33,10 @@ export type PossibleWhereClauseFilters =
   | "$lte"
   | "$within"
   | "$intersects"
-  | "$startsWith";
+  | "$startsWith"
+  | "$containsAllTermsInOrder"
+  | "$containsAnyTerm"
+  | "$containsAllTerms";
 
 // We need to conditional here to force the union to be distributed
 type MakeFilter<K extends PossibleWhereClauseFilters, V> = K extends string ? {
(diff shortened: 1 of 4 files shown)


packages/client/src/query/WhereClause.ts
```diff
@@ -113,12 +128,18 @@ export type GeoFilter_Intersects = {
...
 export type GeoFilter = GeoFilter_Within | GeoFilter_Intersects;
 
-type FilterFor<PD extends ObjectTypePropertyDefinition> = PD["type"] extends
-  "string" ? StringFilter
-  : PD["type"] extends "geopoint" | "geoshape" ? GeoFilter
-  : PD["type"] extends "datetime" | "timestamp" ? DatetimeFilter
-  : PD["type"] extends "boolean" ? BooleanFilter
-  : NumberFilter; // FIXME we need to represent all types
+type FilterFor<PD extends ObjectTypePropertyDefinition> =
+  PD["multiplicity"] extends true
```
ssanjay1, line 132: I thought it was clearer to do the multiplicity check right here and split out into the different array filters, but could also do that check in the individual `StringFilter`, `GeoFilter`, `DateTimeFilter` etc. types instead

mfedderly: I think you need `()` around the array cases, the ternary syntax gets weird. Even if its correct as is it would be clearer with parens.

mfedderly (approved):
Approved to unblock but please add the parens to the array cases first

packages/client/src/query/WhereClause.ts
ssanjay1: Yea good call, just added

PR #131, opened 2024-03-13
Minor improvements everywhere

ericanderson:
- @osdk/cli typescript generator learns `--ontologyRid` optional argument to limit which ontologies are considered when you use a `--clientId` that is open ended.
- `apiName` on `ObjectInterfaceBaseDefinition` gets a stronger brand so we can select out the raw string without the brand
- `console.log(someObj.$link)` now prints something like `{ lead: [class ObjectSet], peeps: [class ObjectSet] }` instead of `{}`
- Using `fetchPage` with an interface now properly removes `$rid` unless you requested it (prototype interfaces backend always returns $rid sadly, meaning console.log()/JSON.stringify() does the wrong thing)
- `fetchPage` learns to `augment` interface queries for additional properties
- `OsdkObject` gets the `$objectType` field added to its type (value was already there)
- `Osdk<SomeType, never>` now returns all property types, however unions them with `undefined` since we don't know anything about them. You
- The test data had the spt maps inverted from the way the backend returns them, causing real ontology's we fetch to be represented wrong. This is now fixed

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -44,3 +44,5 @@ export type { NOOP } from "./util/NOOP.js";
 
 export { createMinimalClient } from "./createMinimalClient.js";
 export type { MinimalClient } from "./MinimalClientContext.js";
+
+export { augment } from "./object/fetchPage.js";
--- a/packages/client/src/object/convertWireToOsdkObjects.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects.ts
@@ -38,55 +38,82 @@ const OriginClient = Symbol();
 const UnderlyingObject = Symbol();
 const InterfaceDefinitions = Symbol();
 
+class LinkFetcherProxyHandler<Q extends AugmentedObjectTypeDefinition<any, any>>
+  implements ProxyHandler<any>
+{
+  constructor(
+    private readonly objDef: Q,
+    private readonly primaryKey: any,
+    private readonly client: MinimalClient,
+  ) {}
+
+  get(_target: any, p: string | symbol) {
+    const linkDef = this.objDef.links[p as string];
+    if (linkDef == null) {
+      return;
+    }
+
+    const objectSet = createObjectSet(this.objDef, this.client).where({
+      [this.objDef.primaryKeyApiName]: this.primaryKey,
+    } as WhereClause<Q>).pivotTo(p as string);
+
+    if (!linkDef.multiplicity) {
+      return {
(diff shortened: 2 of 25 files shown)


packages/client/src/index.ts
```diff
@@ -44,3 +44,5 @@ export type { NOOP } from "./util/NOOP.js";
 
 export { createMinimalClient } from "./createMinimalClient.js";
 export type { MinimalClient } from "./MinimalClientContext.js";
+
+export { augment } from "./object/fetchPage.js";
```
mfedderly, line 48: I'm not totally thrilled with this becoming part of the public API, but appreciate that its super hard to get the typings to work out. Perhaps this is covered by the 'make hard things possible' side of things instead of the 'make most things easy' part.

ericanderson: It definitely works without doing this. This is only needed if you want the typesafety on picking properties.

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -38,6 +37,10 @@ const OriginClient = Symbol();
 const UnderlyingObject = Symbol();
 const InterfaceDefinitions = Symbol();
 
+// this is just so our fake link proxy looks nice
+// when you console.log()
+class ObjectSet {}
```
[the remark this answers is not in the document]
ericanderson: Switched to just relying on Object.keys(). Not worth the bloat for console.log() to look pretty

packages/client/src/index.ts
ericanderson: Lets leave this for now

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -38,55 +38,82 @@ const OriginClient = Symbol();
...
+  ownKeys(): ArrayLike<string | symbol> {
+    return Object.keys(this.objDef.links);
+  }
+
+  getOwnPropertyDescriptor(
+    target: any,
+    p: string | symbol,
+  ): PropertyDescriptor | undefined {
+    return {
+      enumerable: true,
```
mfedderly, line 84: I forget does this need to be `this.objDef.links != null`

mfedderly (approved):

packages/client/src/object/convertWireToOsdkObjects.ts
ericanderson: I decided to drop that for now. we can always add back

PR #135, opened 2024-03-17
change application url prompt

ewitkon:
Application url prompt in `create-app` is confusing
`Do you know the URL your production application will be hosted on? This is required to create a production build of your application with the correct OAuth redirect URL.` options are:
`Yes, prefill it for me` - this reads like the app will pre-fill the url for me when we actually want the user to provide the url on the next prompt
`No, I will fill it manually later` - we don't instruct the user where we expect them to fill it. 

This PR suggest a different language for both ahead of a user testing.

### DIFF
--- a/packages/create-app/src/prompts/promptApplicationUrl.ts
+++ b/packages/create-app/src/prompts/promptApplicationUrl.ts
@@ -33,8 +33,12 @@ export async function promptApplicationUrl(
       {
         type: "select",
         options: [
-          { label: "Yes, prefill it for me", value: "yes" },
-          { label: "No, I will fill it in myself later", value: "no" },
+          { label: "Yes, let me fill it here", value: "yes" },
+          {
+            label:
+              "No, I will set `VITE_FOUNDRY_REDIRECT_URL` variable in `.env.production` file later",
+            value: "no",
+          },
         ],
       },
       // Types for "select" are wrong the value is returned rather than the option object
--- a/packages/create-app/src/prompts/promptApplicationUrl.ts
+++ b/packages/create-app/src/prompts/promptApplicationUrl.ts
@@ -35,8 +35,7 @@ export async function promptApplicationUrl(
         options: [
           { label: "Yes, let me fill it here", value: "yes" },
           {
-            label:
-              "No, I will set `VITE_FOUNDRY_REDIRECT_URL` variable in `.env.production` file later",
+            label: "No, I will set `VITE_FOUNDRY_REDIRECT_URL` variable later",
             value: "no",
           },
         ],
--- a/packages/create-app/src/prompts/promptApplicationUrl.ts
+++ b/packages/create-app/src/prompts/promptApplicationUrl.ts
@@ -36,7 +36,7 @@ export async function promptApplicationUrl(
           { label: "Yes, let me fill it here", value: "yes" },
           {
             label:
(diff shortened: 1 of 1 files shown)


packages/create-app/src/prompts/promptApplicationUrl.ts
```diff
@@ -33,8 +33,12 @@ export async function promptApplicationUrl(
...
         type: "select",
         options: [
-          { label: "Yes, prefill it for me", value: "yes" },
-          { label: "No, I will fill it in myself later", value: "no" },
+          { label: "Yes, let me fill it here", value: "yes" },
+          {
+            label:
+              "No, I will set `VITE_FOUNDRY_REDIRECT_URL` variable in `.env.production` file later",
+            value: "no",
+          },
```
tzyl, line 41: For the no label I would be more hesitant on referencing the exact place to change because that depends on the template (for example nextjs uses a `NEXT_PUBLIC_` prefix). The READMEs that are generated for the template do [explain this](https://github.com/palantir/osdk-ts/blob/3cfe65fb7a795cbe35f4a4d51fc51ccfadbe32e7/packages/create-app/templates/template-react/README.md.hbs#L29-L31) and we could technically make this label aware of the template that was chosen as well but I think it would be enough to defer to the README.

```suggestion
          { label: "Yes, let me fill it here", value: "yes" },
          { label: "No, I will fill it in myself later", value: "no" },
```

ewitkon: @tzyl How about just referencing the name of the variable?

tzyl: The static variable name here is still incorrect as it would need to switch depending on the template selected in the previous prompt. Another suggestion is after the user selected "no" to put another line with the extra detail telling them to follow the generated README instructions later:

```
✔ Do you know the URL your production application will be hosted on? This is required to create a production build of your application with the correct OAuth redirect URL.
No, I will fill it in myself later

You have chosen not to fill in the URL your production application will be hosted on. Follow the instructions in the generated project README once you are ready to create a production build of your application.
```

ewitkon: Ended up suggesting 
`No, I will set it up later based on the instructions in the README`
One line is simple

packages/create-app/src/prompts/promptApplicationUrl.ts
```diff
@@ -33,8 +33,12 @@ export async function promptApplicationUrl(
       {
         type: "select",
         options: [
-          { label: "Yes, prefill it for me", value: "yes" },
-          { label: "No, I will fill it in myself later", value: "no" },
+          { label: "Yes, let me fill it here", value: "yes" },
```
tzyl, line 36:
```suggestion
          { label: "Yes, let me fill it in here", value: "yes" },
```

packages/create-app/src/prompts/promptApplicationUrl.ts
```diff
@@ -33,8 +33,12 @@ export async function promptApplicationUrl(
...
+          {
+            label:
+              "No, I will set it up later based on the instructions in the README",
```
tzyl, line 39:
```suggestion
              "No, I will fill it in later following the instructions in the generated README",
```

tzyl (approved):

PR #142, opened 2024-03-21
Deprecate __apiName, __rid, and __primaryKey

ssanjay1:


packages/legacy-client/src/client/baseTypes/OntologyObject.ts
```diff
@@ -20,12 +20,19 @@ export interface OntologyObject<
   T extends string = string,
   P extends NonNullable<ParameterValue> = NonNullable<ParameterValue>,
 > {
+  /** @deprecated use $rid */
   __rid: string;
+  /** @deprecated use $apiName */
   __apiName: T;
+  /** @deprecated use $primaryKey */
   __primaryKey: P;
+  $rid: string | undefined;
```
ericanderson, line 29: I know I said we should `| undefined` this but now im second guessing that. I feel like many people will see this and be like "ugh thats annoying" and just add a `!` to their code. Then when they switch to 2.0 when its actually not likely to be there, they will now realize things are broken.

ericanderson (approved):
One thing we should fix but preemptive approval

packages/legacy-client/src/client/baseTypes/OntologyObject.ts
ssanjay1: Ah ok, I'm fine removing the undefined because I do see the point of people just hacking around it, but then we're basically saying 2.0 will _definitely_ break things for anyone using the rids. Is that ok?

PR #145, opened 2024-03-22
Add fetchpage with errors

ssanjay1:
Add result wrapper around fetch page result for 2.0


packages/client/src/object/errors/handlers/ExecuteActionErrorHandler.ts
```diff
@@ -0,0 +1,359 @@
...
+  ParametersNotFound,
+  PropertiesNotFound,
+  PropertyApiNameNotFound,
+  PropertyBaseTypeNotSupported,
+  UnknownParameter,
+} from "../ApiErrors.js";
+import type { PalantirApiError } from "../Errors.js";
+import { DefaultErrorHandler } from "./DefaultErrorHandler.js";
+
+export class ExecuteActionErrorHandler extends DefaultErrorHandler {
```
ssanjay1, line 43: Do not need all these error handlers, basically just ported over everything from 1.1 for right now

packages/client/src/object/fetchPage.ts
```diff
@@ -129,6 +136,16 @@ export type FetchPageResult<
   >
 >;
 
+export type FetchPageResultWithErrors<
```
ssanjay1, line 139: Should probably rename this type, as its just `FetchPageResult` without the `Promise` wrapper

packages/client/src/object/fetchPage.ts
```diff
@@ -192,6 +209,50 @@ async function fetchInterfacePage<
...
+      );
+      result.data = await convertWireToOsdkObjects(
+        client,
+        result.data as OntologyObjectV2[], // drop readonly
+        interfaceType.apiName,
+        !args.includeRid,
+      );
+      return result as any;
+    }, // FIXME: not for interfaces
+    e =>
```
ssanjay1, line 247: Need to handle interface load errors seaprately

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -173,4 +173,19 @@ describe("ObjectSet", () => {
 
     expect(employee.$primaryKey).toBe(stubData.employee1.employeeId);
   });
+
+  it(" object set union works with fetchPageWithErrors", async () => {
+    const objectSet = client(Employee);
+    const unionedObjectSet = objectSet.union(objectSet);
+    let iter = 0;
+    const result = await unionedObjectSet.fetchPageWithErrors();
```
ssanjay1, line 181: This is what using it would look like in practice

packages/client/src/object/fetchPage.ts
```diff
@@ -205,12 +243,46 @@ export async function fetchPageInternal<
   args: FetchPageArgs<Q, L, R, A> = {},
 ): FetchPageResult<Q, L, R> {
   if (objectType.type === "interface") {
-    return await fetchInterfacePage(client, objectType, args, objectSet) as any; // fixme
+    return await fetchInterfacePage(
```
ssanjay1, line 246: Not sure why this got reformatted, but no change here otherwise

packages/client/src/object/Result.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type ErrorResult = { error: Error; value?: never };
+export type OkResult<V> = { value: V; error?: never };
+export type Result<V> = OkResult<V> | ErrorResult;
+
+export function isOk<X>(a: Result<X>): a is OkResult<X> {
+  return ("value" in a);
+}
+
+export function isError<X>(a: Result<X>): a is ErrorResult {
```
ssanjay1, line 25: I know we don't really need this, but added it to try and match up with 1.1's as much as possible to help people migrating who may have been using it

packages/client/src/object/fetchPage.ts
ericanderson: dprint will preserve newline whitespace often if it exists. lets set this back to the way it was before merging.

packages/client/src/object/fetchPage.ts
```diff
@@ -122,11 +125,17 @@ export type FetchPageResult<
...
 > = Promise<
-  PageResult<
-    ObjectOrInterfacePropertyKeysFrom2<Q> extends L
-      ? (DefaultToFalse<R> extends false ? Osdk<Q> : Osdk<Q, "$all" | "$rid">)
-      : (DefaultToFalse<R> extends false ? Osdk<Q, L> : Osdk<Q, L | "$rid">)
-  >
+  FetchPageWithErrorsResult<Q, L, R>
+>;
+
+export type FetchPageWithErrorsResult<
```
ericanderson, line 131: Should we call this `FetchPageResult` like it was before but drop the `Promise<>` part? That way you can compose like below?

packages/client/src/object/fetchPage.ts
```diff
@@ -205,12 +229,46 @@ export async function fetchPageInternal<
...
+      args,
+      objectSet,
+    ) as any; // fixme
   } else {
     return await fetchObjectPage(client, objectType, args, objectSet) as any; // fixme
   }
 }
 
+/** @internal */
+export async function fetchPageWithErrorsInternal<
```
ericanderson, line 244: Do we need these extra methods? Could fetchPageOrError() just call fetchPage but with the wrapResult call?

packages/client/src/object/wrapResult.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { Result } from "./Result.js";
+
+export async function wrapResult<T>(
```
ericanderson, line 19: If we do the suggestion i give above where its just a single new method that wraps the old method, we could just inline this I think?

packages/client/src/object/fetchPage.ts
ericanderson: Just realized you said the same thing as me last week!

packages/client/src/object/fetchPage.ts
ssanjay1: We definitely do not, I initially kept them separate because of all the error handling stuff we had before so wasn't sure if I'd need to do some special stuff for interfaces, but without all that can totally just wrap the `fetchPage`

packages/client/src/object/wrapResult.ts
ssanjay1: Done

packages/client/src/object/fetchPage.ts
```diff
@@ -203,14 +203,37 @@ export async function fetchPageInternal<
...
+  L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
+  R extends boolean,
+  A extends Augments,
+>(
+  client: MinimalClient,
+  objectType: Q,
+  objectSet: ObjectSet,
+  args: FetchPageArgs<Q, L, R, A> = {},
+): Promise<Result<FetchPageResult<Q, L, R>>> {
+  try {
```
ssanjay1, line 226: Now inlined the wrapResult method here

packages/client/src/objectSet/ObjectSet.ts
```diff
@@ -44,7 +45,15 @@ export interface MinimalObjectSet<Q extends ObjectOrInterfaceDefinition> {
...
-  ) => FetchPageResult<Q, L, R>;
+  ) => Promise<FetchPageResult<Q, L, R>>;
+
+  fetchPageWithErrors: <
+    L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
+    R extends boolean,
+    const A extends Augments,
+  >(
+    args?: FetchPageArgs<Q, L, R, A>,
+  ) => Promise<Result<FetchPageResult<Q, L, R>>>;
```
ssanjay1, line 56: A lot of return types changed just because I know have to wrap with promise since it's not a part of the type

ericanderson (approved):

PR #147, opened 2024-03-26
Omniapi support

ericanderson:


ericanderson:
Dropping this code from this PR but I want to be able to quickly refind if I need to:
```
const marker = Symbol("omniapi");
export type OmniResource<
  KK,
> = KK extends Record<infer T extends string, (...args: any[]) => Promise<any>>
  ? {
    type: "omniapi";
    methods: {
      [K in T]: [
        method: number,
        path: string,
        hasBody?: 1,
        contentType?: string,
        responseContentType?: string,
      ];
    };
    [marker]?: KK;
  }
  : never;

export type OmniResourceMethods<T extends OmniResource<any>> = NonNullable<
  T[typeof marker]
>;
```

packages/client/src/Client.ts
```diff
@@ -51,11 +51,15 @@ export interface Client<O extends OntologyDefinition<any>>
 // Once we migrate everyone off of using the deprecated parts of `Client` we can rename this to `Client`.
 export interface FutureClient {
   <
-    Q extends ObjectOrInterfaceDefinition | ActionDefinition<any, any, any>,
+    Q extends
```
ssanjay1, line 54: nit: Seems to be a lot of places where the type generic signatures got reformatted, it's fine but just flagging if this was unintentional

packages/omniapi-generator/src/generateResource.ts
```diff
@@ -0,0 +1,306 @@
...
+  return `[${constParts.join(",")}]`;
+}
+
+export function generateRuntimeConst(resource: Resource) {
+  let ret = "";
+
+  for (const method of resource.staticOperations) {
+    // Remove resource name from the function name if it ends with it
+    const methodName = method.name; // method.verb ?? method.name;
+    // const requestType = method.requestBody?.body.requestType;
```
ssanjay1, line 158: Is this supposed to be commented out?

packages/omniapi/src/public/Datasets_Branch.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+import type {
+  ClientContext as $ClientContext,
+  OmniMethod as $OmniMethod,
+} from "@osdk/shared.net";
+import { omniFetch as $omniFetch } from "@osdk/shared.net";
+import type * as $C from "../generated/components.js";
+
+const _listBranches: $OmniMethod<
```
ssanjay1, line 24: These are all generated right? (everything in public)

packages/client/src/createClient.test.ts
```diff
@@ -62,7 +62,7 @@ describe(createClient, () => {
       ontologyRid: "",
       userAgent: "",
     };
-    const stack = "foo.bar";
+    const stack = "https://foo.bar";
```
ssanjay1, line 65: Why do we need to add the protocol now?

ssanjay1 (approved):
Seems fine to me, nothing odd really jumped out

packages/omniapi/src/public/Datasets_Branch.ts
ericanderson: yeah

packages/client/src/createClient.test.ts
ericanderson: Because when you are doing a local dev, you may not be doing HTTPs so we are just going to force it to be configured properly from the onset

packages/omniapi-generator/src/generateResource.ts
ericanderson: Thanks I refactored some of this

PR #152, opened 2024-03-27
Log API name clashes in Foundry SDK Generator

ssanjay1:
To help with our migration to 2.0, we will start logging if any OSDK's generated entity api names clash with our own OSDK types


packages/foundry-sdk-generator/src/generate/commandUtils.ts
```diff
@@ -22,3 +22,371 @@ export function transformArrayArg(arg?: string[]) {
   }
   return arg;
 }
+
+export const osdkTypes = new Set([
```
ssanjay1, line 26: Literally just copied and pasted what we export from the root of legacy client, since we will start exporting all the 2.0 types from root in 1.3 version of OSDK

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
```diff
@@ -128,6 +132,28 @@ export async function generatePackage(
...
+      ...Object.keys(sanitizedOntology.sharedPropertyTypes),
+    ]
+  ) {
+    if (osdkTypes.has(type)) {
+      clashCount++;
+    }
+  }
+
+  if (clashCount > 0) {
+    consola.log("API names clashed: ", clashCount);
```
ssanjay1, line 155: Not actually sure if this will log it where we want

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
```diff
@@ -128,6 +132,28 @@ export async function generatePackage(
...
+
+  const sanitizedOntology = sanitizeMetadata(ontology);
+  let clashCount = 0;
+  for (
+    const type of [
+      ...Object.keys(sanitizedOntology.actionTypes),
+      ...Object.keys(sanitizedOntology.interfaceTypes),
+      ...Object.keys(sanitizedOntology.objectTypes),
+      ...Object.keys(sanitizedOntology.queryTypes),
+      ...Object.keys(sanitizedOntology.sharedPropertyTypes),
```
ericanderson, line 146: because its a cli, we don't need to do anything about this but i would probably have done something like an inner function instead of the above for:
```
function count(x: Record<string, any>) {
  for (const type of Object.keys(x) {
    if (osdkTypes.has(type)) clashCount++;
  }
}
const clashCount = count(santizedOntology.actionTypes) + count(sanitizedOntology.interfaceTypes) +...;
```
Due to it avoiding making the extra array. Another option may have been do to a for loop over a for loop leveraging `Object.keys(santitizedOntology)` for the outter.

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
ericanderson: If this cli uses consola then its probably okay? Maybe use `consola.error` or `consola.warn`? Do we think the user will actually see this when run from the UI?

packages/foundry-sdk-generator/src/generate/commandUtils.ts
ericanderson: nit: select lines, cmd+p, "sort"

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
ssanjay1: Ah I see, I'll modify anyways to use the inner function, I like that better as well

packages/foundry-sdk-generator/src/generate/commandUtils.ts
ssanjay1: Done

ssanjay1:
Closing because we decided to do the logging directly in 3PA: https://github.palantir.build/foundry/third-party-application-service/pull/2504

PR #154, opened 2024-03-28
Add fetch page without result wrapper

ssanjay1:
Add a `fetchPage` method in 1.1 that doesn't wrap results, and just throws errors

Once this merges, we can merge in https://github.com/palantir/osdk-ts/pull/139 to deprecate `page` and rename `fetchPageWithErrors`

### DIFF
--- a/packages/legacy-client/src/client/net/loadObjectsPage.ts
+++ b/packages/legacy-client/src/client/net/loadObjectsPage.ts
@@ -44,29 +44,14 @@ export async function loadObjectsPage<
 ): Promise<Result<Page<T>, LoadObjectSetError>> {
   return wrapResult(
     async () => {
-      const page = await loadObjectSetV2(
-        createOpenApiRequest(client.stack, client.fetch),
-        client.ontology.metadata.ontologyApiName,
-        mapObjectSetBody(
-          objectSetDefinition,
-          orderByClauses,
-          selectedProperties,
-          options,
-        ),
-      ) as {
-        data: WireOntologyObjectV2<T["__apiName"]>[];
-        nextPageToken?: string;
-      };
-
-      return {
-        data: page.data.map(object =>
-          convertWireToOsdkObject(
-            client,
-            object,
-          ) as unknown as T
-        ),
-        nextPageToken: page.nextPageToken,
-      };
+      return loadObjectsPageOrThrows(
+        client,
+        objectApiName,
+        objectSetDefinition,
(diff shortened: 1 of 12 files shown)


packages/legacy-client/src/client/net/loadObjectsPage.ts
```diff
@@ -77,6 +62,43 @@ export async function loadObjectsPage<
   );
 }
 
+export async function loadObjectsPageOrThrows<
```
ssanjay1, line 65: Not sure how we feel about the `orThrows`, I used that while writing this up just for me to keep track of things, but happy to rename or just remove that suffix

packages/legacy-client/src/client/net/pageLinkedObjects.ts
```diff
@@ -67,3 +70,35 @@ export async function pageLinkedObjects<T extends OntologyObject>(
...
+  client: ClientContext<OntologyDefinition<any>>,
+  sourceApiName: string,
+  primaryKey: any,
+  linkTypeApiName: string,
+  options?: {
+    pageSize?: number;
+    pageToken?: string;
+  },
+): Promise<Page<T>> {
+  const response = createPageIteratorOrThrow<T>(
```
ssanjay1, line 84: Felt it was cleaner to keep as a separate function, but can easily inline as well if we want

packages/legacy-client/src/client/objectSets/OsdkObjectSet.test.ts
```diff
@@ -59,7 +60,7 @@ describe("OsdkObjectSet", () => {
   });
 
   it("exposes descriptions", () => {
-    const os = createBaseTodoObjectSet(client);
+    const os = createBaseObjectSet(client, "Todo");
```
ssanjay1, line 63: These all changed only because I made a generic helper instead

packages/legacy-client/src/client/net/loadObjectsPage.ts
ericanderson: Seems fine in 1.1 since the default is nothing throws. In v2, i would go for the opposite

packages/legacy-client/src/client/net/pageLinkedObjects.ts
ericanderson: for v1 i care less about such things

ericanderson (approved):
tiny nits

PR #156, opened 2024-03-28
Public Oauth Client for v2

ericanderson:
Features:
* Slimmed down as much as possible
* Return type is both a token provider and has functions on it
* Uses https://github.com/panva/oauth4webapi/ under the covers, which has been "certified" and is the recommended web library for javascript on https://oauth.net/code/nodejs/
* Has built in support to remove the need for things like https://github.com/palantir/osdk-ts/blob/main/examples/example-react/src/AuthenticatedRoute.tsx
  * The signIn() flow automatically happens when you try to get a token that is expired/doesn't exist.
  * You can provide a custom page to redirect to first, so you can still present a pre-login page.
  * The result is that this should now be useful even without react router.


examples-extra/todoapp/src/useTodos.tsx
```diff
@@ -1,129 +1,66 @@
...
-    throw result.error;
+import { ObjectTypeDefinition } from "@osdk/api";
+
+declare global {
+  interface ArrayConstructor {
+    /**
+     * Creates an array from an async iterator or iterable object.
+     * @param iterableOrArrayLike An async iterator or array-like object to convert to an array.
+     */
+    fromAsync<T>(
```
ssanjay1, line 14: isn't fromAsync a built in feature? Why do we need to redeclare this here?

packages/client/src/index.ts
```diff
@@ -25,6 +25,8 @@ export * as Objects from "./object/index.js";
 export type { OsdkObject } from "./OsdkObject.js";
 export type { Osdk } from "./OsdkObjectFrom.js";
 
+export type { WhereClause } from "./query/WhereClause.js";
```
ssanjay1, line 28: I think you may need to rebase because I _think_ you already exported this in a PR last week right?

packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -0,0 +1,341 @@
...
+  revocationRequest,
+  validateAuthResponse,
+} from "oauth4webapi";
+import invariant from "tiny-invariant";
+import { TypedEventTarget } from "typescript-event-target";
+import type { Events, PublicOauthClient } from "./PublicOauthClient.js";
+import { throwIfError } from "./throwIfError.js";
+import type { Token } from "./Token.js";
+
+const storageKey = "asdfasdfdhjlkajhgj";
```
ssanjay1, line 42: Is this right? Is this just supposed to be some random string for the key?

packages/client/src/ontology/loadFullObjectMetadata.ts
```diff
@@ -61,12 +61,8 @@ function makeConjureContext(
   client: MinimalClient,
   servicePath: string,
 ): ConjureContext {
-  const baseUrl = client.stack.startsWith("https://")
```
ssanjay1, line 64: Why are we able to assume that client.stack starts with https now?

ssanjay1 (approved):

examples-extra/todoapp/.env.local.sample
```diff
@@ -1,3 +1,3 @@
-VITE_FOUNDRY_URL=
-VITE_FOUNDRY_CLIENT_ID=
-VITE_FOUNDRY_CLIENT_SECRET=
\ No newline at end of file
+VITE_FOUNDRY_URL=${TS_OSDK_EXAMPLE_TODOAPP_FOUNDRY_URL}
```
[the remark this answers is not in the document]
ericanderson: I did, it allows me to keep the environment variables set in my profile so that when i have multiple checkouts, this keeps working.

examples-extra/todoapp/src/foundryClient.ts
```diff
@@ -1,23 +1,24 @@
...
+const auth = createPublicOauthClient(
+  import.meta.env.VITE_FOUNDRY_CLIENT_ID,
+  // import.meta.env.VITE_FOUNDRY_URL,
+  "http://localhost:8080",
+  "http://localhost:8080/auth/callback",
+  true,
+);
 
-export const foundryClient2 = createClient(
+export const $ = createClient(
```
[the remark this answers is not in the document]
ericanderson: This was something mfedd and I discussed doing and suggesting others do, so I am trying it out on one of our internal projects

examples-extra/todoapp/src/useTodos.tsx
ericanderson: Its not in the types yet, so while we can count on modern browsers having it and polyfills taking care of it, it was only merged a few weeks ago into ts's main branch

packages/oauth/src/createPublicOauthClient.ts
ericanderson: the value really doesn't matter :) As long as we keep using the same key all is well.

PR #157, opened 2024-04-01
Remove thin client and methods from public api

ericanderson:
We don't need to figure out how to support this workflow any time soon, so supporting the APIs would be a mistake until we do.


packages/client/src/index.ts
```diff
@@ -14,37 +14,27 @@
...
-export type { OsdkObject } from "./OsdkObject.js";
-export type { Osdk } from "./OsdkObjectFrom.js";
-
-export type { WhereClause } from "./query/WhereClause.js";
-
 export { createClientContext } from "@osdk/shared.net";
 export { isOk } from "@osdk/shared.net";
 export type { ResultOrError } from "@osdk/shared.net";
 export type { ClientContext } from "@osdk/shared.net";
-
```
ssanjay1, line 34: Nit: were these line removals intentional?

ssanjay1 (approved):

packages/client/src/index.ts
ericanderson: They were so that all the exports would sort

PR #162, opened 2024-04-03
Extract package generation

ericanderson:
For now this still requires running `(p)npm prepack` after installing deps. In a future version we can just have this skip the compile step all together which will make this slimmer and faster.

### DIFF
--- a/packages/generator/src/generateClientSdkPackage.ts
+++ b/packages/generator/src/generateClientSdkPackage.ts
@@ -1,0 +1,204 @@
+/*
+ * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


packages/generator/src/generateClientSdkPackage.ts
```diff
@@ -0,0 +1,204 @@
...
+    ],
+  };
+}
+
+async function writeJson(
+  minimalFs: MinimalFs,
+  filePath: string,
+  body: unknown,
+) {
+  // consola.info(`Writing ${filePath}`);
```
ssanjay1, line 198: nit: do we want to get rid of these?

packages/generator/src/index.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
+export { generateClientSdkPackage as __UNSTALE_generateClientSdkPackage } from "./generateClientSdkPackage";
```
ssanjay1, line 17: nit: Assuming you meant UNSTABLE?

ssanjay1 (approved):

packages/generator/src/generateClientSdkPackage.ts
ericanderson: This was only put in here for me for the hackweek we had 2 weeks ago

PR #163, opened 2024-04-04
Rename bulk actions to batch actions

ssanjay1:
Rename batch actions


packages/legacy-client/src/client/actions/actions.test.ts
```diff
@@ -116,6 +118,16 @@ describe("Actions", () => {
           ]
         >();
 
+      expectTypeOf<Parameters<typeof bulkActions.createTask>>()
```
ssanjay1, line 121: Added these just to make sure `BulkAction` types and `BatchAction` types are equivalent

packages/generator/src/v1.1/generateMetadataFile.ts
```diff
@@ -143,7 +144,9 @@ export async function generateMetadataFile(
 export interface Ontology extends ClientOntology<typeof Ontology> {
     objects: Objects;
     actions: Actions;
+    /** @deprecated use batchActions */
```
ssanjay1, line 147: Deprecated in codegen as well

packages/legacy-client/src/client/actions/createActionProxy.ts
```diff
@@ -123,3 +128,45 @@ export function createBulkActionProxy<
   ) as BulkActions<O>;
   return proxy;
 }
+
+export function createBatchActionProxy<
```
ericanderson, line 132: Is this identical to the createBulkActionProxy?

ericanderson (approved):
One c/p conern.

packages/legacy-client/src/client/actions/createActionProxy.ts
ssanjay1: Its pretty much the same, but inside of `createBulkActionProxy` it's checking for all of the Bulk types e.g. `Op extends BulkActionExecutionOptions`, `WrappedBulkActionReturnType`, etc. and returns as `BulkActions<O>` rather than `BatchActions<O>` which are all the same I guess, but I figured I should still just keep them separate

PR #166, opened 2024-04-09
Switch site commands to public API

tzyl:
Remove generated/handwritten internal API code for site commands and switch to handwritten public API code. In the future we can move this to be generated once we have the publishing set up.

I left the site command as is under `unstable` until we test this out a bit and we can remove from `unstable` in a separate small PR


packages/cli/src/net/third-party-applications/getWebsiteDeployment.mts
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    const result = await fetch(url);
+    return result.json();
+  } catch (e) {
+    // Revisit this error handling in the API
+    if (
+      e instanceof PalantirApiError
+      && e.errorName === "WebsiteDeploymentNotFound"
+    ) {
+      return undefined;
```
tzyl, line 40: Changed this to handle this error here for now as we don't want to exit in this case

packages/cli/src/commands/site/deploy/siteDeployCommand.mts
```diff
@@ -88,41 +76,36 @@ export default async function siteDeployCommand(
...
     consola.success(`Deployed ${siteVersion} successfully`);
+    const domain = deployment.subdomains[0];
     if (domain != null) {
       logSiteLink(
         "View live site:",
         `https://${domain}`,
       );
     }
   } else {
+    const deployment = await thirdPartyApplications.getWebsiteDeployment(
```
tzyl, line 104: There's one difference here where the deployment contains the domains meaning if no version is currently deployed we can no longer show the preview URL with the uploaded version

packages/cli/src/net/createFetch.mts
```diff
@@ -64,8 +64,15 @@ function handleFetchError(e: unknown): Promise<Response> {
   }
 
   let message = e.message;
-  if (e.errorName === "Artifacts:ArtifactAlreadyExists") {
-    message = "Version already exists";
+  if (e.errorName === "CannotDeleteDeployedWebsiteVersion") {
+    message = "The site version to delete is live and cannot be deleted";
+    tip = "Run the `site version set` command to change the live site version";
```
zeyadkhaled, line 69: Probably unset is easier here as it doesn't require a version passed?

zeyadkhaled (approved):
I tried all commands and it looks like nothing is broken except for what you pointed out.

On a side thought, now that we have getWebsiteDeployment we have the potential to check before deleting is the version exists or if its the live version and we can utilize that for a better UX? perhaps not showing the `Are you sure you want to delete?` and showing a specific message? 
That being said I think what we have with the Error and the tip is pretty good, so just throwing it out there.

packages/cli/src/net/createFetch.mts
tzyl: Yeah we may want to discourage people from using that command as the default though given that it impacts the live website (although we do have a confirmation prompt for the destructive change now)

PR #170, opened 2024-04-09
Fix query params based omni api calls

ericanderson:


ericanderson:
No changelog because we haven't shipped the omni change so its not needed.

packages/omniapi-generator/src/generateResource.ts
```diff
@@ -132,12 +132,18 @@ function generateOperationArray(method: StaticOperation) {
   const requestType = method.requestBody?.body.requestType;
   const bodyInfo = getReqBodyInfo(requestType);
   const respInfo = getResponseInfo(method.response);
+  const paramInfo = getParamInfo(method.parameters);
+  paramInfo.QUERY.length;
```
ssanjay1, line 136: Can we remove this line?

packages/omniapi/src/public/Datasets_Branch.ts
```diff
@@ -26,7 +26,7 @@ const _listBranches: $OmniMethod<
     datasetRid: $C.DatasetRid,
     $queryParams?: { pageSize?: $C.PageSize; pageToken?: $C.PageToken },
   ) => Promise<$C.ListBranchesResponse>
-> = [0, "/v1/datasets/{0}/branches"];
+> = [0, "/v1/datasets/{0}/branches", 2];
```
ssanjay1, line 29: I'm confused as to why these numbers have been appended to the end in these generated files?

ssanjay1 (approved):

packages/omniapi/src/public/Datasets_Branch.ts
ericanderson: They tell the runtime if there is query or header params. Its just bitwise math.

packages/omniapi-generator/src/generateResource.ts
ericanderson: OOps, yes thank you

PR #180, opened 2024-04-09
Generate package.json.hbs so dependency bumps work

ericanderson:
Whether its dependabot or manual package bumps, because the way the create-app verification works, manual intervention is required.

Instead, we now generate the `package.json.hbs` files on the fly from the committed versions in `/examples/` to avoid this problem. Now if a developer or dependabot updates a version, it should be updated in fresh versions of create-app also.

I also found it annoying that I couldn't see why things were failing in the `example-generator` package, so I integrated `jest-diff` so that you can see whats up when files are not the same.


ericanderson:
Abandoning this approach for now

PR #188, opened 2024-04-10
Change release process and automate version updates

ericanderson:
## Changes
- Restore CHANGELOG.md files from CHANGELOG_OLD.md
- Restore build scripts for changesets
- Adds check that you did in fact add changesets
- Adds snapshot builds
- Use globs for ignore in changeset config
- Remove the changelog.yml which should stop internal bot
- Migrates unreleased changes to the changeset format
- Auto updates the versions committed in code when they change

## Script to convert yml files to markdown changes (just assumes patches)
```sh

gfind -name '*.yml' -path '*/@unreleased/*' -not -path '*/node_modules/*' -exec sh -c 'echo { \"filename\": \"{}\", \"body\": &&  yq -o=json {} && echo }'  \; \
    | jq  '{
        pkg: .filename | match("packages\/(.*?)\/").captures[0].string, 
        desc: .body.[.body.type].description, 
        f: .filename | match("@unreleased/(.*?).yml").captures[0].string
      }' \
    | jq --slurp 'group_by(.f)[] | {
        f: ((.[0].f) + ".md"), 
        desc: (.[0].desc), 
        packages: [.[] | "@osdk/" + .pkg]
      }' \
    | jq -r '
        "\ncat <<EOT >> .changeset/\(.f)\n" +
        (("---\n"
        + (.packages | map("\"" + . + "\": patch") | join("\n")) 
        + "\n---\n\n" 
        + .desc) )
        + "\nEOT"
        ' | bash
```


packages/version-updater/scripts/postVersioning.mjs
```diff
@@ -0,0 +1,139 @@
...
+const workspaceDirPath = getWorkspaceDirPath();
+const clientPackageVersion = getClientPackageVersion();
+
+updateConstVariable(
+  path.join(
+    workspaceDirPath,
+    "packages",
+    "client",
+    "src",
+    "createClient.ts",
```
ericanderson, line 36: Note to self, update to Client.ts

packages/client/src/createClient.ts
```diff
@@ -118,7 +118,9 @@ function isOntologyMetadata(
   return "ontologyRid" in ontology || "ontologyApiName" in ontology;
 }
 
-const MaxOsdkVersion = "0.14.0";
+// BEGIN: THIS IS GENERATED CODE. DO NOT EDIT.
+const MaxOsdkVersion = "0.15.0";
```
ssanjay1, line 122: Woahh, this is generated

packages/version-updater/scripts/postVersioning.mjs
ssanjay1: Do you want to leave a comment in the code?

packages/version-updater/scripts/postVersioning.mjs
```diff
@@ -0,0 +1,139 @@
...
+  "ExpectedOsdkVersion",
+  clientPackageVersion,
+);
+
+/**
+ * @param {string} filePath
+ * @param {string} variableName
+ * @param {string} value
+ */
+function updateConstVariable(filePath, variableName, value) {
```
ssanjay1, line 60: nit: doesn't look like you're actually using `value` in the function, can probably just use `clientPackageVersion` directly if you want

packages/version-updater/src/junk.test.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { describe, expect, it } from "vitest";
+
+describe("anything", () => {
```
ssanjay1, line 19: Do we need this test file?

.github/workflows/ci.yml
```diff
@@ -19,7 +22,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v4
         with:
-          fetch-depth: 2
+          fetch-depth: 0
```
ssanjay1, line 25: Why did this get reduced?

ssanjay1 (approved):

packages/version-updater/scripts/postVersioning.mjs
ericanderson: Im going to update it after my other PR goes in (where you noted the duplication in createClient.ts and Client.ts)

packages/version-updater/src/junk.test.ts
ericanderson: yes the tests fail if you have no tests.

.github/workflows/ci.yml
ericanderson: It actually got increased. We need all the history to perform the check for whether or not changesets were added

packages/version-updater/scripts/postVersioning.mjs
ericanderson: Great eye. Thank you. I intended to.

PR #194, opened 2024-04-12
Make sure we publish with tags for snapshots

ericanderson:


.github/workflows/snapshot.yml
```diff
@@ -45,6 +45,6 @@ jobs:
         run: pnpm exec turbo run build
       
       - name: Publish results
-        run: pnpm exec changeset publish --no-git-tag --snapshot ${GITHUB_REF_NAME//\//__}
+        run: pnpm exec changeset publish --no-git-tag --tag next-${GITHUB_REF_NAME//\//__}
```
ericanderson, line 48: I'm doing two things here. 
1.  We need to use `--tag` instead of `--snapshot` for publish.
2. On the off chance that `$GITHUB_REF_NAME` is empty/not defined, we still end up publishing as `next-` so we cannot accidentally publish as `latest`. This will have the effect of the next branch publishing as `next-next` but i think thats a reasonable compromise.

PR #209, opened 2024-04-19
Cut major version of foundry sdk generator

ssanjay1:
Add change logs to cut major version of foundry sdk generator


ssanjay1:
Closing since we merged https://github.com/palantir/osdk-ts/pull/208 that doesn't bump legacy client

PR #214, opened 2024-04-23
Implement subscriptions per the v1 api behavior

ericanderson:
- Supports multiple subscriptions now
- Fixes issues with where clauses for equality
- Fixes issues where apiname != id_name in OSS.

Also:
- Introduced an optional argument to pass a Pino logger to the client.
  - Tip: You can get nice output like so `node build/js/index.mjs | npx pino-pretty`


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -72,74 +105,107 @@ export class ObjectSetListenerWebsocket {
...
-    const requestId = crypto.randomUUID();
-    const expiry = setTimeout(() => {
-      this.#expire(requestId);
-    }, ONE_DAY_MS);
-    this.#listeners.set(requestId, { listener, objectSet, expiry });
-    this.#subscribe(requestId, objectSet);
+  ): Promise<() => void> {
+    globalThis.crypto ??= (await import("node:crypto")).webcrypto as any;
+    const sub: Subscription = {
+      listener: fillOutListener<any>(listener),
```
styu, line 165: why is this `any` vs `Q` ?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -72,74 +105,107 @@ export class ObjectSetListenerWebsocket {
...
+   * - Recreates temporary object set
+   * - Triggers a full subscribe message
+   *
+   * @returns
+   */
+  async #initiateSubscribe(sub: Subscription) {
+    if (sub.expiry) {
+      clearTimeout(sub.expiry);
+    }
+    sub.expiry = setTimeout(() => this.#expire(sub), ONE_DAY_MS);
```
styu, line 197: should the expiration time be configurable? (with this as a fallback)

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -151,59 +217,87 @@ export class ObjectSetListenerWebsocket {
...
+  }
+
+  #sendSubscribeMessage() {
+    // If two calls to `.subscribe()` happen at once (or if the connection is reset),
+    // we may have multiple subscriptions that don't have a subscriptionId yet,
+    // so we filter those out.
+    const readySubs = [...this.#subscriptions.values()].filter(isReady);
+    if (readySubs.length === 0) return;
+
+    const id = crypto.randomUUID();
```
styu, line 241: why do you do the fallback  in `subscribe` but not here?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -151,59 +217,87 @@ export class ObjectSetListenerWebsocket {
...
+    // responses come back as an array of subIds, so we need to know the sources
+    this.#pendingSubscriptions.set(id, readySubs);
+
+    // every subscribe message "overwrites" the previous ones that are not
+    // re-included, we we have to reconsistute the entire list of subscriptions
```
styu, line 246:
```suggestion
    // re-included, so we have to reconstitute the entire list of subscriptions
```

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -151,59 +217,87 @@ export class ObjectSetListenerWebsocket {
...
+    // the websocket connection in that case.
+    if (this.#maybeDisconnectTimeout) {
+      clearTimeout(this.#maybeDisconnectTimeout);
     }
+    this.#maybeDisconnectTimeout = setTimeout(() => {
+      this.#maybeDisconnectTimeout = undefined;
+      if (this.#subscriptions.size === 0) {
+        this.#cycleWebsocket();
+      }
+    }, 15_000 /* ms */);
```
styu, line 300: how did you determine 15 seconds?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -151,59 +217,87 @@ export class ObjectSetListenerWebsocket {
...
-    } else {
-      // TODO backend does not yet have an unsubscribe message payload
+    this.#subscriptions.delete(sub.subscriptionId);
+
+    // If we have no more subscriptions, we can disconnect the websocket
+    // however we should wait a bit to see if we get any more subscriptions.
+    // For example, when switching between react views, you may unsubscribe
+    // in the old view and subscribe in the new view. We don't need to re-establish
+    // the websocket connection in that case.
+    if (this.#maybeDisconnectTimeout) {
```
styu, line 292: if someone unsubscribed twice, is it bad that this timeout "resets" the subsequent times unsubscribe was called?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Because it's going into a generic storage for all subscriptions. The type information is effectively erased at that point and the use of generics here is to make it easier for the consumer not us.

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: The temporary object set we create is set to 1 day. In reality this would be hidden by gateway but we have to deal in these directly for now.

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: There are two types of ids at play. A requestId for initiating the subscription, chosen by the consumer of the api. The result is tied to this value. And gives an array of the second type: subscriptionId. There is one of these for each subscription entry but the server controls them.

styu: i understand the different IDs, but im asking about why you fallback to node crypto in the `subscribe`function  but not here when you call `crypto`

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Since my use case was the concept of changing views I figured most page activity from transitions would settle after a few seconds. I gave fifteen for buffer.

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Suppose you have one sub and remove it. 10s elapses. Then a new sub starts. After 4s we unsub. If we keep the original timer then after only 1s of no subs the connection is closed, instead of the 15s buffer I was trying to allow.

ericanderson (commented):
Responses

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Oh! Because it assigns to globalThis. It sure to be set after that as thats the public entry point and you need to call it before this method is ever called. Will add a comment

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
styu: comment?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
styu: ah ok - comment? :)

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Ah, nothing bad will happen. If the backend wants to protect against idle websocket connections they should handle that separetly. I'll guard too.

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Will be in next update

styu (approved):

PR #216, opened 2024-04-24
Add batch actions for 2.0 synatx

ssanjay1:
Add batch action support to 2.0


packages/client/src/actions/applyAction.ts
```diff
@@ -16,58 +16,93 @@
...
+      ? response.edits
+      : undefined) as ActionReturnTypeForOptions<Op>;
+  } else {
+    const response = await applyActionV2(
+      addUserAgent(client, action),
+      client.ontologyRid,
+      action.apiName,
+      {
+        parameters: remapActionParams(
+          parameters as OsdkActionParameters<AD["parameters"]>,
```
ssanjay1, line 83: So I had to cast here because even though I did the `Array.isArray` check above, it wasn't able to tell that this would be the non array type I defined above.

ericanderson (approved):

PR #246, opened 2024-04-29
WIP: Modifying $ property names

ssanjay1:
Modified properties so that any built in OSDK property name is prefixed with $, look at osdkExample.tsx for detailed syntax change examples. This change should also make discoverability easier, as users can just type in $ and see what properties are available for a certain function

image


ericanderson (approved):
Approving although we might want to support both $ and not while we migrate our internal alpha uses.

ssanjay1:
> Approving although we might want to support both $ and not while we migrate our internal alpha uses.

Talked offline, we will be just merging as is and owning the migration of internal alpha uses on our end

PR #249, opened 2024-04-30
Update site APIs

tzyl:
Some renames to endpoints and objects and the ability to load subdomains back even if no deployed version set


zeyadkhaled (approved):
Tested the CLI and all functionality is working + Changes look good!

PR #253, opened 2024-04-30
Add fetchOne

ssanjay1:
Adds fetchOne which allows you to get a single object, without a Result wrapper

### DIFF
--- a/packages/legacy-client/src/client/net/getOnlyLinkedObject.ts
+++ b/packages/legacy-client/src/client/net/getOnlyLinkedObject.ts
@@ -15,11 +15,14 @@
  */
 
 import type { OntologyDefinition } from "@osdk/api";
-import type { ClientContext } from "@osdk/shared.net";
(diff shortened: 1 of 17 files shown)


packages/legacy-client/src/client/net/getOnlyLinkedObject.ts
```diff
@@ -79,3 +82,26 @@ export async function getOnlyLinkedObject<
...
+  sourcePrimaryKey: NonNullable<ParameterValue>,
+  targetLinkType: string,
+): Promise<T> {
+  const result = await listLinkedObjectsWithoutErrors<T>(
+    client,
+    sourceObjectType,
+    sourcePrimaryKey,
+    targetLinkType,
+  );
+  if (result.length !== 1) {
```
ssanjay1, line 100: Pulled this error throwing from how we handle fetching single objects in 2.0

ericanderson (approved):

PR #273, opened 2024-05-03
Generate omni api v2

ericanderson:
**Advice for review:**
I broke this down into separate commits:
1. The first is just moving `@osdk/foundry` -> `@osdk/internal.foundry` and updating related changes.
2. The second is all the new code generation
3. The third is actually generating the code (`@osdk/foundry`, `@osdk/foundry.security`, `@osdk/foundry.thirdpartyapplications`)
4. A quick example to be sure it works

The 'v2' style now generates one package per Omni namespace and each exports esm namespaces with methods to be used via the foundry sdk.


ericanderson:
No changeset, we aren't really releasing anything here.

packages/foundry-generator/src/generateResource.ts
```diff
@@ -288,7 +273,3 @@ function getResponseInfo(response: Response) {
     return { componentType: "void", mimeType: "" };
   }
 }
-
-function quoteMimeTypeOrEmpty(s: string) {
```
ericanderson, line 292: Moved

packages/foundry-generator/src/generateResource.ts
```diff
@@ -183,13 +175,6 @@ interface PayloadInfo {
   componentType: string | undefined;
 }
 
-function getCleanedUpJsdoc(doc?: Documentation) {
```
ericanderson, line 186: Moved

packages/foundry-generator/src/generateResource.ts
```diff
@@ -77,15 +78,6 @@ function generateMethods(resource: Resource) {
   return out;
 }
 
-function generateMethodJsdoc(method: StaticOperation) {
```
ericanderson, line 80: Moved

packages/foundry-generator/src/generateResource2.ts
```diff
@@ -0,0 +1,143 @@
...
+import type { Resource } from "./model/Resource.js";
+import type { StaticOperation } from "./model/StaticOperation.js";
+import { addAll } from "./util/addAll.js";
+import { writeCode } from "./writeCode.js";
+
+export async function writeResource2(
+  resource: Resource,
+  filePath: string,
+  model: Model,
+) {
```
ericanderson, line 33: This file is alot like `generateResources` except it uses this new `model` which makes it easier to manage cross package dependencies.

PR #280, opened 2024-05-06
Regenerate conjure

ericanderson:
- Makes lint-staged ugly on commit (but not fail vscode)
- Upgrades to latest conjure-lite
- Fixes/improves scripts, including resetting files that have only changed by copyright year.
- Regenerates.

No changelog, its not user facing.


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+TERM=dumb npx lint-staged
```
ssanjay1, line 4: I may not have searched google hard enough, but it seems like this just disables color output right? Does that really choke on big changes?

scripts/generateConjure.sh
```diff
@@ -25,13 +25,33 @@ function generateConjure() {
...
 generateConjure "com.palantir.ontology" "ontology-metadata-api" "${SCRIPT_DIR}/../packages/client"
+
+# Reset git changes if the generated files are only changed by copyright year
+git status --porcelain --untracked-files=no | while read line; do
+    status=$(echo "$line" | awk '{print $1 }')
+    file=$(echo "$line" | awk '{print $2 }')
+
+    if [[ "$status" == "D" ]]; then
+        # do nothing
+        bnitgub="da"
```
ssanjay1, line 44: where do you come up with these names lol

scripts/generateConjure.sh
```diff
@@ -25,13 +25,33 @@ function generateConjure() {
...
+    else 
+        v=$(git diff -U0 $file | tail +5 | md5sum)
+        expected="ce0376e23ce2aa3d0dcd56c2befc2993  -"
```
ssanjay1, line 47: So this is the hash we'd expect if only the copyright year changes?

ssanjay1 (approved):

scripts/generateConjure.sh
ericanderson: Yes

scripts/generateConjure.sh
ericanderson: I needed something for the if statement so bash wouldnt be mad.

.husky/pre-commit
ericanderson: It disables the ansi control code which seems to upset vscode

PR #286, opened 2024-05-07
Always include trailing slash for registry urls in .npmrc

tzyl:
The sample instructions suggest `npm install` however when using other package managers such as `pnpm install` pnpm does not correctly pick up the registry url for auth unless there is a trailing slash

### DIFF
--- a/packages/create-app/src/generate/generateNpmRc.ts
+++ b/packages/create-app/src/generate/generateNpmRc.ts
@@ -21,7 +21,12 @@ export function generateNpmRc({
   osdkPackage: string;
   osdkRegistryUrl: string;
 }): string {
-  const withoutProtocol = osdkRegistryUrl.replace(/^https:\/\//, "");
+  // pnpm requires a trailing slash in .npmrc
+  // https://github.com/pnpm/pnpm/issues/5941
+  const withTrailingSlash = osdkRegistryUrl.endsWith("/")
+    ? osdkRegistryUrl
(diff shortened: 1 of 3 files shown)


packages/create-app/src/generate/generateNpmRc.ts
```diff
@@ -21,7 +21,12 @@ export function generateNpmRc({
...
-  const withoutProtocol = osdkRegistryUrl.replace(/^https:\/\//, "");
+  // pnpm requires a trailing slash in .npmrc
+  // https://github.com/pnpm/pnpm/issues/5941
+  const withTrailingSlash = osdkRegistryUrl.endsWith("/")
+    ? osdkRegistryUrl
+    : osdkRegistryUrl + "/";
+  const withoutProtocol = withTrailingSlash.replace(/^https:\/\//, "");
   return `//${withoutProtocol}:_authToken=\${FOUNDRY_TOKEN}\n`
-    + `${osdkPackage.split("/")[0]}:registry=${osdkRegistryUrl}\n`;
+    + `${osdkPackage.split("/")[0]}:registry=${withTrailingSlash}\n`;
```
zeyadkhaled, line 31: minor nit: I know this existed from before, but we can replace this with a descriptive name to fit with the rest

```
const packageName = osdkPackage.split("/")[0];
```
```suggestion
    + `${packageName}:registry=${withTrailingSlash}\n`;
```

zeyadkhaled (approved):

PR #288, opened 2024-05-07
Add fetchOneWithErrors to 2.0

ssanjay1:
Add fetchOneWithErrors, which functions the same as fetchOne, but adds a Result wrapper

### DIFF
--- a/packages/client/src/object/fetchSingle.ts
+++ b/packages/client/src/object/fetchSingle.ts
@@ -24,6 +24,7 @@ import {
   type FetchPageArgs,
   type SelectArgToKeys,
 } from "./fetchPage.js";
+import type { Result } from "./Result.js";
 
 export async function fetchSingle<
   Q extends ObjectOrInterfaceDefinition,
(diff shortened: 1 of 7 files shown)


packages/client/src/object/fetchSingle.ts
```diff
@@ -57,3 +58,31 @@ export async function fetchSingle<
...
+  >
+> {
+  try {
+    const result = await fetchSingle(client, objectType, args, objectSet);
+    return { value: result as any };
+  } catch (e) {
+    if (e instanceof Error) {
+      return { error: e };
+    }
+    return { error: e as Error };
```
ericanderson, line 86: Is this right?

ericanderson: Should we have a wrapper function that can take a call and create a result based on if something throws?

ssanjay1: Hmm I'm confused, I was just emulating what we did for like `fetchPageWithErrors` https://github.com/palantir/osdk-ts/blob/d183d92b3fe0b2fa8016ea94d44eebc1e21fa2fc/packages/client/src/object/fetchPage.ts#L229

ericanderson: Okay, lets go with it for now.

ericanderson (approved):

PR #292, opened 2024-05-08
rename user agents

ssanjay1:
Renaming user agents to be uniform with python user agent


ericanderson (commented):
Just want to confirm we want this to affect legacy-client now?

ssanjay1:
> Just want to confirm we want this to affect legacy-client now?

Yea pretty sure we do, because this will be to help with collecting telemetry

ericanderson (approved):

PR #297, opened 2024-05-08
Batch link support

ericanderson:
```ts
const { data: employees } = await client(Employee).fetchPage({
  pageSize: 100,
});

logger.debug("Fetching the bulk links");

for await (
  const {
    object,
    linkApiName,
    otherObjectApiName,
    otherObjectPk,
  } of client.__UNSTABLE_getBulkLinks(employees, [
    "ventures",
    "amishsSyncGroup",
  ])
) {
  logger.info(
    `Found link ${object.$objectType}:${object.$primaryKey} <- (${linkApiName}) -> ${otherObjectApiName}:${otherObjectPk}`,
  );
}
```


packages/client/src/createClient.ts
```diff
@@ -71,10 +70,11 @@ export function createClientInternal(
   }
 
   const client: Client = Object.defineProperties<Client>(
-    clientFn as UnstableClient,
+    clientFn as Client,
     {
       ctx: {
         value: clientCtx,
+        writable: false,
```
ssanjay1, line 77: Why did we need to add this?

packages/client/src/ontology/OntologyProvider.ts
```diff
@@ -32,7 +32,7 @@ export interface OntologyProvider {
    */
   getObjectDefinition: (
     apiName: string,
-  ) => Promise<ObjectTypeDefinition<any>>;
+  ) => Promise<ObjectTypeDefinition<any> & { rid: string }>;
```
ssanjay1, line 35: Just double checking, we're only adding rid here for internal use when we want to get object defs for batch links right?

ssanjay1 (approved):
Approved, left some comments/questions

packages/client/src/__unstable/createBulkLinksAsyncIterFactory.ts
```diff
@@ -0,0 +1,190 @@
...
+        ontologyBranchRid: undefined,
+        owningRid: undefined,
+        reportUsage: undefined,
+        workstateRid: undefined,
+      },
+      responseOptions: {
+        includeObjectSetEntities: true,
+        includeUsageCost: false,
+      },
+      pageSize: 1000,
```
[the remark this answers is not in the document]
ericanderson: Im not sure. We can iterate on this part later. We need feedback on this anyway.

packages/client/src/createClient.ts
ericanderson: I was trying to protect caches a bit better but I think this is superseeded by other changes I made for the omniapi anyway.

packages/client/src/ontology/OntologyProvider.ts
ericanderson: Correct. Thats why I didn't add it to the actual `ObjectTypeDefinition`, especially as thats in the api package and shared with `legacy-client`. We should not need this in an ideal world.

ericanderson (commented):
Notes

PR #320, opened 2024-05-13
Foundry SDK Improvements

ericanderson:

PR #321, opened 2024-05-13
Use the new omni apis for client

ericanderson:
Blocks on #320

Broken into two commits to make review easier. Its a pretty small change over all but it does generate a lot of code.


packages/platform-sdk-generator/src/model/Model.ts
```diff
@@ -80,8 +82,11 @@ export class Model {
           mapObjectValues(dt.object.properties, p => this.getType(p.type)),
         );
       case "asyncOperation":
-      case "binary":
+        return new BuiltinType({ type: "any", any: {} });
         throw new Error("Method not implemented.");
```
ssanjay1, line 86: Do you still want this throw here?

packages/platform-sdk-generator/src/model/Resource.ts
```diff
@@ -18,6 +18,6 @@ import type { Component } from "./Component.js";
 import type { StaticOperation } from "./StaticOperation.js";
 
 export interface Resource {
-  component: Component;
+  component: string;
```
ssanjay1, line 21: Why is this now typed as a string?

ssanjay1 (approved):
Approved, couple of quick questions

packages/platform-sdk-generator/src/model/Resource.ts
ericanderson: You arent required to have a real component it turns out. I was only using this for the name anyway.

PR #331, opened 2024-05-20
Update tutorial template to not use any OSDK types

zeyadkhaled:
- Removed all references to concrete types from the OSDK
- Removed previously added types in the one.dot.one package 
- Updated all type definitions to use the Mock types

I tested an app generated through the template
- The initial mock types work without loading anything from the Ontology
- Once updated with all the snippets from the Lessons, the application flow works as expected and all actions are applied correctly


packages/create-app/templates/template-tutorial-todo-app/src/TaskList.tsx.hbs
```diff
@@ -1,10 +1,10 @@
-import { OsdkTodoProject } from "{{osdkPackage}}/ontology/objects";
```
tzyl, line 1: For all of these files I think we can remove the `.hbs` extension as they no longer need templating (it doesn't make a big difference but maybe makes it clearer that there isn't anything special about these files)

packages/create-app/templates/template-tutorial-todo-app/src/mocks.ts
```diff
@@ -0,0 +1,147 @@
...
+  await delay();
+  for (const project of projects) {
+    const idx = project.tasks.findIndex((t) => t.id === id);
+    if (idx !== -1) {
+      project.tasks.splice(idx, 1);
+    }
+  }
+}
+
+function project(mockProject: MockProject): MockProject {
```
tzyl, line 129: Does this function do anything anymore?

tzyl (approved):

packages/create-app/templates/template-tutorial-todo-app/src/mocks.ts
zeyadkhaled: Yeah it is no longer needed 👍 Good catch

PR #340, opened 2024-05-21
Update global layout styles

tzyl:
This PR moves some global layout styles that are applied on the body element to another intermediate container element. There shouldn't be any effect on the visual impact but this makes the templates work better with other layouts such as adding a banner at the top of the page in the body where before the row flex with center item placement would mess things up.

### DIFF
--- a/examples/example-vue/src/Home.vue
+++ b/examples/example-vue/src/Home.vue
@@ -14,21 +14,15 @@ const queryApiNames = Object.keys(client.ontology.queries);
   <div class="methods">
     <div>
       <h2>Objects ({{ objectApiNames.length }})</h2>
-      <pre v-for="objectApiName in objectApiNames" key="objectApiName">
-        client.ontology.objects.{{ objectApiName }}
-      </pre>
+      <pre v-for="objectApiName in objectApiNames" key="objectApiName">client.ontology.objects.{{ objectApiName }}</pre>
     </div>
     <div>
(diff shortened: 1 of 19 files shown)


examples/example-vue/src/Home.vue
```diff
@@ -14,21 +14,15 @@ const queryApiNames = Object.keys(client.ontology.queries);
   <div class="methods">
     <div>
       <h2>Objects ({{ objectApiNames.length }})</h2>
-      <pre v-for="objectApiName in objectApiNames" key="objectApiName">
-        client.ontology.objects.{{ objectApiName }}
-      </pre>
+      <pre v-for="objectApiName in objectApiNames" key="objectApiName">client.ontology.objects.{{ objectApiName }}</pre>
```
tzyl, line 17: The whitespace here is meaningful for vue as this is plain HTML rather than JSX where the leading and trailing whitespace in the tag would normally be stripped out

zeyadkhaled, line 17: Nothing changed here I think, just eslint doing its thing?

zeyadkhaled (approved):

examples/example-vue/src/Home.vue
zeyadkhaled: Disregard as you answered [here](https://github.com/palantir/osdk-ts/pull/340#discussion_r1608510972)

PR #351, opened 2024-05-28
Add helper for async iter

ssanjay1:
Adding a helper function so users can do an `all()` equivalent with `asyncIter()`


ssanjay1:
Closing as we will just make a docs change

PR #352, opened 2024-05-28
Add helper for asyncIter

ssanjay1:
Add helper to asyncIter so users can emulate `all()`


ssanjay1:
Closing, as we will just make a docs change

PR #353, opened 2024-05-28
Prepare for ts 5.5 by adding return definitions

ericanderson:


packages/shared.net.fetch/src/createFetchHeaderMutator.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 export function createFetchHeaderMutator(
-  fetchFn: typeof fetch = fetch,
+  fetchFn: typeof fetch | undefined = fetch,
```
ssanjay1, line 18: Why the `| undefined` here?

packages/shared.test/src/mock-ontology/mockMetadata.ts
```diff
@@ -15,4 +15,5 @@
  */
 
 export const MOCK_ORIGIN = "https://mock.com";
-export const MOCK_BASE_URL = `${MOCK_ORIGIN}/api/v2/ontologies/`;
+export const MOCK_BASE_URL: `${typeof MOCK_ORIGIN}/api/v2/ontologies/` =
```
ssanjay1, line 18: I thought we don't need to specify types on literal types for isolatedDeclarations?

ssanjay1: Oh wait, is it because its a template literal type?

ssanjay1 (approved):
Approved, left a couple of questions

packages/shared.test/src/mock-ontology/mockMetadata.ts
ericanderson: Yeah, its being strict here. I think stricter than necessary. I don't need the type checker to generate: 
```ts
export const MOCK_ORIGIN: "https://mock.com";
export const MOCK_BASE_URL: `${MOCK_ORIGIN}/api/v2/ontologies/`;
```

packages/maker/src/api/defineOntology.ts
```diff
@@ -26,7 +26,24 @@ export let namespace: string;
 export async function defineOntology(
   ns: string,
   body: () => void | Promise<void>,
-) {
+): Promise<
+  {
```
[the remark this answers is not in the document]
ericanderson: Probably but since its just my hack week project im not terribly concerned.

Issue #266, opened 2024-05-02
Confirm GH token discovery as part of createReleasePr


ericanderson:
Right now it says trying but then moves on. If a later failure happens its not clear if it was due to getting the token or not. Outputting that we got the token successfully would help.

PR #354, opened 2024-05-28
Log GitHub token being found in createReleasePr

bryantpark04:
Also replace "github cli" with "GitHub CLI" in log messages.

This PR addresses #266.


ssanjay1 (approved):
Yay!

bryantpark04:
Considered both only logging if the token was not found in env and the "trying" message was logged, or always logging, and decided to always log after discussing with Sanjay.

ssanjay1 (approved):

PR #366, opened 2024-05-31
Experimental subscription no longer fires outOfDate on first sub

### LINKED ISSUE #348: Prevent initial onOutOfDate in experimental subscription
Notes:

The experimental v1 subscribe api (and possibly future/final APIs) sends all subscriptions when you need to modify/add/remove any of them. And the v1  experimental also requires us to do a few book keeping network calls before we can actually subscribe. 

So you can end up in the following situation:

```ts
os1.subscribe(...);
// soon after
os2.subscribe(...)
```

The first subscription will come back successful and we can chose to not fire the out of date callback. However right after that, another subscribe request will come back. Now we have a problem. If this subscribe came back because the earlier ObjectSet expired and so we had to resubscribe, then we definitely need to fire the out of date handler. But if it came back because we just added a subscription, we don't want to do anything. Finally, we need to handle the case that we may have previously had a successful subscription but that we had to reconnect the websocket.

ericanderson:
Fixes #348

Also improves some subtle behavior quirks discovered when writing tests.

### DIFF
--- a/packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
+++ b/packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
@@ -425,7 +425,7 @@ function createMockWebSocketConstructor(
       }),
       close: vi.fn(),
       _eventEmitter: eventEmitter,
-      readyState: WebSocket.CONNECTING,
+      readyState: OriginalWebSocket.CONNECTING,
     } satisfies RawWebSocketPlus as any as MockedWebSocket;
     // ^ we only implement some things but the type system wants to think its the full deal,
     // thus the satisfies plus the cast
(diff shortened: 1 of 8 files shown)


packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -0,0 +1,494 @@
...
+
+        vi.runAllTicks();
+
+        // Make sure we cleaned up
+        expectEqualRemoveAndAddListeners(ws);
+      });
+
+      it("only sends a single request", async () => {
+        respondSuccessToSubscribe(ws, subReq1);
+        // actually this is broken FIXME
```
ssanjay1, line 196: What exactly is broken here?

ssanjay1 (approved):
Looks good to me, left one question

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
ericanderson: There are scenarios where you can get multiple send requests, I just didnt mock the timing of them all out.

PR #367, opened 2024-05-31
Attachments work V2

ssanjay1:
Added attachment functionality to 2.0.

**Attachment Reading**
```
const attachmentContent = await client(object1.attachment).fetchContents();
const attachmentText = await attachmentContent.text();
```

**Attachment Metadata Reading**
` const attachmentMetadata = await client(object1.attachment).fetchMetadata();`

**Attachment Uploading**
_For now, you can only upload attachments by calling an action_
```
const attachment = new AttachmentUpload(blob, "file1.txt");
await client(actionTakesAttachment)({
      attachment,
    });
```


packages/client/src/util/toDataValue.ts
```diff
@@ -25,25 +31,48 @@ import { isWireObjectSet } from "./WireObjectSet.js";
...
   }
 
   // attachments just send the rid directly
   if (isAttachment(value)) {
     return value.rid;
   }
 
+  // For uploads, we need to upload ourselves first to get the RID of the attachment
+  if (isAttachmentUpload(value)) {
+    const attachment = await Ontologies.Attachments.uploadAttachment(
```
ssanjay1, line 59: This seemed like the best place to put this to follow the existing pattern. Only downside is had to modify everything else in this to be async, and nothing else really needs the client context.

packages/client/src/util/toDataValue.ts
```diff
@@ -25,25 +31,48 @@ import { isWireObjectSet } from "./WireObjectSet.js";
...
+      client,
+      value.data,
+      {
+        filename: value.fileName,
+      },
+      {
+        "Content-Length": value.data.size.toString(),
+        "Content-Type": value.data.type,
```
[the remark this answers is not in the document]
ssanjay1: Mm yea good call, lemme do that

packages/client/src/object/Attachment.ts
```diff
@@ -14,12 +14,23 @@
  * limitations under the License.
  */
 
-import type { Attachment as IAttachment } from "@osdk/client.api";
+import type {
+  Attachment as IAttachment,
+  AttachmentUpload as IAttachmentUpload,
+} from "@osdk/client.api";
 
 export class Attachment implements IAttachment {
```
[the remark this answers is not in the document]
ssanjay1: Yea I was trying to get more SA on why we did this as well, found this old PR you discussed this on: https://github.com/palantir/osdk-ts/pull/21. I just followed the pattern of what we had so far for the additions I made, it seems like it was mainly for the data conversions because just checking for `{rid: string}` was risky

packages/client/src/object/Attachment.ts
```diff
@@ -14,12 +14,23 @@
...
   constructor(public rid: string) {}
 }
 
+export class AttachmentUpload implements IAttachmentUpload {
+  constructor(public data: Blob, public fileName: string) {}
```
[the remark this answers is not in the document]
ssanjay1: Ah these are good points thanks for flagging that, I will try reworking

packages/client/src/util/toDataValue.ts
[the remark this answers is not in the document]
ssanjay1: Will address this in a FLUP

packages/client/src/object/attachment.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  it("reads attachment metadata succesfully", async () => {
+    const result = await client(
+      MockOntology.objects.objectTypeWithAllPropertyTypes,
+    )
+      .where({ id: stubData.objectWithAllPropertyTypes1.id }).fetchPage();
+
+    const object1 = result.data[0];
+    expect(object1.attachment).toBeDefined();
+    const attachmentMetadata = await client(object1.attachment!).readMetadata();
```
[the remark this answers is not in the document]
ssanjay1: Hmm potentially, I've kept as is for now, but if we do that follow up, could be worth adding. I feel like what's nice about this syntax is it makes it slightly more clear we're doing another network call with the client?

packages/client.api/src/object/Attachment.ts
```diff
@@ -17,3 +17,19 @@
 export interface Attachment {
   rid: string;
 }
+
+export interface AttachmentUpload extends Blob {
```
ericanderson, line 21: JSDoc that mentions that the `File` object from the W3C FileApi https://www.w3.org/TR/FileAPI/#file-section satisifies this?

packages/client/src/actions/applyAction.ts
```diff
@@ -93,31 +100,35 @@ export async function applyAction<
...
-    acc[key] = toDataValue(value);
-    return acc;
-  }, parameterMap);
+  const remappedParams = Object.entries(params).reduce(
+    async (promisedAcc, [key, value]) => {
+      const acc = await promisedAcc;
+      acc[key] = await toDataValue(value, client);
+      return acc;
+    },
+    Promise.resolve(parameterMap),
```
ericanderson, line 118: This is a fascinating trick that might be too clever. Compare to straight forward:
```ts
const parameterMap: { [parameterName: string]: unknown } = {};
for (const key of Object.keys(params)) {
  parameterMap[key] = await toDataValue(value, client);
}
```

packages/client/src/createAttachmentReader.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import type {
+  Attachment,
+  AttachmentMetadata,
+  AttachmentSignature,
+} from "@osdk/client.api";
+import { Ontologies } from "@osdk/internal.foundry";
+import type { MinimalClient } from "./MinimalClientContext.js";
+
+export function createAttachmentReader(
```
ericanderson, line 25: At a future date we may want to make an internal-only class that implements this or a prototype so we don't produce a ton of churn on "hidden classes".

packages/client/src/object/Attachment.ts
```diff
@@ -23,3 +26,17 @@ export class Attachment implements IAttachment {
...
+}
+
+export function createAttachmentUpload(
+  data: Blob,
+  name: string,
+): IAttachmentUpload {
+  const attachmentUpload = Object.create(data);
+  Object.assign(attachmentUpload, { name });
+
+  return attachmentUpload as IAttachmentUpload;
```
ericanderson, line 41: This could probably just be `return Object.create(data, { name: {value: name} });` or maybe slightly more friendly: `return Object.create(data, { name: {value: name, enumerable: true} });`

The enumerable isn't needed for `in`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties

packages/client/src/util/toDataValue.ts
```diff
@@ -25,25 +31,48 @@ import { isWireObjectSet } from "./WireObjectSet.js";
  *
  * @see DataValue for the expected payloads
  */
-export function toDataValue(value: any): DataValue {
+export async function toDataValue(
+  value: any,
```
ericanderson, line 35: nit: `unknown` instead of any in a place like this to prevent mistakes.

ericanderson (approved):
Non blocking comments

PR #368, opened 2024-06-03
Add support for binary type body inputs with Omni API

ssanjay1:
Modified codegen so that we add support for binary type body inputs. We accept blobs for now

Split into two commits for ease of review:
1. Modifying generation
2. The actual codegen


packages/internal.foundry.core/src/_components.ts
```diff
@@ -100,12 +100,7 @@ export interface LteQueryV2 {
 /**
  * Log Safety: UNSAFE
  */
-export interface DeleteLinkRule {
-  linkTypeApiNameAtoB: LinkTypeApiName;
-  linkTypeApiNameBtoA: LinkTypeApiName;
-  aSideObjectTypeApiName: ObjectTypeApiName;
-  bSideObjectTypeApiName: ObjectTypeApiName;
-}
+export type ObjectPrimaryKey = Record<PropertyApiName, PropertyValue>;
```
ssanjay1, line 103: I did not do anything to modify this, or some of the other items below, assuming the spec we generate from was just modified?

ericanderson (approved):

PR #372, opened 2024-06-04
Fix $rid access issue

ssanjay1:
We had a bug in our Osdk<> type where if you requested a rid along with other fields, the rid would not be populated on the object type.

### DIFF
--- a/packages/client/src/OsdkObjectFrom.ts
+++ b/packages/client/src/OsdkObjectFrom.ts
@@ -123,7 +123,8 @@ export type Osdk<
   }
   // We are hiding the $rid field if it wasnt requested as we want to discourage its use
   & (IsNever<P> extends true ? {}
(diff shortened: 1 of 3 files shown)


packages/client/src/OsdkObjectFrom.ts
```diff
@@ -123,7 +123,8 @@ export type Osdk<
   }
   // We are hiding the $rid field if it wasnt requested as we want to discourage its use
   & (IsNever<P> extends true ? {}
-    : P extends "$rid" ? { $rid: string }
+    : string extends P ? {}
```
ssanjay1, line 126: This is not as permissive, so passing in `any` will not give you access to the `rid` property unless you explicitly ask for it. This matches up with the behavior with our existing type tests which is why I kept it like this for now.

ericanderson (approved):

PR #378, opened 2024-06-05
Interfaces are synthetic.

### LINKED ISSUE #304: Convert $as objects to virtual objects
Roughly:

```ts
const someObj = (await client(ThingObject).fetchPage()).data[0];
const someObjAsFoo = someObj.$as(FooInterface);

someObj.$as(Foo) === someObjAsFoo;
someObj === someObjAsFoo.$as(ThingObject);
```

Goal is to make it possible for updates (optimistic or not) to update the same underlying object. For today, we don't really let you update these things so it does not matter but keeping that option available is important.

ericanderson:
Tip to reviewer: I had to change all package.json to make this work. I broke this into two commits to make it easier on you.

## Interfaces are synthetic
...AND
- Removed the overly complex ontology cache skipping logic.
- Removes code for doing "full ontology load" from the server as it has not been needed for awhile
- Introduces a way to update the underlying values of an object.
- Adds a pretty console.log() if you're on nodejs.
  image

Example of the synthetic interfaces and their connection to objects:

```ts
expect(obj.fullName).toEqual("Steve");
const objAsFoo = obj.$as(FooInterface);
(obj as any).$updateInternalValues({
  fullName: "Bob",
});
expect(obj.fullName).toEqual("Bob");
expect(objAsFoo.fooSpt).toEqual("Bob");


expect(obj).toBe(objAsFoo.$as(Employee));
expect(objAsFoo).toBe(obj.$as(FooInterface));
```

Fixes #304

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
@@ -71,17 +71,22 @@ export function createOsdkObject<
   objectDef: Q,
   rawObj: OntologyObjectV2,
 ) {
-  const kk = Object.create(objectPrototypeCache.get(client, objectDef), {
-    [RawObject]: {
-      value: rawObj,
-      writable: true, // so we can allow updates
+  // We use multiple layers of prototypes to maximize reuse and also to keep
+  // [RawObject] out of `ownKeys`. This keeps the code in the proxy below simpler.
+  const objectHolderPrototype = Object.create(
+    objectPrototypeCache.get(client, objectDef),
+    {
+      [RawObject]: {
+        value: rawObj,
+        writable: true, // so we can allow updates
+      },
     },
-  });
+  );
 
   // we separate the holder out so we can update
-  // the underlying data without having to return a new
+  // the underlying data without having to return a new object
   // we also need the holder so we can customize the console.log output
-  const holder: ObjectHolder<Q> = Object.create(kk);
+  const holder: ObjectHolder<Q> = Object.create(objectHolderPrototype);
 
   const osdkObject: any = new Proxy(holder, {
     ownKeys(target) {
--- a/packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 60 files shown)


packages/client/src/object/convertWireToOsdkObjects/getDollarAs.ts
```diff
@@ -0,0 +1,102 @@
...
+      if (newDef === objDef.apiName) {
+        return this[UnderlyingOsdkObject];
+      }
+
+      // this is sufficient to determine if we implement the interface
+      if (objDef.interfaceMap?.[newDef] == null) {
+        throw new Error(`Object does not implement interface '${newDef}'.`);
+      }
+
+      const def = objDef[InterfaceDefinitions][newDef];
```
ssanjay1, line 75: What is this second check for again? Like why do we need both this and the one above

ssanjay1: There was a comment in the old code about interface links, not sure if thats still relevant

packages/client/src/object/convertWireToOsdkObjects/getDollarAs.ts
```diff
@@ -0,0 +1,102 @@
...
+      if (!def) {
+        throw new Error(`Object does not implement interface '${newDef}'.`);
+      }
+      newDef = def.def as NEW_Q;
+    } else if (newDef.apiName === objDef.apiName) {
+      return this[UnderlyingOsdkObject];
+    }
+
+    if (newDef.type === "object") {
+      throw new Error(`'${newDef.apiName}' is not an interface.`);
```
ssanjay1, line 85: Should we update this error message so its more clear that its not an interface implemented by this object type? Could see someone getting confused if they pass in a random object type thats not an interface and doesn't match the underlying object, and the error would just say the issue is because `it's not an interface`

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -0,0 +1,137 @@
...
+);
+
+export function createOsdkObject<
+  Q extends FetchedObjectTypeDefinition<any, any>,
+>(
+  client: MinimalClient,
+  objectDef: Q,
+  rawObj: OntologyObjectV2,
+) {
+  const kk = Object.create(objectPrototypeCache.get(client, objectDef), {
```
ssanjay1, line 74: super nit: could we use a clearer variable name please?

ssanjay1 (approved):

packages/client/src/object/convertWireToOsdkObjects/getDollarAs.ts
ericanderson: So the first one is an invariant check that the backend gave us the data we need. The second one is an invariant check that the code that loads the object also loaded the interface data. 

In theory they should both fail. 

We do use the new definition that we assign here in this call below: `createOsdkInterface(underlying, newDef)` which is part of the caching strategy for the proxy handler. Thats why I removed the comment, as it was no longer valid.

packages/client/src/object/convertWireToOsdkObjects/getDollarAs.ts
ericanderson: Technically its true but yeah we can be slightly better here. Good eye.

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
ericanderson: Appreciate you keeping me honest :)

PR #387, opened 2024-06-10
OSDK learns __EXPERIMENTAL_strictNonNull...

ericanderson:
to throw, drop objects, or return `| undefined` for properties, allowing for correct type safety.

## Additional info

Right now, it is not possible to generate an sdk with nullable types (besides the one we do for the primary key which can't be null), however this now lays the ground work so that when we do implement those, we can leverage strict types. 🎉 

## Demos

image

image

### DIFF
--- a/packages/client/src/Definitions.test.ts
+++ b/packages/client/src/Definitions.test.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectTypePropertyDefinition } from "@osdk/api";
+import { describe, expectTypeOf, it } from "vitest";
+import type { OsdkObjectPropertyType } from "./Definitions.js";
+
+describe("OsdkObjectPropertyType", () => {
+  describe("{ nullable: false } property", () => {
+    const nonNullDef = {
+      type: "string",
+      nullable: false,
+    } satisfies ObjectTypePropertyDefinition;
+
+    it("is `| undefined` for `false`", () => {
+      expectTypeOf<OsdkObjectPropertyType<typeof nonNullDef, false>>()
+        .toEqualTypeOf<string | undefined>();
+    });
+
+    it("is not `| undefined` for `true`", () => {
+      expectTypeOf<OsdkObjectPropertyType<typeof nonNullDef, true>>()
+        .toEqualTypeOf<string>();
+    });
+  });
+
+  describe("{ nullable: true } property", () => {
+    const nullableDef = {
+      type: "string",
+      nullable: true,
+    } satisfies ObjectTypePropertyDefinition;
+
+    it("is | undefined for `false`", () => {
+      expectTypeOf<OsdkObjectPropertyType<typeof nullableDef, false>>()
+        .toEqualTypeOf<string | undefined>();
+    });
+
+    it("is `| undefined` for `true`", () => {
+      expectTypeOf<OsdkObjectPropertyType<typeof nullableDef, true>>()
+        .toEqualTypeOf<string | undefined>();
+    });
+  });
+});
--- a/packages/client/src/OsdkObjectFrom.test.ts
+++ b/packages/client/src/OsdkObjectFrom.test.ts
@@ -30,6 +30,24 @@ describe("ConvertProps", () => {
       expectTypeOf<ConvertProps<FooInterface, Employee, "$all">>()
         .toEqualTypeOf<"fullName">();
     });
+
+    it("converts non-strict nullness for handles single prop", () => {
+      expectTypeOf<
+        ConvertProps<
+          FooInterface,
+          Employee,
+          | "fooSpt"
+          | "$notStrict"
+        >
+      >()
+        .toEqualTypeOf<"fullName" | "$notStrict">();
+    });
+    it("converts non-strict nullness for $all", () => {
+      expectTypeOf<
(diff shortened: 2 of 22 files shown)


packages/client/src/object/FetchPageArgs.ts
```diff
@@ -20,14 +20,20 @@ import type {
   ObjectOrInterfaceDefinition,
   ObjectOrInterfacePropertyKeysFrom2,
 } from "@osdk/api";
+
+export type StrictNonNull = false | "throw" | "drop";
```
ssanjay1, line 24: This type name confused me a bit at first because it made it sound like it was just enforcing properties on an object or non-null, when in reality its actually looking at the nullability field on each parameter and deciding whether to adhere to it or not. Can we rename to something like `StrictNullability` or basically something that makes it more clear this about whether or not you're respecting the nullabilty on the parameter type?

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -118,7 +166,53 @@ function reframeAsObjectInPlace(
...
+function isConforming(
+  client: MinimalClient,
+  def:
+    | InterfaceDefinition<any, any>
+    | ObjectTypeDefinition<any, unknown>,
+  obj: OntologyObjectV2,
+  propsToCheck: readonly string[],
+) {
+  for (const propName of propsToCheck) {
+    if (process.env.NODE_ENV !== "production") {
```
ssanjay1, line 178: I'm probably missing something here, but wouldn't you only want to log this when you find something where its non-nullable and the prop is null? So like inside the if on line 190?

packages/client/src/query/aggregations/AggregationResultsWithGroups.ts
```diff
@@ -31,13 +31,8 @@ export type AggregationResultsWithGroups<
...
     $group: {
       [P in keyof G & keyof Q["properties"]]: G[P] extends
-        { $ranges: GroupByRange<number>[] }
-        ? { startValue: number; endValue: number }
-        : G[P] extends { $ranges: GroupByRange<string>[] }
-          ? { startValue: string; endValue: string }
-        : OsdkObjectPropertyType<
-          Q["properties"][P]
-        >;
+        { $ranges: GroupByRange<infer T>[] } ? { startValue: T; endValue: T }
```
ssanjay1, line 34: Thanks for cleaning that up

packages/client/src/Definitions.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      type: "string",
+      nullable: true,
+    } satisfies ObjectTypePropertyDefinition;
+
+    it("is | undefined for `false`", () => {
+      expectTypeOf<OsdkObjectPropertyType<typeof nullableDef, false>>()
+        .toEqualTypeOf<string | undefined>();
+    });
+
+    it("is `| undefined` for `\"throw\"`", () => {
```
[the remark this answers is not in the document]
ssanjay1: super nit: Reading the rest of the pr, makes sense why you said throw here as thats the default for the nullabilty adherence type, but in the context of this test feels like saying true would make more sense

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -46,29 +51,85 @@ export async function convertWireToOsdkObjects(
...
 
-    const osdkObject: any = createOsdkObject(client, objectDef, rawObj);
-    ret.push(interfaceApiName ? osdkObject.$as(interfaceApiName) : osdkObject);
+    conforming &&= isConforming(client, objectDef, rawObj, objProps);
+
+    let osdkObject = createOsdkObject(client, objectDef, rawObj);
+    if (interfaceApiName) osdkObject = osdkObject.$as(interfaceApiName);
+
+    allResults.push(osdkObject);
+    if (conforming) conformingResults.push(osdkObject);
```
ssanjay1, line 103: Could you maybe short circuit here for the throw case? Like if it's `throw`, as soon as you hit one non-conforming object you can just throw the error right there and not bother converting the rest of the objects?

packages/client/src/object/FetchPageArgs.ts
```diff
@@ -20,14 +20,20 @@ import type {
...
+export type StrictNonNullDefault = "throw";
```
ssanjay1, line 25: Is this safe to default to throw? This may break existing usage right, should this be opt-in instead of opt-out?

packages/client/src/objectSet/ObjectSet.ts
```diff
@@ -94,17 +101,23 @@ export interface ObjectSet<Q extends ObjectOrInterfaceDefinition>
 
   pivotTo: <L extends LinkNames<Q>>(type: L) => ObjectSet<LinkedType<Q, L>>;
 
-  fetchOne: Q extends ObjectTypeDefinition<any>
-    ? <L extends ObjectOrInterfacePropertyKeysFrom2<Q>>(
+  fetchOne: Q extends ObjectTypeDefinition<any> ? <
+      L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
+      R extends boolean,
+      S extends StrictNonNull = StrictNonNullDefault,
```
ssanjay1, line 107: Hmm, so if I give it `drop`, this could return an empty result right? Could we maybe just hide/restrict the drop option for `fetchOne` for now? Feels like this could get confusing

ssanjay1 (approved):
Overall looks cool, approving to unblock, but left some comments

packages/client/src/OsdkObjectFrom.ts
```diff
@@ -42,31 +51,46 @@ export type ConvertProps<
...
       (
-        TO["interfaceMap"][ApiNameAsString<FROM>][
-          P extends "$all"
-            ? keyof FROM["properties"] extends
-              keyof TO["interfaceMap"][ApiNameAsString<FROM>]
-              ? keyof FROM["properties"]
-            : never
-            : DropRidAndAll<P>
-        ]
+        UnionIfTrue<
```
ericanderson, line 54: @ssanjay1 I ended up adding more tests and discovered this needed to be updated.

packages/client/src/OsdkObjectFrom.test.ts
```diff
@@ -117,18 +117,42 @@ describe("Osdk", () => {
     });
   });
 
+  describe("$notStrict", () => {
```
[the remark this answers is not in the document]
ssanjay1: Ah you're totally right, missed that

packages/client/src/object/FetchPageArgs.ts
ssanjay1: Ok that's fair

PR #389, opened 2024-06-11
Do not restart auth flow on failed auth callback

tzyl:
If the final token request to complete the authorization_code for public clients fails, currently the client resets state and restarts the whole oauth flow with a new login redirect. If there is a persistent error causing the token request to fail this will cause a repeated cycle from the auth callback back to a new login redirect. Instead, surface the error in the token request for application code to handle which by default in the create-app templates displays the error in text on the AuthCallback component

### DIFF
--- a/packages/legacy-client/src/oauth-client/PublicClient/PublicClientAuth.ts
+++ b/packages/legacy-client/src/oauth-client/PublicClient/PublicClientAuth.ts
@@ -142,10 +142,11 @@ export class PublicClientAuth implements Auth {
       } catch (e) {
         // eslint-disable-next-line no-console
         console.warn(
-          "Failed to get OAuth2 token using PCKE, removing PCKE and starting a new auth flow",
+          "Failed to get OAuth2 token using PCKE, removing PCKE",
           e,
         );
         localStorage.removeItem(this.palantirPcke);
+        throw e;
       }
     }
 
--- a/.changeset/great-geese-fail.md
+++ b/.changeset/great-geese-fail.md
@@ -1,0 +1,5 @@
+---
+"@osdk/legacy-client": patch
+---
+
+Do not restart auth flow on failed auth callback


nadprana (approved):

ericanderson (dismissed):
Blocking merge while i research

ericanderson:
As I look over the auth flow and how oauth works, I believe this change causes a problem in scenarios:

1. User loads the page, unauthenticated.
2. User is redirected to oauth login
3. User gets distracted and closes the page without logging in.
4. User comes back to the page and is not logged in but now they have an old pkce in local storage.
5. The exception throwing you added causes signIn to fail and an exception is thrown.

What is supposed to happen now? Is the user just in a broken state?

ericanderson (commented):
Unblocking for session storage change

packages/legacy-client/src/oauth-client/PublicClient/PublicClientAuth.ts
```diff
@@ -33,7 +33,7 @@ import {
 
 export class PublicClientAuth implements Auth {
   private palantirRefreshToken = "palantir_refresh_token" as const;
-  private palantirPcke = "palantir_pcke" as const;
+  private palantirPkce = "palantir_pkce" as const;
```
tzyl, line 36: The rename here and the move from local to session storage does leave behind old values but given that this state should only live for the lifetime of one auth flow there shouldn't be a risk of breaking a user's login

packages/legacy-client/src/oauth-client/PublicClient/PublicClientFlowProvider.ts
```diff
@@ -75,6 +75,12 @@ export async function getTokenWithCodeVerifier(
...
   const parsedUrl = new URLSearchParams(url.substring(url.indexOf("?")));
+
+  const error = parsedUrl.get("error");
+  if (error) {
+    throw new Error("Error received from server: " + error);
+  }
+
   const code = parsedUrl.get("code");
   if (!code) {
     throw new Error("Code parameter missing in URL: " + url);
```
tzyl, line 86: If we receive the callback with the error in the url query param then we would have failed here in the past so I included the actual error if it's present so in the default template the user will at least see the real reason for the code parameter being missing
image

packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -228,7 +228,7 @@ export function createPublicOauthClient(
       if (process?.env?.NODE_ENV !== "production") {
         // eslint-disable-next-line no-console
         console.warn(
-          "Failed to get OAuth2 token using PCKE, removing PCKE and starting a new auth flow",
+          "Failed to get OAuth2 token using PKCE, removing PKCE and starting a new auth flow",
```
tzyl, line 231: Note that the 2.0 oauth is a bit out of sync now but applied this rename across the repo

ericanderson: Thank you

ericanderson (approved):

PR #392, opened 2024-06-13
Attachments Work V3

ssanjay1:
Modified attachment syntax for fetching data, now instead of using the client, attachment properties on objects will have this functionality exposed already:

```
const result = client(objectType).fetchPage();
const metadata = result.data[0].myAttachment.fetchMetadata();
const contents = result.data[0].myAttachment.fetchContents();
```

If you already have a rid, you can also reconstruct an Attachment object as follows:
```
import {createAttachmentFromRid} from "@osdk/client; 
const attachment = createAttachmentFromRid(client, "myAttachmentRid");
```

### DIFF
--- a/packages/client.api/src/mapping/DataValueMapping.ts
+++ b/packages/client.api/src/mapping/DataValueMapping.ts
@@ -14,13 +14,16 @@
  * limitations under the License.
  */
 
-import type { Attachment, AttachmentUpload } from "../object/Attachment.js";
+import type {
+  AttachmentUpload,
+  AttachmentWrapper,
+} from "../object/Attachment.js";
 
 /**
  * Map from the DataValue type to the typescript type that we return
  */
 export interface DataValueWireToClient {
-  attachment: Attachment;
+  attachment: AttachmentWrapper;
   boolean: boolean;
   byte: number;
   datetime: string;
@@ -40,7 +43,7 @@ export interface DataValueWireToClient {
  * Map from the DataValue type to the typescript type that we accept
  */
 export interface DataValueClientToWire {
-  attachment: Attachment | AttachmentUpload;
+  attachment: string | AttachmentUpload;
   boolean: boolean;
   byte: number;
   datetime: string;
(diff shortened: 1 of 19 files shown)


packages/client/src/Client.ts
```diff
@@ -46,8 +42,6 @@ export interface Client extends SharedClient<MinimalClient> {
   <Q extends ActionDefinition<any, any, any>>(
     o: CheckVersionBound<Q>,
   ): ActionSignatureFromDef<Q>;
-
-  (o: Attachment): AttachmentSignature;
```
ssanjay1, line 50: Removing ability to read attachments directly with client

packages/client/src/createAttachmentFromRid.ts
```diff
@@ -14,27 +14,24 @@
...
-  Attachment,
-  AttachmentMetadata,
-  AttachmentSignature,
-} from "@osdk/client.api";
+import type { Attachment } from "@osdk/client.api";
 import { Ontologies } from "@osdk/internal.foundry";
 import type { MinimalClient } from "./MinimalClientContext.js";
 
-export function createAttachmentReader(
+export function createAttachmentFromRid(
```
ssanjay1, line 21: This helper function can be used for people to reconstruct an attachment object with just a rid and client

packages/client/src/util/toDataValue.ts
```diff
@@ -49,11 +44,6 @@ export async function toDataValue(
     return Promise.all(promiseArray);
   }
 
-  // attachments just send the rid directly
-  if (isAttachment(value)) {
```
ssanjay1, line 53: We don't need this check anymore as we just let users directly pass in the rid as a string (which we just pass through to the wire)

packages/client.api/src/mapping/DataValueMapping.ts
```diff
@@ -40,7 +40,7 @@ export interface DataValueWireToClient {
  * Map from the DataValue type to the typescript type that we accept
  */
 export interface DataValueClientToWire {
-  attachment: Attachment | AttachmentUpload;
+  attachment: string | AttachmentUpload;
```
ssanjay1, line 43: Now we just let users give us the rid as a `string`, since we deconstruct and send only the string to the wire anyways

ericanderson (approved):

PR #393, opened 2024-06-13
Add support for ordering on $select clauses in aggregations

bryantpark04:
Opening draft PR while I start writing tests


packages/client/src/query/aggregations/AggregationResultsWithoutGroups.ts
```diff
@@ -16,25 +16,37 @@
...
-  AC extends AggregationClause<Q>,
+  AC extends UnorderedAggregationClause<Q> | OrderedAggregationClause<Q>,
 > = {
-  [P in keyof Q["properties"] as SubselectKeys<AC, P>]: AC[P] extends
-    readonly string[] | string ? {
-      [Z in StringArrayToUnion<AC[P]>]: Z extends "approximateDistinct" ? number
-        : OsdkObjectPropertyType<Q["properties"][P]>;
-    }
-    : never;
+  [
```
bryantpark04, line 28: This whole process of extracting a nested object structure from a flattened "property:metric" union is pretty messy. Is there a better way of doing this?

packages/client/src/object/AggregateOptsThatErrors.ts
```diff
@@ -15,9 +15,25 @@
  */
 
 import type { ObjectOrInterfaceDefinition } from "@osdk/api";
+import type { SingleKeyObject } from "type-fest";
 import type { AggregateOpts } from "../query/aggregations/AggregateOpts.js";
+import type { UnorderedAggregationClause } from "../query/aggregations/AggregationsClause.js";
 import type { GroupByClause } from "../query/aggregations/GroupByClause.js";
 
+export type AggregateOptsThatErrorsAndDisallowsOrderingWithMultipleGroupBy<
```
bryantpark04, line 23: I tried a few different approaches for enforcing the "can't order on $select if there are multiple $groupBy clauses" restriction in the type system, and this method of adding another wrapped type layer around the `AggregateOptsThatErrors` type just to check this restriction was the easiest way to do it. 

@ericanderson was this an okay approach?

packages/client/src/internal/conversions/legacyToModernSingleAggregationResult.ts
```diff
@@ -36,8 +40,10 @@ export function legacyToModernSingleAggregationResult<
         "assumed we were getting a `${key}.${type}`",
       );
       if (!(parts[0] in accumulator)) {
+        // @ts-expect-error TODO fixme
```
bryantpark04, line 43: Need to figure out why this broke

packages/client/src/query/aggregations/AggregationResultsWithoutGroups.ts
ericanderson: Yeah so creating helper types starts to expose places where things can clean up. I ended up with this:

```ts
type ExtractPropName<T extends string> = T extends `${infer PREFIX}:${string}`
  ? PREFIX
  : T extends "$count" ? T
  : never;
type ExtractMetricForPropName<T, PropName extends string> = T extends
  `${PropName}:${infer METRIC_NAME}` ? METRIC_NAME : never;

export type AggregationResultsWithoutGroups<
  Q extends ObjectOrInterfaceDefinition<any, any>,
  AC extends UnorderedAggregationClause<Q> | OrderedAggregationClause<Q>,
> = {
  [TPropName in ExtractPropName<keyof AC & string>]: TPropName extends "$count"
    ? number
    : {
      [TMetricName in ExtractMetricForPropName<keyof AC & string, TPropName>]:
        TMetricName extends "approximateDistinct" ? number
          : (OsdkObjectPropertyType<Q["properties"][TPropName]>);
    };
};
```
I think this works but didn't check extensively.

A big difference here is I dropped the `as` in the mapped type and instead just directly returned the values, allowing me to reuse the propname and metricname.

Other places in the codebase where we have to do the `as` syntax, is because it allows us to retain things like jsdoc from the mapped types. However here we are mapping object literals that have no interesting information. `{ "foo:bar": true }` has no jsdoc to map, so we can simplify the syntax.

Now, if we wanted to some how keep the jsdoc descriptions of properties, we would have to map over the keys of Q["properties"] which would be doable. This is psuedocode I did not test:

```ts
export type AggregationResultsWithoutGroups<
  Q extends ObjectOrInterfaceDefinition<any, any>,
  AC extends UnorderedAggregationClause<Q> | OrderedAggregationClause<Q>,
> = {
  [ TPropName in keyof Q["properties"] ]: {
    [ K in ExtractMetricForPropName<keyof AC & string, TPropName>]: 
        TMetricName extends "approximateDistinct" ? number
          : (OsdkObjectPropertyType<Q["properties"][TPropName]>);
  }
  & {
    [ K in keyof AC as K extends "$count" ? K : never]: number;
};
```

packages/client/src/object/aggregate.test.ts
```diff
@@ -398,6 +425,202 @@ describe("aggregate", () => {
     });
   });
 
+  it("works with $orderBy (no groups)", async () => {
+    const mockFetch: Mock = vi.fn();
```
ssanjay1, line 429: You may want to throw this boilerplate setup code for the fetch in a `beforeAll`

packages/client/src/internal/conversions/modernToLegacyAggregationClause.ts
```diff
@@ -15,31 +15,42 @@
...
+
+    // k is property:metric
+    const colonPos = k.lastIndexOf(":");
+    const property = k.slice(0, colonPos);
+    const metric = k.slice(colonPos + 1);
+
+    return [
+      {
+        type: metric as StringAggregateOption | NumericAggregateOption,
+        name: `${property}.${metric}`,
```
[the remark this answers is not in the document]
bryantpark04: I think the property.metric string is what gets sent to the wire as part of the request body, so the reformatting from property:metric to property.metric is necessary

packages/client/src/internal/conversions/modernToLegacyAggregationClause.ts
```diff
@@ -15,31 +15,42 @@
...
+        field: property,
+        direction: directionFieldMap(v),
```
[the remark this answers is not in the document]
bryantpark04: Ah that's fascinating, I didn't know. Thanks for the heads up! Fixed it

packages/client/src/query/aggregations/AggregationResultsWithoutGroups.ts
bryantpark04: Tested this and it seems to work. I'll refactor this - thanks!

bryantpark04: I don't think we need the $count part in this because the $count gets &'d in with the AggregationCountResult type

bryantpark04: It seemed simpler to just remove the AggregationCountResult instead, so I just did that

packages/client/src/internal/conversions/modernToLegacyAggregationClause.ts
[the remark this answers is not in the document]
ericanderson: It is whats in the request body, but the backend doesn't care what this value is. It will simply return a result that matches that value.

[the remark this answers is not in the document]
ericanderson: We can leave as is for now

ericanderson (commented):
You have my blessing if you have Saurav's blessing

packages/client/src/internal/conversions/modernToLegacyAggregationClause.ts
```diff
@@ -15,31 +15,44 @@
...
-          field: k,
+          type: "count",
+          name: "count",
+          direction: directionFieldMap(aggregationType),
         };
-      });
-    } else {
+      }
+
+      // k is property:metric
```
ssanjay1, line 43: nit: this comment is outdated

packages/client/src/object/aggregate.test.ts
```diff
@@ -133,53 +144,58 @@ const mockOntology = {
 type mockOntology = typeof mockOntology;
 interface MockOntology extends mockOntology {}
 
-describe("aggregate", () => {
-  it("works", async () => {
-    const mockFetch: Mock = vi.fn();
+let mockFetch: Mock;
+let clientCtx: MinimalClient;
+
+beforeEach(() => {
```
ssanjay1, line 150: If you don't modify anything in here (fetch function, client etc.), you can just use `beforeAll` which is only called once before everything

ssanjay1 (approved):
Approved, couple of unblocking comments

bryantpark04:
> Approved, couple of unblocking comments

What kind of cruel joke is this

image

PR #396, opened 2024-06-16
Ewitkon/add aip function to aip now

ewitkon:
This PR adds a new template to be used with AIP Now AIP edition.


ericanderson:
Stale and I think was done in another PR? Feel free to reopen if im wrong

PR #399, opened 2024-06-17
Refactor more types into client.api

ssanjay1:
Moved groupby and aggregation types into client.api, which should make some of the queries work easier as well


packages/client.api/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@osdk/api": "workspace:^",
+    "@osdk/internal.foundry": "workspace:^",
```
ericanderson, line 33: If these types are re-exported (implicitly or explicitly) then they cannot be a devDependency without also being a peer dependency (or just a regular dependency).

If we are using this package for more than types, then making it a dev dependency also means it gets inlined into this package by tsup so that might also not be desirable.

packages/client.api/src/index.ts
```diff
@@ -25,6 +25,34 @@ export type {
...
+  WhereClause,
+} from "./aggregate/WhereClause.js";
+export { DistanceUnitMapping } from "./aggregate/WhereClause.js";
+export type { OsdkObjectPropertyType } from "./Definitions.js";
+export type {
+  AllGroupByValues,
+  GroupByClause,
+  GroupByRange,
+} from "./groupby/GroupByClause.js";
+export { DurationMapping } from "./groupby/GroupByClause.js";
```
ericanderson, line 55: So much API surface :/

We definitely need to see what we can prune, either cause it just doesnt need to be exported or if we dont need all the exact types elsewhere. 

I'm actually quite glad we have a simple way to see the back compat surface, I just want to make sure its manageable.

packages/client/src/OsdkObjectFrom.ts
```diff
@@ -19,8 +19,11 @@ import type {
   ObjectOrInterfaceDefinition,
   ObjectTypeDefinition,
 } from "@osdk/api";
-import type { OsdkBase, OsdkObjectPrimaryKeyType } from "@osdk/client.api";
-import type { OsdkObjectPropertyType } from "./Definitions.js";
+import type {
+  OsdkBase,
```
ericanderson, line 23: I think this might be a good example of something we don't want in the surface per se. Should we just be exporting `Osdk<>` from `@osdk/client.api` instead? Does that remove the need to export some of these?

Non blocking comment of course, just something for us to look at. Could be now. Could be later.

packages/client/src/object/ObjectDefinitions.test.ts
```diff
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 
-import type { Attachment } from "@osdk/client.api";
+import type { Attachment, OsdkObjectPropertyType } from "@osdk/client.api";
```
ericanderson, line 17: If this test only tests things in client.api it should move to client.api

packages/client/src/object/aggregate.test.ts
```diff
@@ -15,15 +15,17 @@
  */
 
 import type { ObjectTypeDefinition, OntologyDefinition } from "@osdk/api";
+import type {
+  AggregateOpts,
+  AggregateOptsThatErrors,
+  GroupByClause,
+} from "@osdk/client.api";
```
ericanderson, line 22: I wonder if we should be splitting part of this test out?

packages/client/src/query/aggregations/AggregationsClause.test.ts
```diff
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 
+import type { AggregatableKeys, AggregationClause } from "@osdk/client.api";
```
ericanderson, line 17: This test should move?

packages/client/src/query/aggregations/AggregationsResults.test.ts
```diff
@@ -14,9 +14,11 @@
  * limitations under the License.
  */
 
+import type {
+  AggregationResultsWithoutGroups,
+  AggregationsResults,
+} from "@osdk/client.api";
```
ericanderson, line 20: This test should move?

packages/client/src/query/aggregations/GroupByClause.test.ts
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
+import type { GroupByClause } from "@osdk/client.api";
```
ericanderson, line 17: Move test?

ericanderson (approved):
Conceptually fine. Left some notes to fix up before merging.

packages/client/src/object/aggregate.test.ts
ssanjay1: Can look in a FLUP, bryant's PR also touches this a lot so can wait until after that

PR #408, opened 2024-06-19
Move objectset into client.api

ssanjay1:
Refactored so objectSet type is now in client.api


packages/client.api/src/index.ts
```diff
@@ -76,3 +76,40 @@ export type { LinkedType, LinkNames } from "./util/LinkUtils.js";
 export type { NOOP } from "./util/NOOP.js";
 
 export {};
+export type {
```
ssanjay1, line 79: Added a lot here, as mentioned in previous PRs, definitely should look through this and make sure we prune things we don't want to expose before releasing

ericanderson: Nit can you group these with the the above exports, leaving a space with the `export {};` last and run cmd+shift+p organize imports?

ericanderson (approved):

mfedderly (approved):
👋  approving without review to appease policybot

PR #415, opened 2024-06-20
Add api-extractor, fix turbo, fix refactor

ericanderson:
Originally, this started as adding api-extractor so that we can see how the API surface changes over time. This lead me to realize we had some inconsistencies in our turbo config. Then to fix those I ended up also upgrading turbo. Then I found a problem in a recent refactor that prevented this from going in. So this fixes all of those.

For posterity, api-extractor has a very annoying problem that you can have:

```ts
// packages/foo/src/foo.ts
import { bar } from "@osdk/bar";
// do stuff
```

You can then move it to the bar package and not change the contents:
```ts
// packages/bar/src/foo.ts
import { bar } from "@osdk/bar";
// do stuff
```

TS is happy. All is fine. But the api-checker hates this and really wants you to `import { bar } from "./bar.js"`.

In the future, as we expand use of the api-checker, this will just be an annoyance for whoever is doing the edits because the checker will prevent us from merging it in.


.monorepolint.config.mjs
```diff
@@ -32,6 +32,7 @@ const DELETE_SCRIPT_ENTRY = { options: [undefined], fixValue: undefined };
 const nonStandardPackages = [
   "mytsup",
   "tsconfig",
+  "@osdk/api-extractor",
```
ericanderson, line 35: Just the config file to extend

packages/client.api/etc/client.api.report.api.md
```diff
@@ -0,0 +1,397 @@
+## API Report File for "@osdk/client.api"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
```
ericanderson, line 3: Generated file which lets us see:
- what was exported
- what was referenced without being exported
- what external packages are part of our public api
- which types are undocumented
- which types are public/internal/beta

packages/client.api/package.json
```diff
@@ -17,21 +17,26 @@
     }
   },
   "scripts": {
+    "api-checker": "api-extractor run --local --verbose --diagnostics",
+    "api-docs": "api-documenter markdown -i build/api-extractor --output-folder api-docs",
```
ericanderson, line 21: Only added this here for now. Will add to a few more in the future.

packages/client.api/src/aggregate/AggregationsClause.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import type { ObjectOrInterfaceDefinition } from "@osdk/api";
-import type { ValidAggregationKeys } from "@osdk/client.api";
+import type { ValidAggregationKeys } from "./AggregatableKeys.js";
```
ericanderson, line 18: Fix the api-checker bug caused by refactor

packages/client.api/src/aggregate/AggregationResultsWithoutGroups.ts
```diff
@@ -15,11 +15,11 @@
  */
 
 import type { ObjectOrInterfaceDefinition } from "@osdk/api";
+import type { OsdkObjectPropertyType } from "../Definitions.js";
 import type {
   OrderedAggregationClause,
   UnorderedAggregationClause,
-} from "@osdk/client.api";
-import type { OsdkObjectPropertyType } from "../Definitions.js";
+} from "./AggregationsClause.js";
```
ericanderson, line 22: Fix the api-checker bug caused by refactor

packages/client.test.ontology/turbo.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": ["//"],
```
ericanderson, line 2: Improving build perf by separating the codegen dependencies to per project. You'll see a handful of turbo.json files in here because of that.

turbo.json
```diff
@@ -1,17 +1,16 @@
...
-  "pipeline": {
+  "ui": "stream",
+  "tasks": {
     "codegen": {
-      "inputs": ["ontology.json"],
-      "outputs": ["src/generatedNoCheck/**/*", "src/generatedNoCheck2/**/*"],
-      "dependsOn": ["^transpile", "^typecheck"]
-    },
-
+      "outputLogs": "new-only"
```
ericanderson, line 6: Limit output when cached.

turbo.json
```diff
@@ -1,17 +1,16 @@
...
+    }, // empty for overriding
```
ericanderson, line 7: No default dependencies to optimize build graph

turbo.json
```diff
@@ -101,12 +123,24 @@
...
     "transpile": {
-      "dependsOn": ["mytsup#typecheck", "tsconfig#typecheck", "codegen"],
+      "outputLogs": "new-only",
+      "dependsOn": ["mytsup#typecheck", "tsconfig#typecheck", "^transpile", "codegen"],
       "inputs": ["src/**", "tsup.config.js", "tsconfig.json", "tsconfig.cjs.json"],
-      "outputs": ["dist/**", "lib/**", "build/**/*.{js,cjs,js.map,cjs.map}"]
+      "outputs": [
+        "dist/**",
+        "lib/**",
+        "build/{esm,cjs,browser}/**/*.{js,cjs,js.map,cjs.map}"
```
ericanderson, line 132: Get explicit about outputs so we can decouple typecheck from transpile.

turbo.json
```diff
@@ -133,7 +168,15 @@
...
     "check": {
-      "dependsOn": ["//#check-mrl", "lint", "transpile", "typecheck", "test", "check-attw"]
+      "dependsOn": [
+        "//#check-mrl",
+        "lint",
+        "transpile",
+        "typecheck",
+        "test",
+        "check-attw",
+        "api-checker"
```
ericanderson, line 172: Add api-checker if it exists.

monorepo/mytsup/tsup.mjs
```diff
@@ -40,14 +40,15 @@ export default async (options, ourOptions) => {
...
-    clean: true,
+    clean: false, // we do this ourselves so its granular
     silent: true,
     sourcemap: true,
     splitting: true,
     minify: false, // !options.watch,
     onSuccess: async () => {
       console.log("👍");
     },
+    keepNames: false,
```
ericanderson, line 51: optimization to build output.

monorepo/mytsup/tsup.mjs
```diff
@@ -40,14 +40,15 @@ export default async (options, ourOptions) => {
       ),
       TARGET: "node",
     },
-    clean: true,
+    clean: false, // we do this ourselves so its granular
```
ericanderson, line 43: This was deleting the entire output directory, which meant that typecheck ALWAYS had to run after transpile. If you ran typecheck. then did a transpile run, you would lose your typecheck output.

ssanjay1 (approved):
Looks generally fine to me, approved!

PR #422, opened 2024-06-25
Update ontology as code to use new shapes as specified by Alex

ericanderson:
Blocks on #421


packages/client.unstable/src/blockDataIr.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export interface OntologyIrSharedPropertyType
+  extends Omit<SharedPropertyType, "rid">
+{}
+
+export interface OntologyIrInterfaceLinkType
+  extends Omit<InterfaceLinkType, "rid">
+{}
+
+export type OntologyIrInterfaceTypeBlockDataV2 = ReplaceKeys<
```
ssanjay1, line 72: I'm assuming you used interfaces rather than types for the above ones for speed, but confused why this one is still a type?

packages/maker/src/api/defineOntology.ts
```diff
@@ -67,63 +61,140 @@ export async function defineOntology(
...
+      return asdf(type);
+  }
+}
+
+/**
+ * Helper function to avoid duplication
+ * @param type
+ * @returns
+ */
+function asdf<T extends string>(
```
ssanjay1, line 196: can we rename this function please?

ssanjay1 (approved):
Approved with one question and one nonblocking nit

packages/client.unstable/src/blockDataIr.ts
ericanderson: oops, missed one.

packages/maker/src/api/defineOntology.ts
ericanderson: 😬 that's embarrassing. when i refactor i just throw up a ton of crap and worry about renaming later

PR #423, opened 2024-06-27
change action invocation to require `.applyAction`

### LINKED ISSUE #373: Add .invoke() for actions

ericanderson:
Fixes #373


packages/client.api/src/actions/Actions.ts
```diff
@@ -67,13 +67,11 @@ export type OsdkActionParameters<
   : PartialBy<NotOptionalParams<X>, NullableProps<X>>;
 
 export type ActionSignatureFromDef<T extends ActionDefinition<any, any, any>> =
-  NonNullable<T["__OsdkActionType"]> extends never
-    ? ActionSignature<T["parameters"]>
-    : NonNullable<T["__OsdkActionType"]>;
-
-export type Actions<O extends OntologyDefinition<any>> = {
```
ericanderson, line 74: Dead code

ericanderson:
@ssanjay1 I know we discussed breaking batch out from non-batch but I left it alone for now because it complicates the type generation and would require internal users to have to regenerate.

I also went with `applyAction` instead of just `apply` as the old return value was a function which meant it already had a `.apply` method on it, which could be confusing. I also like that when you read the code you can tell that an action is getting called.

ssanjay1 (approved):

PR #424, opened 2024-06-27
Creates defineInterfaceLinkConstraint

ericanderson:
and modifies syntax to reflect feedback & readability.

Blocks on ##422

### DIFF
--- a/packages/maker/src/api/defineInterfaceLinkConstraint.ts
+++ b/packages/maker/src/api/defineInterfaceLinkConstraint.ts
@@ -24,20 +24,11 @@ type Many = { type: InterfaceType; many: ApiNameOrMeta; one?: never };
 type One = { type: InterfaceType; one: ApiNameOrMeta; many?: never };
 
 export function defineInterfaceLinkConstraint(
-  opts: {
+  { from, to }: {
     from: Many;
     to: One;
-  },
-): unknown;
-export function defineInterfaceLinkConstraint(
-  opts: {
+  } | {
     from: One;
-    to: One | Many;
-  },
-): unknown;
-export function defineInterfaceLinkConstraint(
-  { from, to }: {
-    from: One | Many;
     to: One | Many;
   },
 ) {
--- a/packages/maker/src/api/defineInterfaceLinkConstraint.ts
+++ b/packages/maker/src/api/defineInterfaceLinkConstraint.ts
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


packages/maker/src/api/defineOntology.ts
```diff
@@ -99,29 +95,16 @@ function convertToWireOntology(
...
-    extendsInterfaces: [],
-    links: [],
-    status: {
-      active: true,
-      type: "active",
-    },
+    ...interfaceType,
+    properties: Object.values(interfaceType.properties)
+      .map<OntologyIrSharedPropertyType>((spt) => convertSpt(spt)),
+    // these are omitted from our internal types but we need to readd them for the final json
```
ssanjay1, line 104: nit: *read

packages/maker/src/api/defineInterfaceLinkConstraint.ts
```diff
@@ -0,0 +1,105 @@
...
+import invariant from "tiny-invariant";
+import type { InterfaceType } from "./types.js";
+
+type Meta = { apiName: string; displayName?: string; description?: string };
+type ApiNameOrMeta = string | Meta;
+
+type Many = { type: InterfaceType; many: ApiNameOrMeta; one?: never };
+type One = { type: InterfaceType; one: ApiNameOrMeta; many?: never };
+
+export function defineInterfaceLinkConstraint(
```
ssanjay1, line 26: I'm confused, do you still need these functions? (lines 26-36)

ssanjay1 (approved):
Approved with question

packages/maker/src/api/defineOntology.ts
ericanderson: re-add

packages/maker/src/api/defineInterfaceLinkConstraint.ts
ericanderson: Yes those are overloads. We don't support many-to-many. I could rewrite this I guess.

packages/maker/src/api/defineOntology.ts
ssanjay1: oh that would make more sense lol

packages/maker/src/api/defineInterfaceLinkConstraint.ts
ssanjay1: Ok but the third function you wrote with the invariants throw an error for many-to-many right? Did you just want to overload it so that the function params also reflect that restriction?

ericanderson: Yeah I wanted to be sure if you called it from JS or did an `any` that it was helpful. And I wanted the types to reflect it. But I rewrote it in my update (about to push) since it can just be a union.

PR #435, opened 2024-07-02
WIP Timeseries

ssanjay1:


packages/client.api/src/OsdkObjectFrom.ts
```diff
@@ -139,7 +140,7 @@ export type Osdk<
     /** @deprecated use $primaryKey */
     __primaryKey: Q extends ObjectTypeDefinition<any>
       ? OsdkObjectPrimaryKeyType<Q>
-      : unknown;
+      : any;
```
ssanjay1, line 143: Had to type this as any because unknown doesn't work when we have an interface type. It used to work when `PropertyValueClientToWire` basically also evaluated to unknown, so it'd be an unknown to unknown check, but now its unknown to `string|boolean|number|attachment...`

packages/client/src/createTimeseriesProperty.ts
```diff
@@ -0,0 +1,171 @@
...
+      type: "relative",
+      startTime: {
+        when: "BEFORE",
+        value: body.$before,
+        unit: TimeseriesDurationMapping[body.$unit],
+      },
+    }
+    : {
+      type: "relative",
+      endTime: {
```
ssanjay1, line 165: It made the most sense to me to set this as an endtime if the query was after, e.g. give me all the high temperatures 2 weeks after today, will give all the temps from today up to 2 weeks later

packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -167,7 +167,7 @@ describe("convertWireToOsdkObjects", () => {
       clientCtx,
       [objectFromWire],
       undefined,
-    )) as Osdk<Employee>[];
+    )) as unknown as Osdk<Employee>[];
```
ssanjay1, line 170: Had to do this because `Osdk<>` type is pretty complex now so it has a hard time casting from `Osdk<ObjectOrInterfaceDefinition>` to `Osdk<myConcreteType>`

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -50,7 +50,7 @@ export function createOsdkInterface<
   };
 
   if (process.env.TARGET !== "browser") {
-    Object.setPrototypeOf(interfaceHolder, OsdkCustomInspectPrototype);
+    Object.setPrototypeOf(interfaceHolder, null);
```
ssanjay1, line 53: So this is what we discussed, are we still good with this for now?

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -38,7 +39,7 @@ import type { PropertyDescriptorRecord } from "./PropertyDescriptorRecord.js";
 const objectPrototypeCache = createClientCache(
   function(client, objectDef: FetchedObjectTypeDefinition<any, any>) {
     return Object.create(
-      process.env.target !== "browser" ? OsdkCustomInspectPrototype : null,
+      process.env.target !== "browser" ? null : null,
```
ssanjay1, line 42: Can remove the ternary if we just decide to set the prototype to null

packages/client/src/object/object.test.ts
```diff
@@ -58,7 +58,18 @@ describe("OsdkObject", () => {
 
       // we should get the employee we requested
       const employee = result.data[0];
-      expect(employee).toEqual(asV2Object(stubData.employee1));
```
ssanjay1, line 61: Had to revise this because the employeeStatus property now gets functions

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -109,47 +109,44 @@ describe("ObjectSet", () => {
...
-          "class": "Red",
-          "employeeId": 50032,
-          "employeeStatus": "TimeSeries<String>",
-          "fullName": "Jack Smith",
-          "office": "LON",
-          "startDate": "2015-05-15",
-        },
-      ]
-    `);
+    expect(employees).toMatchObject([
```
ssanjay1, line 112: Test checks the same things, just revised so I could tell it to expect anything for the timeseries property

packages/client/src/util/streamutils.ts
```diff
@@ -0,0 +1,123 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const START_TOKEN = new Uint8Array([123, 34, 100, 97, 116, 97, 34, 58, 91]); // `{"data":[`
+const OBJECT_OPEN_CHAR_CODE = 123; // '{'
+const OBJECT_CLOSE_CHAR_CODE = 125; // '}'
+
+export async function* parseStreamedResponse(
```
ssanjay1, line 21: Pulled this parsing helper from the legacy implementation

packages/client/src/createTimeseriesProperty.ts
```diff
@@ -0,0 +1,171 @@
...
+        objectApiName,
+        primaryKey,
+        propertyName,
+        query,
+      );
+    },
+  };
+}
+
+async function getAllTimeSeriesPoints<T extends string | number>(
```
ericanderson, line 85: A few questions:
1.  Why isn't this function implemented by calling `iterateTimeSeriesPoints`? It feels duplicative and hard to maintain as two parallel code paths.
2. Should we be exposing a `getAll` here?

packages/client/src/createTimeseriesProperty.ts
ericanderson: Makes sense. Can we add an invariant check in this function when not in production mode too? That would help anyone using pure JS (or throwing `as any` at their code) as well as serve as an extra validation to our own logic.

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
ericanderson: Lets revert this change for now. It only makes sense to do this in non-browser. If it rears its head again, we can come up with a better long term solution.

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
ericanderson: Lets restore this one too

packages/client/src/util/streamutils.ts
ericanderson: Cool with this for now as we know it works but I'm curious if there is something a bit less custom that works out of the box for browsers? Def for a follow up.

ericanderson (approved):

packages/client.api/src/index.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import { TimeseriesDurationMapping } from "./timeseries/timeseries.js";
```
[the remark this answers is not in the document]
ssanjay1: Ah we very much do not, good catch

packages/client/src/createTimeseriesProperty.ts
```diff
@@ -0,0 +1,171 @@
...
+  parseStreamedResponse,
+} from "./util/streamutils.js";
+
+export function createTimeseriesProperty<T extends number | string>(
+  client: MinimalClient,
+  objectApiName: string,
+  primaryKey: any,
+  propertyName: string,
+): TimeSeriesProperty<T> {
+  return {
```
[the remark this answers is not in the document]
ssanjay1: Sure will FLUP

packages/client/src/createTimeseriesProperty.ts
ssanjay1: 1. Totally right, I'll do that
2. We had a getAll in legacy so I exposed for parity, I don't have enough data on usage here though to see if thats how people use it or not

PR #437, opened 2024-07-03
Fix client exports

ssanjay1:
When we refactored client.api, we accidentally stopped exporting certain entities from client, which broke internal usecases. Fixing with this


bryantpark04 (approved):
LGTM

PR #441, opened 2024-07-08
Interface inherited properties are now generated

ericanderson:
Also generating JSdoc for properties for both objects and interfaces.

### DIFF
--- a/examples-extra/basic/sdk/src/generatedNoCheck/ontology/objects/BoundariesUsState.ts
+++ b/examples-extra/basic/sdk/src/generatedNoCheck/ontology/objects/BoundariesUsState.ts
@@ -11,9 +11,22 @@ export interface BoundariesUsState
   primaryKeyApiName: 'usState';
   primaryKeyType: 'string';
   properties: {
+    /**
+     *   display name: 'Geometry10M',
+     *   description: geoshape
+     */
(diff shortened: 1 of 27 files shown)


examples-extra/basic/sdk/src/generatedNoCheck/ontology/objects/BoundariesUsState.ts
```diff
@@ -11,9 +11,22 @@ export interface BoundariesUsState
   primaryKeyApiName: 'usState';
   primaryKeyType: 'string';
   properties: {
+    /**
+     *   display name: 'Geometry10M',
+     *   description: geoshape
+     */
     geometry10M: PropertyDef<'geoshape', 'nullable', 'single'>;
+    /**
+     * (no ontology metadata)
```
ssanjay1, line 20: Do you feel like it'd be better to just not have anything here? Not sure how much no ontology metadata adds besides cluttering the generated files

ssanjay1 (approved):

examples-extra/basic/sdk/src/generatedNoCheck/ontology/objects/BoundariesUsState.ts
ericanderson: I went with this so there wouldn't be confusion about there possibly being a bug that there was no ontology data. Happy for us to change this in a follow up.

PR #442, opened 2024-07-09
Change representation of ont as code interface link types from both sides to just one

aep000:


packages/maker/src/api/overall.test.ts
```diff
@@ -312,109 +279,7 @@ describe("Ontology Defining", () => {
                   "icon": undefined,
                 },
                 "extendsInterfaces": [],
-                "links": [
```
ericanderson, line 315: Looks like the syntax is not creating anything for the links block?

ericanderson: Oops, expanded diff to see its just the one side.

ericanderson (approved):

PR #457, opened 2024-07-09
Change ClientCache key to be a field of the client

bryantpark04:


ericanderson (approved):
There is a weakmap for MinimalClient in ObjectSetListenerWebsocket that needs to be updated.

packages/client/src/MinimalClientContext.ts
```diff
@@ -19,11 +19,16 @@ import type { Logger } from "pino";
 import type { ObjectSetFactory } from "./objectSet/ObjectSetFactory.js";
 import type { OntologyProvider } from "./ontology/OntologyProvider.js";
 
+declare const tag: unique symbol;
+
+export type ClientCacheKey = {} & { readonly [tag]: void };
```
bryantpark04, line 24: is `void` the right type for the tag field? not sure if it matters since we're never actually assigning to it, just casting a `{}`

ericanderson: yeah its fine. the goal is just to make sure creating one requires manual effort (cast) and not that any object is assignable.

PR #459, opened 2024-07-09
add changeset config flag

ssanjay1:
Trying to fix dependency bumping issues


ericanderson:
Given our other changes I think its safe to abandon this.

PR #464, opened 2024-07-10
MRL enforces tilde not caret for internal deps

ericanderson:


.monorepolint.config.mjs
```diff
@@ -145,6 +201,8 @@ function getTsconfigOptions(baseTsconfigPath, opts) {
   };
 }
 
+// function make
```
ssanjay1, line 204: Do we need this?

.monorepolint.config.mjs
```diff
@@ -485,3 +546,37 @@ package you do so at your own risk.
...
+            context.addError({
+              message:
+                `Disallowed dependency \`${dep}: ${version}. Did you mean 'workspace:~'?`,
+              file: context.getPackageJsonPath(),
+            });
+          }
+        }
+      }
+    },
+    id: "asdf",
```
ssanjay1, line 573: nit: Maybe rename these?

ssanjay1 (approved):

PR #477, opened 2024-07-11
Restructure create app to make it easier to upgrade deps

ericanderson:
Changes:
- Re-enable diff contents output when things don't match so its easier to debug whats wrong
- Shift each template for create-app to its own package 
- Use codegen to package the template contents into a single variable so its easy to import from @osdk/create-app
- Use the package.json (not just the package.json.hbs) for the dependencies of the template. This allows dependabot to update the relevant values, without us needing to update the template by hand.

Thankfully, the existence of the `examples` directory to validate the output of the generators means I was able to ensure this did not change the output!

Note for reviewer:
- All of the `@osdk/create-app.template.*` folders are just:
  - migration of `packages/create-app/templates/X` -> `packages/create-app.template.X` (except X changed slightly).
  - removal of concrete dependencies from the `templates/package.json.hbs` file and put them in `package.json`
  - boilerplate
- All the real meat of the change exists in `packages/create-app/src` and the new package that has the codegen `packages/create-app.template-packager`.


.monorepolint.config.mjs
```diff
@@ -324,6 +326,30 @@ function standardPackageRules(shared, options) {
...
+
+This package contains templates for \`@osdk/create-app\`.
+
+The dependencies will come from this package's \`package.json\` (excluding \`@osdk/create-app.template-packager\`) and the rest of template is filled out from the \`templates\` directory.
+
+NOTE: DO NOT EDIT THIS README BY HAND. It is generated by monorepolint.
+`;
+        },
+      },
+    }),
```
ericanderson, line 344: Unified README's.

.vscode/settings.json
```diff
@@ -5,7 +5,7 @@
     "editor.defaultFormatter": "dprint.dprint"
   },
   "[typescript]": {
-    "editor.defaultFormatter": "vscode.typescript-language-features"
+    "editor.defaultFormatter": "dprint.dprint"
```
ericanderson, line 8: I'm not sure how this got changed but its doing the wrong formatting until now.

packages/create-app.template-packager/src/index.ts
```diff
@@ -0,0 +1,105 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { findUp } from "find-up";
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+import serialize from "serialize-javascript";
+
```
ericanderson, line 21: Code gen for each `@osdk/create-app.template.*/src/generatedNoCheck/`

packages/create-app/src/run.ts
```diff
@@ -79,16 +79,38 @@ export async function run(
 
   consola.info(`Copying files into project directory`);
```
ericanderson, line 80: Begin actual meat change for the template generation. Just writes from a map instead of cp from a directory.

packages/example-generator/src/run.ts
```diff
@@ -141,12 +142,49 @@ async function checkExamples(
         `Found ${compareResult.differences} differences in ${exampleId} please generate examples again.`,
       );
       consola.error(compareResult.diffSet?.filter(d => d.state !== "equal"));
+
```
ericanderson, line 145: Output proper diffs when things don't match

ssanjay1 (approved):

PR #478, opened 2024-07-12
Add final method call to request context header

bryantpark04:


packages/client.api/src/objectSet/ObjectSet.ts
```diff
@@ -51,6 +51,7 @@ export interface MinimalObjectSet<Q extends ObjectOrInterfaceDefinition>
     S extends NullabilityAdherence = NullabilityAdherenceDefault,
   >(
     args?: FetchPageArgs<Q, L, R, A, S>,
+    calledByAsyncIter?: boolean,
```
bryantpark04, line 54: Ah I need to figure out how to do this without changing the external API

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -141,11 +146,23 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
     asyncIter: async function*(): AsyncIterableIterator<Osdk<Q, "$all">> {
       let $nextPageToken: string | undefined = undefined;
       do {
-        const result = await base.fetchPage({ $nextPageToken });
```
bryantpark04, line 144: Was there a reason that we call `base.fetchPage` here instead of calling the unbound `fetchPageInternal`?

PR #484, opened 2024-07-14
Add interface support for foundry-sdk-generator

ssanjay1:


packages/foundry-sdk-generator/src/generate/betaClient/generatePackageJson.ts
```diff
@@ -22,32 +22,70 @@ export async function generatePackageJson(options: {
...
-    main: "./index.js",
-    types: "./index.d.ts",
-    exports: {
-      ".": {
-        types: "./index.d.ts",
-        script: {
-          types: "./dist/bundle/index.d.ts",
-          default: "./dist/bundle/index.esm.js",
+  const packageJson = options.beta
+    ? {
```
ssanjay1, line 29: I am not sure about this part yet

packages/foundry-sdk-generator/package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@osdk/api": "workspace:~",
     "@osdk/client": "workspace:~",
+    "@osdk/client.api": "workspace:~",
```
ssanjay1, line 35: Added this so that the generated packages from tests would correctly populate client.api version in the package.jsons

packages/foundry-sdk-generator/src/generate/betaClient/generatePackageJson.ts
```diff
@@ -22,32 +22,52 @@ export async function generatePackageJson(options: {
...
+          default: "./index.js",
         },
-        default: "./index.js",
       },
-      "./ontology/objects": {
-        types: "./ontology/objects/index.d.ts",
-        default: "./ontology/objects/index.js",
+      dependencies: packageDeps,
+    }
+    : {
```
ericanderson, line 50: The fallback block is unchanged right?

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -51,6 +52,12 @@ export class OntologyMetadataResolver {
       ) => expectedEntities.objectTypes.has(objectTypeApiName.toLowerCase())),
     );
 
+    const filteredInterfaceTypes = Object.fromEntries(
```
ericanderson, line 55: Can we only add these for 2.x? That way you have to get all the way to 2.0 to use them and we can reduce 1.x support burden.

packages/foundry-sdk-generator/src/generate/betaClient/generatePackageJson.ts
ssanjay1: Yes, besides me pulling out the dependencies into that helper function

ericanderson (approved):

bryantpark04 (approved):

PR #492, opened 2024-07-17
Introduce spell checking into builds

ericanderson:


.junk.cspell.json
```diff
@@ -0,0 +1,120 @@
+{
+  "$schema": "https://raw.githubusercontent.com/streetsidesoftware/cspell/main/cspell.schema.json",
```
ssanjay1, line 2: Why is this file labeled junk?

ssanjay1 (approved):

.junk.cspell.json
ericanderson: Oops, because I originally made it this as a config file but found it annoying to keep making all the changes for new dictionaries so then I broke it up into being code driven but didnt want to lose my work in case i did something wrong. Didn't mean to commit this file. Good catch.

PR #500, opened 2024-07-19
Fix primary key types

ssanjay1:


etc/client.api.report.api.md
```diff
@@ -483,7 +483,7 @@ export type Osdk<Q extends ObjectTypeDefinition<any> | InterfaceDefinition<any,
     __apiName: Q["apiName"] & {
         __OsdkType?: Q["apiName"];
     };
-    __primaryKey: Q extends ObjectTypeDefinition<any> ? OsdkObjectPrimaryKeyType<Q> : unknown;
+    __primaryKey: Q extends ObjectTypeDefinition<any> ? OsdkObjectPrimaryKeyType<Q> : string | number | boolean;
```
ericanderson, line 486: I know its technically correct that it can be a boolean but its just so absurd (a table with 2 rows) that it just won't happen.

ericanderson (approved):

PR #507, opened 2024-07-22
Add more exports to client

ssanjay1:


packages/client/src/index.ts
```diff
@@ -27,16 +27,19 @@ export type {
...
 
 export type { Client } from "./Client.js";
 export { createClient } from "./createClient.js";
 export { createPlatformClient } from "./createPlatformClient.js";
 
 export { createAttachmentFromRid } from "./createAttachmentFromRid.js";
+export { createAttachmentUpload } from "./object/AttachmentUpload.js";
 
 export { ActionValidationError } from "./actions/ActionValidationError.js";
-export { isOk } from "./ResultOrError.js";
```
ssanjay1, line 41: So we originally exporting this which I don't think we're actually using anywhere? https://github.com/palantir/osdk-ts/blob/3ad243443517d98469ed47959beb5ae5612fc67f/packages/client/src/ResultOrError.ts#L21 changed it to use this: https://github.com/palantir/osdk-ts/blob/3ad243443517d98469ed47959beb5ae5612fc67f/packages/client.api/src/object/Result.ts#L21

packages/client/src/index.ts
```diff
@@ -27,16 +27,19 @@ export type {
...
+
+export { isOk } from "@osdk/client.api";
 export { PalantirApiError } from "@osdk/shared.net.errors";
 
 export type { Client } from "./Client.js";
 export { createClient } from "./createClient.js";
 export { createPlatformClient } from "./createPlatformClient.js";
 
 export { createAttachmentFromRid } from "./createAttachmentFromRid.js";
+export { createAttachmentUpload } from "./object/AttachmentUpload.js";
```
ssanjay1, line 42: Need this if someone wants to upload a blob to an action. The `AttachmentUpload` type should also accept the File object from FileApi but we need to provide this helper for now because that doesn't come standard yet

packages/client/src/index.ts
```diff
@@ -27,16 +27,19 @@ export type {
   OsdkActionParameters,
   OsdkObject,
   PageResult,
+  Result,
   WhereClause,
 } from "@osdk/client.api";
+
+export { isOk } from "@osdk/client.api";
```
ssanjay1, line 34: See below on why this got swapped

packages/client/src/index.ts
```diff
@@ -27,16 +27,19 @@ export type {
   OsdkActionParameters,
   OsdkObject,
   PageResult,
+  Result,
```
ssanjay1, line 30: If someone uses `fetchPageWithError` the type is wrapped with `Result`, so thought it best to expose here

ericanderson (approved):

PR #509, opened 2024-07-23
Replace instances of /v2/security/ with /v2/admin/

bryantpark04:
Because regenerating platform SDK is broken right now


bryantpark04:
Proper fix in #510

PR #510, opened 2024-07-23
Fix foundry.security -> foundry.admin break

bryantpark04:
Fix SDK generation and regen SDK


packages/platform-sdk-generator/src/model/Model.ts
```diff
@@ -175,7 +175,10 @@ export class Model {
   }
 
   #addComponent(c: ir.Component) {
-    const ns = this.#namespaces.get(c.namespace ?? "");
+    const nsName = (c.namespace === undefined || c.namespace === "Core")
```
ssanjay1, line 178: lets do `==null`, read https://pages.github.palantir.build/palantir/documentation/#/gotham/web-standards/typescript/explicitly-check-null

bryantpark04, line 178: this is the SDK generation fix to handle the new Core namespace

ssanjay1 (approved):
Approving to unblock, let's address comments if possible

packages/platform-sdk-generator/src/model/Model.ts
bryantpark04: Will add in the legacy client long aggregation fix PR

PR #515, opened 2024-07-24
Rename to _gitignore inside templates

ewitkon:
similar to #160 we need to do this for the tutorial templates


zeyadkhaled (approved):
I think it is due to npm not the template bundle, so can update the PR title/Description accordingly. https://github.com/palantir/osdk-ts/pull/160

PR #525, opened 2024-07-25
Marks things as @internal to clean up API requirements

### LINKED ISSUE #527: OSDK packages do not compile correctly out of the box due to phantom dependencies on type dependencies
## Phantom Dependencies

Many packages in this repo publish publish TypeScript declaration files (`.d.ts`) with import statements to other dependencies.

For example, `@osdk/client` contains an import on `type-fest`.

```ts
// node_modules/.pnpm/@osdk+client@0.19.0/node_modules/@osdk/client/build/types/SatisfiesSemver.d.ts

import type { GreaterThan, GreaterThanOrEqual, IsEqual, LessThan } from "type-fest";
export type VersionString<Major extends number, Minor extends number, Patch extends number> = `${Major}.${Minor}.${Patch}`;
export type SatisfiesSemver<E extends VersionString<any, any, any>, C extends VersionString<any, any, any>> = E extends VersionString<infer ExpMajor, infer ExpMinor, infer ExpPatch> ? C extends VersionString<infer CurMajor, infer CurMinor, infer CurPatch> ? (IsEqual<CurMajor, ExpMajor> extends false ? false : LessThan<CurMinor, ExpMinor> extends true ? false : GreaterThan<CurMinor, ExpMinor> extends true ? true : GreaterThanOrEqual<CurPatch, ExpPatch>) : false : false;
//# sourceMappingURL=SatisfiesSemver.d.ts.map
```

A lot of these imported packages are not declared as dependencies. In this example, `@osdk/client` does not declare `type-fest` as a dependency. It's only present under the `devDependencies` of this package.

https://github.com/palantir/osdk-ts/blob/cd37441676b64b71ea674f48ab5d3b4e90962c7e/packages/client/package.json#L78

This means:

- Consumers of `@osdk/client` will see it fail to compile out of the box, forcing them to enable TypeScript's [`skipLibCheck`](https://www.typescriptlang.org/tsconfig/#skipLibCheck) option.
- If the consumer happens to have `type-fest` installed elsewhere in their repo, they might **get a different version** of this dependency than expected.

Here's a TypeScript compilation error due to a different version of `type-fest` being used in a consuming repo than expected.

Screenshot 2024-07-25 at 12 55 01 PM

This example isn't isolated to `@osdk/client` and `type-fest`. I believe many of the `devDependencies` of the packages in this repo should be under `dependencies` instead.

## Example: Problem

The undeclared dependencies (i.e. phantom dependencies) above can cause confusing type assignment and overload selection problems. I discovered this issue when debugging a problem where TypeScript kept choosing the wrong overload for `Client` when passing in an action.

```ts
    <Q extends (ObjectTypeDefinition<any, any> & VersionBound<any>)>(o: CheckVersionBound<Q>): ObjectSet<Q>;
    <Q extends ActionDefinition<any, any, any>>(o: CheckVersionBound<Q>): ActionSignatureFromDef<Q>;
}
```

Instead of choosing the `ActionDefinition` overload, TypeScript kept choosing the `ObjectTypeDefinition` overload and failing to compile. This ended up being due to `CheckVersionBound<Q>` evaluating to `any` under `skipLibCheck: true`. This is definitely incorrect.

Here's the first argument to `Client` evaluating to `any`:

Screenshot 2024-07-25 at 1 05 10 PM

## Suggestions

To improve type checking soundness for TypeScript consumers, can the OSDK packages declare their type-only imports as dependencies?

To check which dependencies need to be declared, I'd recommend importing `@osdk/client` (and other packages) into a new repo and ensure it compiles without `skipLibCheck` enabled.

## Caveats

Type-only dependencies will cause JavaScript-only consumers to download more bytes than necessary. However, most consumers of this repo will be using TypeScript. I think the improvements to type safety far outweigh a smaller dependency footprint.

ericanderson:
Fixes #527


etc/client.api.report.api.md
```diff
@@ -9,7 +9,6 @@ import type { ActionParameterDefinition } from '@osdk/api';
 import type { BBox } from 'geojson';
 import type { BrandedApiName } from '@osdk/api';
 import type { InterfaceDefinition } from '@osdk/api';
-import type { IsNever as IsNever_2 } from 'type-fest';
```
bryantpark04, line 12: what does renaming the types do that fixes the API leaks?

ericanderson: Nothing. Moving type-fest to a regular dependency is the real fix.

packages/client.api/src/object/FetchPageResult.ts
```diff
@@ -18,24 +18,23 @@ import type {
...
-import type { IsNever } from "type-fest";
 import type { DefaultToFalse } from "../definitions/LinkDefinitions.js";
-import type { Osdk } from "../OsdkObjectFrom.js";
+import type { IsNever, Osdk } from "../OsdkObjectFrom.js";
 import type { PageResult } from "../PageResult.js";
 import type { NullabilityAdherence } from "./FetchPageArgs.js";
 
-/** @internal exposed for a test */
+/** exposed for a test */
 export type RespectNullability<S extends NullabilityAdherence> = S extends false
```
ssanjay1, line 27: Wait why are we exposing these types?

ericanderson: Because anything marked @internal is no longer being included in the output. This is removing things like references to `@osdk/internal.*` from `@osdk/client`.

ssanjay1 (approved):

PR #530, opened 2024-07-25
Upgrade to vitest2

ericanderson:


.monorepolint.config.mjs
```diff
@@ -78,6 +78,7 @@ const esmOnlyPackages = [
   "@osdk/e2e.generated.catchall",
   "@osdk/e2e.sandbox.catchall",
   "@osdk/example-generator",
+  "@osdk/foundry-sdk-generator",
```
ericanderson, line 81: The generator is no longer "custom" and is just a basic esm only now.

.monorepolint.config.mjs
```diff
@@ -241,7 +242,7 @@ function getTsconfigOptions(baseTsconfigPath, opts) {
  *  legacy: boolean,
  *  esmOnly?: boolean,
  *  customTsconfigExcludes?: string[],
- *  tsVersion?: "^5.5.2"|"^4.9.5",
+ *  tsVersion?: typeof LATEST_TYPESCRIPT_DEP | "^4.9.5",
```
ericanderson, line 245: Just preventing future annoying behavior. Typescript doesn't actually check the .mjs files so everything was working here with any string but we were getting red in the IDE. This will at least prevent that mismatch.

.monorepolint.config.mjs
```diff
@@ -484,13 +485,12 @@ NOTE: DO NOT EDIT THIS README BY HAND. It is generated by monorepolint.
     }),
 
     ...standardPackageRules({
-      includePackages: ["@osdk/foundry-sdk-generator"],
+      includePackages: ["@osdk/e2e.test.foundry-sdk-generator"],
```
ericanderson, line 488: foundry-sdk-generator is no longer bespoke due to its generated code and e2e tests. But of course the e2e test is still bespoke at least for the generated code. It needs to be excluded because for some reason we are generating an entire package into this space. My goal with this PR was to migrate us to vitest2, not rethink this entire test suite.

packages/client/src/object/Cache.test.ts
```diff
@@ -89,20 +89,14 @@ describe("AsyncCache", () => {
   });
 
   describe("race checks", () => {
-    let factoryFn: Mock<[MinimalClient, string], Promise<any>>;
+    let factoryFn: Mock<AsyncFactory<any, any>>;
```
ericanderson, line 92: Honestly, this is a way better type api for vitest (taking the typeof a function instead of an array of args and a return value). Alas, it required minor changes in this file

packages/client/src/object/Cache.test.ts
```diff
@@ -274,10 +268,10 @@ describe("AsyncCache", () => {
           i++
         ) {
           if (asyncCacheSpies.get.mock.calls[i][1] === key) {
-            expect(asyncCacheSpies.get.mock.results[i].type).toBe(
-              "throw",
+            expect(asyncCacheSpies.get.mock.settledResults[i].type).toBe(
```
ericanderson, line 271: vitest in 1.x was very very strange wrt to how it handled functions that returned promises. In the old way, if the returned promise was resolved/rejected, the value itself was treated like the return value from the function. Otherwise it was treated as pending. Of course a function that returns a Promise could throw unrelated to that promise and you might want to test for that. So they changed the API to be far far more logical in terms of whats happening in the underlying system but that also broke all these carefully crafted tests.

`settledResults` is a lot like the former `results` except it only fills out AFTER the promises have been resolved/rejected. In almost all of our tests, this was a pure replacement but in some cases we had to change things.

packages/client/src/object/Cache.test.ts
```diff
@@ -396,8 +390,8 @@ describe("AsyncCache", () => {
               ]);
 
               itReturnsForAsyncGet([
-                expect.any(Promise),
-                expect.any(Promise),
+                undefined,
```
ericanderson, line 393: Before we could get an object that appeared to be a Promise from `.results` when the object was result was pending. So these `excpect.any(Promise)` would technically work. However now they split the settledResults out and as those first two promises are still pending, the result is undefined in vitest.

packages/e2e.test.foundry-sdk-generator/generateMockOntology.js
```diff
@@ -14,21 +14,29 @@
  * limitations under the License.
  */
 
+// @ts-check
```
ericanderson, line 17: This file is the old test setup file from `foundry-sdk-generator` except now its run as a script. I've taken care to try to keep as many lines as they were to keep the diff down. Thus you'll see things like `setup` and `teardown` still being used below so that indentation stays the same.

packages/e2e.test.foundry-sdk-generator/generateMockOntology.js
```diff
@@ -14,21 +14,29 @@
...
-import { rm } from "fs/promises";
-import { join } from "path";
-import { GeneratePackageCommand } from "../generate/index.js";
+import * as fs from "node:fs/promises";
+import * as path from "node:path";
+import { fileURLToPath } from "node:url";
 
-const dir = `${__dirname}/../generatedNoCheck/`;
-export async function setup() {
+const dir = path.join(
```
ericanderson, line 24: ESM doesn't have `__dirname`. This worked in vitest 1 because in reality vitest was converting all code to CJS before executing it and CJS has `__dirname`. However the tests are actually being run as ESM now and thus this breaks.

packages/e2e.test.foundry-sdk-generator/generateMockOntology.js
```diff
@@ -14,21 +14,29 @@
...
+  path.dirname(fileURLToPath(import.meta.url)),
+  "src",
+  "generatedNoCheck",
+);
+async function setup() {
   apiServer.listen();
 
-  try {
-    await rm(join(dir, "@test-app"), { recursive: true });
```
ericanderson, line 27: In the old code, this could throw causing the second rm to not run.

packages/e2e.test.foundry-sdk-generator/generateMockOntology.js
```diff
@@ -14,21 +14,29 @@
...
-    // Only needed for regenerations
-  }
+  const testAppDir = path.join(dir, "@test-app");
+  const testApp2Dir = path.join(dir, "@test-app2");
+
+  await rmRf(testAppDir);
+  await rmRf(testApp2Dir);
+
+  await safeStat(testAppDir, "should not exist");
+  await safeStat(testApp2Dir, "should not exist");
```
ericanderson, line 39: After upgrading to vitest 2, on initial runs of this test suite, everything would fail due to not being able to find the generated code. And then later when I was testing to find out what was happening I realized that if you bailed on `setup` early (after deleting) that it ALSO was failing. So I added some safety checks to be sure that things were being deleted (they were).

Turns out that in vitest 1, the global setup would run before any forks for the test runners would happen but in vitest 2, the forks happen right away. There also seemed to be some caching of the filesystem on initial load such that no matter what, on the first run, even though I could confirm we fully regenerated prior to the test running, the test would still think those files didn't exist.

I figured there was no harm in keeping the extra sanity checks after I chased the behavior down, so it remains.

packages/e2e.test.foundry-sdk-generator/generateMockOntology.js
```diff
@@ -107,10 +117,54 @@ export async function setup() {
...
   apiServer.close();
 }
+
+await setup();
+await teardown();
+
+/**
+ * @param {string} testAppDir
+ */
+async function rmRf(testAppDir) {
```
ericanderson, line 136: Split the deletion out so its more readable up above and so that the first failure doesn't cause the be ignored.

packages/e2e.test.foundry-sdk-generator/generateMockOntology.js
```diff
@@ -107,10 +117,54 @@ export async function setup() {
...
+    // Only needed for regenerations
+  }
+}
+
+/**
+ * @param {string} filePath
+ * @param {"should exist" | "should not exist"} type
+ * @returns
+ */
+async function safeStat(filePath, type) {
```
ericanderson, line 151: `fs.stat` throws if the file does not exist, which i generally don't care for as it makes ones code far more difficult to read, especially when you expect to be able to test for the non-existence of something. This helper both logs out whats going on as well as acting as an assert to clean up the rest of the codebase

packages/e2e.test.foundry-sdk-generator/package.json
```diff
@@ -9,34 +9,35 @@
...
     },
     "./*": {
-      "require": "./build/cjs/public/*.cjs",
       "browser": "./build/browser/public/*.js",
       "import": "./build/esm/public/*.js"
     }
   },
   "scripts": {
-    "check-attw": "monorepo.tool.attw both",
+    "check-attw": "monorepo.tool.attw esm",
```
ericanderson, line 21: This was a `both` before but didn't need to be. This simplifies logic and reduces build steps

packages/e2e.test.foundry-sdk-generator/package.json
```diff
@@ -9,34 +9,35 @@
...
     "transpile": "monorepo.tool.transpile",
-    "typecheck": "monorepo.tool.typecheck both"
+    "typecheck": "monorepo.tool.typecheck esm"
   },
   "dependencies": {
     "@osdk/foundry-sdk-generator": "workspace:~",
     "execa": "^9.3.0"
   },
   "devDependencies": {
+    "@osdk/legacy-client": "workspace:~",
```
ericanderson, line 36: Because we execute some of the generated code we need to have access to its dependencies in this project.

packages/e2e.test.foundry-sdk-generator/src/GeneratePackageCommand.test.ts
```diff
@@ -14,18 +14,18 @@
  * limitations under the License.
  */
 
+import { __testSeamOnly_NotSemverStable__GeneratePackageCommand as GeneratePackageCommand } from "@osdk/foundry-sdk-generator";
```
ericanderson, line 17: When moving the test code from a sibling to a cousin, we needed to be able to expose a test seam for GeneratePackageCommand. This name seemed scary enough for anyone that tries to use this package (no one should to be clear).

packages/e2e.test.foundry-sdk-generator/src/legacy-client/loadObjects.test.ts
```diff
@@ -123,7 +122,7 @@ describe("LoadObjects", () => {
 
     const emp = assertOkOrThrow(result);
 
-    expectTypeOf(emp).toEqualTypeOf<{
+    expectTypeOf(emp).branded.toEqualTypeOf<{
```
ericanderson, line 125: Because these were not actually being source code verified before (see below) this wasn't correct. `__apiName` became a branded string awhile ago. Which means these types arent TECHNICALLY correct. `.branded` looses things up seemingly stripping away brands allowing this check to keep working

packages/e2e.test.foundry-sdk-generator/src/legacy-client/search.test.ts
```diff
@@ -446,10 +447,12 @@ describe("SearchObjects", () => {
     const employees = assertOkOrThrow(result);
     expect(employees.data.length).toEqual(1);
     expect(employees.data[0].fullName).toEqual("John Doe");
-    expectTypeOf(employees.data[0]).toEqualTypeOf<{
+    expectTypeOf(employees.data[0]).branded.toEqualTypeOf<{
       readonly fullName: string | undefined;
       readonly __primaryKey: number;
       readonly __apiName: "Employee";
+      readonly $primaryKey: number;
```
ericanderson, line 454: The e2e tests in `foundry-sdk-generator` were not being typechecked before. For some reason they were excluded from the tsconfig. Thus when we did things like add `$primaryKey` to the output, we didn't know we broke tests. I didn't exclude them after the move so I had to fix the tests too

packages/e2e.test.foundry-sdk-generator/tsconfig.json
```diff
@@ -7,5 +7,8 @@
   "include": [
     "./src/**/*"
   ],
+  "exclude": [
+    "./src/generatedNoCheck/**/*"
```
ericanderson, line 11: This code that is generated just doesnt work in THIS project for type checking. It wasnt type checked before and isnt now to be consistent.

packages/e2e.test.foundry-sdk-generator/vitest.config.mts
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+import { dirname, join } from "path";
+import { fileURLToPath } from "url";
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    alias: {
```
ericanderson, line 23: this file basically migrated but it changed enough that git thinks its new

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -419,7 +419,7 @@ function createMockWebSocketConstructor(
         eventEmitter.removeEventListener.bind(eventEmitter),
       ) as any,
 
-      send: vi.fn((a, _b) => {
+      send: vi.fn((a, _b: any) => {
```
ssanjay1, line 422: Why'd you need to add a type for _b?

ssanjay1 (approved):
Approved, with one small question

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
ericanderson: in the old way of the type being args and return value, there was only one way to resolve the function. But now that its `typeof foo` and there  can be overloads, it can't infer properly what this value is because in one case its missing and in another it was defined. This was fine before, we didnt use it and i just told it about the defined value in the type but we don't have that any more.

.monorepolint.config.mjs
ericanderson: turns out this cant be true so i will be fixing this

PR #539, opened 2024-07-29
Manually restore omni versions and prevent future issues

ericanderson:
The omni's got hit by the peerDependencies paradox (for lack of a better name). We already oops over to 2.x for them and almost went to 3.x so this gets us back to a stable state and they won't major rev with client updates any more.

### DIFF
--- a/packages/foundry.admin/CHANGELOG.md
+++ b/packages/foundry.admin/CHANGELOG.md
@@ -1,24 +1,5 @@
 # @osdk/foundry.admin
 
-## 3.0.0-beta.1
-
-### Patch Changes
-
-- Updated dependencies [f91cd58]
-  - @osdk/client@0.21.0-beta.1
-  - @osdk/foundry.core@3.0.0-beta.1
-
-## 3.0.0-beta.0
-
-### Patch Changes
-
-- Updated dependencies [2deb4d9]
-- Updated dependencies [e54f413]
-- Updated dependencies [6387a92]
-- Updated dependencies [651c1b8]
-  - @osdk/client@0.21.0-beta.0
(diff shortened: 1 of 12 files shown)


packages/foundry.admin/CHANGELOG.md
```diff
@@ -1,24 +1,5 @@
...
-
-### Patch Changes
-
-- Updated dependencies [2deb4d9]
-- Updated dependencies [e54f413]
-- Updated dependencies [6387a92]
-- Updated dependencies [651c1b8]
-  - @osdk/client@0.21.0-beta.0
-  - @osdk/foundry.core@3.0.0-beta.0
-
```
ericanderson, line 21: Removing changelogs that don't matter

packages/foundry.admin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@osdk/foundry.admin",
-  "version": "3.0.0-beta.1",
```
ericanderson, line 3: Restoring version back to 2.0.0 so we don't ship this as 3.0.0 when we version next

packages/platform-sdk-generator/src/generatePlatformSdkv2.ts
```diff
@@ -224,11 +224,6 @@ const BASE_PACKAGE_JSON = {
     "typecheck": "tsc-absolute",
   },
   "dependencies": {},
-  "peerDependencies": {
```
ericanderson, line 227: Don't generate with peer deps for new packages of the platform sdk

ssanjay1 (approved):

PR #546, opened 2024-07-30
Change query invocation to applyQuery

ssanjay1:


packages/client/src/createClient.ts
```diff
@@ -54,6 +54,19 @@ class ActionInvoker<Q extends ActionDefinition<any, any, any>>
   applyAction: (...args: any[]) => any;
 }
 
+class QueryInvoker<Q extends QueryDefinition<any, any>>
```
ssanjay1, line 57: Basically kept in parity with what you did for actions

packages/client/src/ontology/OntologyProvider.ts
```diff
@@ -69,7 +70,8 @@ export interface OntologyProvider {
   maybeSeed: (
     definition:
       | ObjectOrInterfaceDefinition
-      | ActionDefinition<string, string, any>,
+      | ActionDefinition<string, string, any>
+      | QueryDefinition<any, any>,
```
ssanjay1, line 74: Added queries for ontology provider seeding

packages/client/src/createClient.ts
```diff
@@ -54,6 +54,19 @@ class ActionInvoker<Q extends ActionDefinition<any, any, any>>
...
+  implements QuerySignatureFromDef<Q>
+{
+  constructor(
+    clientCtx: MinimalClient,
+    queryDef: QueryDefinition<any, any>,
+  ) {
+    this.applyQuery = applyQuery.bind(undefined, clientCtx, queryDef);
+  }
+
+  applyQuery: (...args: any[]) => any;
```
ssanjay1, line 67: Using `applyQuery`, open to other suggestions though. Other ideas I thought of:

- executeQuery
- applyFunction
- executeFunction
Reason I even suggest `function` is because that's what our docs refer to queries as

packages/client/src/createClient.ts
ericanderson: Sweet. Thank you

ericanderson (approved):
Approved on condition of renaming the call per slack

PR #550, opened 2024-08-01
Deprecate legacy entities

ssanjay1:
Deprecating legacy entities

**What we deprecated**

- All base object set calls
- FoundryClient constructor
- Public and Confidential OAuth client constructors
- Any special property methods (e.g. for attachments and timeseries)
- All actions and queries


ericanderson (approved):
Approved but needs changeset

ericanderson:
Going to close this because its stale and we can reopen if we need it.

PR #555, opened 2024-08-06
Start adding Value Type scaffolding for maker

nathklei:
Adds ability to define value types, base property types that can have constraints and other metadata attached to allow reuse among property types and strong typing guarantees.

Example:
"Phone Number" is a value type, which defines a specific format of phone numbers.
"Phone Number" value type is the base type of "Home Phone", "Work Phone" and "Cell "Phone" shared property types.

Value types may be referenced by their name/version pair - this is slightly different from other primitives which are referenced just be name, as value types are not mutable within a given version (e.g. if the format of "Phone Number" is changed, a new version must be released and historical versions may still be referenced by property types).


packages/maker/src/api/defineValueType.ts
```diff
@@ -0,0 +1,46 @@
...
+import invariant from "tiny-invariant";
+import { valueTypes } from "./defineOntology.js";
+import type { ValueTypeBaseType, ValueTypeReference } from "./types.js";
+
+export function defineSharedPropertyType(
+  opts: {
+    apiName: string;
+    version: string;
+    type: ValueTypeBaseType;
+    array?: boolean;
```
nathklei, line 26: bugfix: actually use

ericanderson:
Going to close this. feel free to reopen if you need it later.

PR #551, opened 2024-08-01
Initial api namespace support

ericanderson:

PR #559, opened 2024-08-13
Simplify generated action params and mark readonly

ericanderson:
Chains to #551. Broken into 2 commits (cpu/human) for review simplicity.

Notable changes:
- `{{actionApiName}}$Params` is deprecated in favor of `ActionParams${{actionApiName}}`.
- All generated `{{actionApiName}}$Params` objects are now exported from generated code.
- All `{{actionApiName}}$Params` are marked as `readonly`.
- Some types that are now only needed in `@osdk/client` have been moved back out of `@osdk/client.api`.
- Generated `ActionParams${{actionApiName}}` are simpler and do not rely on type mapping for the keys, the array'ness, nor multiplicity.
- `AttachmentUpload.name` is now `readonly`.


packages/client.api/src/actions/Actions.ts
```diff
@@ -43,53 +33,29 @@ export type ApplyActionOptions =
...
->(
-  args: A,
-  options?: OP,
-) => Promise<
-  ActionReturnTypeForOptions<OP>
->;
+/**
+ * Helper types for converting action definition parameter types to typescript types
+ */
+export namespace ActionParam {
```
ericanderson, line 39: I know it may seem like I'm anti-namespace but I am actually okay with them if they are only types.

packages/client.api/src/actions/Actions.ts
```diff
@@ -43,53 +33,29 @@ export type ApplyActionOptions =
 
 export type ApplyBatchActionOptions = { $returnEdits?: boolean };
 
-type BaseType<APD extends ActionParameterDefinition<any, any>> =
```
ericanderson, line 46: All of this read moved to @osdk/client

packages/client.api/src/index.ts
```diff
@@ -119,14 +117,14 @@ export type {
...
 export type {
   TimeSeriesPoint,
   TimeSeriesProperty,
   TimeSeriesQuery,
 } from "./timeseries/timeseries.js";
-export { TimeseriesDurationMapping } from "./timeseries/timeseries.js";
-
 export type { LinkedType, LinkNames } from "./util/LinkUtils.js";
 export type { NOOP } from "./util/NOOP.js";
+export type { PartialBy } from "./util/PartialBy.js";
```
ericanderson, line 128: Exposed for @osdk/client. I think this actually goes away once I do the same treatment to queries that I just did to actions.

packages/client/src/actions/actions.test.ts
```diff
@@ -143,12 +143,18 @@ describe("actions", () => {
     const clientBoundActionTakesAttachment = client(
       actionTakesAttachment,
     ).applyAction;
-    expectTypeOf<Parameters<typeof clientBoundActionTakesAttachment>[0]>()
```
ericanderson, line 146: Sadly, after converting the params to take a `ReadonlyArray<X>` this started evaluating the second half of the union to `never[]`. 

Since we actually only need assignability in the opposite direction, I flipped it in the tests.

packages/client/src/actions/applyAction.ts
```diff
@@ -14,21 +14,77 @@
...
-import { Ontologies, OntologiesV2 } from "@osdk/internal.foundry";
+import { OntologiesV2 } from "@osdk/internal.foundry";
 import type { MinimalClient } from "../MinimalClientContext.js";
 import { addUserAgentAndRequestContextHeaders } from "../util/addUserAgentAndRequestContextHeaders.js";
 import { augmentRequestContext } from "../util/augmentRequestContext.js";
+import type { NullableProps } from "../util/NullableProps.js";
 import { toDataValue } from "../util/toDataValue.js";
 import { ActionValidationError } from "./ActionValidationError.js";
 
+type BaseType<APD extends Pick<ActionParameterDefinition<any, any>, "type">> =
```
ericanderson, line 41: All these types moved back from @client.api needed a home and this seemed as good as any.

packages/generator/src/shared/generatePerActionDataFiles.ts
```diff
@@ -66,21 +67,22 @@ export async function generatePerActionDataFiles(
 
       const { parameters, ...actionDefSansParameters } = fullActionDef;
 
-      const actionDefIdentifier = `ActionDef$${action.shortApiName}`;
-      const paramsDefIdentifier = `${actionDefIdentifier}$Params`;
-      const paramsIdentifier = `${action.shortApiName}$Params`;
+      // const actionDefIdentifier = `ActionDef$${action.shortApiName}`;
```
ssanjay1, line 70: Assuming you don't want these comments

examples-extra/docs_example/src/generatedNoCheck/ontology/actions/moveOffice.ts
```diff
@@ -36,21 +35,39 @@ export type ActionDef$moveOffice$Params = {
...
+  /**
+   * A list of all office names
+   */
+  readonly officeNames?: ReadonlyArray<ActionParam.PrimitiveType<'integer'>>;
+}
+
+/**
+ * @deprecated Use `ActionParams$moveOffice`
+ */
+export type moveOffice$Params = ActionParams$moveOffice | ReadonlyArray<ActionParams$moveOffice>;
```
ssanjay1, line 61: Little confused by these, these are using your new types right, just unioning the array or singular parameters case. Why do we want to deprecate this?

ssanjay1: Like I know its the old identifier, but seems fine because it's using you're new types?

ssanjay1 (approved):
Approved with one question and one nit

examples-extra/docs_example/src/generatedNoCheck/ontology/actions/moveOffice.ts
ericanderson: I didn't want to break the internal code base right out. This should  make it easier for us to upgrade to an alpha and regex fix the internal usages in separate steps.

PR #578, opened 2024-08-16
Cleans up generated queries

ericanderson:


packages/generator/src/GenerateContext/EnhancedQuery.ts
```diff
@@ -22,10 +22,28 @@ export class EnhancedQuery extends EnhancedBase<QueryTypeV2> {
...
   get output() {
     return this.og.output;
   }
+
+  get paramsIdentifier() {
+    return `QueryParams$${this.shortApiName}`;
+  }
+
+  get definitionIdentifier() {
+    return `ActionDef$${this.shortApiName}`;
```
ssanjay1, line 43: This should probably be `QueryDef` right?

ssanjay1 (approved):
Approved, but left one comment that I think we should fix before merging

packages/generator/src/GenerateContext/EnhancedQuery.ts
ericanderson: Nice. Wasnt used out of file so nothign broke but good to have right. Thank you for catching that.

PR #601, opened 2024-08-28
Add Exact Distinct Aggregations to Object Generation

beloiual:
## Before this PR

With the added support of Exact Distinct Aggregations in the API-Gateway, customers have been wanting to get accurate Exact cardinalities from their OSDKs. This was not supported however in the ts-osdk

## After this PR

With this PR, you can now perform Exact Distinct Aggregations in the same way you perform Approximate Distinct Aggregations.

Use:
```
const AircraftAggregate = await client.ontology.objects.Aircraft
    .aggregate(obj => ({
        metricName: obj.lifetimeHours.exactDistinct(),
    }))
    .compute()
```
In the example you can call on each property a aggregation 

==COMMIT_MSG==
Add Exact Distinct Aggregations to Object Generation
==COMMIT_MSG==

## Are Docs needed?

Need to add docs in Dev Console


packages/legacy-client/src/client/objectSets/aggregations/metrics/DistinctCountAggregatableProperty.ts
```diff
@@ -20,15 +20,25 @@ import type {
 } from "../Aggregations.js";
 import { MetricValueType } from "./metrics.js";
 
-export interface ApproximateDistinctCountAggregatableProperty {
```
ericanderson, line 23: I think this is technically a breaking change.

packages/e2e.generated.1.1.x/src/generatedNoCheck/internal/@foundry/ontology-runtime/dist/aggregations/metrics/ApproximateDistinctCountAggregatableProperty.ts
```diff
@@ -1,7 +0,0 @@
-import { ApproximateDistinctCountAggregatableProperty as OG_ApproximateDistinctCountAggregatableProperty } from '@osdk/legacy-client';
```
ericanderson, line 1: I think this is a breaking change.

packages/legacy-client/src/index.ts
```diff
@@ -18,7 +18,6 @@ export {
   ActionExecutionMode,
   ActionResponse,
   ActionValidationResult,
-  ApproximateDistinctCountAggregatableProperty,
```
ericanderson, line 21: Yeah, this definitely a breaking change.

ericanderson (changes requested):

ericanderson (approved):

PR #605, opened 2024-08-29
Add TS compatible React template 

ewitkon:
TS2.0 introduce not just new syntax but also a bit of a different file structure and auth process, if you generate using `create-app` a TS2.0 package, the output project does not compile, this PR fixes this.

1. Add a `create-app.template.react.2.0` - template
2. Update `generate-template` to use a 2.0 compatible package 
3. add `useBeta` option to the create-app cli which will only show templates which supports beta


ewitkon:
Currently blocked on:
https://palantir.slack.com/archives/C062JTAA9AP/p1724964583171239

```
Do we have a @osdk/e2e.generated.1.1.x equivalent for 2.0? currently I can't import ontology from it
import { Ontology } from "@osdk/e2e.generated.1.1.x";
2.  Can we make actions and queries enumerable like objects?
  const objectApiNames = Object.keys(Ontology.objects);
  const actionApiNames = Object.keys(client(Ontology.actions));
  const queryApiNames = Object.keys(client(Ontology.queries));
this works for objects but doesn't compile for actions and queries (edited) 
```

ewitkon:
> Currently blocked on: https://palantir.slack.com/archives/C062JTAA9AP/p1724964583171239
> 
> ```
> Do we have a @osdk/e2e.generated.1.1.x equivalent for 2.0? currently I can't import ontology from it
> import { Ontology } from "@osdk/e2e.generated.1.1.x";
> 2.  Can we make actions and queries enumerable like objects?
>   const objectApiNames = Object.keys(Ontology.objects);
>   const actionApiNames = Object.keys(client(Ontology.actions));
>   const queryApiNames = Object.keys(client(Ontology.queries));
> this works for objects but doesn't compile for actions and queries (edited) 
> ```

@tzyl  This ^ is no longer blocking but we need to think what we want to show in the default home screen. - Currently we don't support enumerating the objects, actions and functions like before 2.0.

@ekang98 for SA, depending how hard is it to add such enumeration

nadprana (approved):
Approving on the useBeta tag while we wait for the template changes

tzyl (commented):
Some high level comments:

- If we are going with the option flag can you make it consistent as `--beta` rather than `--useBeta`?
- Please can you make sure the template is up to date with latest changes for example with the env test https://github.com/palantir/osdk-ts/pull/607
- I would suggest avoid prompting for beta and only allow this as an entrypoint when specified through the command line directly as this isn't really a user facing concept the user will know how to answer with zero context so it can just default to false

ewitkon:
> Some high level comments:
> 
> * If we are going with the option flag can you make it consistent as `--beta` rather than `--useBeta`?
> * Please can you make sure the template is up to date with latest changes for example with the env test [Include default tests for .env.production in templates #607](https://github.com/palantir/osdk-ts/pull/607)
> * I would suggest avoid prompting for beta and only allow this as an entrypoint when specified through the command line directly as this isn't really a user facing concept the user will know how to answer with zero context so it can just default to false

Done in https://github.com/palantir/osdk-ts/pull/634

dev-docs/design-rationale.md
```diff
@@ -6,23 +6,23 @@ The v1 of the OSDK demonstrated a significant amount of value for statically acc
 
 ### Build size
 
-The original OSDK involved generating full implementations of code for each object/link/property etc. This meant the size of the OSDK scaled linearly with the size of the ontology, which could get quite large. Further, the design meant that the entirety of the generated OSDK had to be loaded into the browser/javascript engine's memory for usage. 
+The original OSDK involved generating full implementations of code for each object/link/property etc. This meant the size of the OSDK scaled linearly with the size of the ontology, which could get quite large. Further, the design meant that the entirety of the generated OSDK had to be loaded into the browser/javascript engine's memory for usage.
```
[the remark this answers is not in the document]
ewitkon: Fixed.

examples-extra/docs_example/src/App.tsx
```diff
@@ -1,10 +1,10 @@
-import { useState } from 'react'
-import reactLogo from './assets/react.svg'
-import viteLogo from '/vite.svg'
-import './App.css'
+import viteLogo from "/vite.svg";
```
ssanjay1, line 1: Is there a reason you changed these docs/example files?

ssanjay1 (approved):

examples-extra/docs_example/src/App.tsx
ewitkon: Fixed

PR #609, opened 2024-09-04
Split actions and batch actions and simplify usage

### LINKED ISSUE #558: Split actions and batch actions in 2.0

ericanderson:
Fixes #558

### DIFF
--- a/packages/generator/src/GenerateContext/EnhancedAction.ts
+++ b/packages/generator/src/GenerateContext/EnhancedAction.ts
@@ -36,7 +36,7 @@ export class EnhancedAction extends EnhancedBase<ActionTypeV2> {
   }
 
   get paramsIdentifier() {
-    return `${this.shortApiName}.Parameters`;
+    return `${this.shortApiName}.Params`;
   }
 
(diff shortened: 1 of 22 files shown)


packages/generator/src/GenerateContext/EnhancedAction.ts
```diff
@@ -36,7 +36,7 @@ export class EnhancedAction extends EnhancedBase<ActionTypeV2> {
   }
 
   get paramsIdentifier() {
-    return `${this.shortApiName}.Parameters`;
+    return `${this.shortApiName}.Params`;
```
ssanjay1, line 39: Fine with the rename, but just wondering why?

ssanjay1 (approved):

packages/generator/src/GenerateContext/EnhancedAction.ts
ericanderson: Got annoyed in the internal repo typing out the long string. Imagine if we used `Properties` instead of `Props` for all of our react components?

PR #610, opened 2024-09-04
Fix object property selection inference

ericanderson:
Chains to #558 

We were in a scenario where the following would compile but should fail:
```ts
function asdf(client: Client): Promise<Osdk<Employee>> {
  return client(Employee).fetchOne(1, {
    $select: ["employeeId"],
  });
}
```

### DIFF
--- a/packages/client.api/src/objectSet/ObjectSet.ts
+++ b/packages/client.api/src/objectSet/ObjectSet.ts
@@ -201,9 +201,9 @@ export interface ObjectSet<
    * Fetches one object with the specified primary key, without a result wrapper
    */
   readonly fetchOne: Q extends ObjectTypeDefinition<any> ? <
-      L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
-      R extends boolean,
-      S extends false | "throw" = NullabilityAdherenceDefault,
+      const L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
+      const R extends boolean,
+      const S extends false | "throw" = NullabilityAdherenceDefault,
     >(
       primaryKey: PropertyValueClientToWire[Q["primaryKeyType"]],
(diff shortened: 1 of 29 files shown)


packages/client/src/Client.ts
```diff
@@ -39,7 +40,7 @@ export type CheckVersionBound<Q> = Q extends VersionBound<infer V> ? (
 export interface Client extends SharedClient<MinimalClient> {
   <Q extends ObjectTypeDefinition<any, any>>(
     o: Q,
-  ): Q["objectSet"];
+  ): unknown extends Q["objectSet"] ? ObjectSet<Q> : Q["objectSet"];
```
ericanderson, line 43: Fixes an edge case with an older generation not working.

packages/client.api/src/objectSet/ObjectSet.ts
```diff
@@ -201,9 +201,9 @@ export interface ObjectSet<
    * Fetches one object with the specified primary key, without a result wrapper
    */
   readonly fetchOne: Q extends ObjectTypeDefinition<any> ? <
-      L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
-      R extends boolean,
-      S extends false | "throw" = NullabilityAdherenceDefault,
+      const L extends ObjectOrInterfacePropertyKeysFrom2<Q>,
```
ssanjay1, line 204: This is to fix that edge case you found in return statements right?

ssanjay1 (approved):

packages/client.api/src/objectSet/ObjectSet.ts
ericanderson: Yeah. This made tsc be less permissive when trying to align a defined return value from a function with the inferred generics it used.

PR #611, opened 2024-09-04
Add support for extra nullable field on properties

ericanderson:
One of the really old internal first uses hard coded all properties not to be null. I would love for the gateway to expose nullability to us but if we can't then we can at least make it possible to specify in the raw ontology.json file.

Thats what this does. I may hold off until I know if we will add this to gateway proper (and if not perhaps give it an unlikely name so it won't collide in the future).

### DIFF
--- a/packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
+++ b/packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
@@ -25,7 +25,7 @@ import type {
 } from "@osdk/gateway/types";
 
 export function wirePropertyV2ToSdkPropertyDefinition(
-  input: PropertyV2 | SharedPropertyType,
+  input: (PropertyV2 | SharedPropertyType) & { nullable?: boolean },
   isNullable: boolean = true,
 ): ObjectTypePropertyDefinition {
   switch (input.dataType.type) {
@@ -50,7 +50,7 @@ export function wirePropertyV2ToSdkPropertyDefinition(
         multiplicity: false,
         description: input.description,
         type: objectPropertyTypeToSdkPropertyDefinition(input.dataType),
-        nullable: isNullable,
+        nullable: input.nullable == null ? isNullable : input.nullable,
       };
     case "array": {
       return {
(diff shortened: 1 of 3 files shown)


packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
```diff
@@ -25,7 +25,7 @@ import type {
 } from "@osdk/gateway/types";
 
 export function wirePropertyV2ToSdkPropertyDefinition(
-  input: PropertyV2 | SharedPropertyType,
+  input: (PropertyV2 | SharedPropertyType) & { nullable?: boolean },
   isNullable: boolean = true,
```
nihalbhatnagar, line 29: Will we eventually be able to remove the isNullable prop entirely, or are there always cases where we need to specify nullability for certain properties

ssanjay1: I'd imagine so eventually once the backend gives us all the necessary info we could. If you look through our repo for usages of this function, you'll see that either we pass in 

1. blanket `True` for nullable
2. `!(v2 && objectTypeWithLink.objectType.primaryKey === key),` which basically just says if this isn't a primary key, then it could potentially be null

But, once we can expose info on the backend in the property definitions, we can look at each input property and see if its set to nullable or not without just assuming everything is nullable on our end.

ssanjay1 (approved):

packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
ericanderson: Exactly. So with this hack I can at least upgrade really old internal osdk 2.0 usages to use the cli generation to:
1. Generate the code + fetch the ontology.json
2. Run a script to force the properties to `nullable: false` which they were effectively doing by hand
3. Re generate the code without fetching

Otherwise I have to litter their code base with `obj.prop!`. At the end of the day its the EXACT same net effect (we are ignoring the `| undefined` but I don't have to put in a massive delta.

PR #618, opened 2024-09-06
Improve ActionResults types

### LINKED ISSUE #576: Improve types of actionresults
1. Always exposed `editedObjectTypes`
2. remove `edits` from `ObjectEdits`
3. spread `ObjectEdit` into arrays in `ObjectEdits` (`addedObjects`, `modifiedObjects`, `addedLInks`)

nihalbhatnagar:
Closes #576

### DIFF
--- a/packages/client/src/actions/actions.test.ts
+++ b/packages/client/src/actions/actions.test.ts
@@ -65,17 +65,19 @@ describe("actions", () => {
     expectTypeOf<typeof result>().toEqualTypeOf<ActionEditResponse>();
     expect(result).toMatchInlineSnapshot(`
       {
+        "addedLinks": [],
         "addedLinksCount": 0,
         "addedObjectCount": 2,
-        "deletedLinksCount": 0,
-        "deletedObjectsCount": 0,
-        "edits": [
+        "addedObjects": [
           {
             "objectType": "Office",
             "primaryKey": "NYC",
             "type": "addObject",
           },
         ],
+        "deletedLinksCount": 0,
+        "deletedObjectsCount": 0,
+        "modifiedObjects": [],
         "modifiedObjectsCount": 0,
         "type": "edits",
       }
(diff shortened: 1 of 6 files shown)


packages/client/src/actions/applyAction.ts
```diff
@@ -202,3 +209,28 @@ async function remapBatchActionParams<
...
+      modifiedObjects: [],
+    };
+    for (const edit of editResponses.edits) {
+      if (edit.type === "addLink") {
+        remappedActionResponse.addedLinks.push(edit);
+      } else if (edit.type === "addObject") {
+        remappedActionResponse.addedObjects.push(edit);
+      } else if (edit.type === "modifyObject") {
+        remappedActionResponse.modifiedObjects.push(edit);
+      }
```
nihalbhatnagar, line 232: Do we need an assertNever here?

packages/client/src/actions/applyAction.ts
```diff
@@ -202,3 +209,28 @@ async function remapBatchActionParams<
...
   return remappedParams;
 }
+
+function remapActionResponse(
+  response: SyncApplyActionResponseV2 | BatchApplyActionResponseV2,
+): ActionEditResponse | undefined {
+  const editResponses = response?.edits;
+  if (editResponses?.type === "edits") {
+    const { edits, ...editResponse } = editResponses;
+    const remappedActionResponse: ActionEditResponse = {
```
nihalbhatnagar, line 219: Not sure if there is a cleaner way to do this

packages/client/src/actions/actions.test.ts
```diff
@@ -218,25 +220,28 @@ describe("actions", () => {
...
+            "objectType": "Office",
+            "primaryKey": "NYC",
+            "type": "modifyObject",
+          },
+        ],
+        "type": "edits",
+      }
+  `);
   });
 });
```
nihalbhatnagar, line 247: FLUP: Might be good to write some tests/stubs that can handle the return of largeScaleEdits to flesh out the tests

packages/client/src/actions/actions.test.ts
```diff
@@ -65,18 +65,20 @@ describe("actions", () => {
     expectTypeOf<typeof result>().toEqualTypeOf<ActionEditResponse>();
     expect(result).toMatchInlineSnapshot(`
       {
-        "addedLinksCount": 0,
-        "addedObjectCount": 2,
```
ericanderson, line 69: This count is different. Was the mock creating bad data before?

packages/client/src/actions/actions.test.ts
ericanderson: Good idea!

packages/client/src/actions/applyAction.ts
ericanderson: Can you elaborate on your concern more?

packages/client/src/actions/applyAction.ts
ericanderson: It wouldn't hurt. Or an `else { invariant(false, "Unknown edit type") }`.

packages/client/src/actions/applyAction.ts
nihalbhatnagar: I was a bit worried while that creating the object and casting it to the ActionEditResponse type, which is the client type rather than the wire type, but looking back I think this is the only way and most typesafe way to do this

packages/client/src/actions/actions.test.ts
nihalbhatnagar: I'm not sure why this is 2, will look into it

nihalbhatnagar: Looks like the stub was mistakenly returning 1 even when only 1 edit was applied, went ahead and fixed it even though we dont touch it

ericanderson (approved):
Pre-emptive approve

ssanjay1 (approved):

PR #619, opened 2024-09-09
Adding value types to maker

aep000:
Adds support for value types. 

Includes TONS of generated files for the data types. Not sure all of them are necessary. Maybe we should prune them. 

Meat of the PR is in the maker package with a few real changes to the blockDataIr files.


packages/maker/src/api/defineOntology.ts
```diff
@@ -59,10 +67,35 @@ export async function defineOntology(
...
+
+function convertOntologyToValueTypeIr(
+  ontology: Ontology,
+): OntologyIrValueTypeBlockData {
+  return {
+    valueTypes: Object.values(ontology.valueTypes).map<
+      OntologyIrValueTypeBlockDataEntry
+    >(definitions => ({
+      metadata: {
+        apiName: definitions[0].apiName,
```
nathklei, line 84: do we order them to have "newest" version first, I guess?

packages/maker/src/api/defineSpt.ts
```diff
@@ -25,6 +26,7 @@ export function defineSharedPropertyType(
     array?: boolean;
     description?: string;
     displayName?: string;
+    valueType?: ApiNameValueTypeReference;
```
nathklei, line 29: should we add an invariant that the value type doesn't exist, similar to the SPT already exists?

also - if we provide valueType, we can omit type/array right? Rather than force repeating them

aep000: Well we collapse them based on version. I guess we should check if you are adding 2 spts at the same version

ericanderson (dismissed):

packages/client.unstable/src/index.ts
```diff
@@ -36,10 +36,26 @@ export { bulkLoadOntologyEntities } from "./generated/ontology-metadata/api/Onto
...
+export type { DataConstraint } from "./generated/type-registry/api/DataConstraint.js";
+export type { DataConstraintWrapper } from "./generated/type-registry/api/DataConstraintWrapper.js";
+export type { ExampleValue } from "./generated/type-registry/api/ExampleValue.js";
+export type { FailureMessage } from "./generated/type-registry/api/FailureMessage.js";
+export type { ValueTypeApiName } from "./generated/type-registry/api/ValueTypeApiName.js";
+export type { ValueTypeDataConstraint } from "./generated/type-registry/api/ValueTypeDataConstraint.js";
+export type { ValueTypeDisplayMetadata } from "./generated/type-registry/api/ValueTypeDisplayMetadata.js";
+export type { ValueTypeStatus } from "./generated/type-registry/api/ValueTypeStatus.js";
+export type { ValueTypeVersion } from "./generated/type-registry/api/ValueTypeVersion.js";
+
```
ericanderson, line 49: Thank you for only exporting what was required (and types only it seems). Keeps the tree shaking very effective.

ericanderson (approved):

PR #630, opened 2024-09-09
Delete legacy client

ericanderson:
Chains to #629


packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
```diff
@@ -82,12 +72,7 @@ export async function generatePackage(
   };
 
   if (!options.beta) {
-    await generateClientSdkVersionOneDotOne(
-      ontology,
-      `typescript-sdk/${options.packageVersion} ${USER_AGENT}`,
-      hostFs,
-      packagePath,
-    );
+    throw new Error("Only beta is supported in this version");
```
ssanjay1, line 75: Once we have the ability to just get our beta releases into dev console, we can just remove this flag right?

packages/cli.cmd.typescript/src/generate/handleGenerate.mts
```diff
@@ -256,17 +254,15 @@ async function generateSourceFiles(
...
-    : generateClientSdkVersionOneDotOne)(
-      ontology,
-      getUserAgent(args.version),
-      fs,
-      args.outDir,
-      args.packageType,
-      args.ontologyApiNamespace,
-      args.apiNamespaceMap,
-    );
+  await generateClientSdkVersionTwoPointZero(
```
ssanjay1, line 257: Not related to this, but looking at `handleGenerate` in this file, should we still have a beta flag for it?

ssanjay1 (approved):

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
ericanderson: Correct, this is just hold over

packages/cli.cmd.typescript/src/generate/handleGenerate.mts
ericanderson: good call , we can clean that up now too

PR #636, opened 2024-09-10
Autofill Blob Headers in Platform SDK

nihalbhatnagar:


packages/client/src/ontology/loadFullObjectMetadata.ts
nihalbhatnagar, line 1: Had to make this change do to a change over the wire. Underlying API URL's remained consistent

ericanderson: Makes sense, they are still evolving some of these.

packages/client/src/util/toDataValue.ts
```diff
@@ -55,10 +55,6 @@ export async function toDataValue(
       {
         filename: value.name,
       },
-      {
-        "Content-Length": value.size.toString(),
-        "Content-Type": value.type,
```
ericanderson, line 60: So obviously in this case, we just want to use whatever the blob's content-type is. After all, we're asking it.

However, the foundry platform sdk is used by customers too and they may at some point end up with a blob that has a content type they don't want. Lets say they get an upload for a json file but they need the content type in the system to be something specific? 

Therefore, I would like to raise the question: Should we make `Content-Type` optional and fill it with the blob value if not provided?

nihalbhatnagar: I see the use case, and I probably agree with the solution. The alternative would be to just enforce the users attach a Blob with the correct type attached, but that's probably more overhead than we should pass to the consumers and also requires them to make a copy. I don't think there's a scenario in which the Blob's actual attached content type is transmitted through the wire or used in the generated SDK's, so this should be fine

ericanderson (approved):

packages/foundry.core/src/_components.ts
```diff
@@ -233,8 +233,7 @@ export type TimeUnit =
   | "DAYS"
   | "WEEKS"
   | "MONTHS"
-  | "YEARS"
-  | "QUARTERS";
+  | "YEARS";
```
ssanjay1, line 236: Could you track as a FLUP to see if we need to remove support for `QUARTERS` in our internal time unit types?

packages/platform-sdk-generator/src/generateResource2.ts
```diff
@@ -66,21 +68,38 @@ async function generateMethods(resource: Resource, model: Model) {
...
+        && shouldFillBlobHeaders
+        ? blobHeaders.autofill
+        : ""
+    }return $foundryPlatformFetch($ctx, _${methodName}, ${
+      shouldFillBlobHeaders ? blobHeaders.returnArgs : "...args"
+    }); }
 
     `;
   }
+
```
ssanjay1, line 102: nit:  can we remove this newline

ssanjay1 (approved):

PR #640, opened 2024-09-10
Fetching interfaces learns subselect by object type

ericanderson:


packages/e2e.sandbox.catchall/src/runInterfacesTest.ts
```diff
@@ -31,6 +31,16 @@ export async function runInterfacesTest() {
 
   const qqq = await client(FooInterface).where({ name: { $ne: "Patti" } });
 
+  const fooLimitedToEmployees = await client(FooInterface).fetchPage({
+    $__EXPERIMENTAL_selectedObjectTypes: ["Employee"],
```
ssanjay1, line 35: So we can't really do any checks to make sure they're not subselecting down to an invalid object type can we?

ssanjay1 (approved):
Approved with one question

packages/e2e.sandbox.catchall/src/runInterfacesTest.ts
ericanderson: We might be able to do something in the future but we also might want to leave this open ended. Solving the immediate need by making it experimental

PR #643, opened 2024-09-11
SLS dependencies are optional

ericanderson:


packages/e2e.sandbox.catchall/src/index.ts
```diff
@@ -33,7 +32,6 @@ const testSubscriptions = false;
 
 async function runTests() {
   try {
-    await runFoundryPlatformApiTest();
```
ssanjay1, line 36: Why'd we delete this?

ssanjay1 (approved):

packages/e2e.sandbox.catchall/src/index.ts
ericanderson: The apis that test used are no longer generated any more. We have other tests for the platform api so no loss.

PR #648, opened 2024-09-11
Add Exact-Distinct Aggregations

nihalbhatnagar:


packages/client.api/src/aggregate/AggregationsResults.test.ts
```diff
@@ -77,9 +79,11 @@ type Q = AggregationResultsWithoutGroups<
   }["objects"]["Todo"],
   {
     "locationCity:approximateDistinct": "unordered";
+    "locationCity:exactDistinct": "unordered";
     "id:max": "unordered";
     "id:sum": "unordered";
     "text:approximateDistinct": "unordered";
+    "text:Distinct": "unordered";
```
nihalbhatnagar, line 86: I'm not fully sure if the tests in this file are doing anything, but I added them here just in case I was wrong and could be used for the future. Can we remove this file if its not being used?

ericanderson: Yeah they maybe tested something at one time but clearly do nothing now.

.changeset/wild-cats-hug.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/client.api": minor
+"@osdk/client": minor
+---
+
+Add Exact Distinct Aggregations
```
ericanderson, line 6: Nit, these are going to be written as bullets on the release notes so are generally better as phrases/sentences than "titles". e.g. `Adds exactDistinct to aggregations`

ericanderson (approved):
Preemptive approval

packages/client.api/src/aggregate/AggregationsResults.test.ts
ssanjay1: Yea, @nihalbhatnagar let's maybe do any file deletions as a FLUP to this PR, so we can keep this PR to just adding the new functionality.

ssanjay1 (approved):
LGTM!

packages/client.api/src/aggregate/AggregationsResults.test.ts
nihalbhatnagar: Cool I'll do that

PR #653, opened 2024-09-12
legacy types were removed

ericanderson:


packages/shared.net/src/createClientContext.ts
```diff
@@ -14,28 +14,25 @@
...
   createFetchOrThrow,
   createRetryingFetch,
 } from "@osdk/shared.net.fetch";
-import type { ClientContext } from "./ClientContext.js";
 
 /**
  * The goal of the thin client is to provide a way to tree shake as much as possible.
  */
 
-export function createClientContext<
```
ssanjay1, line 30: Ah cool, this addresses my question from the previous PR, looks like we are removing it

ssanjay1 (approved):

PR #655, opened 2024-09-13
Nearly complete removal of gateway

ericanderson:


packages/shared.test/src/handlers/util/handleOpenApiCall.ts
```diff
@@ -94,30 +91,39 @@ export function handleOpenApiCall<
...
   );
 
   return http
     [captured.method.toLowerCase() as Lowercase<typeof captured.method>](
-      `https://stack.palantir.com/api${captured.endPoint}`,
+      captured.endPoint,
       authHandlerMiddleware(async (info) => {
         try {
-          const result = await restImpl(
+          const result: any = await restImpl(
```
nihalbhatnagar, line 126: Why do we need this?

nihalbhatnagar (approved):

packages/shared.test/src/handlers/util/handleOpenApiCall.ts
ericanderson: The line on 130 produces an error: `The left-hand side of an 'instanceof' expression must be of type 'any', an object type or a type parameter.ts(2358)` without it.

PR #670, opened 2024-09-16
[main] Dont organize imports

ericanderson:


packages/generator/src/GenerateContext/EnhancedOntologyDefinition.ts
```diff
@@ -110,12 +111,15 @@ export class EnhancedOntologyDefinition {
           );
         }
 
-        return new ForeignType(
+        const ret = this.#foreignTypes[fullApiName] ?? new ForeignType(
```
ericanderson, line 114: We need to memozie so we don't emit duplicate imports

packages/generator/src/shared/getObjectImports.ts
```diff
@@ -25,8 +25,15 @@ export function getObjectImports(
 ) {
   return Array.from(objects).filter(obj => obj.fullApiName !== curApiName)
     .map(obj => {
-      return `import type { ${obj.getDefinitionIdentifier(v2)} as ${
-        obj.getImportedDefinitionIdentifier(v2)
-      } } from "${obj.getImportPathRelTo("./" + currentFilePath)}";`;
+      const defId = obj.getDefinitionIdentifier(v2);
+      const importedId = obj.getImportedDefinitionIdentifier(v2);
+
+      const nameOrAlias = defId === importedId
```
ericanderson, line 31: Make it so we don't produce an `as` import if its not needed.

packages/generator/src/v2.0/UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst.ts
```diff
@@ -126,43 +126,15 @@ export function __UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst(
   };
 
   function getV2Types() {
-    return `import type {
-      AggregateOpts as $AggregateOpts,
-      AggregateOptsThatErrorsAndDisallowsOrderingWithMultipleGroupBy as $AggregateOptsThatErrorsAndDisallowsOrderingWithMultipleGroupBy,
```
ericanderson, line 131: Remove a ton of types that weren't being used but were being emitted (and cleaned up  by organize imports)

examples-extra/docs_example/tsconfig.json
```diff
@@ -16,7 +16,7 @@
 
     /* Linting */
     "strict": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
ssanjay1, line 19: Why are we turning this off?

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -245,22 +250,22 @@ export function queryParamJsDoc(
...
-  for (const fqObjectApiName of objectTypes) {
-    const obj = enhancedOntology.requireObjectType(fqObjectApiName);
-    ret += `import type { ${obj.getDefinitionIdentifier(true)} as ${
-      obj.getImportedDefinitionIdentifier(true)
-    } } from "${obj.getImportPathRelTo(filePath)}";\n`;
-  }
-
-  return ret;
-}
+// function getObjectImports(
```
ssanjay1, line 253: Why are we commenting this out?

ssanjay1 (approved):
Approved, one question and one thing pointed out

examples-extra/docs_example/tsconfig.json
ericanderson: Because the generated code now produces unused imports which causes this to fail. Its not a big deal for fully built packages because they get the sdk from npm and so this would never hit. But if you generate in place it does.

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
ericanderson: Oops. Meant to delete that. We had two of these it turned out and I changed this to use the other one.

PR #674, opened 2024-09-17
Add env variable tests to beta template

ewitkon:
Add the recent added env variable test to the React beta template

### DIFF
--- a/packages/create-app.template.react.beta/templates/package.json.hbs
+++ b/packages/create-app.template.react.beta/templates/package.json.hbs
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "test": "vitest run",
     "preview": "vite preview"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


packages/create-app.template.react.beta/templates/package.json.hbs
```diff
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "test": "vitest run",
```
tzyl, line 10: @ewitkon Please can you also:

- Include vitest as a devDependency. This doesn't currently fail on the examples/ package I believe because in this monorepo it already has vitest available from the root but it will fail to run the test command if the template is generated standalone and doesn't have vitest installed (you can test it yourself!)

https://github.com/palantir/osdk-ts/pull/607/files#diff-d6e10d0a434db65208f3fe5909966245621cfa32c51cf7ff18339dc4c46b6de9R42

- Add the generated README note about the test

https://github.com/palantir/osdk-ts/pull/607/files#diff-8a6884e5ccca52be7d7cc412ef9b64248dcba40d2b3deb4f07adea18be1adba2R31

PR #677, opened 2024-09-17
Emitted action definition is minimal size now

ericanderson:


packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -14,22 +14,32 @@
...
+> {
+  type: "action";
+  apiName: A;
+  osdkMetadata: OsdkMetadata;
+  __DefinitionMetadata?: RawActionDefinition<A, K> & {
+    signatures: T_signatures;
+  };
+}
+
+export interface RawActionDefinition<
```
ssanjay1, line 33: `RawActionDefinition` doesn't feel like the right term here, I want to say `Full` but also understand the trepidations there

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -14,22 +14,32 @@
...
 export interface ActionDefinition<
   A extends string,
   K extends string,
-  T = never, // used in client 2
+  T_signatures = never,
+> {
+  type: "action";
+  apiName: A;
+  osdkMetadata: OsdkMetadata;
+  __DefinitionMetadata?: RawActionDefinition<A, K> & {
```
ssanjay1, line 28: Not blocking, mostly because I'm having a hard time thinking of better names, but `_DefinitionMetadata` feels confusing here

packages/generator/src/v2.0/generatePerActionDataFiles.ts
```diff
@@ -188,16 +172,43 @@ export async function generatePerActionDataFiles(
   
           }
 
+          
           ${jsDocBlock.join("\n")}
           */
-          export type ${action.shortApiName} = ${action.shortApiName}.Signatures;
+          export interface ${action.shortApiName} extends ActionDefinition<"${action.shortApiName}", ${uniqueApiNamesString}, ${action.shortApiName}.Signatures> {
```
ssanjay1, line 178: Just double checking that hovering over `client(myAction)` will give the right docs?

ssanjay1 (approved):
Approved, with couple of questions

packages/generator/src/v2.0/generatePerActionDataFiles.ts
ericanderson: It does.

PR #678, opened 2024-09-17
Display metadata

nihalbhatnagar:


packages/client.api/src/DisplayMetadata.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+type ObjectTypeVisibility = "NORMAL" | "PROMINENT" | "HIDDEN";
+
+type BlueprintIcon = {
+  color: string;
+  name: string;
+};
+
+type Icon = BlueprintIcon;
```
nihalbhatnagar, line 24: This is probably specific to DisplayMetadata but we only support blueprint icons from gateway right now, variable here for easy access later on

nihalbhatnagar:
Yet to write specific tests but was able to run e2e tests on my end - will create for final types

etc/client.api.report.api.md
```diff
@@ -268,6 +268,15 @@ export interface DataValueWireToClient {
 // @public (undocumented)
 export type DefaultToFalse<B extends boolean | undefined> = false extends B ? false : undefined extends B ? false : true;
 
+// @public (undocumented)
+export type DisplayMetadata = {
+    displayName: string;
+    pluralDisplayName: string;
+    description?: string;
+    icon: Icon;
+    visibility: ObjectTypeVisibility;
```
ericanderson, line 277: This is good but for the end game we might just want to add these two things to the `ObjectDefinition`

PR #680, opened 2024-09-17
Abandon UnstableClient idea in favor of symbol references

### LINKED ISSUE #718: Move unstable methods back into Client/ObjectSet
We want them hidden behind a symbol to reference the function so they are hard to consume but removes the casting pain we have seen internally.

ericanderson:
Fixes #718


packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -227,6 +230,19 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
         ) as Result<Osdk<Q>>;
       }
       : undefined) as ObjectSet<Q>["fetchOneWithErrors"],
+
+    [__EXPERIMENTAL__NOT_SUPPORTED_YET_subscribe]: (
```
ericanderson, line 234: Example usage

packages/e2e.sandbox.catchall/src/runSubscriptionsTest.ts
```diff
@@ -38,9 +41,13 @@ export function runSubscriptionsTest() {
 
   client(Employee).where({
```
ericanderson, line 42: Example usage

packages/client/src/Client.ts
```diff
@@ -40,14 +49,19 @@ export type CheckVersionBound<Q> = Q extends VersionBound<infer V> ? (
...
       }
-    // Q
   )
   : Q;
 
 export interface Client extends SharedClient<MinimalClient> {
   <Q extends MinObjectDef<any, any>>(
     o: Q,
-  ): CompileTimeMetadata<Q>["objectSet"];
+  ): unknown extends CompileTimeMetadata<Q>["objectSet"] ? ObjectSet<Q>
```
ssanjay1, line 58: Why did we need this change?

packages/client/src/Client.ts
```diff
@@ -71,6 +85,34 @@ export interface Client extends SharedClient<MinimalClient> {
...
   >;
+
+  /**
+   * WARNING. THIS METHOD IS EXPERIMENTAL AND NOT SUPPORTED YET.
+   *
+   * It may change at any time and does not follow semantic versioning. Use at your own risk.
+   *
+   *  @alpha
+   */
+  readonly [__EXPERIMENTAL__NOT_SUPPORTED_YET__preexistingObjectSet]: <
```
ssanjay1, line 96: This pre-existing objectset thing is new right? What use case are we adding this for?

etc/client.api.report.api.md
```diff
@@ -475,6 +475,10 @@ export type ObjectMetadata = {
 
 // @public (undocumented)
 export interface ObjectSet<Q extends ObjectOrInterfaceDefinition = any, _UNUSED = any> extends MinimalObjectSet<Q> {
+    // Warning: (ae-forgotten-export) The symbol "EXPERIMENTAL_ObjectSetListener" needs to be exported by the entry point index.d.ts
+    //
+    // @alpha
+    readonly [__EXPERIMENTAL__NOT_SUPPORTED_YET_subscribe]: (listener: EXPERIMENTAL_ObjectSetListener<Q>) => () => unknown;
```
ssanjay1, line 481: Any reason you didn't just want to call these all `__EXPERIMENTAL_subscribe` etc.?

ssanjay1 (approved):
Approved to unblock, with questions

packages/client/src/Client.ts
ericanderson: I think this is similar to the issue you saw recently where the type complexity ends up failing us. If I recall it was no longer able to determine if the primaryKey was the same type

packages/client/src/Client.ts
ericanderson: Its not new, we had it before. Its for people who have a rid for an object set they want to start with.

etc/client.api.report.api.md
ericanderson: scarier the better! Also we can rename it as much as we want so if you like it smaller we can do that later.

packages/client/src/Client.ts
ssanjay1: Oh ok sounds good then

packages/client/src/Client.ts
ssanjay1: Ah yea that was the same issue I ran into

PR #688, opened 2024-09-18
Add fetchMetadata

nihalbhatnagar:
Adds fetchMetadata function to client


packages/client.api/src/Definition.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import type {
+  ActionDefinition,
+  InterfaceDefinition,
+  ObjectTypeDefinition,
+  QueryDefinition,
+} from "@osdk/api";
+
+export type Definition =
```
nihalbhatnagar, line 24: I need this type in order to type the parameters for client. I can't find a type like this already in the api. I'm not sure if this is the correct name or if I should name it something like MetadataDefinition.

packages/client/src/fetchMetadata.ts
```diff
@@ -0,0 +1,127 @@
...
+    : Q extends QueryDefinition<any, any, any> ? QueryMetadata
+    : never
+> => {
+  if (definition.type === "object") {
+    return fetchObjectMetadata(client, definition) as any;
+  } else if (definition.type === "interface") {
+    return fetchInterfaceMetadata(client, definition) as any;
+  } else if (definition.type === "action") {
+    return fetchActionMetadata(client, definition) as any;
+  } else {
```
nihalbhatnagar, line 53: I have to use else instead of else if here even though this is the only other option because otherwise I get ts errors saying I cannot return undefined. Is there a clever way to get around this?

packages/client/src/Client.ts
```diff
@@ -49,6 +56,16 @@ export interface Client extends SharedClient<MinimalClient> {
...
     o: Q,
   ): QuerySignatureFromDef<Q>;
+
+  fetchMetadata<
+    Q extends Definition,
+  >(o: Q): Promise<
+    Q extends ObjectTypeDefinition<any, any> ? ObjectMetadata
+      : Q extends InterfaceDefinition<any, any> ? InterfaceMetadata
+      : Q extends ActionDefinition<any, any, any> ? ActionMetadata
+      : Q extends QueryDefinition<any, any, any> ? QueryMetadata
```
ericanderson, line 66: I'm wondering after we get my thinned out objects and co merged in, if we should just be returning the `*Definition` for the metadata? So this woul dbe `Q extends MinObjectDef ? ObjectTypeDefinition<any,any> : ...`

packages/client/src/fetchMetadata.ts
ericanderson: Do the `else if { ... } else { throw "Should never happen": }` ?

packages/client.api/src/Definition.ts
ericanderson: Hmm. I'm on the fence about having a type like this for minimal use

packages/client.api/src/object/ObjectMetadata.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+type ObjectTypeVisibility = "NORMAL" | "PROMINENT" | "HIDDEN";
+
+type BlueprintIcon = {
```
ericanderson, line 19: If there is another type of icon like `StandardWindowsIcon` how will you know its any different from `BlueprintIcon`?

packages/client.api/src/object/ObjectMetadata.ts
```diff
@@ -0,0 +1,34 @@
...
+  color: string;
+  name: string;
+};
+
+type Icon = BlueprintIcon;
+
+export type ObjectMetadata = {
+  description?: string;
+  displayName: string;
+  icon: Icon;
```
ericanderson, line 29: Are we guaranteed to always get an icon here? What if its a type we can't support and we want to return undefined in that case?

packages/client/src/fetchMetadata.ts
nihalbhatnagar: Ah yeah that will work, but is there not a way to avoid throwing an error and have it fully checked by the compiler

packages/client/src/Client.ts
nihalbhatnagar: Yeah we can do that, my goal with this was to be able to easily control what we expose to the user. Based on what we discussed yesterday I dont want to return apiName, osdk metadata, or anything else thats on your MinObjectDef type because thats at compile. Thats mainly it, although if there's a case where we want something exposed on the type but not as a concrete value that adds to the argument of keeping it separate.

packages/client.api/src/Definition.ts
nihalbhatnagar: Ok, I can just inline the type in the 2 places that need it.

packages/client.api/src/object/ObjectMetadata.ts
nihalbhatnagar: Icon from gateway is just a union consisting of only the Blueprint type [shown here.](https://github.com/palantir/osdk-ts/blob/5a7443582c7e72acd0c389f840eba5813d1fe681/packages/internal.foundry.core/src/_components.ts#L680C1-L685C57) I was thinking we should keep parity with this for an easy update if it changes

packages/client.api/src/object/ObjectMetadata.ts
nihalbhatnagar: If I understand what you are saying because Icon is a union of just Blueprint right now we are guaranteed to get just a blueprint icon and the types match up. If they dont, ts will throw a flag and we can add the type to our own union

packages/client/src/fetchMetadata.ts
nihalbhatnagar: Never mind, I was just a bit confused, was basically looking for an assertNever here

packages/client/src/fetchMetadata.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (definition.type === "object") {
+    return fetchObjectMetadata(client, definition) as any;
+  } else if (definition.type === "interface") {
+    return fetchInterfaceMetadata(client, definition) as any;
+  } else if (definition.type === "action") {
+    return fetchActionMetadata(client, definition) as any;
+  } else if (definition.type === "query") {
+    return fetchQueryMetadata(client, definition) as any;
+  } else {
+    throw new Error("Type checking should be exhaustive");
```
ssanjay1, line 60: Let's do something more along the lines of `not implemented` , basically something that would be more beneficial for a user

packages/client/src/FetchMetadata.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      async () => "myAccessToken",
+    );
+  });
+
+  afterAll(() => {
+    apiServer.close();
+  });
+
+  it("fetches object metadata correctly", async () => {
+    client.fetchMetadata;
```
ssanjay1, line 59: You need this line?

ssanjay1: Same with the rest of the tests

packages/client/src/FetchMetadata.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        "pluralDisplayName": "Employees",
+        "rid": "ri.ontology.main.object-type.401ac022-89eb-4591-8b7e-0a912b9efb44",
+        "visibility": "NORMAL",
+      }
+    `);
+  });
+
+  it("fetches interface metadata correctly", async () => {
+    client.fetchMetadata;
+    const objectMetadata = await client.fetchMetadata($Interfaces.FooInterface);
```
ssanjay1, line 84: nit: for each test, maybe name the variables accordingly

so here, `interfaceMetadata`

packages/client/src/Client.ts
ericanderson: Its probably easier for us to maintain the parity. The osdkmetadata is optional and the apIName they already know but it doesnt end the world to include it.

packages/client.api/src/object/ObjectMetadata.ts
ericanderson: Agree but without the `type` field, if the `StandardWindowsIcon` interface also has the name and color, you won't know which to do

packages/client.api/src/object/ObjectMetadata.ts
ericanderson: Yeh if they add another type this version can keep working because we can just return undefined instead of the wire value until we update the code.

ericanderson (approved):
approving for followup

packages/client.api/src/object/ObjectMetadata.ts
nihalbhatnagar: Cool yeah I understand, added a check here to conditionally return undefined

ericanderson (approved):

ssanjay1 (approved):

PR #711, opened 2024-09-23
Expose total count

ssanjay1:


packages/client/src/object/fetchPage.ts
```diff
@@ -280,5 +280,6 @@ export async function fetchObjectPage<
       args.$__EXPERIMENTAL_strictNonNull,
     ),
     nextPageToken: r.nextPageToken,
-  }) as Promise<FetchPageResult<Q, L, R, S>>;
+    totalCount: r.totalCount,
+  }) as unknown as Promise<FetchPageResult<Q, L, R, S>>;
```
ssanjay1, line 284: So I'm not sure if I messed up, but as soon as I added this new field I think it just became too much for the type system because it started complaining about the primary key types in data not matching. So I had to do this

ericanderson (approved):
I dont want to block on the cast but this is our best chance of understanding whats wrong given how simple a change this is.

PR #712, opened 2024-09-23
MinQueryDef

### LINKED ISSUE #300: Explore emitting strings for ontology types that are referenced.
As we push further into complex use cases, we may want syntax like this:

```ts
client(TypeA).union(client(TypeB)).where({
  [TypeA]: {
    ...
  },
  [TypeB]: {
    ...
  }
});
```

Given the current runtime types for `TypeA`/`TypeB` are concrete objects, this isn't possible and instead may require something like:

```ts
  client(TypeA).union(client(TypeB)).where({
    ...filter(TypeA, { ... }),
    ...filter(TypeB, { ... }),
});
```

I don't actually mind this, but it needs to be an explicit choice we make. This is similar to the `augment()` issue with interfaces.

ericanderson:
Fixes #300


packages/client/src/ontology/StandardOntologyProvider.ts
```diff
@@ -65,7 +67,17 @@ export const createStandardOntologyProviderFactory: (
       return loadInterfaceDefinition(client, key);
     }
 
-    function makeGetter<N extends ObjectOrInterfaceDefinition>(
+    async function loadQuery(
```
ssanjay1, line 70: Little confused what this is for. So is this equivalent to the fetchMetadata stuff we'll have for objects?

ssanjay1 (approved):

PR #714, opened 2024-09-23
Update return type for action edits

nihalbhatnagar:


packages/client.api/src/actions/ActionResults.ts
```diff
@@ -40,16 +40,12 @@ interface LargeScaleObjectEdits {
...
-  | ({ type: "addObject" } & ObjectReference)
-  | ({ type: "modifyObject" } & ObjectReference)
-  | ({
-    type: "addLink";
-    linkTypeApiNameAtoB: string;
-    linkTypeApiNameBtoA: string;
-    aSideObject: ObjectReference;
-    bSideObject: ObjectReference;
-  });
+type LinkReference = {
```
ericanderson, line 43: nit for followup, types that can be interfaces are generally better as such. Its a bit easier on the typescript compiler

ericanderson (approved):

PR #728, opened 2024-09-25
Remove CJS from shared.client

ericanderson:
Existed for the old world where we supported both legacy client and client in repo. So now this package is consistent with the rest of the repo.

Bonus: because its no longer a mixed package, vite does not need special handling any more, which should simplify supporting people.


ssanjay1 (approved):
Yay

PR #730, opened 2024-09-25
Updates action and query definitions to match desired metadata return values

nihalbhatnagar:


packages/e2e.generated.api-namespace.local/src/generatedNoCheck/ontology/queries/getTask.ts
```diff
@@ -46,6 +47,7 @@ export type getTask = getTask.Signature;
 
 export const getTask: getTask.Definition = {
```
nihalbhatnagar, line 48: It seems that we still use the old .Definition instead of the new style, will address in a FLUP PR

packages/client/src/Client.ts
```diff
@@ -23,22 +24,16 @@ import type {
...
-  Osdk,
-  QueryMetadata,
-} from "@osdk/client.api";
+import type { MinimalObjectSet, ObjectSet, Osdk } from "@osdk/client.api";
 import type {
   __EXPERIMENTAL__NOT_SUPPORTED_YET__getBulkLinks,
   __EXPERIMENTAL__NOT_SUPPORTED_YET__preexistingObjectSet,
 } from "@osdk/client.api/unstable";
 import type { SharedClient } from "@osdk/shared.client";
+import { an } from "vitest/dist/reporters-B7ebVMkT.js";
```
nihalbhatnagar, line 36: oops

ericanderson (approved):

PR #736, opened 2024-09-25
Audit deps

ericanderson:
Chains to #733

After doing #733, I was looking at the deps we used and if they were correct. In doing so, I found some dead code and I found what should prevent future issues with deps shifting in foundry-sdk-generator.


packages/client.api/package.json
```diff
@@ -36,7 +36,6 @@
   "devDependencies": {
     "@microsoft/api-documenter": "^7.25.3",
     "@microsoft/api-extractor": "^7.47.0",
-    "@osdk/api": "workspace:~",
```
ericanderson, line 39: Already in regular deps

packages/client.test.ontology/package.json
```diff
@@ -27,13 +27,11 @@
...
   "dependencies": {
+    "@osdk/api": "workspace:*",
     "@osdk/client.api": "workspace:*"
   },
-  "peerDependencies": {
-    "@osdk/api": "workspace:*"
-  },
+  "peerDependencies": {},
   "devDependencies": {
-    "@osdk/api": "workspace:~",
```
ericanderson, line 36: Already in regular deps

packages/client/package.json
```diff
@@ -37,7 +37,7 @@
     "@osdk/generator-converters": "workspace:~",
     "@osdk/internal.foundry.core": "workspace:~",
     "@osdk/internal.foundry.ontologiesv2": "workspace:~",
-    "@osdk/shared.client": "workspace:~",
+    "@osdk/shared.client": "workspace:^",
```
ericanderson, line 40: All references for `@osdk/shared.client` should be `^` so that we have maximum change of cross SDK generation/platform sdks working together.

This is okay because we should basically not ever touch that package without extreme care.

packages/client/package.json
```diff
@@ -65,7 +65,6 @@
     "jest-extended": "^4.0.2",
     "msw": "^2.3.4",
     "p-defer": "^4.0.1",
-    "p-event": "^6.0.1",
```
ericanderson, line 68: unused

packages/foundry-sdk-generator/package.json
```diff
@@ -29,24 +29,23 @@
...
+    "@osdk/shared.net": "workspace:*",
+    "@rollup/plugin-commonjs": "25.0.7",
+    "@rollup/plugin-node-resolve": "15.2.3",
+    "consola": "3.2.3",
+    "find-up": "7.0.0",
+    "rollup": "3.29.1",
+    "rollup-plugin-polyfill-node": "0.13.0",
+    "ts-morph": "19.0.0",
+    "typescript": "5.5.3",
+    "yargs": "17.7.2"
```
ericanderson, line 48: By shifting all of foundry-sdk-generator to fixed packages, we should not have surprises where typescript gets upgraded underneath us and breaks our generator

ssanjay1 (dismissed):

ssanjay1 (dismissed):

ssanjay1 (approved):

PR #738, opened 2024-09-26
Use verbs for Platform SDK

jsmith:


ericanderson:
@jsmith im really wondering if this is a good idea until we change major again.

jsmith:
Hmm I feel like asking the handful of users of the TS platform SDK to swap the function names they are using will be way easier now than asking this during the next major version bump. Going from the operation name -> verb is not always straightforward. Unless you're saying just hold off until you the platform SDK out of beta?

ericanderson:
Discussed offline. We almost certainly have other breaks since we started releasing this, so this isn't the straw that breaks the camel. 

Additional thought: we never gave this to anyone in a "supported" manner so while this will be a surprise and frustrating we can properly support semver following the next ga of these packages.

ericanderson (approved):

ssanjay1 (approved):

PR #745, opened 2024-09-26
update dependencies if they exist

ericanderson:


packages/cli.cmd.typescript/package.json
```diff
@@ -23,6 +23,7 @@
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
+    "test": "vitest",
```
ssanjay1, line 26: this needs to be vitest run right? or else vitest defaults to watch mode?

ericanderson: Good call!

ssanjay1 (dismissed):

ssanjay1 (approved):

ssanjay1 (approved):

PR #748, opened 2024-09-26
remove client.api

ericanderson:


ericanderson:
Not only did I edit all the changesets I also added one, so no clue why the changeset check is failing

.changeset/thick-mice-sit.md
```diff
@@ -1,6 +1,6 @@
 ---
 "@osdk/shared.test": patch
-"@osdk/client.api": patch
+"@osdk/api": patch
```
ssanjay1, line 3: I think error is coming from here maybe, since there is two @osdk/apis?

ssanjay1 (approved):

PR #750, opened 2024-09-26
Audit API Package

ssanjay1:
Went through and removed any unused types.

Here are types that are only used internally for tests that I have kept for now:
* ValidAggregationkeys
* OsdkObjectPropertyType
* DefaultToFalse
* PropertyKeys
* ConvertProps
* PageResult
    * This is also used in dev console doc snippets

### DIFF
--- a/etc/api.report.api.md
+++ b/etc/api.report.api.md
@@ -17,6 +17,8 @@ export interface ActionDefinition<A extends string, K extends string, T_signatur
     __DefinitionMetadata?: ActionCompileTimeMetadata<T_signatures> & ActionMetadata<A, K>;
     // (undocumented)
     apiName: A;
+    // Warning: (ae-forgotten-export) The symbol "OsdkMetadata" needs to be exported by the entry point index.d.ts
+    //
     // (undocumented)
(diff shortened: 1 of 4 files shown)


etc/api.report.api.md
```diff
@@ -183,29 +184,20 @@ export interface AsyncIterArgs<Q extends ObjectOrInterfaceDefinition, K extends
...
 export interface Attachment {
     fetchContents(): Promise<Blob>;
+    // Warning: (ae-forgotten-export) The symbol "AttachmentMetadata" needs to be exported by the entry point index.d.ts
     fetchMetadata(): Promise<AttachmentMetadata>;
     // (undocumented)
     rid: string;
 }
 
-// @public (undocumented)
-export interface AttachmentMetadata {
```
ericanderson, line 192: Do you think someone might want to use this type?

ericanderson (approved):

etc/api.report.api.md
ssanjay1: Talked offline, going to withhold this type for now, we can easily add later if needed

PR #754, opened 2024-09-26
Refactors for reduced exports

ericanderson:


packages/api/src/timeseries/timeseries.ts
```diff
@@ -67,17 +67,16 @@ export interface TimeSeriesPoint<T extends string | number> {
...
-    /**
-     * Queries the first point of the Timeseries
-     */
-    public getFirstPoint: () => Promise<TimeSeriesPoint<T>>,
-    /**
-     * Queries the last point of the Timeseries
-     */
-    public getLastPoint: () => Promise<TimeSeriesPoint<T>>,
-    /**
+export interface TimeSeriesProperty<T extends number | string> {
```
ssanjay1, line 70: Why did we switch this back to an interface? I remember you mentioning that we wanted to make this a class since its slightly more performant?

packages/client/src/createTimeseriesProperty.ts
```diff
@@ -28,117 +28,70 @@ import {
...
 } from "./util/streamutils.js";
 
-export function createTimeseriesProperty<T extends number | string>(
-  client: MinimalClient,
-  objectApiName: string,
-  primaryKey: any,
-  propertyName: string,
-): TimeSeriesProperty<T> {
-  const getFirstPoint = async () => {
+export class TimeSeriesPropertyImpl<T extends number | string>
```
ssanjay1, line 31: Hmm ok, so you're making this a class instead. Ok yea still confused what the benefit here is

packages/client/src/internal/conversions/modernToLegacyAggregationClause.ts
```diff
@@ -45,7 +40,15 @@ export function modernToLegacyAggregationClause<
 
       return [
         {
-          type: metric as StringAggregateOption | NumericAggregateOption,
+          type: metric as
+            | "approximateDistinct"
```
ssanjay1, line 44: Is it not better if we had these aggregateoptions as a shared type like before? In case we need to modify/add an option?

ssanjay1 (approved):
Approved to unblock, had a couple of questions

packages/api/src/timeseries/timeseries.ts
ericanderson: Yea, I should have been more clear and I should have looked more closely at the revised PR.

I want the consumer to have an interface but I don't want them constructing it. The class, since it just took the functions and assigned them to itself was no better than a raw object the way it was. The way it is now, we aren't constructing additional function objects per call, nor additional scopes that must be kept by the engine.

packages/client/src/internal/conversions/modernToLegacyAggregationClause.ts
ericanderson: Sadly adding an option isn't going to do anything because you can always make this cast: https://www.typescriptlang.org/play/?#code/DYUwLgBAzgXNYCcCWA7A5gbgFBahAvBABQCsEApBAEwH6ECMAlBAPwQBEAFkuxHOwCMAnmBDtsWAMYB7FFEgAzadP4ppABzEFoEAIZ52azeyA

So really we are exposing types we have to keep back compat on that aren't serving us any good.

ssanjay1: ah ok makes sense

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -20,40 +20,67 @@ import type {
...
-> {
+export interface ActionMetadata {
   type: "action";
-  apiName: A;
+  apiName: string;
   description?: string;
   displayName?: string;
-  parameters: Record<any, ActionParameterDefinition<K, any>>;
-  modifiedEntities?: Partial<Record<K, ActionModifiedEntity>>;
+  parameters: Record<any, ActionMetadata.Parameter<any>>;
```
nihalbhatnagar, line 28: Do we never need to specify this?

ericanderson: If you mean the K, it seems to work just fine without it.

PR #756, opened 2024-09-27
Shared client uses .js extension now

ericanderson:


packages/monorepo.tool.attw/bin/check-attw.sh
```diff
@@ -31,7 +32,7 @@ errors=$(cat build/attw.json | jq --arg check_type "$BUILD_TYPE" '
         ) | map_values(if . | length == 0 then empty else . end)
     ;
 
-    filterIssues(.problems) | if (. == {}) then empty else . end
+    filterIssues(.problems | {}) | if (. == {}) then empty else . end
```
ericanderson, line 35: This is because now we have zero problems and was causing jq to npe

nihalbhatnagar (approved):

PR #758, opened 2024-09-27
refactors

ericanderson:


packages/client/src/object/object.test.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type { Osdk } from "@osdk/api";
+import type { Osdk, PropertyKeys } from "@osdk/api";
```
ssanjay1, line 17: Why the new import here?

ericanderson: accident. can remove in follow up. has no effect on output

ssanjay1: Sounds good

ssanjay1 (approved):

PR #775, opened 2024-09-30
Remove shared.client to remove risk of shipping it

ericanderson:
There is no changeset because the old dependencies of `workspace:^` are identical to `^1.0.0` after they were shipped and so this is a NOOP change.


.monorepolint.config.mjs
```diff
@@ -642,7 +643,7 @@ NOTE: DO NOT EDIT THIS README BY HAND. It is generated by monorepolint.
       ],
       options: {
         dependencies: {
-          "@osdk/shared.client": "workspace:^",
+          "@osdk/shared.client": "^1.0.0",
```
ssanjay1, line 646: 1.0.1?

ericanderson:
We need to cherry pick this to release/2.0.x

ssanjay1 (approved):

PR #798, opened 2024-10-02
Tighten dependencies for fewer runtime surprises

ericanderson:
# @osdk/client
Reduces the spread of dependencies for `@osdk/client` such that the following packages will now be exact match:
- `@osdk/generator-converters`
- `@osdk/internal.foundry.core`
- `@osdk/internal.foundry.ontologiesv2`

In the event that a consumer ends up with mixed client versions, this may result in some slight duplication but I think it is worth the minimal cost to avoid problems, especially since we shake away a lot of ontologiesv2 and core is nearly all types.

Specifically, these upgrading beyond what we compiled to could result in updates to the gateway apis causing runtime breaks for older gateways.

# @osdk/api

Removes an unused dependency. No changeset needed.

# @osdk/cli

Bundles all `@osdk/` deps

# @osdk/generator-converters

Locks dependency on `@osdk/internal.foundry.core` to align with client.


nihalbhatnagar (dismissed):

packages/cli/package.json
```diff
@@ -30,8 +30,6 @@
   },
   "dependencies": {
     "@arethetypeswrong/cli": "^0.15.2",
-    "@osdk/generator": "workspace:~",
```
nihalbhatnagar, line 33: I assume were good to remove this because its going through .cmd.typescript? If so, does .cmd.typescript have the strict dep on the generator without the ~?

ericanderson: Yeah this was directly unused and transitively used by the cmd.typescript.

And since all cli osdk things are being bundled, the line in cmd.typescript is irrelevant (we dont even publish cmd.typescript)

nihalbhatnagar (approved):

PR #802, opened 2024-10-03
Fix proxy handler

ssanjay1:
We ran into an issue where someone deleted the `$title` property on their object and they ran into this error:
`TypeError: 'getOwnPropertyDescriptor' on proxy: trap reported non-configurability for property '$title' which is either non-existent or configurable in the proxy target`.

This is because our handler for `getOwnPropertyDescriptor` was violating this invariant: 
**_A property cannot be reported as non-configurable, unless it exists as a non-configurable own property of the target object. That is, if [Reflect.getOwnPropertyDescriptor()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/getOwnPropertyDescriptor) returns undefined or configurable: true for the property on target, then the trap must not return configurable: false._**

This is now fixed by returning `undefined` in the default case, rather than `{enumerable:false}` which will also set `configurable` to false

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects.test.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects.test.ts
@@ -77,6 +77,31 @@ describe("convertWireToOsdkObjects", () => {
     ]);
   });
 
+  it("stringifies properties on objects and interfaces correctly", async () => {
+    const { data: [employee] } = await client(Employee).fetchPage();
+    const { data: [employee2] } = await client(Employee).where({
+      $and: [{ employeeId: { $gt: 50030 } }, { employeeId: { $lt: 50032 } }],
+    }).fetchPage();
+
+    // Should not have $title
+    expect(JSON.stringify(employee)).toMatchInlineSnapshot(
+      `"{"employeeId":50030,"fullName":"John Doe","office":"NYC","class":"Red","startDate":"2019-01-01","employeeStatus":{},"$apiName":"Employee","$objectType":"Employee","$primaryKey":50030}"`,
+    );
+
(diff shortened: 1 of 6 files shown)


packages/client/src/object/object.test.ts
```diff
@@ -57,11 +57,11 @@ describe("OsdkObject", () => {
 
       // we should get the employee we requested
       const employee = result.data[0];
+      expect(JSON.stringify(employee)).toBeDefined();
```
ssanjay1, line 60: Deleted title property from this object and making sure stringify works (this is the function that failed for our consumers).

ssanjay1: Also added tests above in createOsdkObject.test

packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -78,6 +77,31 @@ describe("convertWireToOsdkObjects", () => {
...
   });
 
+  it("stringifies properties on objects and interfaces correctly", async () => {
+    const { data: [employee] } = await client(Employee).fetchPage();
+    const { data: [employee2] } = await client(Employee).where({
+      $and: [{ employeeId: { $gt: 50030 } }, { employeeId: { $lt: 50032 } }],
+    }).fetchPage();
+
+    // Should not have $title
+    expect(JSON.stringify(employee)).toMatchInlineSnapshot(
```
ericanderson, line 87: nit in the future if you `JSON.stringify(x, undefined, 2)` you get something much easier to read/review

ericanderson (approved):

PR #805, opened 2024-10-03
Improve types

ericanderson:
Following some confusion about type assignability and that the `Osdk<>` type was causing a bit of complexity,  I thought it might make sense to further refine the types.

Further we can adapt what we have as well such that:
- `Osdk` object still remains the primary object used but generated code refers to the `Instance` or `OsdkObject`.

### DIFF
--- a/packages/api/src/OsdkObject.ts
+++ b/packages/api/src/OsdkObject.ts
@@ -17,6 +17,9 @@
 import type { PropertyValueWireToClient } from "./mapping/PropertyValueMapping.js";
 import type { PrimaryKeyTypes } from "./ontology/PrimaryKeyTypes.js";
 
+/**
+ * @deprecated Use OsdkBase
+ */
 export type OsdkObject<N extends string> = {
   readonly $apiName: N;
   readonly $objectType: string;
--- a/packages/api/src/OsdkObjectFrom.ts
+++ b/packages/api/src/OsdkObjectFrom.ts
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
-import type { OsdkObjectLinksObject } from "./definitions/LinkDefinitions.js";
+import type { IsAny } from "type-fest";
+import type {
+  DefaultToFalse,
+  OsdkObjectLinksObject,
+} from "./definitions/LinkDefinitions.js";
+import type { NullabilityAdherence } from "./object/FetchPageArgs.js";
 import type { UnionIfTrue } from "./object/FetchPageResult.js";
 import type { InterfaceDefinition } from "./ontology/InterfaceDefinition.js";
 import type {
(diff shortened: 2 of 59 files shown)


packages/generator/src/v2.0/generatePerInterfaceDataFiles.ts
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+import path from "node:path";
+import type { GenerateContext } from "../GenerateContext/GenerateContext.js";
+import { formatTs } from "../util/test/formatTs.js";
+import { __UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst } from "./UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst.js";
+
+/** @internal */
+export async function generatePerInterfaceDataFiles(
```
ssanjay1, line 23: All these generatePer... files are unchanged from original implementation right? Just refactored into new files?

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -70,7 +70,9 @@ export async function generatePerQueryDataFilesV2(
     await formatTs(`
     ${
       Object.values(ontology.queryTypes).map(query =>
-        `export * from "${query.getImportPathRelTo(relOutDir)}";`
+        `export {${query.shortApiName}} from "${
```
ssanjay1, line 73: Ah, what is this change for?

ssanjay1 (commented):
Left some questions/comments

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
```diff
@@ -14,76 +14,18 @@
...
 import { sanitizeMetadata } from "../shared/sanitizeMetadata.js";
-import { formatTs } from "../util/test/formatTs.js";
 import { verifyOutDir } from "../util/verifyOutDir.js";
 import type { WireOntologyDefinition } from "../WireOntologyDefinition.js";
 import { generateOntologyMetadataFile } from "./generateMetadata.js";
 import { generatePerActionDataFiles } from "./generatePerActionDataFiles.js";
+import { generatePerInterfaceDataFiles } from "./generatePerInterfaceDataFiles.js";
+import { generatePerObjectDataFiles } from "./generatePerObjectDataFiles.js";
 import { generatePerQueryDataFilesV2 } from "./generatePerQueryDataFiles.js";
-import { __UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst } from "./UNSTABLE_wireInterfaceTypeV2ToSdkObjectConst.js";
```
nihalbhatnagar, line 28: Appreciate the clean up here!

packages/api/src/OsdkObjectFrom.ts
```diff
@@ -137,44 +142,90 @@ type UnderlyingProps<
...
-    : string extends P ? {}
-    : "$rid" extends P ? { readonly $rid: string }
-    : {});
-
-export type OsdkObjectOrInterfaceFrom<
-  Q extends ObjectOrInterfaceDefinition,
-  P extends string = PropertyKeys<Q>,
-> = Osdk<Q, P>;
+
+export namespace Osdk {
```
nihalbhatnagar, line 187: I'm not sure I fully understand the issues we were facing with the initial OSDK/why focusing on MyObject.Instance using what used to be OsdkObjectOrInterfaceFrom helps here

ericanderson: `OsdkObjectOrInterfaceFrom` was a type we should have deleted. It was only used in the experimental subscription code, everything else was just `Osdk<>`.

The problem with keeping everything `Osdk<>` is that it leaves us in the world where we still have very complex types to evaluate what properties are available. Switching to this new type vastly simplifies the work the compiler was doing. However, we shipped 2.0 so we cannot JUST change `Osdk<>` because thats a breaking change. So this change makes it so that existing code that uses the `Osdk<>` object will keep working (at the continued expense of costly evaluation in the compiler) AND it makes it so that you can also use `Osdk<>` in the same shape as `Osdk.Instance<>` due to the back compat changes. This allows us in 3.x to later tell people to just use `Osdk<>` instead of `.Instance` if we so choose when we can completely delete the fallback behavior.

Because 2.0.0 has not gone to many people except internally and a few beta customers, this loss of `Osdk<>` is only going to be noticeable by those few people. Everyone else will just learn `Osdk.Instance`.

packages/api/src/OsdkObject.ts
```diff
@@ -17,6 +17,9 @@
 import type { PropertyValueWireToClient } from "./mapping/PropertyValueMapping.js";
 import type { PrimaryKeyTypes } from "./ontology/PrimaryKeyTypes.js";
 
+/**
+ * @deprecated This type should have been removed. You probably want OsdkBase
```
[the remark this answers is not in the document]
ericanderson: It was only used in a few places in the client code and we had alternatives we could have used and should have. It just slipped through the cracks.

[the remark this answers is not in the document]
ericanderson: Yea

ssanjay1 (approved):

PR #806, opened 2024-10-03
Add infrastructure for reference updates

nihalbhatnagar:


packages/api/src/objectSet/EXPERIMENTAL_ObjectSetListener.ts
```diff
@@ -26,10 +27,9 @@ export interface EXPERIMENTAL_ObjectSetListener<
   onChange?: (objects: Array<OsdkObjectOrInterfaceFrom<O>>) => void;
 
   /**
-   * An objects backing datasource has changed and a new value is provided
+   * A property on an object that is backed by another datasource has received changes
    */
-  // TODO: Implement
-  onReferenceUpdate?: (updates: Array<OsdkObjectOrInterfaceFrom<O>>) => void;
+  onReferenceUpdate?: (updates: Array<ReferenceUpdate<O>>) => void;
```
ericanderson, line 32: I am wondering why I would want to get an array of things to deal with vs something that looks like a partial object?

nihalbhatnagar: If I understand what you're saying, Saurav and I talked about this yesterday, since the GTSR property types are going to look like the timeseries properties we already have, we can't really return the actual value as a part of the object where the property is then typed with the actual value instead of the shape of the handlers

packages/api/src/object/ReferenceUpdates.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { PrimaryKeyType } from "../OsdkBase.js";
+
+export type ReferenceUpdate<T extends ObjectOrInterfaceDefinition> = {
+  apiName: T["apiName"];
+  primaryKey: PrimaryKeyType<T>;
+  property: PropertyKeys<T>;
+  value?: ReferenceValue;
+};
+
+type ReferenceValue = GeotimeSeriesValue;
```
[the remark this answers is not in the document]
nihalbhatnagar: I think I definitely need to add a type to the GTSR value here, but did you mean anything else?

PR #807, opened 2024-10-04
Fix queries that return arrays

ssanjay1:
There was a mismatch of inputs for remapping query responses when we had arrays and other nested response types, which caused us to return the client context instead of the actual return data

### DIFF
--- a/packages/client/src/queries/applyQuery.ts
+++ b/packages/client/src/queries/applyQuery.ts
@@ -116,9 +116,9 @@ async function remapQueryResponse<
     const withoutMultiplicity = { ...responseDataType, multiplicity: false };
     for (let i = 0; i < responseValue.length; i++) {
       responseValue[i] = await remapQueryResponse(
-        responseValue[i],
-        withoutMultiplicity,
         client,
+        withoutMultiplicity,
+        responseValue[i],
         definitions,
       );
     }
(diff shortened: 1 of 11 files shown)


packages/client/src/queries/applyQuery.ts
```diff
@@ -116,9 +116,9 @@ async function remapQueryResponse<
     const withoutMultiplicity = { ...responseDataType, multiplicity: false };
     for (let i = 0; i < responseValue.length; i++) {
       responseValue[i] = await remapQueryResponse(
-        responseValue[i],
-        withoutMultiplicity,
         client,
```
ssanjay1, line 119: This passed through fine before because `responseValue[i]` is typed any and that matched up with `MinimalClient` and the third param accepts type any (which is `DataValue`) so it took client :(

packages/shared.test/src/stubs/queries.ts
```diff
@@ -223,6 +224,16 @@ export const queryTypeAcceptsThreeDimensionalAggregationResponse:
     },
   };
 
+export const queryTypeReturnsArrayRequest: ExecuteQueryRequest = {
+  parameters: {
+    country: ["Brad", "George", "Ryan"],
```
nihalbhatnagar, line 229: Nit but can we rename this

nihalbhatnagar: "country" to "people"

packages/client/src/queries/applyQuery.ts
ericanderson: Maybe we could use a branded type for these to force us to double check these / protect against refactors? Like `responseValue[i] as BrandedResponseValue` or whatever.

ericanderson (approved):

packages/client/src/queries/applyQuery.ts
ssanjay1: Yea was thinking about that, can address in a FLUP

PR #809, opened 2024-10-07
Codegen fixes

ssanjay1:


packages/client/src/objectSet/toConjureObjectSet.ts
```diff
@@ -74,12 +74,17 @@ export function toConjureObjectSet(
         ),
       });
     case "searchAround":
+    case "asBaseObjectTypes":
```
ssanjay1, line 77: New objectset types were added

packages/foundry.streams/src/_components.ts
```diff
@@ -59,6 +59,29 @@ export interface Dataset {
...
+  records: Array<_Record>;
+  viewRid?: ViewRid;
+}
+
+/**
+ * A record to be published to a stream.
+ *
+ * Log Safety: DO_NOT_LOG
+ */
+export type _Record = Record<string, any>;
```
ssanjay1, line 83: So they declared a type called Record, which is why I had to do some weird overrides in our codegen to not conflict with TS record types

packages/platform-sdk-generator/src/addPackagesToPackageJson.ts
```diff
@@ -26,7 +26,7 @@ export async function addPackagesToPackageJson(
     packageJson[section],
     Object.fromEntries(packagesToAdd.map(a => [a, "workspace:*"])),
   );
-  packageJson.dependencies["@osdk/shared.client"] = "workspace:~";
+  packageJson.dependencies["@osdk/shared.client"] = "1.0.1";
```
ssanjay1, line 29: We removed shared.client so just made it concrete dep on 1.0.1 that we released

packages/platform-sdk-generator/src/generatePlatformSdkv2.ts
```diff
@@ -161,7 +161,7 @@ export async function generateComponents(
       continue;
     }
     out += component.getDeclaration(ns.name);
-    ret.push(component.name);
+    ret.push(component.name === "Record" ? "_Record" : component.name);
```
ssanjay1, line 164: To deconflict Record

packages/platform-sdk-generator/src/generateResource2.ts
```diff
@@ -58,7 +58,9 @@ async function generateMethods(resource: Resource, model: Model) {
   const referencedTypes = new Set<Component>();
 
   for (const method of resource.operations) {
-    const methodName = (method.verb === "delete") ? method.name : method.verb;
+    const methodName = (method.verb === "delete" || method.verb === "load")
```
ssanjay1, line 61: The new loadObject set endpoints for interfaces also use verb load so they conflict in codegen. Used name instead to deconflict

packages/platform-sdk-generator/src/model/Component.ts
```diff
@@ -67,9 +67,11 @@ export class Component extends Type {
    * 
    * Log Safety: ${component.safety}
    */
-  export ${isAlias ? "type" : "interface"} ${component.locator.localName} ${
-      isAlias ? " = " : ""
-    }`;
+  export ${isAlias ? "type" : "interface"} ${
+      component.locator.localName === "Record"
```
ssanjay1, line 71: More record type deconflicting

packages/platform-sdk-generator/src/model/Model.ts
```diff
@@ -94,6 +94,8 @@ export class Model {
         throw new Error("Method not implemented.");
       case "binary":
         return new BinaryType();
+      case "string":
```
ssanjay1, line 97: Maps can have a key of raw string type, we never ran into this before but the new stream apis added something with a Map<string, othertype> so had to add string type here

packages/platform-sdk-generator/src/model/Type.ts
```diff
@@ -22,7 +22,9 @@ export abstract class Type {
   abstract getTsReferenceString(localNamespace: string | undefined): string;
 
   getPropertyDeclaration(name: string, localNamespace: string | undefined) {
-    return `${name}: ${this.getDeclaration(localNamespace)}`;
+    return `${name}: ${
```
ericanderson, line 25: This part of this change makes me a little worried. I am guessing that changing `Component`'s `getTsReferenceString` to check for `Record` might be the more stable change here. I think you can then also replace the change to `Component`'s `getDeclaration` to just call its own `getTsReferenceString` when getting the name.

ericanderson (changes requested):

packages/platform-sdk-generator/src/generateResource2.ts
ssanjay1: Discussing offline

packages/platform-sdk-generator/src/model/Type.ts
ssanjay1: Done

packages/platform-sdk-generator/src/model/Component.ts
```diff
@@ -49,10 +49,14 @@ export class Component extends Type {
   }
 
   getTsReferenceString(localNamespace: string) {
+    const revisedName = this.component.locator.localName.replace(
```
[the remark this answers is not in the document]
ericanderson: Im thinking a `if (this.component.locator.localName === "Record")` probably would have sufficed :)

ericanderson (approved):

packages/platform-sdk-generator/src/model/Component.ts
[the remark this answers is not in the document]
ssanjay1: *face palms*

ericanderson (changes requested):
We can close this right

ssanjay1:
Deleted platform APIs

PR #813, opened 2024-10-07
Geotime API support

ssanjay1:
Chains on https://github.com/palantir/osdk-ts/pull/809


packages/shared.test/src/handlers/loadObjectsEndpoints.ts
```diff
@@ -653,10 +692,12 @@ async function handleStreamPoints(
       | BaseAPIError
     >
   >[0],
+  useGeotime?: boolean,
 ) {
   const requestBody = await req.request.json();
-  const streamPointsResp =
-    streamPointsRequestHandlers[stableStringify(requestBody)];
+  const streamPointsResp = useGeotime
```
ssanjay1, line 698: The new API is designed to support normal timeseries as well, so once I add that support, we won't need to bifurcate and can just call the same API without needing to worry about backend implementation

packages/client/src/objectSet/toConjureObjectSet.ts
```diff
@@ -102,6 +107,10 @@ export async function getObjectSetBaseType(objectSet: ObjectSet) {
     case "searchAround":
       // TODO: Get a mapping using the link
       throw new Error("not implemented");
+    case "asBaseObjectTypes":
```
ssanjay1, line 110: This was from the platform API updates. PR

packages/client/src/util/timeseriesUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+
+import type { TimeSeriesQuery } from "@osdk/api";
+import { TimeseriesDurationMapping, TimeSeriesPoint } from "@osdk/api";
+import type { TimeRange } from "@osdk/internal.foundry.core";
+import { it } from "vitest";
+import { iterateReadableStream, parseStreamedResponse } from "./streamutils.js";
+
+export function getTimeRange(body: TimeSeriesQuery): TimeRange {
```
ssanjay1, line 23: getTimeRange functionality the same, just refactored to new file

packages/client/src/util/timeseriesUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+      type: "relative",
+      endTime: {
+        when: "AFTER",
+        value: body.$after!,
+        unit: TimeseriesDurationMapping[body.$unit],
+      },
+    };
+}
+
+export async function* asyncIterPointsHelper<
```
ssanjay1, line 50: Added this helper since both geotime and normal timeseries need this

packages/e2e.sandbox.catchall/src/runGeotimeSeriesTest.ts
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+import {
+  FintrafficAis,
+  RhemmingsObjectWithGtsrProperty2,
+} from "@osdk/e2e.generated.catchall";
+import { danubeClient } from "./client.js";
+
+export async function runGeotimeSeriesReferenceTests() {
```
ssanjay1, line 23: Added some tests to hit actual backend, using Danube tho

etc/api.report.api.md
```diff
@@ -433,6 +433,21 @@ export type GeoFilter_Within = {
...
+    // Warning: (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
+    // Warning: (tsdoc-malformed-inline-tag) Expecting a TSDoc tag starting with "{@"
+    // Warning: (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
+    asyncIterValues: (query?: TimeSeriesQuery) => AsyncGenerator<TimeSeriesPoint<T>>;
+    // Warning: (tsdoc-malformed-inline-tag) Expecting a TSDoc tag starting with "{@"
+    // Warning: (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
+    getAllValues: (query?: TimeSeriesQuery) => Promise<Array<TimeSeriesPoint<T>>>;
+    getLatestValue: () => Promise<TimeSeriesPoint<T>>;
+    // (undocumented)
+    lastFetchedPoint: TimeSeriesPoint<T> | undefined;
```
nihalbhatnagar, line 448: Don't we need to expose timestamp here too?

nihalbhatnagar: I'm blind

packages/api/src/timeseries/timeseries.ts
```diff
@@ -105,3 +105,45 @@ export interface TimeSeriesProperty<T extends number | string> {
     query?: TimeSeriesQuery,
   ) => AsyncGenerator<TimeSeriesPoint<T>>;
 }
+
+export interface GeotimeSeriesProperty<T extends GeoJSON.Point> {
```
ericanderson, line 109: Lets make everything in here `readonly`.

packages/api/src/timeseries/timeseries.ts
```diff
@@ -105,3 +105,45 @@ export interface TimeSeriesProperty<T extends number | string> {
...
+  /**
+   * Queries the last point of the Timeseries
+   */
+  getLatestValue: () => Promise<TimeSeriesPoint<T>>;
```
ericanderson, line 113: Should we call this `fetchLatestValue`?

packages/api/src/timeseries/timeseries.ts
```diff
@@ -105,3 +105,45 @@ export interface TimeSeriesProperty<T extends number | string> {
     query?: TimeSeriesQuery,
   ) => AsyncGenerator<TimeSeriesPoint<T>>;
 }
+
+export interface GeotimeSeriesProperty<T extends GeoJSON.Point> {
+  /**
+   * Queries the last point of the Timeseries
```
ericanderson, line 111: c/p error `Timeseries`?

packages/api/src/timeseries/timeseries.ts
```diff
@@ -105,3 +105,45 @@ export interface TimeSeriesProperty<T extends number | string> {
...
+     * @param query - a query representing either an absolute or relative range of time
+     * @example
+     *  const points = await employee.employeeStatus?.getAllPoints({
+        $after: 1,
+        $unit: "month",
+      });
+     */
+  getAllValues: (
+    query?: TimeSeriesQuery,
+  ) => Promise<Array<TimeSeriesPoint<T>>>;
```
ericanderson, line 125: `fetchAllValues()`?

ericanderson: or maybe just `fetchValues()` ?

packages/api/src/timeseries/timeseries.ts
```diff
@@ -105,3 +105,45 @@ export interface TimeSeriesProperty<T extends number | string> {
...
+     * @example
+     *  const iterator = employee.employeeStatus?.asyncIter({
+        $after: 1,
+        $unit: "month",
+      });
+      for await (const point of iterator) {
+          // Handle time series point
+      }
+     */
+  asyncIterValues: (
```
ericanderson, line 139: should this just be `asyncIter`?

packages/cli.cmd.typescript/package.json
```diff
@@ -30,8 +30,8 @@
     "@arethetypeswrong/cli": "^0.15.2",
     "@osdk/cli.common": "workspace:~",
     "@osdk/generator": "workspace:~",
-    "@osdk/internal.foundry.core": "0.2.0",
-    "@osdk/internal.foundry.ontologiesv2": "0.2.0",
+    "@osdk/internal.foundry.core": "0.3.0-beta.0",
```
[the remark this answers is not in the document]
ericanderson: yay! Never version syncs them to the non internal too which should make this much less work in the future

packages/client/src/createGeotimeSeriesProperty.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.#triplet = [objectApiName, primaryKey, propertyName];
+  }
+
+  public async getLatestValue() {
+    const latestPointPromise = OntologiesV2.TimeSeriesValueBankProperties
+      .getLatestValue(
+        this.#client,
+        await this.#client.ontologyRid,
+        ...this.#triplet,
+      ) as Promise<TimeSeriesPoint<T>>;
```
ericanderson, line 56: Just because `as` COULD cause issues, I wonder if its better to just set the return type of the function to `Promise<TimeSeriesPoint<T>>` instead? If its not assignable it will error.

ericanderson: Do we run any risk of `getLatestValue` later returning additional information we didn't intend?

packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -85,7 +86,7 @@ describe("convertWireToOsdkObjects", () => {
 
     // Should not have $title
     expect(JSON.stringify(employee)).toMatchInlineSnapshot(
-      `"{"employeeId":50030,"fullName":"John Doe","office":"NYC","class":"Red","startDate":"2019-01-01","employeeStatus":{},"$apiName":"Employee","$objectType":"Employee","$primaryKey":50030}"`,
+      `"{"employeeId":50030,"fullName":"John Doe","office":"NYC","class":"Red","startDate":"2019-01-01","employeeStatus":{},"employeeLocation":{},"$apiName":"Employee","$objectType":"Employee","$primaryKey":50030}"`,
```
ericanderson, line 89: Hmm. Is this how we want this value to serialize?

packages/shared.test/src/stubs/geotimeseriesrequests.ts
```diff
@@ -0,0 +1,126 @@
...
+  StreamTimeSeriesValuesRequest,
+  StreamTimeSeriesValuesResponse,
+  TimeSeriesPoint,
+} from "@osdk/internal.foundry.core";
+import stableStringify from "json-stable-stringify";
+
+const pointPrimaryKey1: string = JSON.stringify({
+  primaryKey: "50030",
+  propertyName: "employeeLocation",
+});
```
ericanderson, line 29: Thank you for adding all of this

ericanderson (commented):
Mostly naming questions. I know the names match what we shipped already I'm just wondering if they are confusing?

Not approving cause i know youre going to push another up anyway

packages/client/src/object/convertWireToOsdkObjects.test.ts
ssanjay1: Mmm, I just did parity with how normal time series does it. We could serialize it as the trackId and integration RID perhaps? Like that's what we get over the wire, though then I feel like timeseries should also be updated to have the timeseries ID

packages/api/src/timeseries/timeseries.ts
ssanjay1: Should I do this to normal timeseries property as well?

packages/client/src/createGeotimeSeriesProperty.ts
ssanjay1: I can change the return type, and no I don't forsee atleast adding anything new like that. Like it will always be a point of {time, value:GEOJSON of some form}

ssanjay1: If we need something drastically different, I think it'd just have to be a new endpoint

packages/api/src/timeseries/timeseries.ts
ericanderson: yeah

packages/api/src/timeseries/timeseries.ts
ssanjay1: Yup, good catch

packages/api/src/timeseries/timeseries.ts
ssanjay1: Spoke offline, keeping as is for now

packages/shared.test/src/handlers/util/handleOpenApiCall.ts
```diff
@@ -59,7 +59,7 @@ export type ExtractBody<
 
 export type ExtractResponse<
   X extends ((...args: any[]) => Promise<any>),
-> = Awaited<ReturnType<X>> & {};
+> = Awaited<ReturnType<X>>;
```
ssanjay1, line 62: Had to remove this so we could allow `|undefined` types in return, like we do for geotime series points, because we could fetchLatestPoint on a track with no data, which the API would return no data for.

packages/client/src/object/convertWireToOsdkObjects.test.ts
ssanjay1: If we feel strongly about this, can do this as a FLUP

packages/cli.cmd.typescript/package.json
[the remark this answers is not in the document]
ericanderson: *newer

packages/client/src/object/convertWireToOsdkObjects.test.ts
ericanderson: The case I want to think about is nextjs, where you load the thing on the backend and then send it to the frontend by serializing it. I think this is okay especially if we add last fetched. Honestly, I don't think nextjs is going to be a good platform for something that wants live updates.

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -127,6 +138,20 @@ export function createOsdkObject<
               p as string,
             );
           }
+          if (propDef.type === "geotimeSeriesReference") {
+            const instance = target[CachedPropertiesRef].get(p);
+            if (instance != null) {
+              return instance;
+            }
+            const geotimeProp = new GeotimeSeriesPropertyImpl<GeoJSON.Point>(
```
ssanjay1, line 146: I will do a FLUP for the other properties above to be cached, I wanted to keep this PR just to geotime stuff rn

ericanderson (approved):
Approved but we should add a test for interfaces to make sure everything maps as we expect / caching is right

PR #819, opened 2024-10-08
Fix gateway using ts 5 on 1.3

ericanderson:
Legacy packages use ts 4.9 to compile to ensure compatibility and I fixed up the two places we were using the TS 5 features.


nihalbhatnagar (approved):

packages/gateway/src/generated/openapi/request/index.ts
```diff
@@ -14,4 +14,4 @@
  * limitations under the License.
  */
 
-export type * from "./OpenApiRequest.js";
+export type { OpenApiRequest } from "./OpenApiRequest.js";
```
nihalbhatnagar, line 17: Are these the 5.x features?

ericanderson: The feature was `type *` was introduced in ts 5.

PR #831, opened 2024-10-09
Update.deps

ericanderson:


packages/client/src/object/Cache.test.ts
```diff
@@ -272,7 +272,7 @@ describe("AsyncCache", () => {
               "rejected",
             );
             expect(asyncCacheSpies.get.mock.settledResults[i].value)
-              .toMatchInlineSnapshot(`[Error: aError]`);
+              .toMatchObject(new Error("aError"));
```
ssanjay1, line 275: Oh did that break with new vitest?

ssanjay1 (approved):

packages/client/src/object/Cache.test.ts
ericanderson: yeah, because you arent supposed to inline snapshot in a loop

PR #834, opened 2024-10-09
Fix generation without --beta

ericanderson:


.changeset/lovely-llamas-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/tmp-foundry-sdk-generator": patch
```
ssanjay1, line 2: Shouldn't this be foundry-sdk-generator?

ssanjay1 (approved):

PR #852, opened 2024-10-11
Add reference updates to OSW

nihalbhatnagar:
Allows for reference updates to come through the OSW API. Also adds test framework


packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -226,7 +252,59 @@ describe("ObjectSetListenerWebsocket", async () => {
...
+
+          sendReferenceUpdatesResponse(ws, `${idNum2}`);
+          await listenerPromise;
+          expect(listener.onChange).toHaveBeenCalled();
+          expect(updateReceived).toMatchInlineSnapshot(`
+            {
+              "object": {
+                "$apiName": "Employee",
+                "$objectType": "Employee",
+                "$primaryKey": {
```
nihalbhatnagar, line 292: This is in a weird state right now. The type of the primary key on a reference update is a Record of property type to property value instead of just a value. Since I haven't actually been able to hit the backend, I haven't been able to confirm what this looks like. I suspect I just need to index in via the primary key type. This PR won't merge until I test and resolve this

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -152,11 +152,15 @@ export function createOsdkObject<
             if (instance != null) {
               return instance;
             }
+
             const geotimeProp = new GeotimeSeriesPropertyImpl<GeoJSON.Point>(
               client,
               objectDef.apiName,
               target[RawObject][objectDef.primaryKeyApiName as string],
               p as string,
+              rawValue.type === "geotimeSeriesValue"
```
nihalbhatnagar, line 161: This type should only exist on the raw property value if it comes from a reference value update.

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -230,14 +230,23 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
       properties,
       listener,
     ) => {
+      const referenceProperties = properties.filter(p => {
+        objectType.__DefinitionMetadata?.properties[p].type
+          === "geotimeSeriesReference";
```
nihalbhatnagar, line 235: What's the best way to enforce this type? Should we expose an from WirePropertyTypes in API, or should this be maintained locally.

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
nihalbhatnagar: So the primary key looks funky but will be fixed when I can confirm what the response is

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -126,8 +135,15 @@ describe("ObjectSetListenerWebsocket", async () => {
...
       });
 
+      listenerPromise = new Promise((resolve) => {
+        listenerPromiseResolveFn = resolve;
+      });
+
       listener = {
-        onChange: vi.fn(),
+        onChange: vi.fn((o) => {
+          listenerPromiseResolveFn();
```
nihalbhatnagar, line 144: I need to do this to make sure the test doesn't end before we receive the update

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -175,16 +180,29 @@ export class ObjectSetListenerWebsocket {
...
+
+    const objDef = await this.#client.ontologyProvider.getObjectDefinition(
+      objectType.apiName,
+    );
+
+    const objectProperties = properties.filter((p) =>
+      objDef.properties[p].type !== "geotimeSeriesReference"
+    );
+    const referenceProperties = properties.filter((p) =>
+      objDef.properties[p].type === "geotimeSeriesReference"
```
nihalbhatnagar, line 192: I think my old comment is outdated now, but the question is the same. Where can I best ensure this is kept up to date? There shouldn't be any breaks if it isn't properly mapped(won't ever request it if not in this union and we can request these updates regularly)

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -436,8 +457,42 @@ export class ObjectSetListenerWebsocket {
     const referenceUpdates = payload.updates.filter((update) =>
       update.type === "reference"
     );
+    const osdkObjectsWithReferenceUpdates = await Promise.all(
+      referenceUpdates.map(async (o) => {
+        const osdkObjectArray = await convertWireToOsdkObjects(
+          this.#client,
+          [{
+            __apiName: o.objectType,
+            __primaryKey: Object.values(o.primaryKey)[0],
```
nihalbhatnagar, line 466: `o.primaryKey` is an object that contains a single mapping of the primary key property name to the property value.

packages/api/src/objectSet/EXPERIMENTAL_ObjectSetListener.ts
```diff
@@ -47,6 +47,6 @@ type ObjectUpdate<
   O extends ObjectOrInterfaceDefinition,
   P extends PropertyKeys<O>,
 > = {
-  object: Osdk.Instance<O, never, P>;
+  object: Osdk.Instance<O, "$notStrict", P>;
```
nihalbhatnagar, line 50: I want all properties to be `| undefined` since it's possible if you get a reference update, you won't get an updated value

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -436,8 +457,42 @@ export class ObjectSetListenerWebsocket {
...
+        sub.listener.onChange?.(osdkObject);
+      }
+    }
 
     const osdkObjects = await Promise.all(objectUpdates.map(async (o) => {
+      const keysToDelete = Object.keys(o.object).filter((key) =>
+        sub.requestedReferenceProperties.includes(key)
+      );
+      for (const key of keysToDelete) {
+        delete o.object[key];
```
nihalbhatnagar, line 493: I added this because I wanted the contract here to essentially be if it's undefined, theres no update. Although we do get the property back when it's requested as a reference value, I thought it would be confusing for a user to find this never undefined and there is no latestFetchedValue sometimes, although I don't have a strong opinion here

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -126,8 +135,15 @@ describe("ObjectSetListenerWebsocket", async () => {
         objectSetExpiryMs: OBJECT_SET_EXPIRY_MS,
       });
 
+      listenerPromise = new Promise((resolve) => {
+        listenerPromiseResolveFn = resolve;
+      });
```
ericanderson, line 140: I'm sure we use this pattern elsewhere but could we switch to https://www.npmjs.com/package/p-defer

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: i dont know yet what we will do here but this is fine for now

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: given we know what the primary key is from the metadata should we just do this directly instead of creating the temporary array to index into?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Hmm, if I am showing where a boat is and the backend has timed out the values however it decides to do that and tells me the boat is gone, i think we want it to reflect that?

packages/e2e.sandbox.catchall/src/runSubscriptionsTest.ts
```diff
@@ -14,42 +14,83 @@
...
+  client(OsdkTestObject)[__EXPERIMENTAL__NOT_SUPPORTED_YET_subscribe](
+    [
+      "primaryKey_",
+      "stringProperty",
+    ],
+    {
+      onChange(object) {
+        console.log(
+          "Object with primaryKey ",
+          object.object.primaryKey_,
```
ericanderson, line 33: this is a real property whose name happens to be `.primaryKey_` right?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Maybe we use `null` to reflect that we loaded a value but it does not exist?

packages/e2e.sandbox.catchall/src/runSubscriptionsTest.ts
nihalbhatnagar: yes, thats correct

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
nihalbhatnagar: sure, I can store the primary key name in the subscription name

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
nihalbhatnagar: Isn't this implementation better here since we can await directly on the listener which should resolve pretty quickly?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -195,6 +196,7 @@ export class ObjectSetListenerWebsocket {
     const sub: Subscription<Q, P> = {
       listener: fillOutListener<Q, P>(listener),
       objectSet,
+      primaryKeyPropertyName: objDef.primaryKeyApiName,
```
ericanderson, line 199: a few lines up we load the `objDef` right? I dont think we need to pass this in?

nihalbhatnagar: We pass this in to use in #handleMessage_objectSetChanged, where we use the API name to delete. In #handleMessage_objectSetChanged, we dont load the def

ericanderson (approved):
Seems like a good v0 to experiment with

PR #858, opened 2024-10-14
properly infer return type of ObjectSet.fetchPage

ericanderson:


nihalbhatnagar (approved):

ssanjay1 (dismissed):

packages/api/src/OsdkObjectFrom.test.ts
```diff
@@ -15,28 +15,60 @@
  */
 
 import { describe, expectTypeOf, it } from "vitest";
+import type { NullabilityAdherence } from "../build/esm/index.js";
```
ssanjay1, line 18: Seems like wrong import

ssanjay1 (changes requested):

packages/api/src/OsdkObjectFrom.test.ts
ericanderson: thank you

ssanjay1 (approved):

PR #868, opened 2024-10-15
Use new types for links

ssanjay1:


ericanderson (changes requested):
Need to follow up to fix the return type

ericanderson (approved):

PR #870, opened 2024-10-15
Change to a version of shared client less error prone

ericanderson:


packages/client/src/createMinimalClient.ts
```diff
@@ -70,9 +70,9 @@ export function createMinimalClient(
     "ontologyProvider"
   > as any;
 
-  return Object.assign(minimalClient, {
+  return Object.freeze(Object.assign(minimalClient, {
```
ssanjay1, line 73: Cool, we're doing this to make props non-configurable and non-writable right?

packages/client/src/Client.ts
```diff
@@ -111,8 +111,13 @@ export interface Client extends SharedClient<MinimalClient> {
     objs: Osdk<T>[],
     links: string[],
   ) => AsyncGenerator<BulkLinkResult, void, undefined>;
+
+  [additionalContext]: MinimalClient;
```
ssanjay1, line 115: Ok so we made a symbol internal to client here to hide stuff that isn't need for sharedClient

ssanjay1 (approved):

packages/client/src/createMinimalClient.ts
ericanderson: Yeah, i dont think its needed but its a nice failsafe and we never reassign this stuff

packages/client/src/Client.ts
ericanderson: exactly. And because its internal to the package we are not in trouble of symbol issues

ssanjay1 (dismissed):

ssanjay1 (approved):

ssanjay1 (approved):

PR #876, opened 2024-10-16
Packages use more specific versions instead of indirection through sh…

ericanderson:
…ared.net


packages/shared.net/src/FetchAsJsonFn.ts
```diff
@@ -18,6 +18,7 @@
  * Fetches the given URL, converting the body to JSON, setting the Content-Type header to application/json,
  * and calling .json() on the response.
  */
+/** @deprecated */
 export type FetchAsJsonFn = (
```
ssanjay1, line 22: When did we deprecate this?

ssanjay1 (approved):

packages/shared.net/src/FetchAsJsonFn.ts
ericanderson: its not used but im playing nice with semver

PR #880, opened 2024-10-16
Fixes the shape of the object that we initialize a GTSR with on reference updates

nihalbhatnagar:


packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -149,7 +150,13 @@ export function createOsdkObject<
               target[RawObject][objectDef.primaryKeyApiName as string],
               p as string,
               rawValue.type === "geotimeSeriesValue"
-                ? { time: rawValue.timestamp, value: rawValue.position }
+                ? {
```
nihalbhatnagar, line 153: How could I have avoided this(besides triple checking this)? Is there a better practice here when dealing with any's?

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -287,10 +287,13 @@ describe("ObjectSetListenerWebsocket", async () => {
...
-                    "value": [
-                      100,
-                      200,
-                    ],
+                    "value": {
+                      "coordinates": [
+                        100,
+                        200,
+                      ],
+                      "type": "Point",
```
nihalbhatnagar, line 295: This should catch this error next time if the shape ever changes accidentally

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
ericanderson: You can do `{ foo: "bar" } satisfies BlahInterface` as an extra check

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
ericanderson: Tests are the most valuable here because they tend to prevent regressions the best

ericanderson (approved):

PR #881, opened 2024-10-16
Remove remaining exported symbols

ericanderson:


etc/api.report.api.md
```diff
@@ -95,7 +95,7 @@ export namespace ActionMetadata {
 
 // @public
 export namespace ActionParam {
-    export type ObjectSetType<T extends ObjectTypeDefinition> = BaseObjectSet<T>;
+    export type ObjectSetType<T extends ObjectTypeDefinition> = ObjectSet<T>;
```
ericanderson, line 98: BaseObjectSet relied on a symbol and also didnt really provide much.

etc/api.report.api.md
```diff
@@ -223,7 +223,9 @@ export type Augments = Record<string, string[]>;
 // @public (undocumented)
 export interface BaseObjectSet<Q extends ObjectOrInterfaceDefinition> {
     // (undocumented)
-    readonly [ObjectSetType]: Q;
+    readonly $objectSetInternals?: {
```
ericanderson: Since we don't have the symbol we must now use this to brand.

etc/api.report.api.md
```diff
@@ -223,7 +223,9 @@ export type Augments = Record<string, string[]>;
 // @public (undocumented)
 export interface BaseObjectSet<Q extends ObjectOrInterfaceDefinition> {
     // (undocumented)
-    readonly [ObjectSetType]: Q;
+    readonly $objectSetInternals: {
```
ericanderson, line 226: How we bind the brand now / useful for subscribe

packages/api/src/objectSet/BaseObjectSet.ts
```diff
@@ -16,9 +16,7 @@
 
 import type { ObjectOrInterfaceDefinition } from "../ontology/ObjectOrInterface.js";
 
-declare const ObjectSetType: unique symbol;
-
 // mostly a marker interface to simplify types that must be exported/shared
 export interface BaseObjectSet<Q extends ObjectOrInterfaceDefinition> {
-  readonly [ObjectSetType]: Q;
+  readonly $objectSetInternals: { def: Q };
```
ericanderson, line 21: This allows a client that is newer than the api to still work

packages/api/src/objectSet/BulkLinkResult.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { OsdkBase } from "../OsdkBase.js";
+
+export interface EXPERIMENTAL_BulkLinkResult {
```
ericanderson, line 19: Just moved from `@osdk/client`

packages/client/src/object/convertWireToOsdkObjects/InternalSymbols.ts
```diff
@@ -16,27 +16,33 @@
...
   process.env.MODE !== "production" ? "RawObject" : undefined,
 );
 
+/** @internal */
 export const ClientRef = Symbol(
   process.env.MODE !== "production" ? "ClientRef" : undefined,
 );
 
+/** @internal */
 export const CachedPropertiesRef = Symbol(
```
ericanderson, line 46: Make sure we arent exposting these symbols

packages/e2e.sandbox.catchall/src/public/checkUnstableBulkLinks.ts
```diff
@@ -53,10 +53,13 @@ export async function checkUnstableBulkLinks() {
       linkApiName,
       otherObjectApiName,
       otherObjectPk,
-    } of client[__EXPERIMENTAL__NOT_SUPPORTED_YET__getBulkLinks](employees, [
-      "ventures",
-      "amishsSyncGroup",
-    ])
+    } of client(__EXPERIMENTAL__NOT_SUPPORTED_YET__getBulkLinks).invoke(
```
ericanderson, line 56: switch the style of invocation

packages/client/src/Client.ts
```diff
@@ -70,6 +67,10 @@ export interface Client extends SharedClient, OldSharedClient {
     o: Q,
   ): QuerySignatureFromDef<Q>;
 
+  <Q extends Experiment<"2.0.8">>(
```
nihalbhatnagar, line 70: Are we going to need to update these for every cut?

packages/client/src/createClient.ts
```diff
@@ -108,11 +119,13 @@ export function createClientInternal(
...
       | ObjectOrInterfaceDefinition
       | ActionDefinition<any>
-      | QueryDefinition<any>,
+      | QueryDefinition<any>
+      | Experiment<"2.0.7">,
   >(o: T): T extends ObjectTypeDefinition ? ObjectSet<T>
     : T extends InterfaceDefinition ? MinimalObjectSet<T>
     : T extends ActionDefinition<any> ? ActionSignatureFromDef<T>
     : T extends QueryDefinition<any> ? QuerySignatureFromDef<T>
+    : T extends Experiment<"2.0.7"> ? { invoke: ExperimentFns<T> }
```
nihalbhatnagar, line 128: I don't fully understand why these are 2.0.7 and objectSet is 2.0.8

packages/client/src/Client.ts
ericanderson: Every time we add new experimental APIs or take them away.

packages/client/src/createClient.ts
ericanderson: Mistake on my part. I first did all the things "2.0.7" and then realized the types were actually going to be introduced in 2.0.8 and missed this one. Then it seems the `defineProperties` below with the `as` acted as a an any in effect and didn't catch it.

packages/client/src/createClient.ts
```diff
@@ -131,6 +144,64 @@ export function createClientInternal(
...
         : never) as any;
+    } else if (o.type === "experiment") {
+      switch (o.name) {
+        case __EXPERIMENTAL__NOT_SUPPORTED_YET__getBulkLinks.name:
+          return {
+            invoke: createBulkLinksAsyncIterFactory(
+              clientCtx,
+            ),
+          } as any;
+        case __EXPERIMENTAL__NOT_SUPPORTED_YET__preexistingObjectSet.name:
```
nihalbhatnagar, line 155: Do these as any's have any effect on the types returned from these?

packages/client/src/Client.ts
nihalbhatnagar: I see, so if in 2.0.12 we add an experimental API, we will need to change this to 2.0.12, but are we also going to need to update the rest of the API's to the new version to ensure compatibility?

packages/client/src/createClient.ts
ericanderson: not really, these are all just moved from their old locations so the logic doesnt change and the interface for `Client` actually dictates what the consumer sees.

I went down a rabbit hole of trying to make these super strict but it was getting out of hand and realized that it would be better to follow up later with trying to prevent the future issues.

packages/client/src/createClient.ts
ssanjay1: Do we need to address this?

packages/client/src/Client.ts
ericanderson: We would make this `Experiment<"2.0.8"> | Experiment<"2.0.12">`

ssanjay1 (approved):

PR #882, opened 2024-10-17
Initial Subscribe Sandbox App

nihalbhatnagar:
https://github.com/user-attachments/assets/f16d34ed-6565-41fe-9c2a-dee29202b2db


packages/e2e.sandbox.subscribeapp/src/MapComponent.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+          updateCoordinates(primaryKey, newCoordinates);
+        }
+
+        initializeEntity({
+          id: primaryKey,
+          routeId: object.object.routeId ?? entityInfo[primaryKey].routeId,
+          nextStopId: object.object.nextStopId
+            ?? entityInfo[primaryKey].nextStopId,
+        });
```
nihalbhatnagar, line 147: I have multiple thoughts on this. This seems like the ideal way for a user to update their cache since the properties are coming back as `| undefined`, and they can just fill in the blanks with what is already.  Is it possible a user just does `!` on the returned object, then they receive a reference update, and then it all breaks? I would like to hope that developers follow simple practices such as good handling here, do we need to account for those who won't? However, I am liking the decision we made to include reference updates alongside the rest of these as if you do have proper error handling AND you set up your cache in the right way, the best practice becomes clear. However, the second part of the AND is the bigger worry to me

packages/e2e.sandbox.subscribeapp/src/MapComponent.tsx
```diff
@@ -0,0 +1,177 @@
...
+}
+
+const MapComponent: React.FC = () => {
+  const center: Coordinates = { lat: 40.7128, lng: -74.0060 };
+
+  const [entityInfo, setEntityInfo] = React.useState<EntityInformationMap>({});
+
+  const [objectCoordinates, setObjectCoordinates] = React.useState<
+    ObjectCoordinatesMap
+  >({});
```
nihalbhatnagar, line 45: I separated these out, mainly so I could avoid potentially undefined properties. Users are definitely going to have to deal with this, and the best hope is that they're going to handle this in an optimal way.

packages/e2e.sandbox.subscribeapp/CHANGELOG.md
nihalbhatnagar, line 1: Where do I need to configure this?

packages/e2e.sandbox.subscribeapp/src/MapComponent.tsx
```diff
@@ -0,0 +1,177 @@
...
+          primaryKey,
+          ". Length of map is ",
+          Object.keys(entityInfo).length,
+        );
+
+        if (object.state === "REMOVED") {
+          deleteEntity(entityInfo[primaryKey]);
+        }
+
+        if (object.object.positionId != null) {
```
rodneyeng, line 131: Is this actually something more like event.bus.positionId in context?

nihalbhatnagar: yes, I should enforce better naming conventions since it's a sample app

PR #883, opened 2024-10-17
Remove notstrict

ssanjay1:


packages/api/src/OsdkObjectFrom.ts
```diff
@@ -158,14 +158,6 @@ export type GetPropsKeys<
   : IsAny<P> extends true ? PropertyKeys<Q>
   : P;
 
-export type GetProps<
```
ssanjay1, line 161: I deleted because we didn't need it right now, but I can leave it in case we want it later?

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -295,52 +297,53 @@ describe("ObjectSet", () => {
...
-          expectTypeOf(result.data[0]).branded.toEqualTypeOf<
-            Employee.OsdkObject
-          >();
-          expectTypeOf(result.data[0]).branded.toEqualTypeOf<
-            Employee.OsdkObject<never>
-          >();
-        });
-      });
-    });
+    // describe("strictNonNull: \"drop\"", () => {
```
ssanjay1, line 300: Commented these tests out for now in case we bring back feature in the future

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -375,22 +377,11 @@ describe("ObjectSet", () => {
...
-        `[Error: Unable to safely convert objects as some non nullable properties are null]`,
-      );
-    });
-  });
-
   describe.each(["fetchOne", "fetchOneWithErrors"] as const)("%s", (k) => {
     describe("strictNonNull: false", () => {
       describe("includeRid: true", () => {
         it("returns bad data", async () => {
+          // Look at this
```
ssanjay1, line 384: Yea so technically realized, our types are still permissive to having $ __EXPERIMENTAL_strictNonNull being passed in, do we want to restrict that?

packages/generator/src/v2.0/wireObjectTypeV2ToSdkObjectConstV2.ts
```diff
@@ -224,7 +224,14 @@ ${
             ""}`,
           `$PropType[${JSON.stringify(propertyDefinition.type)}]${
             propertyDefinition.multiplicity ? "[]" : ""
-          }${propertyDefinition.nullable || !strict ? `| undefined` : ""}`,
+          }${
```
ssanjay1, line 227: Enforce primary keys are non-nullable even in normal props

packages/api/src/OsdkObjectFrom.ts
ericanderson: Yeah thats fine

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -295,52 +297,53 @@ describe("ObjectSet", () => {
...
+    //         Employee.OsdkObject<never>
+    //       >();
+    //     });
+    //   });
+    // });
 
     describe("strictNonNull: false", () => {
       describe("includeRid: true", () => {
         it("returns bad data", async () => {
+          // look here
```
ericanderson, line 346: whats this

packages/client/src/objectSet/ObjectSet.test.ts
ericanderson: I am thinking we remove it for now. Its not like we want it showing up for customers right

ericanderson (commented):
On top of the questions you asked, we should also just alias `StrictProps` to `Props` for now instead of regen?

packages/client/src/objectSet/ObjectSet.test.ts
ssanjay1: This was comment I added for myself to point out to you later lol, will delete as I address things

ericanderson (approved):

PR #890, opened 2024-10-21
fix props types

ssanjay1:


packages/api/src/aggregate/WhereClause.ts
```diff
@@ -168,7 +170,21 @@ type FilterFor<PD extends ObjectMetadata.Property> = PD["multiplicity"] extends
     : PD["type"] extends "geopoint" | "geoshape" ? GeoFilter
     : PD["type"] extends "datetime" | "timestamp" ? DatetimeFilter
     : PD["type"] extends "boolean" ? BooleanFilter
-    : NumberFilter); // FIXME we need to represent all types
+    : PD["type"] extends
```
ssanjay1, line 173: Fixed this so we don't default to number filter for properties like geotimeseries or timeseries

packages/api/src/aggregate/WhereClause.ts
```diff
@@ -195,5 +211,41 @@ export type WhereClause<
...
+    [P in PropertyKeys<T>]?:
+      | RequireExactlyOne<
+        Pick<
+          WhereClauseKeys<T, P>,
+          NonNeverKeys<
+            WhereClauseKeys<T, P>
+          >
+        >
+      >
+      | FilterRawType<CompileTimeMetadata<T>["properties"][P]>;
```
ssanjay1, line 223: Use this so people can still pass in raw number, strings,booleans instead of having to do `$eq`

packages/api/src/aggregate/WhereClause.ts
```diff
@@ -195,5 +211,41 @@ export type WhereClause<
...
   };
+
+type WhereClauseKeys<
+  T extends ObjectOrInterfaceDefinition,
+  P extends PropertyKeys<T>,
+> = FilterDollarSignKeys<
+  FlattenUnion<FilterFor<CompileTimeMetadata<T>["properties"][P]>>
+>;
+
+type FlattenUnion<T> = {
```
ssanjay1, line 233: Referred to this https://www.roryba.in/programming/2019/10/12/flattening-typescript-union-types.html#:~:text=unnecessary%20runtime%20cost.-,FlattenUnion,-Now%2C%20the%20main

packages/client/src/internal/conversions/modernToLegacyWhereClause.test.ts
```diff
@@ -26,10 +26,18 @@ describe(modernToLegacyWhereClause, () => {
   describe("api namespaces", () => {
     describe("interfaces", () => {
       it("properly converts shortname to fqApiName", () => {
-        const T: ObjectOrInterfaceDefinition = {
+        const T = {
           type: "interface",
           apiName: "a.Foo",
-        };
+          __DefinitionMetadata: {
```
ssanjay1, line 32: Had to add all this info because our old whereClause type would just default to `NumberFilters` for anything where the types didn't match the rest of our branching logic

packages/api/src/aggregate/WhereClause.ts
```diff
@@ -168,7 +170,21 @@ type FilterFor<PD extends ObjectMetadata.Property> = PD["multiplicity"] extends
     : PD["type"] extends "geopoint" | "geoshape" ? GeoFilter
     : PD["type"] extends "datetime" | "timestamp" ? DatetimeFilter
     : PD["type"] extends "boolean" ? BooleanFilter
-    : NumberFilter); // FIXME we need to represent all types
+    : PD["type"] extends
+      "double" | "integer" | "long" | "float" | "decimal" | "byte"
+      ? NumberFilter
+    :
+      | MakeFilter<"$eq" | "$ne", string>
```
nihalbhatnagar, line 177: Should we use BaseFilter here instead?

nihalbhatnagar: If we don't want the | T then is it possible to add $in?

ssanjay1:
Closed, used EA alternate idea

PR #891, opened 2024-10-22
Adds additional handler to support subscription success notifications

nihalbhatnagar:


packages/api/src/experimental/subscribe.ts
```diff
@@ -22,21 +22,23 @@ import type {
 import type { EXPERIMENTAL_ObjectSetListener } from "../public/unstable.js";
 import type { Experiment } from "./Experiment.js";
 
+type NonEmptyArray<T> = [T, ...T[]];
```
ericanderson, line 25: I'm 80% sure we shouldn't use this type. I've seen this type show up in other code bases where it just ends up proliferating all over the place causing more harm than good.

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -451,7 +422,7 @@ export class ObjectSetListenerWebsocket {
     payload: ObjectSetUpdates,
   ) => {
     const sub = this.#subscriptions.get(payload.id);
-    invariant(sub, `Expected subscription id ${payload.id}`);
+    if (sub == null) return;
```
ericanderson, line 425: The old code assumed you would always be able to find by payload id. Is this no longer the case? Why?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -323,16 +299,11 @@ export class ObjectSetListenerWebsocket {
...
     sub.status = newStatus;
     // make sure listeners do nothing now
     sub.listener = fillOutListener<Q, any>({});
-    if (sub.expiry) {
-      clearTimeout(sub.expiry);
-      sub.expiry = undefined;
-    }
     this.#subscriptions.delete(sub.subscriptionId);
-
+    this.#sendSubscribeMessage();
```
ericanderson, line 306: Why was this added here?

nihalbhatnagar: To unsubscribe from an object set, you need to send another subscribe request without the given subscription request

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
nihalbhatnagar: I forgot to test this, but at least in the new code when we delete the subscription in unsubscribe, without necessarily unsubscribing by sending a new request, this would error because we still received updates for that sub id. Even though we are unsubscribing and this should not happen, I wanted to remove this as a failsafe in case we still receive an update here

packages/api/src/experimental/subscribe.ts
nihalbhatnagar: I see, how should we enforce a non empty array here then? Does it make sense to throw if we receive an empty array? I don't want to accept empty arrays because that will then request all non reference backed properties and some users might become reliant on that. Also curious to know what issues it creates

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: Okay

packages/api/src/experimental/subscribe.ts
nihalbhatnagar: Discussed offline, will add a throw here on a FLUP

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
ericanderson: I see, so the old code kept getting updates (that were dropped because we reset the listener to noop) but this pre-emptively unsubscribes to stop the network traffic.

packages/api/src/experimental/subscribe.ts
```diff
@@ -22,14 +22,12 @@ import type {
...
 
-type NonEmptyArray<T> = [T, ...T[]];
-
 type subscribeFn = <
   Q extends ObjectOrInterfaceDefinition,
   const P extends PropertyKeys<Q>,
 >(
   objectSet: ObjectSet<Q>,
-  properties: NonEmptyArray<P>,
+  properties: Array<P>,
```
ericanderson, line 30: Probably should make this a `ReadOnlyArray<P>` too now that I think about it, so that people who have those in their redux can use it

ericanderson (approved):
Feel free to do followup if you want

packages/api/src/experimental/subscribe.ts
nihalbhatnagar: Will flup this with the other changes

PR #912, opened 2024-10-23
Use fixed versions for create-app client and oauth deps

ericanderson:
`"@osdk/client": "latest"` with `npm install` in a created package either always or often just consults `npm cache`. Even though we think our users get latest they are not. 

This change is not perfect as it requires us to keep updating the version by hand but is superior to what we had.

### DIFF
--- a/packages/create-app/src/run.ts
+++ b/packages/create-app/src/run.ts
@@ -105,12 +105,23 @@ export async function run(
     );
   }
 
+  const ourPackageJsonPath = findUpSync("package.json", {
+    cwd: fileURLToPath(import.meta.url),
+  });
+
+  const ourPackageJsonVersion = ourPackageJsonPath
+    ? JSON.parse(fs.readFileSync(ourPackageJsonPath, "utf-8")).version
+    : undefined;
+
+  const clientVersion = process.env.PACKAGE_CLIENT_VERSION
+    ?? ourPackageJsonVersion;
+
   const templateContext: TemplateContext = {
     project,
     foundryUrl,
     osdkPackage,
     corsProxy,
-    clientVersion: "^2.0.7",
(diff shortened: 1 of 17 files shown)


packages/create-app/src/run.ts
```diff
@@ -110,6 +110,7 @@ export async function run(
     foundryUrl,
     osdkPackage,
     corsProxy,
+    clientVersion: "^2.0.7",
```
nihalbhatnagar, line 113: Will this run into the same cache issue if we don't update this

nihalbhatnagar (dismissed):

packages/create-app/src/run.ts
ericanderson: it will but at least at the point that we wrote this, we know that 2.0.7 should work with the generated code. Otherwise you will see situations where the generated code uses a version of the client that the user doesnt get.

packages/create-app/src/run.ts
```diff
@@ -105,11 +105,23 @@ export async function run(
...
 
+  const ourPackageJsonPath = findUpSync("package.json", {
+    cwd: fileURLToPath(import.meta.url),
+  });
+
+  const ourPackageJsonVersion = ourPackageJsonPath
+    ? JSON.parse(fs.readFileSync(ourPackageJsonPath, "utf-8")).version
+    : undefined;
+
+  const clientVersion = process.env.PACKAGE_CLIENT_VERSION
```
ericanderson, line 116: `process.env.PACKAGE_CLIENT_VERSION` is supposed to be replaced during the build process but the test process does not do that. We can try to rely on either mechanism here to be sure we get a version.

ssanjay1 (approved):

PR #926, opened 2024-10-25
Include API error info in CLI error messages

tzyl:
Include API error info about the public gateway errors https://www.palantir.com/docs/foundry/api/general/overview/errors/ in CLI error messages to help debug. Currently only the network status code and text are exposed, but not any specific detail about the API error.

For the specific case of 404, also include a tip like 401 and 403 to make it clear that this can mean either the resource doesn't exist or the token is missing permissions.

Example error for non-verbose output now:

```
ℹ Palantir OSDK CLI 0.25.0-beta.4                                                                          2:15:35 PM

ℹ Auto version inferred next version to be: 0.109.0-1-g945314d-dirty                                       2:15:35 PM
◐ Zipping site files                                                                                        2:15:35 PM
◐ Uploading site files                                                                                      2:15:35 PM

 ERROR  Failed to fetch 404 Not Found                                                                       2:15:36 PM

{
  "errorCode": "NOT_FOUND",
  "errorName": "ThirdPartyApplicationNotFound",
  "errorInstanceId": "xxx",
  "parameters": {
    "thirdPartyApplicationRid": "xxx"
  }
}

💡 Tip: The resource may not exist or your token may not have the required scopes to load it                2:15:36 PM
```

### DIFF
--- a/packages/cli/src/net/createFetch.mts
+++ b/packages/cli/src/net/createFetch.mts
@@ -61,9 +61,12 @@ function handleFetchError(e: unknown): Promise<Response> {
     tip = "Check your token is valid and has not expired or been disabled";
   } else if (e.statusCode === 403) {
     tip = "Check your token has the required scopes for this operation";
+  } else if (e.statusCode === 404) {
+    tip =
+      "The resource may not exist or your token may not have the required scopes to load it";
   }
 
-  let message = e.message;
+  let message;
   if (e.errorName === "CannotDeleteDeployedVersion") {
     message = "The site version to delete is live and cannot be deleted";
     tip = "Run the `site version set` command to change the live site version";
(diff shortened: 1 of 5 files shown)


packages/shared.net.errors/src/PalantirApiError.ts
```diff
@@ -17,23 +17,23 @@
 export class PalantirApiError extends Error implements PalantirApiError {
   public message: string;
   public errorName?: string;
-  public errorType?: string;
+  public errorCode?: string;
```
tzyl, line 20: This is "errorCode" in the actual body:

https://github.com/palantir/osdk-ts/blob/76a85ef8138543a53f79af9a7d29bab54c9a3384/packages/shared.net.fetch/src/createFetchOrThrow.ts#L55-L63

packages/cli/src/net/createFetch.mts
```diff
@@ -73,6 +76,17 @@ function handleFetchError(e: unknown): Promise<Response> {
     message = "The site version already exists";
   } else if (e.errorName === "VersionNotFound") {
     message = "The site version could not be found";
+  } else {
+    const { errorCode, errorName, errorInstanceId, parameters } = e;
+    // Include extra info about the original API error in CLI error messages
+    // https://www.palantir.com/docs/foundry/api/general/overview/errors/
```
tzyl, line 82: I added this to the message in CLI only so it won't affect any other places that are using the message from PalantirApiError e.g. browser console which may be able to get these details elsewhere

ericanderson (approved):

PR #927, opened 2024-10-25
Add scopes to CLI options

woojiahao:
Currently, there's no way for users to configure the `scopes` parameter of their OSDK application. We want to add a way for them to configure this and have the templates either include or hide the `scopes` parameter as a result

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,4 +24,3 @@
    > Full docs on the `changesets` tool can be found at the [changesets/changesets github repo](https://github.com/changesets/changesets).
 7. If you're curious what the final build output might look like you can run `pnpm build` from root.
 8. Run all lint rules and tests with `pnpm check` from root.
-hi
--- a/README.md
+++ b/README.md
@@ -24,3 +24,4 @@
    > Full docs on the `changesets` tool can be found at the [changesets/changesets github repo](https://github.com/changesets/changesets).
 7. If you're curious what the final build output might look like you can run `pnpm build` from root.
 8. Run all lint rules and tests with `pnpm check` from root.
+hi
--- a/examples/example-next-static-export-sdk-2.x/src/lib/client.ts
+++ b/examples/example-next-static-export-sdk-2.x/src/lib/client.ts
@@ -11,8 +11,8 @@ checkEnv(url, "NEXT_PUBLIC_FOUNDRY_API_URL");
 checkEnv(clientId, "NEXT_PUBLIC_FOUNDRY_CLIENT_ID");
 checkEnv(redirectUrl, "NEXT_PUBLIC_FOUNDRY_REDIRECT_URL");
 const scopes = [
-    "api:read-data",
-    "api:write-data",
+    "api:ontologies-read",
+    "api:ontologies-write",
 ];
 
 function checkEnv(
--- a/examples/example-next-static-export-sdk-2.x/src/lib/client.ts
+++ b/examples/example-next-static-export-sdk-2.x/src/lib/client.ts
@@ -10,6 +10,10 @@ const redirectUrl = process.env.NEXT_PUBLIC_FOUNDRY_REDIRECT_URL;
 checkEnv(url, "NEXT_PUBLIC_FOUNDRY_API_URL");
 checkEnv(clientId, "NEXT_PUBLIC_FOUNDRY_CLIENT_ID");
 checkEnv(redirectUrl, "NEXT_PUBLIC_FOUNDRY_REDIRECT_URL");
+const scopes = [
+    "api:read-data",
+    "api:write-data",
+];
 
 function checkEnv(
   value: string | undefined,
@@ -30,6 +34,10 @@ export const getAuth = () => {
       clientId,
       url,
       redirectUrl,
+      true,
+      undefined,
+      window.location.toString(),
+      scopes,
     );
   }
   return auth;
--- a/packages/create-app/src/cli.ts
+++ b/packages/create-app/src/cli.ts
@@ -120,7 +120,8 @@ export async function cli(args: string[] = process.argv): Promise<void> {
           .option("scopes", {
             type: "string",
             array: true,
-            describe: "Includes a list of scopes for the client call",
+            describe:
+              "List of client-side scopes to be used when creating a client",
           }),
     );
 
--- a/packages/create-app/src/cli.ts
+++ b/packages/create-app/src/cli.ts
@@ -27,6 +27,7 @@ import { promptOsdkPackage } from "./prompts/promptOsdkPackage.js";
 import { promptOsdkRegistryUrl } from "./prompts/promptOsdkRegistryUrl.js";
 import { promptOverwrite } from "./prompts/promptOverwrite.js";
 import { promptProject } from "./prompts/promptProject.js";
+import { promptScopes } from "./prompts/promptScopes.js";
 import { promptSdkVersion } from "./prompts/promptSdkVersion.js";
 import { promptTemplate } from "./prompts/promptTemplate.js";
 import { run } from "./run.js";
@@ -46,6 +47,7 @@ interface CliArgs {
   osdkPackage?: string;
   osdkRegistryUrl?: string;
   corsProxy?: boolean;
+  scopes?: string[];
 }
 
 export async function cli(args: string[] = process.argv): Promise<void> {
@@ -114,6 +116,11 @@ export async function cli(args: string[] = process.argv): Promise<void> {
             type: "boolean",
             describe:
               "Include a CORS proxy for Foundry API requests during local development",
+          })
+          .option("scopes", {
+            type: "string",
+            array: true,
+            describe: "Includes a list of scopes for the client call",
           }),
     );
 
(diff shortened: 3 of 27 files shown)


packages/create-app/src/cli.ts
```diff
@@ -114,6 +116,11 @@ export async function cli(args: string[] = process.argv): Promise<void> {
             type: "boolean",
             describe:
               "Include a CORS proxy for Foundry API requests during local development",
+          })
+          .option("scopes", {
+            type: "string",
+            array: true,
+            describe: "Includes a list of scopes for the client call",
```
woojiahao, line 123: TODO make better description

README.md
```diff
@@ -23,4 +23,4 @@
    >
    > Full docs on the `changesets` tool can be found at the [changesets/changesets github repo](https://github.com/changesets/changesets).
 7. If you're curious what the final build output might look like you can run `pnpm build` from root.
-8. Run all lint rules and tests with `pnpm check` from root.
+8. Run all lint rules and tests with `pnpm check` from root.
```
woojiahao, line 26: I'm not actually sure what this change is

examples/example-next-static-export-sdk-2.x/src/lib/client.ts
```diff
@@ -30,6 +34,10 @@ export const getAuth = () => {
       clientId,
       url,
       redirectUrl,
+      true,
+      undefined,
+      window.location.toString(),
```
woojiahao, line 39: We have to include these because of the order of parameters of the `createPublicOAuthClient` function

packages/create-app/src/cli.ts
```diff
@@ -132,6 +140,7 @@ export async function cli(args: string[] = process.argv): Promise<void> {
   const osdkPackage: string = await promptOsdkPackage(parsed);
   const osdkRegistryUrl: string = await promptOsdkRegistryUrl(parsed);
   const corsProxy: boolean = await promptCorsProxy(parsed);
+  const scopes: string[] | undefined = await promptScopes(parsed);
```
woojiahao, line 143: We want to allow the `scopes` option to be entirely optional to support backward compatability

packages/create-app/src/prompts/promptScopes.ts
```diff
@@ -0,0 +1,39 @@
...
+    return undefined;
+  }
+
+  for (const scope of scopes) {
+    if (!scopeNameRegex.test(scope)) {
+      consola.fail(
+        `Scope name ${scope} can only contain letters, hyphens, underscores, and colons`,
+      );
+      // Following promptOverwrite
+      process.exit(0);
```
woojiahao, line 34: Not entirely sure if we want to immediately terminate

examples/example-next-static-export-sdk-2.x/src/lib/client.ts
zeyadkhaled: I think this is exposing defaults that the user doesn't have to worry about. Should we look into perhaps introducing a new overload of createPublicOauthClient that takes an options dict with defaults? e.g:
function createPublicOauthClient({
  clientId,
  url,
  redirectUrl,
  x = true,
  y = 600,
  z = window.location.toString(),
  scopes = [],
}: OAuthClientOptions)

examples/example-next-static-export-sdk-2.x/src/lib/client.ts
```diff
@@ -10,6 +10,10 @@ const redirectUrl = process.env.NEXT_PUBLIC_FOUNDRY_REDIRECT_URL;
 checkEnv(url, "NEXT_PUBLIC_FOUNDRY_API_URL");
 checkEnv(clientId, "NEXT_PUBLIC_FOUNDRY_CLIENT_ID");
 checkEnv(redirectUrl, "NEXT_PUBLIC_FOUNDRY_REDIRECT_URL");
+const scopes = [
+    "api:read-data",
+    "api:write-data",
+];
```
zeyadkhaled, line 16: Should we wait until the ontologies namespace migration before exposing this?

packages/create-app/src/cli.ts
zeyadkhaled: Instead of it being optional should we instead have a default value e.g `["api:read-data", "api:write-data"]`

examples/example-next-static-export-sdk-2.x/src/lib/client.ts
woojiahao: This was what the older 1.x SDKs used. I would want to clarify why this was changed in 2.0

packages/create-app/src/cli.ts
woojiahao: I think this goes back to the discussion of whether we should be exposing end users to the concept of `scopes` if they aren't "interested". By allowing this to be set as optional, we can avoid generating the templates for users who do not require the `scopes` to avoid confusion. I'm okay with providing default as well but I don't think it strongly aligns with the other places where we do apply these `scopes` display like the instructions on the documentation

examples/example-next-static-export-sdk-2.x/src/lib/client.ts
woojiahao: Well I've already tested the scenarios where the server side is `api:ontologies-X` and the client is `api:X-data` and it works since the endpoints check the post expansion operations. I think it is possible that we just roll the migrations out regardless of client-side changes. I will follow up with the teams on Monday to check on the status of these but I would prefer if information about the scopes is more explicit going forward given that hack week and DevCon is near.

woojiahao: I would also like to point out that migrating these examples from `api:X-data` -> `api:ontologies-X` is part of the migration RFC (very final step) as well. So either way works for me

woojiahao: We've did a quick discussion and we can actually push ahead with these changes here first but changing the example values to be `api:ontologies-read` and `api:ontologies-write` directly. This will not cause any regressions given that the applications would have either fully migrated to these new scopes and still preserve the `api:read-data` and `api:write-data` scopes which work

packages/create-app/src/prompts/promptScopes.ts
zeyadkhaled: Re-prompt for the whole list and show which ones are wrong.

woojiahao: Uploading scopes-parameter.mp4…

This is what it looks like now, had to re-jig the logic a little

examples/example-next-static-export-sdk-2.x/src/lib/client.ts
woojiahao: Decided to keep it how it is given that the current SDK only support this method of client creation

packages/create-app/src/prompts/promptScopes.ts
woojiahao: Also realized that we should de-dupe the scopes. I wrote a custom function to de-dupe it while preserving the order of the elements. There's a shorter way to do it with `reduce` but I felt it's less clear

packages/create-app/src/prompts/promptScopes.ts
```diff
@@ -0,0 +1,64 @@
...
+      default: "api:ontologies-read api:ontologies-write",
+    });
+    scopes = stringScopes.split(" ");
+  }
+
+  return deduplicateScopes(scopes);
+}
+
+// Preserving the originally specified scope order
+function deduplicateScopes(scopes: string[]): string[] {
```
woojiahao, line 53: I wrote this instead of just doing `Array.from(new Set(scopes))` because I wanted to preserve the order of the scopes while de-duping, but okay with just keeping it simple too

packages/create-app/src/prompts/promptScopes.ts
```diff
@@ -0,0 +1,64 @@
...
+    return undefined;
+  }
+
+  while (true) {
+    const allValidScopes = scopes.every(scope => scopeNameRegex.test(scope));
+    if (allValidScopes) {
+      break;
+    }
+
+    const firstFailingScopeIndex = scopes.findIndex(scope =>
```
woojiahao, line 34: Opted with this approach instead of doing a loop since looping would mean keeping a mutable boolean to track if the scopes is valid.

packages/create-app/src/prompts/promptScopes.ts
```diff
@@ -0,0 +1,64 @@
...
+    consola.fail(
+      `Scope name ${failingScope} can only contain letters, hyphens, underscores, and colons`,
+    );
+    const stringScopes = await consola.prompt("Scopes:", {
+      type: "text",
+      placeholder: "api:ontologies-read api:ontologies-write",
+      default: "api:ontologies-read api:ontologies-write",
+    });
+    scopes = stringScopes.split(" ");
+  }
```
zeyadkhaled, line 47: Can we check on all the scopes and just return in the same message to the user all the ones that are incorrect?

zeyadkhaled (approved):
Just a small comment but looks good thank you!

packages/create-app/src/prompts/promptScopes.ts
woojiahao: Fair point, I'll update this

woojiahao: image

This is what it looks like now. we just display all at once

ssanjay1 (approved):
Rubber stamping

PR #928, opened 2024-10-25
Allow configuring uploadOnly through foundry.config.json

tzyl:
The site deploy command supports `--uploadOnly` but this cannot be specified in `foundry.config.json` so if we want to remember this choice for example in Foundry CI templates we have to manage this outside of the config file and set it on the CLI command rather than managing this default project config in a single place.


ericanderson (dismissed):

zeyadkhaled (dismissed):
🚀

zeyadkhaled (approved):

PR #940, opened 2024-10-29
trying new interface APIs

ssanjay1:


examples-extra/docs_example/src/osdkExample.tsx
```diff
@@ -135,6 +135,9 @@ export async function osdkObjectSetExample() {
     $select: ["employeeId", "fullName", "startDate"],
     $orderBy: { "startDate": "asc" },
     $pageSize: 10,
+    // @ts-expect-error
+    $augment: { "Employee": [] },
```
ssanjay1, line 139: If we have the flag on, disallowing augments for now

packages/client/src/object/fetchPage.ts
```diff
@@ -93,28 +93,34 @@ async function fetchInterfacePage<
   args: FetchPageArgs<Q, L, R, any, S>,
   objectSet: ObjectSet,
 ): Promise<FetchPageResult<Q, L, R, S>> {
-  const result = await OntologiesV2.OntologyInterfaces
-    .search(
-      addUserAgentAndRequestContextHeaders(client, interfaceType),
-      await client.ontologyRid,
+  if (args.$__EXPERIMENTAL_useUnstableInterfaces) {
+    // Use new unstable APIs
```
ssanjay1, line 97: Basically just branching to the right API based on what the flag is set to. This way anyone using existing interface stuff wouldn't be affected and since its an optional param in the fetch page it won't require any code changes either

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -69,9 +70,11 @@ export interface AsyncIterArgs<
   R extends boolean = false,
   A extends Augments = {},
   S extends NullabilityAdherence = NullabilityAdherence.Default,
+  U extends boolean = false,
 > extends SelectArg<Q, K, R, S>, OrderByArg<Q, PropertyKeys<Q>> {
-  $augment?: A;
+  $augment?: U extends true ? never : A;
   $__EXPERIMENTAL_selectedObjectTypes?: string[];
+  $__EXPERIMENTAL_useUnstableInterfaces?: U;
```
ssanjay1, line 77: So I liked doing this better than exposing via experimental because I didn't like having to do `client(interfaceExperiemntSymbol)(myInterface).fetchPage...`, especially because then people will need to transition back to the normal syntax afterwards

ssanjay1: Only downside is that if you hit `$` this is easily exposed as an option, but idk we make it very clear its experimental and unstable

PR #942, opened 2024-10-30
Removing augments

ssanjay1:
We are removing the augment capability when fetching interfaces for now, until we have proper API support for this.

### DIFF
--- a/packages/api/src/definitions/LinkDefinitions.ts
+++ b/packages/api/src/definitions/LinkDefinitions.ts
@@ -14,12 +14,7 @@
  * limitations under the License.
  */
 
-import type {
-  FetchPageArgs,
-  SelectArg,
-  SelectArgToKeys,
-} from "../object/FetchPageArgs.js";
-import type { SingleOsdkResult } from "../object/FetchPageResult.js";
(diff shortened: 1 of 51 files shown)


packages/api/src/definitions/LinkDefinitions.ts
```diff
@@ -14,12 +14,7 @@
  * limitations under the License.
  */
 
-import type {
-  FetchPageArgs,
-  SelectArg,
-  SelectArgToKeys,
-} from "../object/FetchPageArgs.js";
-import type { SingleOsdkResult } from "../object/FetchPageResult.js";
```
ssanjay1, line 22: Some of these were also just unused imports

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -80,13 +77,3 @@ export type Augment<
 > = { [K in CompileTimeMetadata<X>["apiName"]]: T[] };
 
 export type Augments = Record<string, string[]>;
-
-export interface FetchInterfacePageArgs<
```
ssanjay1, line 84: This was unused

ericanderson (approved):

PR #945, opened 2024-10-31
Implementing new interface APIs

ssanjay1:
Implemented interface syntax with the new backend APIs.

__Note_: This is currently marked as do not merge so we can wait until the backend pieces are fully merged and ready._


packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -173,7 +179,37 @@ describe("convertWireToOsdkObjects", () => {
     expect(prototypeBefore).not.toBe(prototypeAfter);
   });
 
-  it("updates interface when underlying changes", async () => {
+  it("works even with unknown apiNames - new", async () => {
```
ssanjay1, line 182: Basically, all the tests are the same, I just added duplicates to test the new functionality, while retaining the old stuff since it's still technically supported. Once we rip out the old apis, we can clean up these tests too

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,6 +116,80 @@ export async function convertWireToOsdkObjects(
...
+    );
+    invariant(objectDef, `Missing definition for '${rawObj.$apiName}'`);
+
+    const interfaceToObjMapping = interfaceApiName
+      ? interfaceToObjectTypeMappings[interfaceApiName as InterfaceTypeApiName][
+        rawObj.$apiName
+      ]
+      : undefined;
+
+    const ifaceSelected = interfaceApiName && interfaceToObjMapping
```
ssanjay1, line 148: The map comes to us as `interfaceKey` : `objectPropName` so the keys are interface, values are the object props

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,6 +116,80 @@ export async function convertWireToOsdkObjects(
...
+      ? (selectedProps
+        ? Object.keys(interfaceToObjMapping).filter(
+          val => {
+            selectedProps?.includes(interfaceToObjMapping[val]);
+          },
+        )
+        : [
+          ...Object.values(interfaceToObjMapping),
+          objectDef.primaryKeyApiName,
```
ssanjay1, line 157: All the results from this API come in as the object props, but scoped down to whats on the interface. As a result, if the primary key isn't part of the interface, we don't get the api name of the primary key from the mapping in the result. We also can't just use the keys of the `objectDef` because that would include props not implemented in the interface

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,6 +116,80 @@ export async function convertWireToOsdkObjects(
...
+      : undefined;
+
+    // default value for when we are checking an object
+    let objProps;
+
+    let conforming = true;
+    if (interfaceApiName && ifaceSelected) {
+      invariantInterfacesAsViews(objectDef, interfaceApiName, client);
+
+      objProps = ifaceSelected;
```
ssanjay1, line 168: Again, we can just directly set this without converting because the props come over the wire as the object props, not the interface ones.

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,6 +116,80 @@ export async function convertWireToOsdkObjects(
...
+    } else {
+      objProps = selectedProps ?? Object.keys(objectDef.properties);
```
ssanjay1, line 170: If its not an interface, then we can just directly use the objectDef again

packages/client/src/object/fetchPage.ts
```diff
@@ -58,8 +63,8 @@ export function objectSetToSearchJsonV2(
   expectedApiName: string,
   existingWhere: SearchJsonQueryV2 | undefined = undefined,
 ): SearchJsonQueryV2 | undefined {
-  if (objectSet.type === "base") {
-    if (objectSet.objectType !== expectedApiName) {
+  if (objectSet.type === "base" || objectSet.type === "interfaceBase") {
+    if (objectSet.type === "base" && objectSet.objectType !== expectedApiName) {
```
ssanjay1, line 67: We now have two types of object sets that the old API could theoretically support, so have to check for either in case our users turn the feature flag on.

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -70,6 +70,7 @@ export interface AsyncIterArgs<
   A extends Augments = never,
   S extends NullabilityAdherence = NullabilityAdherence.Default,
 > extends SelectArg<Q, K, R, S>, OrderByArg<Q, PropertyKeys<Q>> {
+  $__UNSTABLE_useOldInterfaceApis?: boolean;
```
ssanjay1, line 73: Users can add this flag if they want to hit the old APIs for some reason

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -112,6 +116,80 @@ export async function convertWireToOsdkObjects(
   return ret;
 }
 
+export async function convertWireToOsdkObjects2(
```
ssanjay1, line 119: Added a new method for now, once we remove the old api functionality, this can be cleaned up to just be this one method.

ssanjay1: The new APIs return all objects with their object prop names, but scoped down to the interface. We also get a map that tells us how to convert from the object props to their respective SPTs

packages/client/src/object/fetchPage.ts
```diff
@@ -93,28 +106,54 @@ async function fetchInterfacePage<
   args: FetchPageArgs<Q, L, R, any, S>,
   objectSet: ObjectSet,
 ): Promise<FetchPageResult<Q, L, R, S>> {
-  const result = await OntologiesV2.OntologyInterfaces
-    .search(
-      addUserAgentAndRequestContextHeaders(client, interfaceType),
-      await client.ontologyRid,
+  if (args.$__UNSTABLE_useOldInterfaceApis) {
```
ssanjay1, line 109: Functionality in this first branch same as before, just flagged off

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -220,15 +298,15 @@ function fixObjectPropertiesInPlace(
     }
 
     // Backend returns as __apiName but we want to stick to $ structure
-    obj.$apiName = obj.__apiName;
+    obj.$apiName ??= obj.__apiName;
```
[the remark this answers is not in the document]
nihalbhatnagar: What a cool operator

ericanderson (commented):
No flaws on skim. let a note

packages/client/src/util/objectSetUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import type { ObjectOrInterfaceDefinition } from "@osdk/api";
+import type { ObjectSet as WireObjectSet } from "@osdk/internal.foundry.core";
+
+export function resolveBaseObjectSetType(
+  objectType: ObjectOrInterfaceDefinition,
+): WireObjectSet {
+  return objectType.type === "interface"
+    ? {
+      type: "interfaceBase",
```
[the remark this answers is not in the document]
ssanjay1: Mmm, the second key differs as well though `interfaceType` vs `objectType`, like I could conditionally set that with a spread operator as well I guess but then the return types get confused

ericanderson (approved):

nihalbhatnagar (approved):

PR #947, opened 2024-10-31
Supports custom formatters in browsers

ericanderson:
Before:
image
image

After:
image
image

To enable: 
image

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
@@ -73,6 +73,65 @@ const objectPrototypeCache = createClientCache(
   },
 );
 
+if (process.env.NODE_ENV !== "production") {
+  const installed = Symbol();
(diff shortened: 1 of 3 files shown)


packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -73,6 +73,65 @@ const objectPrototypeCache = createClientCache(
...
+          // ...Object.keys(raw).map(key => {
+          //   return [
+          //     "li",
+          //     {},
+          //     [
+          //       "span",
+          //       { style: "color: #888888" },
+          //       `${key}: `,
+          //     ],
+          //     ["object", { object: raw[key] }], // ["span", {}, `${key}: `]],
```
nihalbhatnagar, line 126: Did you mean to keep these?

ericanderson: I was on the fence about which style to use so i left it so it would be easy for us to change our mind/experiment

PR #948, opened 2024-10-31
Edit foundry fetch return type to Response instead of Blob

nihalbhatnagar:
New Platform SDKs will ship with this version of shared.net.platformapi

### DIFF
--- a/.changeset/wild-flies-yell.md
+++ b/.changeset/wild-flies-yell.md
@@ -1,0 +1,5 @@
+---
+"@osdk/shared.net.platformapi": minor
+---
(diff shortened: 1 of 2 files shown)


.changeset/wild-flies-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/shared.net.platformapi": minor
```
nihalbhatnagar, line 2: Minor change seems right? Will test this on beta first

ssanjay1 (approved):

PR #949, opened 2024-11-01
More interface args cleanup

ssanjay1:
The new apis do not directly support selecting by object type. While I can simulate this with objectset -> interfaceset intersections, I'm not sure this is the best way to expose this functionality, rather than just letting users directly do an intersect themselves. As I experiment with this more, I want to start with a clean slate first.

There is no internal usage of this, so should be safe to remove.

### DIFF
--- a/packages/api/src/object/FetchPageArgs.ts
+++ b/packages/api/src/object/FetchPageArgs.ts
@@ -70,7 +70,6 @@ export interface AsyncIterArgs<
   A extends Augments = never,
   S extends NullabilityAdherence = NullabilityAdherence.Default,
 > extends SelectArg<Q, K, R, S>, OrderByArg<Q, PropertyKeys<Q>> {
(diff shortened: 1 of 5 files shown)


packages/api/src/object/FetchPageArgs.ts
```diff
@@ -70,7 +70,6 @@ export interface AsyncIterArgs<
   A extends Augments = never,
   S extends NullabilityAdherence = NullabilityAdherence.Default,
 > extends SelectArg<Q, K, R, S>, OrderByArg<Q, PropertyKeys<Q>> {
-  $__EXPERIMENTAL_selectedObjectTypes?: string[];
```
ssanjay1, line 73: Keeping around this type for now in case I want to start adding new things/old things back next week

ericanderson (approved):
Agree with the direction. Things that were inexpensive and low risk to keep pre-GA are no longer such.

PR #958, opened 2024-11-05
Makes properties optional on Subscriptions

nihalbhatnagar:


packages/api/src/objectSet/ObjectSetListener.ts
```diff
@@ -55,3 +55,20 @@ type ObjectUpdate<
...
+/**
+ * Options for subscribing to an ObjectSet.
+ *
+ * @property properties - The properties to request a subscription for. Requesting specific properties limits the possible properties
+ * that can be returned from the subscription. If not provided, all properties will be requested and potentially be returned on updates.
+ *
+ * @property requestLatestGeotimeSeriesValues - If true, any geotime series properties will be updated with the latest values. This can
+ * reduce performance and should be used only if needed.
+ */
+export interface ObjectSetListenerOptions<
```
nihalbhatnagar, line 68: I want to export this from the API similar to other Options types we have, is there anything else to keep in mind?

packages/client/src/createClient.ts
```diff
@@ -186,16 +186,20 @@ export function createClientInternal(
               const P extends PropertyKeys<Q>,
             >(
               objectSet: ObjectSet<Q>,
-              properties: Array<P>,
-              listener: EXPERIMENTAL_ObjectSetListener<Q, P>,
+              listener: ObjectSetListener<Q, P>,
+              opts?: {
+                properties?: Array<P>;
+                requestLatestGeotimeSeriesValues?: boolean;
```
nihalbhatnagar, line 192: I made this ultra specific for right now. I think this is the best option for right now since this is the only property that we can support and it makes it clear what they're going to get back for right now. 

My main concern is that this will cause a break when more of property types are supported here. There are two things I have in mind for that 1. We just rename this to the greatest scope we have or 2. We add another opt specifically for that property type. I'm in favor of 1 I'm just not sure what the next addition will be

packages/e2e.sandbox.catchall/src/runSubscriptionsTest.ts
```diff
@@ -26,15 +26,11 @@ export async function runSubscriptionsTest() {
...
     .subscribe(
-      [
-        "primaryKey_",
-        "stringProperty",
-      ],
       {
         onChange(object) {
           console.log(
             "Object with primaryKey ",
-            object.object.primaryKey_,
```
nihalbhatnagar, line 37: looks weird but this is what the property was called

packages/api/src/objectSet/ObjectSetListener.ts
ericanderson: Seems reasonable to export from the main index.ts

packages/client/src/createClient.ts
ericanderson: From reading the name it is not clear to me what this does. I had to read the code to understand it. 

I think we need to enumerate out what people want to do. 

For `$(Geolocatable).subscribe(myListener)`, should the references be retrieved automatically? What is my likely intent?

What about `$(Geolocatable).subscribe(myListener, { properties: ["location"] })`? Is the answer different this time if `location` is or isnt a geotime?



I'm intentionally not answering these questions myself because I want you to work through the options.

packages/api/src/objectSet/ObjectSetListener.ts
```diff
@@ -55,3 +55,19 @@ type ObjectUpdate<
   object: Osdk.Instance<O, never, P>;
   state: "ADDED_OR_UPDATED" | "REMOVED";
 };
+
+/**
+ * Options for subscribing to an ObjectSet.
+ *
+ * @property properties - The properties to request a subscription for. Requesting specific properties limits the possible properties
+ * that can be returned from the subscription. If not provided, all properties will be requested and potentially be returned on updates.
```
ssanjay1, line 63: Lets update this comment to make it clear that empty array will _also_ give you reference updates (like geotime)

packages/api/src/objectSet/ObjectSetListener.ts
```diff
@@ -55,3 +55,19 @@ type ObjectUpdate<
...
+ *
+ * @property requestLatestGeotimeSeriesValues - If true, any geotime series properties will be updated with the latest values. This can
```
ssanjay1, line 65: Actually, we need to just delete this. That will make things more clear

nihalbhatnagar: Agreed, this was left over.

PR #959, opened 2024-11-06
Try catch calling listeners for subscriptions

nihalbhatnagar:


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -454,7 +454,11 @@ export class ObjectSetListenerWebsocket {
 
     for (const osdkObject of osdkObjectsWithReferenceUpdates) {
       if (osdkObject != null) {
-        sub.listener.onChange?.(osdkObject);
+        try {
```
nihalbhatnagar, line 457: I don't think abstracting this to a method is worth it

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -454,7 +454,11 @@ export class ObjectSetListenerWebsocket {
...
+          sub.listener.onChange(osdkObject);
+        } catch (error) {
+          sub.listener.onError([error]);
```
nihalbhatnagar, line 460: I'm going to FLUP this and add an enum of what caused the error. This would be (excuse the weird names) `SUBSCRIPTION_ERROR` or `HANDLER_ERROR`. This would make more sense if we could add a third that would suggest trying again, but this isn't something we can ensure

ssanjay1 (approved):

PR #960, opened 2024-11-06
Export api through client

ssanjay1:
Exporting all @osdk/api entities through @osdk/client

### DIFF
--- a/etc/client.report.api.md
+++ b/etc/client.report.api.md
@@ -4,10 +4,10 @@
 
 ```ts
 
-import type { ActionDefinition } from '@osdk/api';
+import { ActionDefinition } from '@osdk/api';
 import { ActionEditResponse } from '@osdk/api';
-import type { ActionMetadata } from '@osdk/api';
-import type { ActionParam } from '@osdk/api';
+import { ActionMetadata } from '@osdk/api';
+import { ActionParam } from '@osdk/api';
 import { ActionReturnTypeForOptions } from '@osdk/api';
 import { ActionValidationResponse } from '@osdk/api';
 import { ApplyActionOptions } from '@osdk/api';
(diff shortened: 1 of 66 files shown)


packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
```diff
@@ -124,7 +118,7 @@ export async function generatePackage(
       packageName: options.packageName,
       packagePath,
       packageVersion: options.packageVersion,
-      dependencies: resolvedDependencies,
+      dependencies: [],
```
ssanjay1, line 121: Removed api as a direct dep

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -116,8 +120,9 @@ async function generateV2QueryFile(
   await fs.writeFile(
     path.join(outDir, `${query.shortApiName}.ts`),
     await formatTs(`
-        import type { QueryDefinition , VersionBound} from "@osdk/api";
-        import type { QueryParam, QueryResult } from "@osdk/api";
+        import type { QueryDefinition , QueryParam, QueryResult, VersionBound} from "${
```
ssanjay1, line 123: I'm not sure why these were split up before, just combined. If this was done for a reason lmk and I'll revert

packages/client.test.ontology/generateMockOntology.js
```diff
@@ -57,4 +57,8 @@ await generateClientSdkVersionTwoPointZero(
   },
   outDir,
   "module",
+  undefined,
+  undefined,
+  undefined,
+  true,
```
ssanjay1, line 63: Use internal flag so we avoid circular deps on this one

etc/client.report.api.md
```diff
@@ -4,10 +4,10 @@
 
 ```ts
 
-import type { ActionDefinition } from '@osdk/api';
+import { ActionDefinition } from '@osdk/api';
 import { ActionEditResponse } from '@osdk/api';
-import type { ActionMetadata } from '@osdk/api';
-import type { ActionParam } from '@osdk/api';
+import { ActionMetadata } from '@osdk/api';
+import { ActionParam } from '@osdk/api';
```
ericanderson, line 10: Why don't these say `type` any more?

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
ericanderson: Oversight probably

ericanderson (approved):
Awesome thank you

etc/client.report.api.md
ssanjay1: Mmm, not sure, I'll check

PR #961, opened 2024-11-06
Overload createPublicOauthClient to support object of options

zaini:
* Overload `createPublicOauthClient` function to support passing in an object of options, rather than just taking an bunch of optional params.
* Update JSDocs
* Wasn't sure where to write tests for this
* Open to suggestions on variable renames

### DIFF
--- a/packages/oauth/src/createPublicOauthClient.ts
+++ b/packages/oauth/src/createPublicOauthClient.ts
@@ -103,7 +103,7 @@ export function createPublicOauthClient(
     postLoginPage: postLoginPageUrl = window.location.toString(),
     scopes: scopeList = [],
     fetchFn: fetchFunction = globalThis.fetch,
-    ctxPath: context = "/multipass",
+    ctxPath: contextPath = "/multipass",
   } = processOptions(
     clientIdOrOptions,
     url,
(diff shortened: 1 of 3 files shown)


packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -43,35 +43,85 @@ declare const process: {
...
- * @param scopes
- * @param fetchFn
- * @param ctxPath
- * @returns {PublicOauthClient} which can be used as a token provider
+ * Creates a PublicOauthClient for authentication.
+ *
+ * @param {string | PublicOauthClientOptions} clientIdOrOptions - Either the client_id as a string or an options object
+ * @param {string} [url] - The base URL of your Foundry server
+ * @param {string} [redirectUrl] - The URL configured for redirect in the OAuth configuration on the server
+ * @param {boolean} [useHistory=true] - If true, uses `history.replaceState()`, otherwise uses `window.location.assign()`
```
zaini, line 64: Unfortunately JSDocs in VSCode doesn't seem to support this syntax for default values `[useHistory=true]`

ericanderson:
@zaini I didnt get to test out my changes but I think they are likely good. This is something we need to be sure of so we arent breaking contracts.

ssanjay1 (approved):

PR #965, opened 2024-11-07
Update $in to accept readonly

### LINKED ISSUE #964: Ability to pass `readonly` arrays to `$in` operator
For now, I need to spread my array because it is readonly:

image

nihalbhatnagar:
Fixes #964

### DIFF
--- a/packages/api/src/aggregate/WhereClause.ts
+++ b/packages/api/src/aggregate/WhereClause.ts
@@ -50,7 +50,7 @@ type MakeFilter<K extends PossibleWhereClauseFilters, V> = K extends string ? {
 type BaseFilter<T> =
   | T
   | MakeFilter<"$eq" | "$ne", T>
-  | MakeFilter<"$in", Array<T>>
+  | MakeFilter<"$in", ReadonlyArray<T>>
   | MakeFilter<"$isNull", boolean>;
 
 type StringFilter =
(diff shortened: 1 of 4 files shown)


ssanjay1 (dismissed):

packages/api/src/aggregate/WhereClause.ts
```diff
@@ -50,7 +50,7 @@ type MakeFilter<K extends PossibleWhereClauseFilters, V> = K extends string ? {
 type BaseFilter<T> =
   | T
   | MakeFilter<"$eq" | "$ne", T>
-  | MakeFilter<"$in", Array<T>>
+  | MakeFilter<"$in", ReadonlyArray<T>>
```
ericanderson, line 53: Can we have a tiny test for this? Its the type of thing that would be easy to regress in a year or two with refactorings.

ericanderson (dismissed):
Approved but we should write a test

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -296,6 +296,14 @@ describe("ObjectSet", () => {
     }
   });
 
+  it("allows $in filter with ReadonlyArrays", () => {
```
nihalbhatnagar, line 299: Did not add a changeset for client package

ericanderson (approved):

PR #973, opened 2024-11-12
Fix where clause empty props

ssanjay1:
When we generated object/interface types that had no properties, our where clause type broke down and allowed any arbitrary key. This is now rectified.


nihalbhatnagar (approved):
Lgtm!

PR #975, opened 2024-11-13
Add helpers for extracting date strings

nihalbhatnagar:


packages/client/src/util/datetimeConverters.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import {
+  extractDate,
+  extractDateInLocalTime,
+  extractDateInUTC,
+  generateOffsetUtcString,
+} from "./datetimeConverters.js";
+
+describe("test datetime converters", () => {
+  beforeAll(() => {
+    register("Etc/GMT+5");
```
nihalbhatnagar, line 36: Sets to new york time, not sure why it is +5 and not -5

packages/client/src/util/datetimeConverters.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @returns The date part of the input string
+ */
+export const extractDate = (dateTime: string) => {
+  if (!isoRegex.test(dateTime)) {
+    throw new Error(
+      `Invalid date format. Expected ISO 8601 format, but received: ${dateTime}`,
+    );
+  }
+
+  return dateTime.split(
```
ssanjay1, line 32: minor nit: Maybe could have another helper function for the string splitting logic since you do that in all 3 functions

packages/client/src/util/datetimeConverters.ts
```diff
@@ -0,0 +1,65 @@
...
+const isoRegex =
+  /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})?$/;
+
+/**
+ * Extracts the date from a ISO 8601 formatted date time string. Throws if the input is not in the correct format.
+ * @param dateTime An ISO 8601 formatted date time string
+ * @returns The date part of the input string
+ */
+export const extractDate = (dateTime: string) => {
+  if (!isoRegex.test(dateTime)) {
```
ssanjay1, line 26: Maybe use `invariant` here?

packages/client/src/util/datetimeConverters.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const isoRegex =
+  /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(?:\.\d+)?(?:Z|[+-]\d{2}:\d{2})?$/;
```
ericanderson, line 20: Can we dig into the whole regex security vuln party that the internet has been having for a few years? Does this trigger that? Previously we have ignored such things because they were in our internal code gen and we controlled the inputs but in this case we are providing a library for others to use and cannot guarantee anything

nihalbhatnagar: BLUF from offline, looked into it, the exploit pertains to regex that has a group with a repetition, and that group itself is repeated. We do have repetition in a group `?:\.\d+`, but that group is not repeated. I added a length limiter to prevent arbitrarily long strings to get passed in and hang the regex checker

ericanderson (approved):

ericanderson (approved):

PR #977, opened 2024-11-14
v0 for runtime derived properties

nihalbhatnagar:


packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -245,4 +247,24 @@ export interface ObjectSet<
...
+    >
+  >;
 }
+
+type DerivedPropertyExtendedObjectDefinition<
+  K extends ObjectOrInterfaceDefinition,
+  D extends DeriveClause<K>,
+> = {
+  __DefinitionMetadata: {
+    properties: { [T in Extract<keyof D, string>]: PropertyDef<"integer"> };
```
nihalbhatnagar, line 268: This is the fun part to figure out

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -245,4 +247,24 @@ export interface ObjectSet<
     properties: Array<P>,
     listener: EXPERIMENTAL_ObjectSetListener<Q, P>,
   ) => { unsubscribe: () => void };
+
+  readonly derive: <
```
nihalbhatnagar, line 251: I'm not sure of the best way to gate this. I want this to be on object set since I want to ensure you can do stuff to an object set before it. Is it crazy to leave this ungated on beta(only merged when e2e works ofc)

PR #978, opened 2024-11-15
Add initial views API

styu:
This adds 4 new packages:
- `@osdk/views-api.unstable`: Primarily types that define the contract between the Foundry UI host application (1st party) and the child view (3rd party). This is not meant to be directly consumed by any views code, just used internally in Foundry itself.
- `@osdk/views-client.unstable`: This contains the low level client that sets up the listeners and type-safe message passing between the host and child. This will re-export the types from `@osdk/views-api.unstable` so that consumers only have to worry about relying on `@osdk/views-client*` packages.
- `@osdk/views-react.unstable`: This is a wrapper around `@osdk/views-client.unstable` that sets up the listener + React context with parameters in a component
- `@osdk/e2e.sandbox.views.todo-widget`: This is just a test app that let me make sure that all the types worked as intended

I'm calling things `.unstable` for now while we finalize the API. Also note that most of this diff is in the lockfile (which I ran `pnpm dedupe` on to avoid duplicate vite/etc versions)

### Parameters

To start, we support the following parameter types:
- String
- Number
- Boolean
- Date (represented as ISO date, `YYYY-MM-DD`)
- Timestamp (represented as in ISO format)
- Array of either string, number, boolean, date, or timestamp (no nested arrays)

Parameters are passed from host UI to child view asynchronously, and the types reflect that (see `AsyncValue` in `@osdk/views-api.unstable`).

### Message passing

Messages are passed between host and child via `postMessage`. The client code that we provide will automatically detect the host origin by looking for a specific `meta` tag that the Foundry view sandbox iframe will inject into the HTML for developers. The messages allowed are defined in `@osdk/views-api.unstable`

### Configuration generation
The intent is that to create a view, the dev just needs a parameters file and a vite plugin (that I have yet to write):

```
src/
  entrypoint.tsx             // Entrypoint file
  entrypoint.parameters.ts   // Sibling *.parameters.ts file that default exports a parameter configuration object
```

In `entrypoint.parameters.ts` file, they will need to define it like so:

```ts
import type { ParameterConfig } from "@osdk/views-client.unstable";

const Parameters = {
  parameters: {
    showWarning: {
      displayName: "Show warning callout",
      type: "boolean",
    },
    todoItems: {
      displayName: "Todo items",
      type: "array",
      subType: "string",
    },
  },
} as const satisfies ParameterConfig;
export default Parameters;
```

In theory, the vite plugin will find the default export from the file that is a sibling to the specified entrypoint file and produce a `view-config.json` file in the output directory that will eventually be uploaded to Foundry to save when the view is published.

### Usage

The point of having the parameters defined like this is so that we can use some cute type magic to automatically make sure that the messages and parameter values passed back and forth are strongly typed, yet the developer only has to define the parameters once, in the `*.parameters.ts` file. This means that the code can be as simple as this:

```ts
// entrypoint.tsx
import React from "react";
import { createRoot } from "react-dom/client";
import { App } from "./app.js";
import MainParameters from "./entrypoint.parameters.js";

const root = document.querySelector("body")!;

createRoot(root).render(
  (
    <FoundryView
        config={MainParameters}
      >
        <App />
     </FoundryView>
  ),
);

// app.tsx
import React from "react";
import type MainParameters from "./entrypoint.parameters.js";

export const App: React.FC = () => {
  const { parameterValues, hostEventTarget } = useFoundryViewContext<typeof MainParameters>();
  // showWarning is type { type: "boolean"; value: AsyncValue<boolean> }
  // todoItems is of type { type: "array"; subType: "string"; value: AsyncValue<string[]> }
  const { todoItems, showWarning } = parameterValues;

  useEffect(() => {
    // The FoundryView wrapper already subscribes to parameter events, but if the view wants to do additional things it can also subscribe
    hostEventTarget.addEventListener("host.update-parameters", (event) => {
      // do something
    });
  }, []);
};
```


packages/e2e.sandbox.views.todo-widget/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@osdk/e2e.sandbox.views.todo-widget",
```
styu, line 2: note i haven't actually ran this yet, i mostly made it to make sure all the types are right. I will go after a proper template (+ vite plugin) in a follow up PR

packages/views.client-react.unstable/src/client.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const handleEmitEvent = useCallback((message: ViewMessage<CONFIG>) => {
+    client.sendMessage(message);
+  }, []);
+  return (
+    <FoundryViewContext.Provider
+      value={{
+        emitEvent: handleEmitEvent,
+        parameterValues,
+        // Unfortunately the context is statically defined so we can't use the generic type, hence the cast
+      } as FoundryViewClientContext<ParameterConfig>}
```
styu, line 69: if there's a way to do this without the cast i'm all ears 😬

packages/views.client.unstable/src/client.ts
```diff
@@ -0,0 +1,92 @@
...
+    this.parentWindow = window.top;
+    const metaTag = document.querySelector(
+      `meta[name="${META_TAG_HOST_ORIGIN}"]`,
+    );
+    if (metaTag == null) {
+      throw new Error(
+        "[FoundryViewClient] Missing host origin meta tag "
+          + META_TAG_HOST_ORIGIN,
+      );
+    }
```
styu, line 55: there's an argument to be made that maybe this shouldn't throw in development mode..? on the one hand, it's nice to be consistent in how the code runs whether its in production or not. on the other hand, it means you can't actually develop locally without using the developer mode that Foundry will provide for developing on these views

packages/views.api.unstable/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+export type {
+  ArrayParameterValue,
+  BooleanParameterValue,
+  DateParameterValue,
+  NumericParameterValue,
+  ParameterValue,
+  ParameterValueType,
+  StringParameterValue,
+  TimestampParameterValue,
+} from "./parameters.js";
```
rosscourt, line 45: Should we be exporting all of the array variants too? `StringArrayValue` etc

packages/views.client.unstable/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+  type AsyncParameterValueMap,
+  type AsyncReloadingValue,
+  type AsyncValue,
+  type BooleanParameterValue,
+  type DateParameterValue,
+  type HostMessage,
+  type HostParametersUpdatedMessage,
+  isHostParametersUpdatedMessage,
+  isViewEmitEventMessage,
+  isViewReadyMessage,
```
rosscourt, line 32: Do we need to expose the type guards for view-sourced events on the client library? An event constructor would be useful

packages/views.client.unstable/src/client.ts
rosscourt: Hard to say without thinking about what a disconnected dev/test setup would look like. We might want to provide a stubbed alternative to FoundryViewClient which mocks API calls. Otherwise you'd need a local mock HTML renderer which could provide this value anyway

packages/views.api.unstable/src/utils/asyncValue.ts
```diff
@@ -0,0 +1,52 @@
...
+  type: "not-started";
+}
+
+export interface AsyncLoadingValue {
+  type: "loading";
+}
+
+export interface AsyncLoadedValue<V> {
+  type: "loaded";
+  value: V | undefined;
```
rosscourt, line 27: I still want to spend more time thinking about the `| undefined`s. If we support marking parameters as required we could treat undefined as a failed state, or a default value. One for later

packages/views.client.unstable/src/client.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public unsubscribe() {
+    if (this.listener == null) {
+      return;
+    }
+    window.removeEventListener("message", this.listener);
+  }
+
+  public sendMessage(message: ViewMessage<CONFIG>) {
```
rosscourt, line 89: I think it would make sense to have a event-type specific public methods, i.e. `emitEvent(event: Omit<ViewEmitEventMessage, "type">)`.

If we are auto-sending the ready event it never really makes sense for the dev to resend it. And when `ViewMessage` is extended to fetch-intercept events we should probably keep those off the API.

packages/views.client-react.unstable/src/client.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const FoundryView = <CONFIG extends ParameterConfig>(
+  { children, initialValues, onMessage }: FoundryViewProps<CONFIG>,
+): React.ReactElement<FoundryViewProps<CONFIG>> => {
+  const client = useMemo(() => new FoundryViewClient<CONFIG>(), []);
+  const [parameterValues, setParameterValues] = React.useState<
+    AsyncParameterValueMap<CONFIG>
+  >(initialValues);
+  useEffect(() => {
+    client.subscribe((event) => {
+      if (event.data.type === "host.update-parameters") {
```
rosscourt, line 47: `isHostParametersUpdatedMessage`?

packages/views.api.unstable/src/index.ts
styu: hm good question

packages/views.client.unstable/src/index.ts
styu: yea i can add more

packages/views.client.unstable/src/client.ts
styu: ah, makes sense. i can probably narrow the types so we only explicitly have the ones we want devs to call from the client

packages/views.api.unstable/src/messages/hostMessages.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { AsyncParameterValueMap, ParameterConfig } from "../config.js";
+
+// Interfaces and type guards for messages passed from the host Foundry UI to the child view
+
+export interface HostParametersUpdatedMessage<CONFIG extends ParameterConfig> {
+  type: "host.update-parameters";
+  parameters: AsyncParameterValueMap<CONFIG>;
+}
+
+export type HostMessage<CONFIG extends ParameterConfig> =
```
ericanderson, line 26: I assume you intend this to be a union at some point but I think we should reconsider for this type of project.

You are actually going to be running in a split world where the reality of the host will very likely be newer than the reality of the child. 

This means at some future point you will have the Host thinking there are 2 messages it sends and the client thinking there is only one. 

You may exhaust the union via if or switch and think you have handled everything when you in fact have not. The union will give you false confidence.

If we want to go this way, I think we need to be sure that adding to this union is a semver breaking change and that we communicate with the host which version of the API we are using. We probably need to do this no matter what actually.

ericanderson: I either way we probably need to make a base type that is roughly: `{ type: string, payload: X }`

packages/views.api.unstable/src/messages/viewMessages.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function isViewEmitEventMessage<CONFIG extends ParameterConfig>(
+  event: ViewMessage<CONFIG>,
+): event is ViewEmitEventMessage<CONFIG> {
+  return event.type === "view.emit-event";
+}
+
+export type ViewMessage<CONFIG extends ParameterConfig> =
+  | ViewReadyMessage
+  | ViewEmitEventMessage<CONFIG>;
```
ericanderson, line 51: Here we have the reverse of the above. Since the view will always be the same version of the host or older, you can feel confident that you know the extent of the messages to send and so a union is less of a problem here. Probably still best to create a base type though.

packages/views.api.unstable/src/parameters.ts
```diff
@@ -0,0 +1,80 @@
...
+  | NumericParameterValue
+  | BooleanParameterValue
+  | DateParameterValue
+  | TimestampParameterValue;
+
+export interface GenericArrayParameterValue<T extends PrimitiveParameterType> {
+  type: "array";
+  subType: T;
+  value: AsyncValue<PrimitiveParameterTypes[T][]>;
+}
```
ericanderson, line 54: would it be cleaner to just have a `multiple: true` or similar on the AbstractParameterValue?

ericanderson: I think you probably want an `optional: boolean` on the abstract type too?

packages/views.client.unstable/src/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  type AsyncParameterValueMap,
+  type AsyncReloadingValue,
+  type AsyncValue,
+  type BooleanArrayValue,
+  type BooleanParameterValue,
+  type DateArrayValue,
+  type DateParameterValue,
+  type HostMessage,
+  type HostParametersUpdatedMessage,
+  isHostParametersUpdatedMessage,
```
ericanderson, line 32: nit: break out type and non-type exports.

packages/views.client.unstable/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+) => void;
+
+export interface IFoundryViewClient<CONFIG extends ParameterConfig> {
+  ready: () => void;
+  emit: (
+    message: Extract<ViewMessage<CONFIG>, ViewEmitEventMessage<CONFIG>>,
+  ) => void;
+}
+
+export class FoundryViewClient<CONFIG extends ParameterConfig>
```
ericanderson, line 36: I have mixed feelings about exporting classes. I think they can severely limit what we are capable of doing and open us up to weird things like people trying to extend them when they should really be composing.

If we stick with a class, why would anyone want to use the `IFoundryView` type when they have to deal with a FoundryViewClient type to construct? What are we gaining?

If we go with function based `createFoundryViewClient()` then I would rename `IFoundryViewClient` to just `FoundryViewClient` and `FoundryViewClient` to `FoundryViewClientImpl`. 

I think I am biased towards exporting a function and never using `I` interfaces. I think it feels weird and don't really see it in any code in the wild. Even in this PR we only use it here once even though there are a TON of interfaces.

packages/views.client.unstable/src/client.ts
ericanderson: You definitely need to hard fail in development mode. You want people to realize whats happened easily.

You can also include a lot of extra info in console logs in development mode prior to throwing if you want. I also lean into the `tiny-invariant` library with the babel we have in repo. It rewrites these `invariant(window.top, "You must be running in an iframe")` to be:
```ts 
if (!window.top) {
  if ('production' !== process.env.NODE_ENV) {
    invariant(false, "You must be running in an iframe");
  } else {
    invariant(false);
  }
}
```

So this reduces down to almost nothing in production when the exception is mostly meaningless and non-actionable.

ericanderson: If you take arguments that you want to default in a test environment I highly recommend NOT doing: `constructor(foo: Window = window.top)` as this breaks hard in some runtimes (I forget if node dies on this). Really you want to take an optional and then you can check the optional and assign it to the default in that case. You may also need to change to `globalThis.window?.top`. Definitely do additional research here.

packages/views.client.unstable/src/client.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Notifies the host that this client is ready to receive the first parameter values
+   */
+  public ready() {
+    this.sendMessage({ type: "view.ready" });
+  }
+
+  public subscribe(listener: FoundryHostEventListener<CONFIG>) {
```
ericanderson, line 73: Maybe use `EventTarget` semantics here?

packages/views.client-react.unstable/src/client.tsx
ericanderson: Nope! You have to eat the cast here so others don't have to.

packages/views.client-react.unstable/src/context.ts
```diff
@@ -0,0 +1,43 @@
...
+  emitEvent: () => {},
+  sendReady: () => {},
+  parameterValues: {},
+});
+
+/**
+ * @returns The current FoundryViewClientContext, in the context of your specific parameter configuration
+ */
+export function useFoundryViewContext<CONFIG extends ParameterConfig>() {
+  return useContext(FoundryViewContext) as FoundryViewClientContext<CONFIG>;
```
ericanderson, line 42: If you want "perfection" you would not use `useFoundryViewContext` nor `FoundryViewContext` and instead could do something like this in the consumer's code

```ts
const { FoundryViewContext, useFoundryViewContext } = createFoundryViewReact<MyConfig>();
```

It requires the extra step of indirection but guarantees that you cannot screw up when doing `useFoundryViewContext()` as you would not need to pass a config.

packages/views.client-react.unstable/src/utils/initializeParameters.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import type {
+  AsyncParameterValueMap,
+  ParameterConfig,
+} from "@osdk/views-client.unstable";
+
+/**
+ * Utility function to initialize a map of parameter values to either a loading or not-started loading state
+ */
+export function initializeParameters<CONFIG extends ParameterConfig>(
```
ericanderson, line 25: I have a few thoughts about this:
1. If everyone is basically going to have to call this, why isn't it just done for them? Is there some scenario when someone doesn't want this behavior?
2. Is there a world where the backend doesn't provide all the parameters at once? Why have each value require an async check vs the whole thing?

I think really what I want as a consumer is to do this:
`const { isReady, parameters }  = useFoundryViewClient();` and not worry about checking the async of each value?

I'm also unsure that I would want to be responsible for calling `sendReady()`? Like isn't this hook being run all that is required? If I end up building a complex app am I now in the business of keeping track of which of my uses of `useFoundryViewClient()` did or did not sendReady?

ericanderson (dismissed):
Approving to unblock unstable packages

packages/views.api.unstable/src/parameters.ts
styu: i think `multiple: true` is less legible in my head as to that meaning "array". for optional vs required, im still not sure which way we want to go (depends on how the host frame looks)

packages/views.client.unstable/src/client.ts
styu: `IFoundryView` mostly existed so i could grab the type to use on the context, but there are obviously better ways of doing that. i don't feel strongly one way or the other, so happy to switch to a function

packages/views.client-react.unstable/src/context.ts
styu: hm, if that can be done at the top level (outside a component), will switch that to that, since react context needs to be made outside the component

packages/views.api.unstable/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+export type {
+  AsyncFailedValue,
+  AsyncLoadedValue,
+  AsyncLoadingValue,
+  AsyncNotStartedLoadingValue,
+  AsyncReloadingValue,
+  AsyncValue,
+} from "./utils/asyncValue.js";
```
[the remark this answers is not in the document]
styu: on this, i don't think you gain very much from the namespace, since I also want the generics from `AsyncValue` threaded through correctly

packages/views.api.unstable/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  DateParameterValue,
+  NumericArrayValue,
+  NumericParameterValue,
+  ParameterValue,
+  ParameterValueType,
+  StringArrayValue,
+  StringParameterValue,
+  TimestampArrayValue,
+  TimestampParameterValue,
+} from "./parameters.js";
```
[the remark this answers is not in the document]
styu: i think having an explicit array type is both more legible + makes the type inference for the config easier

packages/views.api.unstable/src/messages/hostMessages.ts
styu: yes, the intent is to be a union. i can update to use a (versioned?) base type 🤔

rosscourt: Should we bake an unknown/unsupported event type concept into the union so that we are encouraged to be explicit on handling that case?

rosscourt (approved):

packages/views.api.unstable/src/messages/viewMessages.ts
```diff
@@ -0,0 +1,94 @@
...
+        payload: infer P;
+      } ? P
+        : never,
+    ) => void;
+  }
+  & {
+    _unknown: (type: string) => void;
+  };
+
+export function visitViewMessage<CONFIG extends ParameterConfig>(
```
styu, line 82: this is meant to be consumed by Foundry's host code, so it's not going to be exported in `-client`

packages/e2e.sandbox.views.todo-widget/src/app.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    hostEventTarget.addEventListener("host.update-parameters", (event) => {
+      console.log("Received event:", event);
+    });
+  }, []);
+
+  const handleAddRow = useCallback(() => {
+    emitEvent("view.emit-event", {
+      eventId: "todo",
```
styu, line 27: i realized i forgot to handle the types for events, but i'll do that in a follow up PR instead

packages/views.api.unstable/src/messages/viewMessages.ts
ericanderson: Should we even put it in here?

packages/views.client.unstable/src/client.ts
```diff
@@ -14,93 +14,106 @@
  * limitations under the License.
  */
 
-import type {
+import {
   HostMessage,
```
ericanderson, line 18: why isn't this a type any more

packages/views.api.unstable/src/parameters.ts
```diff
@@ -35,46 +35,40 @@ export interface AbstractParameterValue<T extends PrimitiveParameterType> {
...
+  export type Date = AbstractParameterValue<"date">;
+  export type Timestamp = AbstractParameterValue<"timestamp">;
 
-export type ParameterValueType = ParameterValue["type"];
+  export type StringArray = GenericArrayParameterValue<"string">;
+  export type NumberArray = GenericArrayParameterValue<"number">;
+  export type BooleanArray = GenericArrayParameterValue<"boolean">;
+  export type DateArray = GenericArrayParameterValue<"date">;
+  export type TimestampArray = GenericArrayParameterValue<"timestamp">;
+  export type Array =
```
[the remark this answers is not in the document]
styu: i think this was easier to get the mapped types to work 🤔 but it was also one of those "got it to work so stopped touching it" situations

packages/views.client.unstable/src/client.ts
styu: there's a constant version string im referencing

packages/views.api.unstable/src/messages/viewMessages.ts
```diff
@@ -15,37 +15,84 @@
...
+  }
+
+  export type Payload<CONFIG extends ParameterConfig> =
+    | Payload.Ready
+    | Payload.EmitEvent<CONFIG>;
+
+  /**
+   * Emit when the child view is ready to start receiving messages from the host Foundry UI
+   */
+  export type Ready = ViewBaseMessage<"view.ready", Payload.Ready>;
```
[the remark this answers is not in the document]
styu: hm ok - i will do this in a follow up since i've started stacking PRs 😅

PR #979, opened 2024-11-15
Timeseries sensor support

ssanjay1:
Adding support for sensor time series types


packages/e2e.sandbox.catchall/src/runTimeseriesTest.ts
```diff
@@ -39,4 +39,9 @@ export async function runTimeseriesTest() {
   ) {
     console.log(point);
   }
+
+  const sensorResult = await client(SotSensor).fetchOne("well_3_pressure");
```
ssanjay1, line 43: Checked sensor objects work hitting backend

examples-extra/docs_example/ontology.json
```diff
@@ -332,7 +332,7 @@
             "description": "TimeSeries of the status of the employee",
             "dataType": {
               "type": "timeseries",
-              "itemType": "string"
```
ssanjay1, line 335: This JSON was incorrect before, so when we converted to timeseries prop we hit the fallback and made this `numericTimeseries` in our generated code. This is now fixed, so the generated code also reflects that

examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Employee.ts
```diff
@@ -25,7 +25,7 @@ export namespace Employee {
   export interface Props {
     readonly class: $PropType['string'] | undefined;
     readonly employeeId: $PropType['integer'];
-    readonly employeeStatus: $PropType['numericTimeseries'] | undefined;
+    readonly employeeStatus: $PropType['stringTimeseries'] | undefined;
```
ssanjay1, line 28: This is correct, we hit fallback before because of incorrect JSON formatting

ericanderson (approved):

PR #980, opened 2024-11-16
Add vite plugin for parsing configuration

styu:
~~// targets https://github.com/palantir/osdk-ts/pull/978~~
~~// targets https://github.com/palantir/osdk-ts/pull/984~~

This adds a vite plugin that will automatically extract view config from a `*.config.ts` file that is imported in the entrypoint file, along with the actual compiled entrypoint file information, and outputs it into a `.palantir/view-config.json` file.

That means that this setup:
image

Automatically produces this file:

image


packages/views.api.unstable/src/index.ts
```diff
@@ -18,10 +18,13 @@ export type {
   AsyncParameterValueMap,
   EventId,
   EventParameterValueMap,
+  ParameterDefinition,
```
styu, line 21: i need this internally to process the manifest

packages/e2e.sandbox.views.todo-widget/src/main.tsx
```diff
@@ -6,7 +6,7 @@ import { Theme } from "@radix-ui/themes";
 import React from "react";
 import { createRoot } from "react-dom/client";
 import { App } from "./app.js";
-import MainConfig from "./main.parameters.js";
+import MainConfig from "./main.config.js";
```
styu, line 9: i decided to rename this since its not just parameters anymore

packages/views.client.unstable/src/host.ts
```diff
@@ -41,7 +41,7 @@ export class FoundryHostEventTarget<
       | null,
     options?: AddEventListenerOptions | boolean,
   ): void {
-    super.addEventListener(type, callback, options);
+    super.addEventListener(type, callback as EventListener, options);
```
styu, line 44: i had to drop the `extends` else the event is `Event | CustomEvent`, which is annoying for consumers to deal with when the type of the event is typesafe

packages/views.api.unstable/src/manifest.ts
```diff
@@ -0,0 +1,36 @@
...
+export interface ViewManifestConfig {
+  rid: string;
+  version: string;
+  entrypointJs: string[];
+  entrypointCss?: string[];
+  parameters: Record<string, ParameterDefinition>;
+  events: Record<string, EventDefinition<ParameterConfig>>;
+}
+
+export const MANIFEST_FILE_LOCATION = ".palantir/view-config.json";
```
tzyl, line 36: Should we follow naming convention of `foundry.config.json`? And make it plural given that it can publish multiple views
```suggestion
export const MANIFEST_FILE_LOCATION = ".palantir/views.config.json";
```

packages/views.api.unstable/src/manifest.ts
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+import type {
+  EventDefinition,
+  ParameterConfig,
+  ParameterDefinition,
+} from "./config.js";
+
+export interface ViewManifest {
```
tzyl, line 23: something like this?
```suggestion
export interface ViewManifestV1 {
  version: "v1"
```

Also, where do you think is the best place for us to formally document this v1 spec, should it live on public docs? I am assuming we want to say as long as you produce this format in the given location you're compatible even if you don't want to use the vite plugin and/or vite

packages/views.vite-plugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+  // Rest of configuration
+});
+```
+
+## Defining configuration
+
+The `@osdk/views-client` package exports a `ViewConfig` type so that you can define a configuration object in the shape that this plugin requires. You can define your config like so:
+
+```js
+// main.config.ts
```
tzyl, line 24: How does this relate to the dedicated parameters file I think you had before? Does it replace it?

packages/views.vite-plugin/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Importing the configuration
+
+Import your `*.config.ts` file in any entrypoint JavaScript/TypeScript file so the plugin can pick it up. Entrypoint files are any that are imported from your HTML file. For example, if you have the following setup:
+
+```
+src/
+  main.tsx
+  main.config.ts
+index.html
```
tzyl, line 63: Can you give an example file structure of what you think a multiple view publishing project would look like to make sure I understand? For example, do they each have an index.html (is this only used for discovering entrypoints?) or are they all just script tags in a single dummy index.html

packages/views.api.unstable/src/manifest.ts
```diff
@@ -0,0 +1,36 @@
...
+  views: Record<string, ViewManifestConfig>;
```
tzyl, line 24: What does the key represent here and is it meaningful? It looks like in the plugin we take the chunk that contains the config name so is it intended to just be a random unique identifier for the view?

```
          viewConfigManifest.views[chunk.name] = viewConfig;
```

Should we ask the user to pick a unique id key themselves in the config? Although we don't actually have that as a real concept of a view id outside of here

packages/views.api.unstable/src/manifest.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface ViewManifestConfig {
+  rid: string;
+  version: string;
+  entrypointJs: string[];
+  entrypointCss?: string[];
+  parameters: Record<string, ParameterDefinition>;
+  events: Record<string, EventDefinition<ParameterConfig>>;
```
tzyl, line 33: Are any of these (looking at parameters/events) specific to workshop or will all host applications be able to provide the view with these? I'm wondering if we need to declare the type of view if there are view type specific configs and then there will be validation on publish that the type matches up with the view rid it publishes to

packages/views.api.unstable/src/manifest.ts
styu: oh fair, i can rename. no strong opinion on `-` vs `.` so i'll just match `foundry.config.json`

I think since this lives on the client side vs literally in the foundry system, i kind of prefer `.palantir`

packages/views.api.unstable/src/manifest.ts
styu: oops i forgot the version, my bad 🤦‍♀️ 
i think public docs is probably best, yeah

packages/views.vite-plugin/README.md
styu: yes, it replaces it. I just renamed it because i realized it has events and parameters, so calling it `.parameters` felt off

packages/views.vite-plugin/README.md
styu: yup can do! for posterity they would each have their own index.html (since that's how vite works)

packages/views.api.unstable/src/manifest.ts
styu: it's not meaningful, it's mostly for bookkeeping inside the vite plugin so i can keep track of individual entrypoints. if someone were to handroll this, they could type whatever they want for that key

packages/views.api.unstable/src/manifest.ts
styu: I would expect any host application to need to support passing in parameters/accepting events, not just Workshop. I think it's ok for that translation to happen internally for now while we just have 1

packages/views.vite-plugin/src/plugin.ts
```diff
@@ -0,0 +1,249 @@
...
+  viewConfigString = viewConfigString.replace(
+    /: '(.+)'/g,
+    ": \"$1\"",
+  );
+
+  // Convert single quote string values in arrays to double quotes
+  viewConfigString = viewConfigString.replace(
+    /: \['(.+)'\]/g,
+    ": [\"$1\"]",
+  );
```
styu, line 246: im sure there's a way to do this in one go but i am an unimaginative regex-er

packages/views.vite-plugin/README.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+And in `index.html`, you import `main.tsx`:
```
rosscourt, line 66: In practice these view repos won't need any HTML files, do you depend on them for the build? They won't get used in the deployed view so I'd want to avoid users thinking they can add things here that effect the deployed view

packages/views.vite-plugin/src/plugin.ts
```diff
@@ -0,0 +1,249 @@
...
+          return;
+        }
+        const entrypoint = jsSourceFileToEntrypointMap[sourceFile];
+        if (entrypoint == null) {
+          return;
+        }
+        const extension = path.extname(source);
+        const filename = path.basename(source, extension);
+        if (filename.endsWith(CONFIG_FILE_SUFFIX)) {
+          // We found a .parameters file that's imported from an entrypoint JS file, save it
```
rosscourt, line 72:
```suggestion
          // We found a .config file that's imported from an entrypoint JS file, save it
```

packages/views.vite-plugin/README.md
```diff
@@ -0,0 +1,163 @@
+# @osdk/views-manifest-vite-plugin
+
+This is a vite plugin that will automatically discover `*.config.(j|t)s` files that are imported into entrypoint files and generates a `.palantir/view-config.json` file for Foundry to read from. This allows developers to write out their configuration for custom views in Foundry in a single, type-safe file and not worry about generating a manifest file needed for publishing new versions of the view within Foundry.
```
rosscourt, line 3: We are fairly confident that nothing else uses the `*.config.js` format right? Or at least in the fact that only considering `.config.js` imports from the entrypoint avoids some set of potential collisions

packages/views.vite-plugin/README.md
styu: yes, that's how vite works - entrypoints are the HTML files

packages/views.vite-plugin/README.md
styu: if we wanted to be super safe we could probably to `.foundry.config.js` ..?

styu: actually i realized it shouldn't matter - esp if we add validation on the parsed config object, we can ignore any other `.config.js` file that doesn't fit what we're looking for

ericanderson (approved):
Approving to unblock. Left a few notes and gave feedback over slack

packages/views.vite-plugin/src/plugin.ts
```diff
@@ -0,0 +1,246 @@
...
+    moduleParsed: (moduleInfo) => {
+      const sourceFile = fileIdToSourceFileMap[moduleInfo.id];
+      if (sourceFile == null) {
+        return;
+      }
+      const entrypointForConfig = configSourceFileToEntrypointMap[sourceFile];
+      if (entrypointForConfig == null) {
+        return;
+      }
+
```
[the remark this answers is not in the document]
styu: i also thought about that - but it seemed a little silly because vite is already compiling the file, i didn't want to double compile it

packages/views.vite-plugin/README.md
styu: hm... im viewing this vite plugin as an atomic unit that can be composed no matter what the rest of your vite config may look like - i think if we wanted to wrap it in an opinionated config i would do that elsewhere.

styu:
im going to merge and address feedback in a FLUP, since i also need to rename everything to `widget`

PR #984, opened 2024-11-18
Implement typesafe events in views API

styu:
This wires up events in the configuration. The key thing with events is that they should only be allowed to reference parameter IDs that are actually defined in the same config. Because we're inferring the config type on the fly, the only way I could get the typechecking to work is to error when creating the client.

For example, the following config will produce the error below:

```
const MainConfig = {
  parameters: {
    headerText: {
      displayName: "Widget title",
      type: "string",
    },
    showWarning: {
      displayName: "Show warning callout",
      type: "boolean",
    },
    todoItems: {
      displayName: "Todo items",
      type: "array",
      subType: "string",
    },
  },
  events: {
    updateHeader: {
      displayName: "Update header",
      parameterIds: ["asdfasdf"], // Will cause error, because this parameter ID does not exist
    },
  },
} as const satisfies ViewConfig;
```

image

### DIFF
--- a/.changeset/empty-poems-hug.md
+++ b/.changeset/empty-poems-hug.md
@@ -1,0 +1,9 @@
+---
+"@osdk/views-client-react.unstable": minor
+"@osdk/views-client.unstable": minor
+"@osdk/views-api.unstable": minor
+"@osdk/client": patch
+"@osdk/api": patch
+---
+
+Add support for events
--- a/.changeset/empty-poems-hug.md
+++ b/.changeset/empty-poems-hug.md
@@ -2,8 +2,6 @@
 "@osdk/views-client-react.unstable": minor
 "@osdk/views-client.unstable": minor
 "@osdk/views-api.unstable": minor
-"@osdk/client": patch
-"@osdk/api": patch
 ---
 
 Add support for events
--- a/packages/views.client-react.unstable/src/client.tsx
+++ b/packages/views.client-react.unstable/src/client.tsx
@@ -16,7 +16,7 @@
 
 import type {
   AsyncParameterValueMap,
-  ParameterConfig,
+  ViewConfig,
 } from "@osdk/views-client.unstable";
 import { createFoundryViewClient } from "@osdk/views-client.unstable";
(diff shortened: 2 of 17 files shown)


.changeset/empty-poems-hug.md
```diff
@@ -0,0 +1,9 @@
+---
+"@osdk/views-client-react.unstable": minor
+"@osdk/views-client.unstable": minor
+"@osdk/views-api.unstable": minor
+"@osdk/client": patch
+"@osdk/api": patch
```
rosscourt, line 6: Should these last two be here?

rosscourt (approved):

.changeset/empty-poems-hug.md
styu: this feels like it came from something on `main`..?

ericanderson: hmm, looks like maybe rebase got you? I bet they cut a beta and you still have this in your tree

packages/views.client-react.unstable/src/client.tsx
```diff
@@ -16,15 +16,17 @@
...
 } from "@osdk/views-client.unstable";
 import { createFoundryViewClient } from "@osdk/views-client.unstable";
 import React, { useEffect, useMemo } from "react";
 import type { FoundryViewClientContext } from "./context.js";
 import { FoundryViewContext } from "./context.js";
 import { initializeParameters } from "./utils/initializeParameters.js";
 
-interface FoundryViewProps<CONFIG extends ParameterConfig> {
+interface FoundryViewProps<
+  CONFIG extends ViewConfig<{ parameters: CONFIG["parameters"] }>,
```
ericanderson, line 28: Were you not getting the types you wanted without this?

ericanderson (approved):

packages/views.client-react.unstable/src/client.tsx
styu: correct, it would not narrow in the event type and instead make every event's parameters be a union of all parameters defined in the event config :(

ericanderson (approved):

PR #991, opened 2024-11-20
Clean up where clause types

ssanjay1:
Cleaning up where clause types to use new ones we created.


packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -15,8 +15,7 @@
  */
 
 import type {
-  GeoFilter_Intersects,
```
ssanjay1, line 18: Cannot remove these two types from api package because that would be a break

ericanderson: So why can we remove them?

ericanderson (approved):

PR #993, opened 2024-11-20
Rename views to widget instead

styu:
This deconflicts with the existing concept of dataset views. I also incorporated some improvements from @ericanderson :

- Adds a `defineConfig` function so that the `as const satisfies WidgetConfig` statement is no longer necessary
- Adds a `emitEvent` function on the client, and renames `emit` to `sendMessage` instead.
- Reworks the parameter values on the React context so that devs can either choose to deal with individual async values, or have the async-ness lumped aggregated together.


packages/widget.client-react.unstable/src/client.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  return (
+    <FoundryWidgetContext.Provider
+      value={{
+        emitEvent: client.emitEvent,
+        hostEventTarget: client.hostEventTarget,
+        asyncParameterValues:
+          asyncParameterValues as unknown as AsyncParameterValueMap<
+            WidgetConfig<ParameterConfig>
+          >,
```
styu, line 160: i could not figure out why typescript didn't think these were assignable..

packages/widget.client-react.unstable/src/context.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @returns The current FoundryWidgetClientContext, in the context of your specific parameter configuration
+ */
+export function useFoundryWidgetContext<
+  C extends WidgetConfig<C["parameters"]>,
+>() {
+  return useContext(
+    FoundryWidgetContext,
+  ) as any as FoundryWidgetClientContext<C>;
```
styu, line 64: same here

packages/e2e.sandbox.todowidget/src/app.tsx
```diff
@@ -26,19 +27,13 @@ export const App: React.FC = () => {
...
-      eventId: "updateTodoItems",
+    emitEvent("updateTodoItems", {
       parameterUpdates: {
-        todoItems: [
-          ...(todoItems.value.type === "loaded"
-            ? todoItems.value.value ?? []
-            : []),
-          newTodoItem,
-        ],
+        todoItems: [...(todoItems ?? []), newTodoItem],
```
ericanderson, line 32: so much better

packages/e2e.sandbox.todowidget/src/app.tsx
```diff
@@ -12,12 +12,13 @@ import {
   TextField,
 } from "@radix-ui/themes";
 import React, { useCallback, useEffect, useState } from "react";
-import { useFoundryContext } from "./context.js";
+import { useWidgetContext } from "./context.js";
 
 export const App: React.FC = () => {
-  const { parameterValues, hostEventTarget, emitEvent } = useFoundryContext();
+  const { parameterValues, parameterLoadingState, hostEventTarget, emitEvent } =
```
ericanderson, line 18: i think `parameterValues` and `parameterLoadingState` is a bit long in the tooth still. maybe `parameters` and `parameters.state` and `parameters.values`

packages/e2e.sandbox.todowidget/src/main.config.ts
```diff
@@ -27,4 +27,4 @@ export default {
       parameterIds: ["todoItems"],
     },
   },
-} as const satisfies ViewConfig;
+});
```
ericanderson, line 30: Also far crispier.

packages/e2e.sandbox.todowidget/src/main.tsx
```diff
@@ -11,13 +11,9 @@ import MainConfig from "./main.config.js";
 const root = document.querySelector("body")!;
```
ericanderson, line 11: Definitely wish we could get rid of this whole file since they will always almost be identical. Same with `index.html`

packages/e2e.sandbox.todowidget/src/secondApp.tsx
```diff
@@ -8,14 +8,12 @@ import {
...
+import React, { useEffect } from "react";
 import type Config from "./second.config.js";
 
 export const App: React.FC = () => {
-  const { parameterValues, hostEventTarget } = useFoundryViewContext.withTypes<
-    typeof Config
-  >()();
-  const [newTodoItem, setNewTodoItem] = useState("");
+  const { parameterValues, parameterLoadingState, hostEventTarget } =
+    useFoundryWidgetContext.withTypes<typeof Config>()();
```
ericanderson, line 16: should this be broken to a higher level to reflect idealized usage?

`const useFoundryWidget = useFoundryWidgetContext.withType<typeof Config>` at the top level?

ericanderson (dismissed):
Approval to unblock progress pending bike shedding in a follow up

packages/e2e.sandbox.todowidget/src/secondApp.tsx
styu: yeah, i was just being lazy lol

ericanderson (approved):

PR #995, opened 2024-11-20
Packages generated with foundry-sdk-generator are now properly esm

ericanderson:
Includes test to make sure the generated output in --beta mode and regular mode does not have exports syntax from commonjs


ssanjay1 (approved):
Thank you for doing this!

PR #997, opened 2024-11-21
[widgets] Add temporary messages for a fetch proxy 

styu:
This does expose the OSDK create client function on the widget client, which is nice because we deal with figuring out the URL for them when within the sandbox.

The end goal is to not have to proxy on the frontend and instead create a proxy on the server so we don't have to endlessly mirror all fetch/browser functionality, but we need this for now to show it working end to end first

### DIFF
--- a/packages/widget.api.unstable/src/messages/widgetMessages.ts
+++ b/packages/widget.api.unstable/src/messages/widgetMessages.ts
@@ -52,7 +52,7 @@ export namespace WidgetMessage {
       id: string;
       url: string;
       method: string;
-      headers: Record<string, string>;
+      headers: Record<string, string> | [string, string][];
       body?: string;
     }
   }
--- a/packages/widget.client.unstable/src/request.ts
+++ b/packages/widget.client.unstable/src/request.ts
@@ -32,9 +32,9 @@ export function serializeRequest(
     id,
     url: input,
     method: init?.method ?? "GET",
-    headers: Object.fromEntries([
+    headers: [
       ...new Headers(init?.headers ?? {}).entries(),
-    ]),
+    ],
     body: body as any,
   };
 }


packages/widget.api.unstable/src/messages/hostMessages.ts
```diff
@@ -30,27 +30,88 @@ export namespace HostMessage {
     export interface UpdateParameters<C extends WidgetConfig<C["parameters"]>> {
       parameters: AsyncParameterValueMap<C>;
     }
+
+    /**
+     * Temporary fetch proxy response
+     * Will be removed in favor of server side proxy
+     */
+    export interface _unstable_FetchResponseSuccess {
```
styu, line 38: i prefixed these with `_unstable_` because i don't think they should be long lived

packages/widget.api.unstable/src/messages/widgetMessages.ts
```diff
@@ -62,11 +75,23 @@ export namespace WidgetMessage {
...
     extends WidgetBaseMessage<"widget.emit-event", Payload.EmitEvent<C>>
   {}
+
+  /**
+   * Temporary fetch proxy
+   * Will be removed in favor of server side proxy
+   */
+  export interface _unstable_FetchRequest extends
+    WidgetBaseMessage<
+      "widget._unstable_fetch-request",
```
rosscourt, line 85: To match the host events
```suggestion
      "widget._unstable.fetch-request",
```

packages/widget.api.unstable/src/messages/widgetMessages.ts
```diff
@@ -80,6 +105,12 @@ export function isWidgetEmitEventMessage<
   return event.type === "widget.emit-event";
 }
 
+export function _unstable_isWidgetFetchMessage<
+  C extends WidgetConfig<C["parameters"]>,
+>(event: WidgetMessage<C>): event is WidgetMessage._unstable_FetchRequest {
+  return event.type === "widget._unstable_fetch-request";
```
rosscourt, line 111: As above
```suggestion
  return event.type === "widget._unstable.fetch-request";
```

packages/widget.client.unstable/src/client.ts
```diff
@@ -94,8 +115,59 @@ export function createFoundryWidgetClient<
...
 
+  // Temporary fetch proxy since the widget iframe isn't allowed to make requests (yet)
+  // This should be replaced by a server-side proxy instead.
+  const fetchProxy: typeof fetch = async (
+    input: RequestInfo | URL,
+    init?: RequestInit | undefined,
+  ) => {
+    const requestId = Math.random().toString(36).substring(7);
+    sendMessageToHost({
+      type: "widget._unstable_fetch-request",
```
rosscourt, line 126:
```suggestion
      type: "widget._unstable.fetch-request",
```

packages/widget.client.unstable/src/client.ts
```diff
@@ -94,8 +115,59 @@ export function createFoundryWidgetClient<
...
   };
 
+  // Temporary fetch proxy since the widget iframe isn't allowed to make requests (yet)
+  // This should be replaced by a server-side proxy instead.
+  const fetchProxy: typeof fetch = async (
+    input: RequestInfo | URL,
+    init?: RequestInit | undefined,
+  ) => {
+    const requestId = Math.random().toString(36).substring(7);
+    sendMessageToHost({
```
rosscourt, line 125: Lets do this last to ensure that the listener has been bound

packages/widget.client.unstable/src/request.ts
```diff
@@ -0,0 +1,59 @@
...
+    id,
+    url: input,
+    method: init?.method ?? "GET",
+    headers: serializeHeaders(init?.headers),
+    body: body as any,
+  };
+}
+
+function serializeHeaders(headers: HeadersInit | undefined) {
+  if (headers === undefined || !(headers instanceof Headers)) {
```
rosscourt, line 41: Apologies I know this is copied from me, but for completeness we could also error if not an instance of Headers, at least then all of the data type assumptions of this code error loudly if the client did something we don't support

rosscourt (approved):

packages/widget.client.unstable/src/request.ts
ericanderson: So `HeadersInit` can be any of the following `string, string][] | Record<string, string> | Headers`. But also you can pass all of those to `new Headers()` to ensure you have a headers object. 

You could also therefore simplify serialization to just `[...new Headers(headers ?? {}).entries()]` and rehydrate on the other side with `new Headers(...)`

packages/widget.client.unstable/src/client.ts
```diff
@@ -94,8 +115,58 @@ export function createFoundryWidgetClient<
...
+  const fetchProxy: typeof fetch = async (
+    input: RequestInfo | URL,
+    init?: RequestInit | undefined,
+  ) => {
+    const requestId = Math.random().toString(36).substring(7);
+
+    return new Promise<Response>((resolve, reject: (error: Error) => void) => {
+      function handleMessage(event: MessageEvent) {
+        const { data } = event;
+        visitHostMessage(event.data, {
```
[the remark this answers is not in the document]
styu: lol i like that the compiler complains if you leave something out!

packages/widget.client.unstable/src/host.ts
```diff
@@ -16,51 +16,55 @@
...
+  removeEventListener<T extends HostMessage<C>["type"]>(
+    type: T,
     callback:
-      | HostMessageEventListener<C, M>
-      | HostMessageEventListenerObject<C, M>
+      | HostMessageEventListener<
+        Extract<HostMessage<C>, { type: T }>["payload"]
+      >
+      | HostMessageEventListenerObject<
+        Extract<HostMessage<C>, { type: T }>["payload"]
```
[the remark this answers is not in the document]
styu: seems so!

ericanderson (dismissed):

ericanderson (approved):

PR #1002, opened 2024-11-21
Expose fuzzy search

ssanjay1:
Exposing option to turn on fuzzy searching


packages/api/src/aggregate/StringFilter.ts
```diff
@@ -17,11 +17,11 @@
 import type { BaseFilterOptions, CatchThemAll } from "./BaseFilter.js";
 import type { Just } from "./Just.js";
 
-interface StringFilterOptions extends BaseFilterOptions<string> {
+export interface StringFilterOptions extends BaseFilterOptions<string> {
   "$startsWith": string;
   "$containsAllTermsInOrder": string;
-  "$containsAnyTerm": string;
-  "$containsAllTerms": string;
+  "$containsAnyTerm": string | { term: string; fuzzySearch?: boolean };
```
nihalbhatnagar, line 23: Can we expose somewhere what fuzzySearch means exactly? Not sure if dev console docs is best or if we need inline.

ssanjay1: Also, removing export for now

nihalbhatnagar (dismissed):

nihalbhatnagar (approved):

PR #1003, opened 2024-11-21
[widgets] Add support for JavaScript script tag types

styu:
I realized when setting up multiple entrypoints that vite expects scripts to be loaded with `<script type="module">`. Since it's not guaranteed that devs will be using vite, we will support either `text/javascript` or `module` when placing widget JS files on the page in the HTML file.

### DIFF
--- a/.changeset/rude-spoons-cheer.md
+++ b/.changeset/rude-spoons-cheer.md
@@ -1,0 +1,6 @@
+---
+"@osdk/widget-api.unstable": minor
+"@osdk/widget-manifest-vite-plugin": minor
+---
+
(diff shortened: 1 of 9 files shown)


.changeset/rude-spoons-cheer.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/widget-api.unstable": minor
```
styu, line 2: im still < 1.x so i just want to bump the minor version

rosscourt (approved):

rosscourt (approved):

PR #1009, opened 2024-11-21
[widgets] Properly list CSS files to load

styu:
Turns out vite will put link tags in the HTML for CSS files required by imported chunks, so I follow those imports to add them to the manifest

### DIFF
--- a/packages/widget.vite-plugin/src/plugin.ts
+++ b/packages/widget.vite-plugin/src/plugin.ts
@@ -245,8 +245,14 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
               imports.includes(chunk.fileName)
               && chunk.viteMetadata?.importedCss.size
             ) {
+              const existingCssFiles =
+                widgetConfigManifest.widgets[entrypointName].entrypointCss?.map(
+                  ({ path }) => path,
+                ) ?? [];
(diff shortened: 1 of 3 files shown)


packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -231,6 +234,24 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
...
+            const [entrypointName, imports] of Object.entries(entrypointImports)
+          ) {
+            if (
+              imports.includes(chunk.fileName)
+              && chunk.viteMetadata?.importedCss.size
+            ) {
+              widgetConfigManifest.widgets[entrypointName].entrypointCss?.push(
+                ...[...chunk.viteMetadata.importedCss].map((css) => ({
+                  path: css,
+                })),
```
rosscourt, line 251: Do we know how many entrypoint's this produces? E.g. is it one per entrypoint / dynamic import etc, rather than one per literal CSS file?

rosscourt (dismissed):

packages/widget.vite-plugin/src/plugin.ts
styu: ah fair, it could be duplicated 🤔 will fix

PR #1010, opened 2024-11-21
Start up verdaccio for e2e

ericanderson:


.github/workflows/ci.yml
```diff
@@ -108,3 +108,48 @@ jobs:
...
 
+  e2e:
+    name: E2E Tests
+    timeout-minutes: 15
+    runs-on: ubuntu-latest
+
+    services:
+      verdaccio:
+        image: verdaccio/verdaccio
+        # options: >-
```
ssanjay1, line 119: Do we want to keep these in case for the future?

PR #1013, opened 2024-11-22
Updates gateway versions and uses Response type for Binary Data

nihalbhatnagar:


.gitignore
```diff
@@ -36,4 +36,6 @@ pnpm-publish-summary.json
 
 api-docs/
 temp/
-tsup.config.bundled_*.mjs
\ No newline at end of file
+tsup.config.bundled_*.mjs
+
+vitest.workspace.ts
```
nihalbhatnagar, line 41: Lets me use vscode vitest extension

packages/client/src/createAttachmentFromRid.ts
```diff
@@ -34,7 +34,7 @@ export function createAttachmentFromRid(
       return OntologiesV2.Attachments.read(
         client,
         rid,
-      ) as Promise<Blob>;
+      );
```
ssanjay1, line 37: Does this not need the cast anymore to `Response`?

ssanjay1 (approved):

PR #1018, opened 2024-11-25
Interfaces for Subscribe

nihalbhatnagar:
Allows for interfaces on object set subscriptions

### DIFF
--- a/packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
+++ b/packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
@@ -35,7 +35,10 @@ import WebSocket from "isomorphic-ws";
 import invariant from "tiny-invariant";
 import type { Logger } from "../Logger.js";
 import type { ClientCacheKey, MinimalClient } from "../MinimalClientContext.js";
-import { convertWireToOsdkObjects } from "../object/convertWireToOsdkObjects.js";
+import {
+  convertWireToOsdkObjects,
+  convertWireToOsdkObjects2,
+} from "../object/convertWireToOsdkObjects.js";
 
 const MINIMUM_RECONNECT_DELAY_MS = 5 * 1000;
 
@@ -65,7 +68,8 @@ interface Subscription<
 > {
   listener: Required<ObjectSetListener<Q, P>>;
   objectSet: ObjectSet;
-  primaryKeyPropertyName: string;
+  interfaceApiName?: string;
+  primaryKeyPropertyName?: string;
   requestedProperties: Array<P>;
   requestedReferenceProperties: Array<P>;
   subscriptionId: string;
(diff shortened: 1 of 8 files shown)


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -432,15 +452,17 @@ export class ObjectSetListenerWebsocket {
     );
     const osdkObjectsWithReferenceUpdates = await Promise.all(
       referenceUpdates.map(async (o) => {
-        const osdkObjectArray = await convertWireToOsdkObjects(
+        const osdkObjectArray = await this.#client.objectFactory(
```
nihalbhatnagar, line 455: I'm using the old interface API's since the our new implementations require and interface to object type properties map. I could manually mock this, but I think it would be best to talk through this @ssanjay1 and just unblock for right now

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -273,12 +286,19 @@ export class ObjectSetListenerWebsocket {
...
+          requestedProperties,
+          requestedReferenceProperties,
+          interfaceApiName,
+        },
       ) => {
         return {
           objectSet: objectSet,
-          propertySet: requestedProperties,
-          referenceSet: requestedReferenceProperties,
+          propertySet: interfaceApiName == null ? requestedProperties : [],
```
nihalbhatnagar, line 298: Empty set because OSW does not currently load SPT mappings. So currently with interfaces, we can say we don't support specifying properties and will load all properties. Reference properties TBD

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -173,31 +177,49 @@ export class ObjectSetListenerWebsocket {
...
+
+      objectProperties = properties.filter((p) =>
+        objDef.properties[p].type !== "geotimeSeriesReference"
+      );
+
+      referenceProperties = properties.filter((p) =>
+        objDef.properties[p].type === "geotimeSeriesReference"
+      );
+    } else {
+      objectProperties = [];
```
nihalbhatnagar, line 204: OSW assumes the names we provide them are the API names for the actual object type implementations instead of the SPT names. Therefore the current API I decided on was assuming if you pass an interface in that you are only specifying reference properties. This does require any casting what is passed in, and I think it's a larger conversation if the TS OSDK supports specifying anything beyond an SPT.

nihalbhatnagar: There are ongoing discussions to allow OSW to take in the SPT names instead of the object type property API names here. I'll also open up the discussion as to how we want to allow requesting reference properties in the finalized API

ericanderson (approved):
Approving to unblock

PR #1019, opened 2024-11-26
Try catch subscription handlers

nihalbhatnagar:


packages/api/src/objectSet/ObjectSetListener.ts
```diff
@@ -45,7 +45,7 @@ export interface ObjectSetListener<
   /**
    * There was a fatal error with the subscription process. The subscription will close or will not be established.
    */
-  onError?: (errors: Array<any>) => void;
+  onError?: (errors: { subscriptionClosed: boolean; error: any }) => void;
```
nihalbhatnagar, line 48: Settled on a boolean here, the alternative was an enum to also be able to display the "I already tried to reconnect for you case" but I think that adds unnecessary complexity

nihalbhatnagar: maybe I need to make it isSubscriptionClosed?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -454,7 +454,12 @@ export class ObjectSetListenerWebsocket {
 
     for (const osdkObject of osdkObjectsWithReferenceUpdates) {
       if (osdkObject != null) {
-        sub.listener.onChange?.(osdkObject);
+        try {
+          sub.listener.onChange?.(osdkObject);
+        } catch (error) {
+          this.#logger?.error(error, "Error in onChange callback");
+          sub.listener.onError({ subscriptionClosed: false, error });
```
ericanderson, line 461: if an error happens while in `onError` the whole thing still breaks, no? We need a fallback in that case.

ericanderson (approved):

PR #1023, opened 2024-11-28
Use api:ontologies-X over api:X-data

woojiahao:
Change usages of `api:read-data` -> `api:ontologies-read` and `api:write-data` -> `api:ontologies-write`


packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -151,6 +151,10 @@ export function createPublicOauthClient(
   const authServer = createAuthorizationServer(ctxPath, url);
   const oauthHttpOptions: HttpRequestOptions = { [customFetch]: fetchFn };
 
+  if (scopes.length === 0) {
```
ericanderson, line 154: guessing this actually belongs in the `processOptionsAndAssignDefaults`?

ericanderson (commented):
A few notes and can you write `pnpm exec changeset` and add a changeset for the oauth and cli.common packages?

woojiahao:
> add a changeset for the oauth and cli.common packages

Hm, I have included it already: https://github.com/palantir/osdk-ts/blob/7fbdc3a00ddeda9f8f398c7fb9a4302b4fd13738/.changeset/funny-rats-tease.md

I think @ssanjay1 mentioned that the CI might be caused by using a fork of the repository, I'll share the relevant thread for that

packages/oauth/src/createPublicOauthClient.ts
woojiahao: Oops, yeah, thanks!

ericanderson (approved):

PR #1030, opened 2024-12-03
Fix deps in @osdk/react on client and api

nihalbhatnagar:


ericanderson (changes requested):
Blocking while we discuss

ericanderson (approved):

PR #1031, opened 2024-12-03
Experiment with no proxy for objects

ericanderson:
# BLUF

This PR removes the proxy objects in favor of simply adding functionality to objects from the wire. Originally this was to make it easier for people when they `console.log()` objects but additionally it seems we get a per object size reduction from `2,324 bytes` per object in this test bed to `1,332 bytes`. This does increase construction of the objects by `13%` which seems reasonable. 

# What does it cost us?

We lose a few things in this exchange:

- The ability to do `$updateInternalValues()` and have it reflect across any object that was `$as()`'d. OSDK objects effectively become immutable in this scenario. I think this is okay, it just removes some future ideas I had.
- An obvious tie in for telemetry metrics on per property usage. Before, in "telemetry mode" we would have been able to just hook into the proxy property access. We could still build this but will need to rethink how it would work.

# What else did we gain?

I think the code is a bit easier to reason about after we remove the proxies. Also the bundled minified client for nodejs (via esbuild) is reduced by about 4k in bytes.

# Other notes

In a separate repo where I originally experimented with this, retrieving properties from a proxy vs directly on an object was about 2x as expensive. So for anyone iterating over all of the objects returned, the 13% construction time increase would be saved on property access.

# Detailed benchmark info

On my machine, `createObjects.ts` benchmark stats prior (1b60b3d3):
```json
      "stats": {
        "time": {
          "std": 20.022638911991578,
          "mean": 542.9110000000001,
          "min": 523,
          "max": 593.92
        },
        "heapUsed": {
          "std": 44488.05771620065,
          "mean": 232406548.8,
          "min": 232355784,
          "max": 232456888
        },
        "rss": {
          "std": 2388346.6567709134,
          "mean": 237371392,
          "min": 233029632,
          "max": 241123328
        }
      }
```

So per object in the test data we were paying about `232406548.8/1000/100 ~= 2,324 bytes` per object.

On my machine, `createObjects.ts` benchmark stats after:
```json
      "stats": {
        "time": {
          "std": 19.36567313573169,
          "mean": 613.712,
          "min": 589.16,
          "max": 653.87
        },
        "heapUsed": {
          "std": 21950.770067585327,
          "mean": 133270992.8,
          "min": 133241184,
          "max": 133293728
        },
        "rss": {
          "std": 2037190.144079202,
          "mean": 158664294.4,
          "min": 155582464,
          "max": 162185216
        }
      }
```

`133270992.8/1000/100 ~= 1,332 bytes` per object.

### DIFF
--- a/packages/client/src/fetchMetadata.test.ts
+++ b/packages/client/src/fetchMetadata.test.ts
@@ -105,6 +105,13 @@ describe("FetchMetadata", () => {
         "primaryKeyApiName": "employeeId",
         "primaryKeyType": "integer",
         "properties": {
+          "class": {
+            "description": "",
+            "displayName": undefined,
+            "multiplicity": false,
+            "nullable": true,
+            "type": "string",
+          },
           "employeeId": {
             "description": undefined,
             "displayName": undefined,
--- a/packages/client/src/object/convertWireToOsdkObjects/ObjectHolder.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/ObjectHolder.ts
@@ -21,7 +21,6 @@ import type { FetchedObjectTypeDefinition } from "../../ontology/OntologyProvide
 import type { DollarAsFn } from "./getDollarAs.js";
 import type { get$link } from "./getDollarLink.js";
 import type {
-  CachedPropertiesRef,
   ClientRef,
   ObjectDefRef,
   RawObject,
@@ -34,12 +33,10 @@ export interface ObjectHolderPrototypeOwnProps {
   readonly [ClientRef]: MinimalClient;
   readonly "$as": DollarAsFn;
   readonly "$link": ReturnType<typeof get$link>;
-  readonly "$updateInternalValues": (newValues: Record<string, any>) => void;
 }
 /** @internal */
 export interface ObjectHolderOwnProperties {
   [RawObject]: OntologyObjectV2;
-  [CachedPropertiesRef]: Map<string | symbol, any>;
 }
 
 /** @internal */
--- a/packages/client/src/object/convertWireToOsdkObjects/ObjectHolder.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/ObjectHolder.ts
@@ -15,7 +15,6 @@
  */
 
 import type { Osdk } from "@osdk/api";
-import type { OntologyObjectV2 } from "@osdk/internal.foundry.core";
 import type { MinimalClient } from "../../MinimalClientContext.js";
 import type { FetchedObjectTypeDefinition } from "../../ontology/OntologyProvider.js";
 import type { DollarAsFn } from "./getDollarAs.js";
@@ -23,25 +22,14 @@ import type { get$link } from "./getDollarLink.js";
 import type {
   ClientRef,
   ObjectDefRef,
-  RawObject,
   UnderlyingOsdkObject,
 } from "./InternalSymbols.js";
 
 /** @internal */
-export interface ObjectHolderPrototypeOwnProps {
+export interface ObjectHolder<Q extends FetchedObjectTypeDefinition> {
+  readonly [UnderlyingOsdkObject]: Osdk<Q>;
   readonly [ObjectDefRef]: FetchedObjectTypeDefinition;
   readonly [ClientRef]: MinimalClient;
   readonly "$as": DollarAsFn;
   readonly "$link": ReturnType<typeof get$link>;
-}
-/** @internal */
-export interface ObjectHolderOwnProperties {
-  [RawObject]: OntologyObjectV2;
-}
-
-/** @internal */
-export interface ObjectHolder<Q extends FetchedObjectTypeDefinition>
-  extends ObjectHolderPrototypeOwnProps, ObjectHolderOwnProperties
-{
-  [UnderlyingOsdkObject]: Osdk<Q>;
 }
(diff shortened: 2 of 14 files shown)


packages/client/src/fetchMetadata.test.ts
```diff
@@ -105,6 +105,13 @@ describe("FetchMetadata", () => {
         "primaryKeyApiName": "employeeId",
         "primaryKeyType": "integer",
         "properties": {
+          "class": {
+            "description": "",
+            "displayName": undefined,
+            "multiplicity": false,
+            "nullable": true,
+            "type": "string",
+          },
```
ericanderson, line 114: We were assuming this property in places but did not actually define it which caused me some problems early on.

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -44,10 +44,10 @@ export function createOsdkInterface<
   underlying: Osdk<Q> & ObjectHolder<Q>,
   interfaceDef: InterfaceMetadata,
 ) {
-  const interfaceHolder: InterfaceHolderOwnProps<Q> = {
-    [UnderlyingOsdkObject]: underlying,
-    [InterfaceDefRef]: interfaceDef,
-  };
+  const interfaceHolder: InterfaceHolderOwnProps<Q> = Object.create(null, {
+    [UnderlyingOsdkObject]: { value: underlying },
```
ericanderson, line 48: I don't want these to be enumerable.

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -67,12 +67,20 @@ function createInterfaceProxyHandler(
...
+        case UnderlyingOsdkObject:
+        case InterfaceDefRef:
+          return Reflect.getOwnPropertyDescriptor(target, p);
+
         case "$primaryKey":
         case "$title":
         case "$objectType":
         case "$rid":
-          return underlying[p] != null
-            ? Reflect.getOwnPropertyDescriptor(underlying, p)
```
ericanderson, line 75: Can't reflect in this version because the underlying is now frozen.

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -123,6 +131,8 @@ function createInterfaceProxyHandler(
         "$primaryKey",
         ...(underlying["$rid"] ? ["$rid"] : []),
         "$title",
+        UnderlyingOsdkObject,
+        InterfaceDefRef,
```
ericanderson, line 135: This is required now that they are no longer configurable/enumerable.

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -140,50 +76,65 @@ export function createOsdkObject<
...
+) {
+  const rawValue = rawObject[p as any];
+  const propDef = objectDef.properties[p as any];
+  if (process.env.NODE_ENV !== "production") {
+    invariant(propDef != null && specialPropertyTypes.has(propDef.type));
+  }
+  {
+    {
+      {
+        {
```
ericanderson, line 130: These just keep the diff simple. We can remove them before merge or in follow up

packages/client/src/object/convertWireToOsdkObjects/ObjectHolder.ts
```diff
@@ -34,12 +33,10 @@ export interface ObjectHolderPrototypeOwnProps {
   readonly [ClientRef]: MinimalClient;
   readonly "$as": DollarAsFn;
   readonly "$link": ReturnType<typeof get$link>;
-  readonly "$updateInternalValues": (newValues: Record<string, any>) => void;
```
ssanjay1, line 37: Ok so now that we're just using the raw object, we don't need this

ericanderson: Yup, this is why I said we have to give up on this functionality to achieve this.

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -140,50 +76,65 @@ export function createOsdkObject<
...
-      },
-      [CachedPropertiesRef]: {
-        value: new Map<string | symbol, any>(),
-        writable: true,
-      },
+  // updates the object's "hidden class/map".
+  Object.defineProperties(rawObj, {
+    [ObjectDefRef]: { value: objectDef, enumerable: false },
+    [ClientRef]: { value: client, enumerable: false },
+    [RawObject]: {
```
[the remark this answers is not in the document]
ericanderson: Nothing. But Im not trying to change ALL the code in the system to demonstrate JUST these objects being raw instead of proxies. The existing contract required both of these things, they just happen to be the same now.

If we go forward, this wouldn't be kept like this, we would clean it all up. But redoing the contract for interfaces would just muddy up this change (and I might as well just remove all proxies at that point since it would be far reaching). 

The goal here was to show that the objects could not be proxies and what that would be like (roughly) so we could discuss taking the next steps.

nihalbhatnagar (approved):

PR #1032, opened 2024-12-04
Starts a benchmarking suite

ericanderson:


benchmarks/tests/primary/src/tests/noop.js
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// @ts-check
+
+import { createMemoryTest, neverOptimizeFunction } from "../helpers.js";
+
+const r = await createMemoryTest(neverOptimizeFunction(async () => {
+  return Math.random();
```
nihalbhatnagar, line 22: What's the use of the noop?

benchmarks/tests/primary/src/helpers.js
```diff
@@ -0,0 +1,132 @@
...
+import { fileURLToPath } from "node:url";
+
+
+export function neverOptimizeFunction(f) {
+    %NeverOptimizeFunction(f);
+    return f;
+}
+
+// dprint-ignore-file
+export function gc(iters=1) {
```
nihalbhatnagar, line 11: Can you explain why more iters would be useful here?

benchmarks/tests/primary/src/tests/createObjects.js
```diff
@@ -0,0 +1,59 @@
...
+  "http://localhost:6723/",
+  "ri.d",
+  () => Promise.resolve("hi"),
+  {},
+);
+
+for (let i = 0; i < WARMUP_REQUESTS; i++) {
+  const page = await client({ type: "object", apiName: "Employee" })
+    .fetchPage();
+  // something to keep the runtime from optimizing away the fetch
```
nihalbhatnagar, line 43: What is it that JS runtime does to optimize?

benchmarks/tests/primary/src/benchmark.js
```diff
@@ -0,0 +1,175 @@
...
+    );
+
+    final.tests[test] = {
+      runs: results,
+      stats: combinedResults,
+    };
+  }
+  multiBar.stop();
+
+  const esBuildStats = await fs.stat(
```
nihalbhatnagar, line 96: Are we seeing any significant differences in build size vs proxy?

benchmarks/tests/primary/src/tests/noop.js
ericanderson: It lets me see how much resources are consumed as a baseline. It was especially helpful for tuning things like warmups and the gc logic

benchmarks/tests/primary/src/helpers.js
ericanderson: Around the web there are varying positions about this. But basically triggering a gc isn't instant from what I can tell. There are also a ton of things happening in the background too. At startup I noticed a lot of variability in how much was collected and this affected the heap diffs. 

Originally I was just running this 10 times in various forms but needed more on startup it seemed. But didn't need the delay at the end as the numbers didn't change much from 1 iteration to 100 later on. 

I assume this is related to internal optimizations of the parse/compile/hit phases as at some point huge amounts of memory get freed up.

benchmarks/tests/primary/src/tests/createObjects.js
ericanderson: If you have a variable that isn't used it will free it earlier. I don't know the depths it goes to. I'm being defensive from experiences in the past

benchmarks/tests/primary/src/benchmark.js
ericanderson: Didn't really check. This is just something I've always wanted.

PR #1035, opened 2024-12-05
RDP's in TS OSDK

nihalbhatnagar:
This PR adds the base for runtime derived properties in the OSDK.

The following are features to be added before the completion of the project but are not covered by this PR
- Calculated properties. The spec for this feature has not been merged in gateway, but we can start exploring the syntax.
- Finalizing names for `selectProperty`, `collectList`, `collectSet`
- Allowing for derived properties whose types are not defined at compile time. 
     - We need to explore how we can let them narrow their types as well as seeing if we can get any type guarantees when the actual object is returned.
     - Our typing breaks if the user has a conditional return with different property types at runtime. For example, `a === b ? integerProperty : stringProperty`. For this PR, enforcing that the user does a simple definition here is fine, but needs to be fixed in the future. Support `any` typed or narrowed properties will have the result of fixing this issue.


packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -254,4 +257,27 @@ export interface ObjectSet<
     listener: ObjectSetListener<Q, P>,
     opts?: ObjectSetListenerOptions<Q, P>,
   ) => { unsubscribe: () => void };
+
+  readonly withProperties: <
+    D extends DeriveClause<Q, any>,
+    P extends ObjectMetadata.Property,
```
nihalbhatnagar, line 263: This generic isn't being captured for some reason, it will show if I type it as DeriveClause<Q,P> within the context of the derive clause, but it isn't being sent to the ExtendedObjectDefinition type

packages/api/src/derivedProperties/DeriveObjectSet.ts
```diff
@@ -0,0 +1,117 @@
...
+
+interface FilterableDeriveObjectSet<
+  Q extends ObjectOrInterfaceDefinition,
+> {
+  readonly where: (
+    clause: WhereClause<Q>,
+  ) => this;
+}
+
+type CollectAggregations = "collectSet" | "collectList";
```
nihalbhatnagar, line 55: Parroting gateway names for now. This probably needs a review into `valueSet` or `valueList`, but the gateway names should be fine for right now

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -436,6 +437,260 @@ describe("ObjectSet", () => {
...
+        "derivedPropertyName": (base) => {
+          // @ts-expect-error
+          base.pivotTo("lead").selectProperty("notAProperty");
+
+          return base.pivotTo("lead").selectProperty("employeeStatus");
+        },
+      });
+    });
+
+    it("propagates derived property type to future object set operations with correct types", () => {
```
nihalbhatnagar, line 528: This test and others with no clause are just for us to ensure that we're typing things correctly.

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,119 @@
...
+  ObjectOrInterfaceDefinition,
+  PropertyKeys,
+} from "../ontology/ObjectOrInterface.js";
+import type {
+  CompileTimeMetadata,
+  PropertyDef,
+} from "../ontology/ObjectTypeDefinition.js";
+import type { LinkedType, LinkNames } from "../util/LinkUtils.js";
+import type { WithPropertyDefinition } from "./WithPropertiesClause.js";
+
```
nihalbhatnagar, line 29: I use inheritance here to make sure we enforce what methods are available,

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,119 @@
...
+export interface WithPropertyObjectSet<Q extends ObjectOrInterfaceDefinition>
+  extends
+    BaseWithPropertyObjectSet<Q>,
+    AggregatableWithPropertyObjectSet<Q>,
+    SingleLinkWithPropertyObjectSet<Q>
+{}
+
+export interface BaseWithPropertyObjectSet<
```
nihalbhatnagar, line 37: This object set only has filter and pivot as aggregations and selectProperty is not available on the base object set.

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,119 @@
...
+        >
+      : P extends "approximateDistinct" | "exactDistinct" | "$count"
+        ? PropertyDef<"integer">
+      : PropertyDef<"double">
+      : V extends "$count" ? PropertyDef<"integer">
+      : never
+  >;
+}
+
+interface SingleLinkWithPropertyObjectSet<
```
nihalbhatnagar, line 105: `BaseObjectSet` extends `SingleLinkWithPropertyObjectSet`. This makes the `selectProperty` method available only for chains of single links, which correlates with the backend. The `pivotTo` operation in `BaseObjectSet` enforces that if we pivot to a many link, we from then on use `ManyLinkWithPropertyObjectSet` which does not have `selectProperty` available and only returns further ManyLink object sets.

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -254,4 +261,36 @@ export interface ObjectSet<
...
+    >
+  >;
 }
+
+type WithPropertiesObjectDefinition<
+  K extends ObjectOrInterfaceDefinition,
+  D extends WithPropertiesClause<K>,
+> = {
+  __DefinitionMetadata: {
+    properties: {
```
nihalbhatnagar, line 282: This is all it seems like we need to propagate the types through all further object set operations. The FLUP here is to make the type of the whole object set extractable and definable so that a user can pass around object sets that are correctly types and do not have to be inferred.

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
```diff
@@ -0,0 +1,125 @@
...
+  objectSet: WireObjectSet,
+  definitionMap: Map<any, DerivedPropertyDefinition>,
+): WithPropertyObjectSet<Q> {
+  const base: WithPropertyObjectSet<Q> = {
+    pivotTo: (link) => {
+      return createWithPropertiesObjectSet(objectType, {
+        type: "searchAround",
+        objectSet,
+        link,
+      }, definitionMap) as any;
```
ssanjay1, line 45: is `as any` the best we can do here?

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -436,6 +437,260 @@ describe("ObjectSet", () => {
...
+
+          return base.pivotTo("lead").selectProperty("employeeId");
+        },
+      });
+    });
+
+    // Executed code fails since we're providing bad strings to the function
+    it.fails("correctly narrows types of aggregate function", () => {
+      client(Employee).withProperties({
+        "derivedPropertyName": (base) => {
```
ssanjay1, line 472: me being annoying nit: but for completeness, lets add an empty string arg as well and make sure that errors

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -436,6 +437,260 @@ describe("ObjectSet", () => {
...
+      client(Employee).withProperties({
+        "derivedPropertyName": (base) =>
+          base.pivotTo("lead").selectProperty("employeeId"),
+      }).where({ "employeeId": { "$eq": 2 } });
+    });
+
+    it("allows fetching derived properties with correctly typed Osdk.Instance types", async () => {
+      const objectWithRdp = await client(Employee).withProperties({
+        "derivedPropertyName": (base) =>
+          base.pivotTo("lead").selectProperty("employeeId"),
```
ssanjay1, line 590: Can we add a test to verify things don't break if we select a property that has all undefined values?

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -436,6 +437,260 @@ describe("ObjectSet", () => {
     });
   });
 
+  describe("Derived Properties Object Set", () => {
+    it("does not allow aggregate or selectProperty before a link type is selected", () => {
+      client(Employee).withProperties({
+        "derivedPropertyName": (base) =>
```
ssanjay1, line 443: Can we also add a test to make sure that within a WithPropertiesClause you can't just return something like `base.pivotTo("lead")` without doing some sort of selection or aggregation?

ssanjay1: Apologies in advance if I missed this somewhere, but don't seem to see that being tested anywhere yet

ssanjay1 (commented):
Overall looking good! Left some minor comments

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,119 @@
...
+> extends AggregatableWithPropertyObjectSet<Q>, BaseWithPropertyObjectSet<Q> {
+  readonly selectProperty: <R extends PropertyKeys<Q>>(
+    propertyName: R,
+  ) => WithPropertyDefinition<CompileTimeMetadata<Q>["properties"][R]>;
+}
+
+interface ManyLinkWithPropertyObjectSet<
+  Q extends ObjectOrInterfaceDefinition,
+> extends AggregatableWithPropertyObjectSet<Q> {
+  readonly pivotTo: <L extends LinkNames<Q>>(
```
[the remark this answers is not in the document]
nihalbhatnagar: Yeah that's correct, which makes sense because for this to work you have to basically be navigating to exactly __one__ object

packages/api/src/derivedProperties/WithPropertiesClause.ts
```diff
@@ -0,0 +1,39 @@
...
+import type {
+  ObjectMetadata,
+  PropertyDef,
+} from "../ontology/ObjectTypeDefinition.js";
+import type { BaseWithPropertyObjectSet } from "./WithPropertyObjectSet.js";
+
+export type WithPropertyDefinition<
+  T extends ObjectMetadata.Property,
+> = {
+  definitionId: unknown;
```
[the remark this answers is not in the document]
nihalbhatnagar: We can type this down to a string, I was wondering if we should type this as unknown to tell the user -> dont touch this

packages/api/src/aggregate/AggregatableKeys.ts
```diff
@@ -36,14 +36,19 @@ type AGG_FOR_TYPE<T> = number extends T ? NumericAggregateOption
 
 export type ValidAggregationKeys<
   Q extends ObjectOrInterfaceDefinition,
+  StringOptions extends string = StringAggregateOption,
```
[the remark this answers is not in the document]
nihalbhatnagar: Hmm yeah this does make sense to me, will refactor

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
nihalbhatnagar: Looks like we don't need it, this was left over from dev

packages/client/src/objectSet/ObjectSet.test.ts
nihalbhatnagar: What do you mean?

packages/client/src/objectSet/ObjectSet.test.ts
nihalbhatnagar: This should be guarded by the type of the lambda, but I will add an explicit test for this behavior

packages/client/src/objectSet/ObjectSet.test.ts
ssanjay1: Basically if a prop returns an undefined value

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,119 @@
...
+  | "min"
+  | "max"
+  | "sum"
+  | "avg"
+  | BaseAggregations
+  | CollectAggregations;
+
+interface AggregatableWithPropertyObjectSet<
+  Q extends ObjectOrInterfaceDefinition,
+> extends FilterableDeriveObjectSet<Q> {
```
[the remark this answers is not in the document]
nihalbhatnagar: Fixed! Agreed this makes it more readable

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,119 @@
...
+
+interface FilterableDeriveObjectSet<
+  Q extends ObjectOrInterfaceDefinition,
+> {
+  readonly where: (
+    clause: WhereClause<Q>,
+  ) => this;
+}
+
+type CollectAggregations = "collectSet" | "collectList";
```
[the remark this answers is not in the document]
nihalbhatnagar: Debating between the above Derive, ow WithPropertyBaseAggregations, or BaseWithPropertyAggregations. We aren't using "derive" as a term, but it is much cleaner than the other 2. Number 3 is kind of confusing, but number 2 would mean they al start the same which I don't want. Stuck with number 1 for conciseness

packages/api/src/aggregate/AggregatableKeys.ts
```diff
@@ -30,25 +34,24 @@ export type NumericAggregateOption =
...
+  ? U extends true ? NumericAggregateOption : NumericDeriveAggregateOption
+  : string extends T
+    ? U extends true ? StringAggregateOption : StringDeriveAggregateOption
   : never;
 
 export type ValidAggregationKeys<
   Q extends ObjectOrInterfaceDefinition,
-  StringOptions extends string = StringAggregateOption,
-  NumericOptions extends string = NumericAggregateOption,
+  R extends "aggregate" | "withPropertiesAggregate" = "aggregate",
```
nihalbhatnagar, line 45: @ssanjay1 Is this what you had in mind?

ssanjay1: Yea that works

packages/api/src/aggregate/AggregatableKeys.ts
```diff
@@ -30,19 +34,23 @@ export type NumericAggregateOption =
   | "approximateDistinct"
   | "exactDistinct";
 
-type AGG_FOR_TYPE<T> = number extends T ? NumericAggregateOption
-  : string extends T ? StringAggregateOption
+type AGG_FOR_TYPE<T, U extends boolean> = number extends T
```
ssanjay1, line 37: nit: rename `U` so its clear what that boolean generic is for

packages/api/src/aggregate/AggregatableKeys.ts
ssanjay1: Though maybe keeping parity with the other names it should be `aggregate | derivedPropertiesAggregate` or something along those lines, basically including the word `derive`

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,121 @@
...
+  PropertyKeys,
+} from "../ontology/ObjectOrInterface.js";
+import type {
+  CompileTimeMetadata,
+  PropertyDef,
+} from "../ontology/ObjectTypeDefinition.js";
+import type { LinkedType, LinkNames } from "../util/LinkUtils.js";
+import type { WithPropertyDefinition } from "./WithPropertiesClause.js";
+
+export interface WithPropertyObjectSet<Q extends ObjectOrInterfaceDefinition>
```
ssanjay1, line 30: Lets be consistent with naming, see some `withProperty` and `derive` here.

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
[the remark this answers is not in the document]
ericanderson: We should probably document that here too.

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
```diff
@@ -0,0 +1,124 @@
...
+  | CollectWithPropAggregations;
+
+export type NumericWithPropAggregateOption =
+  | "min"
+  | "max"
+  | "sum"
+  | "avg"
+  | BaseWithPropAggregations
+  | CollectWithPropAggregations;
+
```
ericanderson, line 72: If I am thinking about whats ideal we need to add some type tests. I'm okay if its in a follow up

ericanderson (commented):
Left a few notes we should fix up. If we don't do the IR I think we should fix up the types as I mention to not carry so much complexity around. 

I am still undecided on the IR or not. I get not wanting to create a complex language (or reproduce what the backend has) but I also feel like there are no surprises there either. IDK, this is a tough one.

nihalbhatnagar:
I updated the types in the API to improve the hover readability. This actually unlocked the ability to super easily have users define a type for their object set with added properties, so now they can write functions that expect certain property types to be on object sets.

This is what the hover over an object set looks like now: 
image

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -485,6 +487,349 @@ describe("ObjectSet", () => {
...
+          Employee,
+          { "hello": PropertyDef<"string", "non-nullable", "single"> }
+        >;
+      });
+
+      it("Allows specifying the type of a variable as the object set with added properties", () => {
+        let objectSet: ObjectSetWithProperties<
+          Employee,
+          { "hello": PropertyDef<"string", "nullable", "single"> }
+        >;
```
nihalbhatnagar, line 750: I'm using the already exported PropertyDef type to have the users define their types. This is an example of how a user would type their function or anything they want

packages/api/src/derivedProperties/WithPropertyObjectSet.ts
nihalbhatnagar: I will follow this up, this PR is a little bloated with tests

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -485,6 +488,285 @@ describe("ObjectSet", () => {
     });
   });
 
+  describe("Derived Properties Object Set", () => {
```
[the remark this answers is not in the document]
nihalbhatnagar: I added some expectTypeOf entries for tests I had already written and my newer tests that use the new `ObjectSetWithProperties` type should also cover a large part of the output testing

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -254,4 +261,45 @@ export interface ObjectSet<
...
+          P["type"],
+          "nullable",
+          P["multiplicity"] extends true ? "array" : "single"
+        >
+        : never;
+    }
+  >;
 }
+
+export type ObjectSetWithProperties<
```
nihalbhatnagar, line 283: This name needs to be discussed. It is very similar to the `WithPropertiesObjectSet` type that we already export. I like this name for this type, so I think we should likely rename the `WithPropertiesObjectSet` to `PropertyDefinitionObjectSet`

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
```diff
@@ -0,0 +1,131 @@
...
+            };
+            break;
+          }
+        default:
+          invariant(
+            false,
+            "Invalid aggregation operation " + aggregationOperation,
+          );
+      }
+      definitionMap.set(definitionId.toString(), {
```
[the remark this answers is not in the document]
nihalbhatnagar: I like this idea for how we'd do the definitionMap, I'll add it to this PR. I'd try and also explore a global cache here that would allow the user to store off the definition in another variable but we'd probably run into typing issues there when we introduce calculated properties that can reference other derived properties. I'll also explore the IR solution here, where I think the biggest benefit is reusability of the type. However, I do think it's quite complicated and I'd like to also see what the demand for this would be. I'll do the IR exploration and other issues in a follow up.

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -512,6 +514,349 @@ describe("ObjectSet", () => {
...
+          "hello": (base) => base.pivotTo("lead").selectProperty("employeeId"),
+        });
+      });
+
+      it("Allows extracting the type of an already constructed object set", () => {
+        const objectSet = client(Employee).withProperties({
+          "hello": (base) => base.pivotTo("lead").selectProperty("fullName"),
+        });
+
+        type extractedType = typeof objectSet;
```
nihalbhatnagar, line 807: Heres how you extract the type from an object set with derived properties on it

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -265,33 +266,39 @@ export interface ObjectSet<
...
-    WithPropertiesObjectDefinition<
-      Q,
-      D
-    >
+  ) => ObjectSetWithProperties<
+    Q,
+    {
+      [K in keyof D]: D[K] extends
+        (baseObjectSet: any) => WithPropertyDefinition<infer P> ? PropertyDef<
+          P["type"],
```
nihalbhatnagar, line 274: The user facing type expects a PropertyDef instead of an ObjectMetadata.Property. I chose to localize it here instead of 2 places when extracting the type from the definition in 2 places

nihalbhatnagar:
Updated Osdk.Instance types look like this:
image
image

nihalbhatnagar:
Updated Builder name looks like this: 
image

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -79,8 +84,15 @@ export interface MinimalObjectSet<Q extends ObjectOrInterfaceDefinition>
...
   ) => Promise<
     PageResult<
-      PropertyKeys<Q> extends L ? Osdk.Instance<Q, ExtractOptions<R, S>>
-        : Osdk.Instance<Q, ExtractOptions<R, S>, L>
+      Osdk.Instance<
+        B,
+        ExtractOptions<R, S>,
+        PropertyKeys<B> extends L ? PropertyKeys<B>
+          : Extract<L, PropertyKeys<B>>,
+        Y extends Record<string, SimplePropertyDef>
```
ssanjay1, line 92: nit: should we possible pull any of this into helper types considering its used throughout this file?

packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -205,12 +205,12 @@ describe("convertWireToOsdkObjects", () => {
       "userAgent",
     );
 
-    let object = {
+    const object = {
```
ssanjay1, line 208: This is just ESlint stuff right?

packages/api/src/OsdkObjectFrom.ts
```diff
@@ -185,12 +187,16 @@ export namespace Osdk {
...
     OPTIONS extends never | "$rid" = never,
     P extends PropertyKeys<Q> = PropertyKeys<Q>,
+    R extends Record<string, SimplePropertyDef> = never,
   > =
     & OsdkBase<Q>
     & Pick<
       CompileTimeMetadata<Q>["props"],
-      GetPropsKeys<Q, P>
+      // If there aren't any additional properties, then we want GetPropsKeys to default to PropertyKeys<Q>
+      GetPropsKeys<Q, P, [R] extends [never] ? false : true>
```
ssanjay1, line 196: Mmm, I'm a bit confused and probably missing something here, what if they had downselected the original object props?

nihalbhatnagar: Yeah thats correct, this is to make sure passing in never still defaults to PropertyKeys<Q> but also allows you to pass in a never if you choose to add an RDP.

packages/client/src/object/convertWireToOsdkObjects.test.ts
nihalbhatnagar: I think so, I don't remember changing this

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -778,125 +778,6 @@ describe("ObjectSet", () => {
...
-          ObjectSet<Employee, {
-            "ouch": "geotimeSeriesReference" | undefined;
-            "hello": "string" | undefined;
-          }>
-        >();
-
-        expectTypeOf(order2).toEqualTypeOf<typeof order1>();
-        expectTypeOf(order1).toEqualTypeOf<typeof order2>();
-      });
-    });
```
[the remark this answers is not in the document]
nihalbhatnagar: I felt that these tests were a repeat of the type tests we had present in the tests located in the api package now, so I wanted to cut down on the clutter

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -114,18 +125,15 @@ interface SingleLinkWithPropertyObjectSet<
 }
 
 /*
- * The ManyLinkWithPropertyObjectSet overrides the pivotTo operation because once we traverse a single link,
+ * The DefaultBuilder overrides the pivotTo operation because once we traverse a single link,
```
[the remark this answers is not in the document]
nihalbhatnagar: Hmm yeah I wrote the name more with a "coming off the tounge nicely" rather than being super informative because I don't imagine any of our users will care about the type of the RDP object set. If the goal is for more readable code, I can add more comments?

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -79,8 +84,15 @@ export interface MinimalObjectSet<Q extends ObjectOrInterfaceDefinition>
     args?: FetchPageArgs<Q, L, R, A, S>,
   ) => Promise<
     PageResult<
-      PropertyKeys<Q> extends L ? Osdk.Instance<Q, ExtractOptions<R, S>>
```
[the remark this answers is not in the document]
nihalbhatnagar: Hmm I may be wrong here but I wrote this so that it defaults to PropertyKeys\<employee\> if it's the full property set, so this should be matching the behavior of the top right?

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -179,14 +205,19 @@ export interface ObjectSet<
...
 
 interface ObjectSetPrime<
   Q extends ObjectOrInterfaceDefinition = any,
-> extends MinimalObjectSet<Q> {
+  B extends ObjectOrInterfaceDefinition = Q,
+  Y extends ObjectSet<B, any> | Record<string, SimplePropertyDef> = ObjectSet<
+    B,
+    any
+  >,
+> extends MinimalObjectSet<Q, B, Y> {
```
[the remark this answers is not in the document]
nihalbhatnagar: Hmm so maybe we should consider merging ObjectSet.fns back into here? Also think we need to discuss MinimalObjectSet because thats really only used for interfaces but they should have feature parity at this point? Maybe we need to just extract out intersect/union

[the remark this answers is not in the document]
ericanderson: Now that Im in here playing with types, one thing that `ObjectSetPrime` got me was knowing for sure what the second param was (removing the chance that its the value from the definition) which I think complicated your types a bit.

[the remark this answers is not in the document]
ericanderson: I think splitting each function out might be the best move here because the different between Minimal vs not vs interface etc is really confusing. Experimenting

ssanjay1 (approved):
Naming stuff seems fine to me now, approving to unblock

PR #1036, opened 2024-12-05
Second round of the benchmarks

ericanderson:
# One benchmark per package

After adding a few benchmarks, having one package got annoying because some of them took a long time and you may not always want to run all of them.

This breaks everything up into independent libraries / packages so you can iterate better.

# Serial and Parallel
This also introduces the concept of serial and parallel benchmarks. Be sure to `turbo benchmark-serial --concurrency 1` when you run those. Sadly we can't tell turbo that certain tasks are serial.

In a future update we can potentially fix this.

# Bundle Benchmarks

I want to be able to see what the output size of our work is across builds. I've started this with `createClient` and `createPublicOauthClient`.

# FUTURE

It would be nice to have a comment like `/runBenchmarks` trigger a benchmark build. This build could potentially run on main and in the repo to compare the results.

It would also be good if we start persisting the bundle file sizes for every build so we can track that.


benchmarks/bundles/simple/src/createPublicOauthClient.js
```diff
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-export interface Result {
-  type: "result";
-  time: number;
-  heapUsed: number;
```
nihalbhatnagar, line 20: Why this is needed?

benchmarks/bundles/simple/src/createPublicOauthClient.js
```diff
@@ -14,9 +14,6 @@
...
-  rss: number;
-}
+import { createPublicOauthClient } from "@osdk/oauth";
+
+export { createPublicOauthClient };
```
nihalbhatnagar, line 19: Why is this needed?

nihalbhatnagar (approved):

PR #1046, opened 2024-12-06
Use workspace version for widget-client-react peer dependency

rosscourt:
This peer dependency being pinned to `^0.1.0` prevents taking equal versions for both `@osdk/widget-client-react.unstable` and `@osdk/widget-client.unstable` as is desired.


rosscourt:
Superseded by #1050

PR #1047, opened 2024-12-07
Add structs read support

ssanjay1:


packages/client/src/object/aggregate.test.ts
```diff
@@ -171,13 +186,7 @@ describe("aggregate", () => {
...
     >(false); // subSelect should hide unused keys
 
-    const grouped = await aggregate(
-      clientCtx,
-      objectTypeWithAllPropertyTypes,
-      {
-        type: "base",
-        objectType: "ToDo",
-      },
+    const grouped = await client(objectTypeWithAllPropertyTypes).aggregate(
```
ssanjay1, line 189: For some reason type tests were having issues when using the aggregate function directly, but were all good when using the user exposed aggregate function. This should be fine since the user facing stuff works, but just flagging

packages/generator-converters/src/wirePropertyV2ToSdkPrimaryKeyTypeDefinition.ts
```diff
@@ -46,6 +46,8 @@ export function wirePropertyV2ToSdkPrimaryKeyTypeDefinition(
     case "float":
     case "geotimeSeriesReference":
     case "mediaReference":
+    case "struct":
```
ssanjay1, line 49: I know we want to ignore these property types instead of throwing errors for unsupported, but saving as a separate PR since structs is higher prio.

packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
```diff
@@ -102,10 +108,23 @@ function objectPropertyTypeToSdkPropertyDefinition(
       } else if (propertyType.itemType?.type === "double") {
         return "numericTimeseries";
       } else return "sensorTimeseries";
+    case "struct": {
+      return propertyType.structFieldTypes.reduce(
+        (structMap: Record<string, SimpleWirePropertyTypes>, structField) => {
+          structMap[structField.apiName] =
+            objectPropertyTypeToSdkPropertyDefinition(
+              structField.dataType,
+            ) as SimpleWirePropertyTypes;
```
ssanjay1, line 117: This is fine because right now, structs can only be field -> simple property type, cannot have nested structs yet in OSS

packages/maker/src/api/defineObject.ts
```diff
@@ -122,7 +123,15 @@ function convertType(
       return {
         type: "timestamp",
       };
-
+    case typeof t.type === "object": {
+      return {
+        type: "struct",
+        structFieldTypes: Object.entries(t.type).map(([apiName, dataType]) => ({
+          apiName,
+          dataType: { type: dataType } as ObjectPropertyType,
```
ssanjay1, line 131: This cast should also be fine because the maker package only uses this for primary key types, which are limited to a select few primitive types

packages/api/src/mapping/PropertyValueMapping.ts
```diff
@@ -71,3 +78,9 @@ export interface PropertyValueClientToWire {
   sensorTimeseries: TimeSeriesProperty<string | number>;
   geotimeSeriesReference: GeotimeSeriesProperty<GeoJSON.Point>;
 }
+export type getWirePropertyValueFromClient<
```
ericanderson, line 81: I find this type being lower case to be a bit confusing

packages/api/src/ontology/WirePropertyTypes.ts
```diff
@@ -15,6 +15,10 @@
  */
 
 export type WirePropertyTypes =
```
ericanderson, line 17: Should this be `WireAndStructPropertyTypes`? That way we don't break `WirePropertyTypes`?

ericanderson (commented):
I think we should try to break less

packages/api/src/mapping/PropertyValueMapping.ts
```diff
@@ -71,3 +78,9 @@ export interface PropertyValueClientToWire {
...
+  T extends
+    | keyof PropertyValueClientToWire
+    | Record<string, keyof PropertyValueClientToWire>,
+> = T extends keyof PropertyValueClientToWire ? PropertyValueClientToWire[T]
+  : T;
```
nihalbhatnagar, line 86: I'm being a bit silly but I'm not understanding why this type works. Why doesn't it need to be nested to be kind of like `T extends keyof PropertyValueClientToWire ? PropertyValueClientToWire[T]
  : {[K in T]: PropertyValueClientToWire[K]};`

packages/client/src/object/aggregate.test.ts
ssanjay1: For some reason, the aggregate function was not capturing clause types correctly and everything was still in generic form (e.g. like it wasn't getting the concrete information passed in `$select: {myProp:asc}`

packages/client/src/object/aggregate.test.ts
```diff
@@ -37,6 +40,9 @@ import {
   type Mock,
   vi,
 } from "vitest";
+import type { getWirePropertyValueFromClient } from "../../../api/build/esm/mapping/PropertyValueMapping.js";
```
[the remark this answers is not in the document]
ssanjay1: Ah good catch

packages/api/src/ontology/WirePropertyTypes.ts
ssanjay1: Talked a bit on slack, just to clarify here as well, the reason you think this is a break is because now instead of just a list of strings, its `SimpleWirePropertyTypes` and the record type? Because technically this boils down to the same thing and is just an additive change, but if someone is using this type, could be confusing since now it referencs something called `SimpleWirePropertyTypes`?

packages/api/src/mapping/PropertyValueMapping.ts
ssanjay1: Ah no I agree, bad convention on my part

packages/api/src/mapping/PropertyValueMapping.ts
ssanjay1: mmm actually, this is a good point. We should still remap the inner types of the struct. We mainly use this type for agg results which isn't supported for structs yet, but I added some tests to make sure this works as intended

nihalbhatnagar (approved):

PR #1051, opened 2024-12-09
Remove crypto from OSDK

nihalbhatnagar:
React-native has issues with dependencies on crypto. Removes our polyfill and usage in object set watcher. Also fixes a small bug that can cause issues with unsubscribing with multiple subscriptions

### DIFF
--- a/packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
+++ b/packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
@@ -141,6 +141,11 @@ export class ObjectSetListenerWebsocket {
     string,
     Subscription<any, any>
   >();
+
+  #endedSubscriptions = new Set<
+    string
+  >();
+
   #maybeDisconnectTimeout: ReturnType<typeof setTimeout> | undefined;
 
   // DO NOT CONSTRUCT DIRECTLY. ONLY EXPOSED AS A TESTING SEAM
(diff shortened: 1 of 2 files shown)


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -275,18 +274,7 @@ export class ObjectSetListenerWebsocket {
     // so we filter those out.
     const readySubs = [...this.#subscriptions.values()].filter(isReady);
 
-    if (readySubs.length === 0) {
```
nihalbhatnagar, line 278: This is not the behavior we want. We want to update that we have no more subscriptions so that we don't unnecessarily take up space over the wire

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -572,6 +565,7 @@ export class ObjectSetListenerWebsocket {
 
   #handleMessage_subscriptionClosed(payload: SubscriptionClosed) {
     const sub = this.#subscriptions.get(payload.id);
+    if (sub == null && this.#endedSubscriptions.has(payload.id)) return;
```
nihalbhatnagar, line 568: Subscription is deleted in unsubscribe

ericanderson (approved):

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -188,7 +188,7 @@ describe("ObjectSetListenerWebsocket", async () => {
           listener,
         );
 
-        subReq1 = await expectSingleSubscribeMessage(ws);
+        subReq1 = await expectSubscribeMessages(ws, 2);
```
nihalbhatnagar, line 191: Empty subscribe messages are now sent

ericanderson (approved):

PR #1054, opened 2024-12-11
Split foundry-config-json package

tzyl:
Based on top of #1052 

In order to use `foundry.config.json` and auto version strategies in both `@osdk/cli` and `@osdk/widget-manifest-vite-plugin`, split out `@osdk/foundry-config-json`

PR #1055, opened 2024-12-11
Use foundry-config-json in widget-manifest-vite-plugin

tzyl:
Based on top of #1054

Replace manual package json auto version strategy and foundry config json parsing in `@osdk/widget-manifest-vite-plugin` with the shared `@osdk/foundry-config-json` package. This means we can use the git describe strategy now for populating the manifest at build time.


packages/example-generator/src/run.ts
```diff
@@ -185,7 +188,12 @@ async function checkExamples(
   resolvedOutput: string,
   tmpDir: tmp.DirResult,
 ): Promise<void> {
-  for (const [template, sdkVersion] of templatesWithSdkVersions(TEMPLATES)) {
+  for (
+    const [template, sdkVersion] of [
+      ...templatesWithSdkVersions(TEMPLATES),
+      ...templatesWithSdkVersions(WIDGET_TEMPLATES),
```
tzyl, line 194: There were a bunch of places here not applying properly to the widget templates so weren't actually beingg enforced at CI time

packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -159,7 +154,7 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
               body += readResult;
             }
           });
-          req.on("end", () => {
+          req.on("end", async () => {
```
tzyl, line 157: Not sure if this works hoping you'll take over for dev mode

packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -508,7 +488,7 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
               }))
               : [],
             rid: entrypointFileIdToConfigMap[chunk.facadeModuleId].rid,
-            version: packageJsonFile.version ?? "0.0.0",
+            version: widgetVersion,
```
tzyl, line 491: This seems to be fine, running build on the examples package does generate the manifest in the dist folder

rosscourt (approved):

rosscourt (approved):

PR #1059, opened 2024-12-12
CLI unstable widgetset commands

tzyl:
Unstable widgetset commands until we have gateway APIs:

```
@osdk/cli unstable widgetset deploy
@osdk/cli unstable widgetset version list
@osdk/cli unstable widgetset version info
@osdk/cli unstable widgetset version delete
```

This is mostly the same as the site commands


packages/cli/src/commands/widgetset/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const command: CommandModule<CliCommonArgs, CommonWidgetSetArgs> = {
+  command: "widgetset",
+  describe: "Manage your widget set",
+  builder: async (argv) => {
+    const config = await configLoader("widgetSet");
+    const widgetSet = config?.foundryConfig.widgetSet.rid;
+    const foundryUrl = config?.foundryConfig.foundryUrl;
+    return argv.options({
+      widgetSet: {
```
tzyl, line 35: Slight rename from `widgetSet.rid` -> `--widgetSet` to be consistent with site commands which map `site.application` -> `--application`. The foundry.config.json field staying as `rid` seems fine as it's obvious from the context what it refers to

packages/cli/src/commands/widgetset/deploy/index.ts
```diff
@@ -0,0 +1,76 @@
...
+import configLoader from "../../../util/configLoader.js";
+import type { CommonWidgetSetArgs } from "../CommonWidgetSetArgs.js";
+import { logWidgetSetDeployCommandConfigFileOverride } from "./logWidgetSetDeployCommandConfigFileOverride.js";
+import type { WidgetSetDeployArgs } from "./WidgetSetDeployArgs.js";
+
+const command: CommandModule<
+  CommonWidgetSetArgs,
+  WidgetSetDeployArgs
+> = {
+  command: "deploy",
```
tzyl, line 30: There are some key differences from site deploy here:
- Version cannot be specified as it's tied to the version built into the manifest
- Project will have to be rebuilt (assuming using vite plugin) after updating git tags or package.json depending on auto version strategy
- The upload target for the sites files is either the widget set itself or optionally a code repository that has been authorized to publish the widget set

packages/cli/src/net/widget-registry/deleteSiteVersion.mts
```diff
@@ -0,0 +1,37 @@
...
+import type { WidgetSetRid } from "../WidgetSetRid.js";
+
+export async function deleteSiteVersion(
+  ctx: InternalClientContext,
+  repositoryRid: WidgetSetRid | StemmaRepositoryRid,
+  version: string,
+): Promise<void> {
+  const fetch = createFetch(ctx.tokenProvider);
+  const url =
+    `${ctx.foundryUrl}/artifacts/api/repositories/${repositoryRid}/contents/release/siteasset/versions/${version}`;
```
tzyl, line 29: These are not widget registry technically but only used here for now and should be replaced by gateway as that will also be mesh comaptible as there will only be one foundry url for the gateway rather than a different one for each service

PR #1063, opened 2024-12-13
Handle Vite plugin injections in Custom Widget Vite plugin

rosscourt:
Vite plugins are able to inject scripts & javascript into the target HTML entrypoint file, which in the case of the React plugin inserts some code as an inline script tag which enables hot module refresh ([source](https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/src/fast-refresh.ts#L23)). As we are server-side rendering our HTML file, we need to capture these injections and get them onto the page before other scripts using a different mechanism.

Vite exposes access to other plugins from our own, and on these plugins we are able to call the `transformIndexHtml` hook (if present). This hook returns descriptors of the scripts that would be injected. The actual return format is more complex than the cases I'm selecting from here, so we may need to expand this in the future. Its well typed & structured but I don't have a sense of how widely used this mechanism is yet, and so want to cap the effort.

This has led me to question whether Vite is the right tool for us to focus on here, given that Vite expects full control of the HTML file whereas we expect to manage it (https://vite.dev/guide/#index-html-and-project-root). We should push forwards with what we have now, but it's worth bearing this in mind and potentially experimenting with out build tools.

Also started to break this code up, which I'll continue to do in FLUPs. Also fixed the `set-settings` call and introduced the `enable` call based on latest API changes.

------

#### Example HTML file compiled by Vite
```html
<!doctype html>
<html lang="en">
  <head>
    <script type="module">
        import RefreshRuntime from "/@react-refresh"
        RefreshRuntime.injectIntoGlobalHook(window)
        window.$RefreshReg$ = () => {}
        window.$RefreshSig$ = () => (type) => type
        window.__vite_plugin_react_preamble_installed__ = true
    </script>
    <script type="module" src="/@vite/client"></script>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Widget: Ontology SDK + React</title>
  </head>
  <body>
    <script type="module" src="/src/main.tsx"></script>
  </body>
</html>
```

#### The contents of the auto-generated `vite-injections.js` in this case:
```javascript
import RefreshRuntime from "/@react-refresh"
RefreshRuntime.injectIntoGlobalHook(window)
window.$RefreshReg$ = () => {}
window.$RefreshSig$ = () => (type) => type
window.__vite_plugin_react_preamble_installed__ = true
```

### DIFF
--- a/packages/widget.vite-plugin/src/plugin.ts
+++ b/packages/widget.vite-plugin/src/plugin.ts
@@ -28,8 +28,15 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 import color from "picocolors";
 import sirv from "sirv";
-import type { Plugin, ResolvedConfig, ViteDevServer } from "vite";
-import { PALANTIR_PATH, SETUP_PATH } from "./constants.js";
+import type {
+  HtmlTagDescriptor,
+  IndexHtmlTransformHook,
+  IndexHtmlTransformResult,
+  Plugin,
+  ResolvedConfig,
+  ViteDevServer,
+} from "vite";
+import { PALANTIR_PATH, SETUP_PATH, VITE_INJECTIONS } from "./constants.js";
 
 export const DIR_DIST = typeof __dirname !== "undefined"
   ? __dirname
@@ -141,6 +148,21 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
         },
       );
 
+      server.middlewares.use(
+        `${server.config.base ?? "/"}${VITE_INJECTIONS}`,
+        async (req, res) => {
+          if (devServer == null) {
+            res.statusCode = 500;
+            res.statusMessage = "Vite server not found.";
+            res.end();
+            return;
+          }
+          res.setHeader("Access-Control-Allow-Origin", "*");
+          res.setHeader("Content-Type", "application/javascript");
+          res.end(await extractInjectedScripts(devServer));
+        },
+      );
+
       // Called by the setup page to start dev mode in Foundry, which queries the appropriate service on the Foundry instance configured in foundry.config.json
       server.middlewares.use(
         `${server.config.base ?? "/"}${PALANTIR_PATH}/finish`,
@@ -159,7 +181,7 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
               body += readResult;
             }
           });
-          req.on("end", () => {
+          req.on("end", async () => {
             if (body.length === 0) {
               res.statusCode = 400;
               res.statusMessage =
@@ -216,9 +238,9 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
               return;
             }
 
-            let url: URL;
+            let foundryUrl: URL;
             try {
-              url = new URL(foundryConfig.foundryUrl);
+              foundryUrl = new URL(foundryConfig.foundryUrl);
             } catch (error) {
               res.statusCode = 500;
               res.statusMessage =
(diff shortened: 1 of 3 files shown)


packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -572,3 +589,107 @@ function extractWidgetConfig(objectExpression: ObjectExpression) {
...
+  const widgetDevModeSettings = {
+    entrypointJs: [
+      `/${VITE_INJECTIONS}`,
+      "/@vite/client",
+      ...entrypointToJsSourceFileMap[entrypointFileName],
+    ].map((file) => ({
+      filePath: `${localhostUrl}${file}`,
+      scriptType: { type: "module", module: {} },
+    })),
+    entrypointCss: [],
```
tzyl, line 609: Is CSS always injected by JS in dev mode?

packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -572,3 +589,107 @@ function extractWidgetConfig(objectExpression: ObjectExpression) {
...
+  };
+  return fetch(
+    `${foundryUrl.origin}/view-registry/api/dev-mode/settings/${widgetRid}`,
```
tzyl, line 612: How are you thinking about this as public API long term? On the deploy side we have gateway APIs but for Artifacts we don't for direct npm access etc.

packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -235,57 +257,52 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
...
-                    config?.logger.error(err);
-                    res.end();
-                  });
-                }
-              })
-              .catch((error) => {
-                res.statusCode = 500;
-                res.statusMessage =
-                  `Unable to start dev mode in Foundry: ${error.message}`;
+            const settingsResponse = await setWidgetSettings(
```
tzyl, line 260: Is it worth logging what's happening? "Setting widget settings" "Enabling dev mode" in the vite dev mode output or is it not really shown

packages/widget.vite-plugin/src/plugin.ts
```diff
@@ -235,57 +257,52 @@ export function FoundryWidgetVitePlugin(options: Options = {}): Plugin {
...
+
+            try {
+              res.setHeader("Content-Type", "application/json");
+              res.end(
+                JSON.stringify({
+                  redirectUrl:
+                    `${foundryUrl.origin}/workspace/custom-views/preview/${foundryConfig.widget.rid}`,
+                }),
+              );
+            } catch (error: any) {
```
tzyl, line 300: What kind of error would this catch? I think in the old code the `.catch` looks like it's targeted on the fetch. Another difference with the `.catch` is it would probably catch if the fetch itself fails to return a response so we may want to add that around the fetch to be the same as before

https://github.com/palantir/osdk-ts/blob/04d5e22299031472d51b1c531a8dbb305aeea3f7/packages/shared.net.fetch/src/createFetchOrThrow.ts#L34-L38

tzyl (approved):

packages/widget.vite-plugin/src/plugin.ts
rosscourt: Yep

packages/widget.vite-plugin/src/plugin.ts
rosscourt: We should align behind public API equivalents of these once we have them!

packages/widget.vite-plugin/src/plugin.ts
rosscourt: Ah yes, the try is in the wrong place

packages/widget.vite-plugin/src/plugin.ts
rosscourt: It self-overwrites, and this only gets triggered when the user clicks the link to their browser so they won't actually see the output

ssanjay1 (approved):
Rubber stamping!

PR #1064, opened 2024-12-13
Fixes attachment uploads in Browser context

nihalbhatnagar:
We are running into issues where passing a JS object where Blob was used as a prototype(e.g using Object.create in createAttachmentUpload) did not work in the browser. The browser seemed to be stringifying or `toString` the object when this object was passed into Fetch. This resulted in the literal string "[object Blob]" being uploaded.


packages/api/src/mapping/PropertyValueMapping.ts
```diff
@@ -50,7 +50,7 @@ export interface PropertyValueWireToClient {
  * Map from the PropertyDefinition type to the typescript type that we accept
  */
 export interface PropertyValueClientToWire {
-  attachment: string | AttachmentUpload;
+  attachment: string | AttachmentUpload | Blob & { readonly name: string };
```
nihalbhatnagar, line 53: This specific mapping not used anywhere, added for sake of consistency

packages/api/src/object/Attachment.ts
```diff
@@ -29,8 +29,9 @@ export interface Attachment {
  * This interface should also accept the File object from
  * the W3C FileApi https://www.w3.org/TR/FileAPI/#file-section
  */
-export interface AttachmentUpload extends Blob {
+export interface AttachmentUpload {
```
nihalbhatnagar, line 32: The alternative is to have AttachmentUpload also accept `Blob & {name: string}`. The current hover looks like this: 
image

packages/example-generator/src/check.test.ts
```diff
@@ -21,9 +21,13 @@ import { run } from "./run.js";
...
-});
+it(
+  "Generates code that matches the files on disk in the examples dir",
+  async () => {
+    await run({
+      outputDirectory: path.join(__dirname, "..", "..", "..", "examples"),
+      check: true,
+    });
+  },
+  { timeout: 15000 },
```
nihalbhatnagar, line 32: Updated this timeout, I'm not sure why this started failing now, but looking at logs previous successful builds it seems like we were already on the edge of 5 seconds, so I don't think this code is necessarily something that caused a major change.

nihalbhatnagar: Log from most recent build: 
image

PR #1068, opened 2024-12-18
Correct widget template package dependency versions

rosscourt:
The dependencies are currently being copied into the template from the `package.json`, resulting in invalid `workspace` versions in the end user's project. The convention used in all other templates is for these in-repo packages to be declared in the handlebars file instead. E.g:
https://github.com/palantir/osdk-ts/blob/5f75268e26e987b52103fca228e93d18894d19e6/packages/create-app.template.react.beta/templates/package.json.hbs#L14-L19

I'm targeting #1065 because then I can also pick up a `^1` version of the Vite plugin and won't need to further bump template versions to pick up fixes for that package.

### DIFF
--- a/packages/example-generator/src/run.ts
+++ b/packages/example-generator/src/run.ts
@@ -330,6 +330,21 @@ const UPDATE_PACKAGE_JSON: Mutator = {
         `"@osdk/oauth": "workspace:*"`,
       )
       .replace(
+        // Use locally generated SDK in the monorepo
+        /"@osdk\/widget-client-react.unstable": "\^.*?"/,
+        `"@osdk/widget-client-react.unstable": "workspace:*"`,
+      )
+      .replace(
+        // Use locally generated SDK in the monorepo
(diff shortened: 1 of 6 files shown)


packages/example-generator/src/run.ts
```diff
@@ -329,6 +329,21 @@ const UPDATE_PACKAGE_JSON: Mutator = {
...
+      .replace(
+        // Use locally generated SDK in the monorepo
+        /"@osdk\/widget-client.unstable": "\^.*?"/,
+        `"@osdk/widget-client.unstable": "workspace:*"`,
+      )
+      .replace(
+        // Use locally generated SDK in the monorepo
+        /"@osdk\/widget.vite-plugin.unstable": "\^.*?"/,
+        `"@osdk/widget.vite-plugin.unstable": "workspace:*"`,
+      )
```
rosscourt, line 346: @ssanjay1 I needed to introduce these replacements for the in-repo generator checks to pass

PR #1069, opened 2024-12-18
Make interface type status configurable

kkauff:
Make interface type status configurable via a string (for "experimental" or "active" only) or using the InterfaceTypeStatus definition struct.

### DIFF
--- a/packages/maker/src/api/defineInterface.ts
+++ b/packages/maker/src/api/defineInterface.ts
@@ -23,7 +23,6 @@ import type {
   InterfaceTypeStatus,
   InterfaceTypeStatus_active,
   InterfaceTypeStatus_deprecated,
-  InterfaceTypeStatus_example,
   InterfaceTypeStatus_experimental,
   PropertyTypeType,
   SharedPropertyType,
@@ -150,8 +149,6 @@ function mapStringToStatus(status: string): InterfaceTypeStatus {
         type: "experimental",
         experimental: {} as InterfaceTypeStatus_experimental,
       };
-    case "example":
-      return { type: "example", example: {} as InterfaceTypeStatus_example };
     case "active":
       return { type: "active", active: {} as InterfaceTypeStatus_active };
     case "deprecated":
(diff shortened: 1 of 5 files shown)


packages/maker/src/api/defineInterface.ts
```diff
@@ -20,20 +20,36 @@ import { defineSharedPropertyType } from "./defineSpt.js";
...
 } from "./types.js";
 
+/**
+ * This status indicates that the interface is reaching the end of its life and will be removed as per the
+ * deadline specified.
+ */
+export interface DeprecatedInterfaceTypeStatus {
+  message: string;
+  deadline: string;
+  replacedBy: InterfaceTypeRid | undefined;
```
aep000, line 38: Generally not a fan of rids inside OAC. I believe this won't even work under the hood, because we removed all Rids and replaced them with API names. Can we remove this for now given its optional?

packages/maker/src/api/defineInterface.ts
```diff
@@ -20,20 +20,36 @@ import { defineSharedPropertyType } from "./defineSpt.js";
...
+}
+
 export function defineInterface(
   opts: {
     apiName: string;
     displayName?: string;
     description?: string;
     icon?: { locator: BlueprintIcon; color: string };
+    status?: string | InterfaceTypeStatus;
```
aep000, line 47: I would prefer instead of exposing the raw InterfaceTypeStatus union we flattened the union to be simply `{type:"deprecated"; message: string; deadline: string;} | {type: "active"} | {type: "experimental"}`

I think this is a nicer UX than base conjure unions.

packages/maker/src/api/defineInterface.ts
```diff
@@ -114,7 +132,7 @@ export function defineInterface(
     extendsInterfaces: extendsInterfaces,
     links: [],
     properties,
-    status: { type: "active", active: {} },
+    status: status,
```
ssanjay1, line 135: nit: should be able to just do `status` here without the :

ssanjay1 (approved):
Approved with one nit

ssanjay1 (approved):

PR #1071, opened 2024-12-18
Fix onOutOfDate handler from firing twice

nihalbhatnagar:


ssanjay1 (approved):

ssanjay1 (commented):
Approved to unblock, do we have any tests that check for this?

nihalbhatnagar:
@ssanjay1 We don't have one specifically, I will write one

PR #1074, opened 2025-01-07
Provide a clean mechanism to lock refresh tokens to sdk versions

### LINKED ISSUE #1073: Tokens with outdated permissions cause 403 errors when user loads new version of app; provide way to get new token with new permissions on load
## Issue

We've had problems in production along the lines of:

- publish new version of app which uses updated permissions (like a new Object Type)
- user who already logged in and has a token with not updated permissions loads new version of app
- user gets a 403 when trying to load the new Object Type because their token is using old scope that does not have perms to query the new Object Type

Is there a way in my TS code/the OSDK flow I could check if the token is using old permissions (or is from an old version of the app) and, if so, refresh the token?

---

## Workaround

HACK: write OSDK user agent (which includes the version) to local storage on sign in. On refresh, compare the current user agent to the one in local storage (if any). If there is a difference, then permissions might have changed, so I sign the user out.

```typescript
const OSDK_USER_AGENT: string =
  $(InstallationConfig).$objectSetInternals.def.osdkMetadata.extraUserAgent
invariant(OSDK_USER_AGENT, 'Failed to find OSDK_USER_AGENT')

const OSDK_USER_AGENT_KEY = 'OSDK_USER_AGENT_KEY'

/**
 * Refresh is run when the page is first loaded; if the user agent has changed or has not been set,
 * log out to get a new token and to set the user agent in local storage.
 */
auth.addEventListener('refresh', () => {
  if (localStorage.getItem(OSDK_USER_AGENT_KEY) === OSDK_USER_AGENT) return

  // eslint-disable-next-line no-console
  console.log('User agent changed, logging out')

  void auth.signOut()
})

auth.addEventListener('signIn', () => {
  localStorage.setItem(OSDK_USER_AGENT_KEY, OSDK_USER_AGENT)
})
```

ericanderson:
Fixes #1073

Example usage would be:
```
import { $osdkMetadata } from "@mygenerated/osdk";

createPublicOauthClient(myClientId, foundryUrl, "/auth/callback", { refreshTokenMarker: $osdkMetadata.extraUserAgent });

```


packages/oauth/src/common.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+  });
+
+  it("should use the old style keys when there is no refresh token marker", () => {
+    const client = {
+      client_id: "hi_mom",
+      $refreshTokenMarker: undefined,
+    };
+
+    const oldLocalStorageKey = `@osdk/oauth : refresh : ${client.client_id}`;
```
ericanderson, line 35: Ensure that people not using this feature won't get logged out on upgrade of `@osdk/oauth`

packages/oauth/src/common.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+  });
+
+  it("should respect the refresh token marker", () => {
+    const client = {
+      client_id: "hi_mom",
+      $refreshTokenMarker: "magic_marker_friend",
+    };
+
+    const localStorageKey =
```
ericanderson, line 61: Make sure that if you do specify the refreshTokenMarker that it is used.

packages/oauth/src/common.ts
```diff
@@ -237,7 +243,7 @@ export function createAuthorizationServer(
     | "revocation_endpoint"
   >
 > {
-  const issuer = `${new URL(ctxPath, url)}`;
+  const issuer = `${new URL(ctxPath, url + "/")}`;
```
ericanderson, line 246: Fix an issue where Tim's tests weren't working if you forgot trailing slash

packages/oauth/package.json
```diff
@@ -24,6 +24,7 @@
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
+    "test": "vitest run --pool=forks",
```
ericanderson, line 27: I think Tim's changes were not properly failing build due to this not being present.

ssanjay1 (dismissed):

nihalbhatnagar (approved):

PR #1078, opened 2025-01-08
Struct search (where clauses)

ssanjay1:


packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -81,14 +84,16 @@ export function modernToLegacyWhereClause<
...
   filterType: "$within" | "$intersects",
+  propertyIdentifier?: PropertyIdentifier,
+  field?: string,
 ): SearchJsonQueryV2 {
   return {
     type: filterType === "$within"
       ? "withinBoundingBox"
       : "intersectsBoundingBox",
+    ...(propertyIdentifier != null && { propertyIdentifier }),
     field,
```
ssanjay1, line 97: Field used to be required, but its now optional along with `PropertyIdentifier`. We only send one or the other

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -119,30 +126,38 @@ function makeGeoFilterPolygon(
...
   objectOrInterface: ObjectOrInterfaceDefinition,
-): SearchJsonQueryV2 {
+  structFieldSelector?: { propertyApiName: string; structFieldApiName: string },
+): any {
   invariant(
     filter != null,
     "Defined key values are only allowed when they are not undefined.",
   );
 
+  const propertyIdentifier: PropertyIdentifier | undefined =
```
ssanjay1, line 138: If we have a struct field selector, use property identifier, otherwise, default back to using field

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -119,30 +126,39 @@ function makeGeoFilterPolygon(
...
-        field = `${objApiNamespace}.${fieldShortName}`;
+        fieldName = `${objApiNamespace}.${fieldShortName}`;
       }
     }
 
     return {
       type: "eq",
-      field,
+      ...(propertyIdentifier != null
+        && { ...propertyIdentifier, propertyApiName: fieldName }),
```
ssanjay1, line 160: Want to make sure for structs we use fqn on interface props

packages/client/src/internal/conversions/modernToLegacyWhereClause.test.ts
```diff
@@ -31,7 +35,12 @@ describe(modernToLegacyWhereClause, () => {
           apiName: "a.Foo",
           __DefinitionMetadata: {
             type: "interface",
-            properties: { "prop": { type: "integer" } },
+            properties: {
+              "prop": { type: "integer" },
+              "prop2": {
+                type: { "innerProp1": "string", "innerProp2": "float" },
```
ssanjay1, line 41: Defined a struct prop on interface

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -119,29 +126,38 @@ function makeGeoFilterPolygon(
...
+        type: "structField",
+        ...structFieldSelector,
+        propertyApiName: fullyQualifyPropName(
+          structFieldSelector.propertyApiName,
+          objectOrInterface,
+        ),
+      }
+      : undefined;
+  const field = structFieldSelector == null
+    ? fullyQualifyPropName(fieldName, objectOrInterface)
```
ssanjay1, line 150: This fixes a bug where we weren't converting fully qualified prop names on interfaces in all cases (like when we use $ filters)

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -81,14 +84,16 @@ export function modernToLegacyWhereClause<
...
   bbox: BBox,
   filterType: "$within" | "$intersects",
+  propertyIdentifier?: PropertyIdentifier,
+  field?: string,
 ): SearchJsonQueryV2 {
   return {
     type: filterType === "$within"
       ? "withinBoundingBox"
       : "intersectsBoundingBox",
+    ...(propertyIdentifier != null && { propertyIdentifier }),
```
ssanjay1, line 96: This is a bit ugly, but did this so that propertyIdentifier only shows up in the return object if its defined, this makes it so we don't need to go update our entire test bed either to include a field which may change in near future. Once we solidify that this is the way forward, I can remove field and clean this up

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -185,16 +206,22 @@ function handleWherePair(
     const withinBody = filter[firstKey] as GeoFilter_Within["$within"];
 
     if (Array.isArray(withinBody)) {
-      return makeGeoFilterBbox(field, withinBody, firstKey);
+      return makeGeoFilterBbox(withinBody, firstKey, propertyIdentifier, field);
```
ssanjay1, line 209: Just swapped order and now send `propertyIdentifier` and `field` now

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -246,10 +294,25 @@ function handleWherePair(
...
+    ...(propertyIdentifier != null && { propertyIdentifier }),
     field,
     value: filter[firstKey] as any,
   };
 }
 
 type DropDollarSign<T extends `$${string}`> = T extends `$${infer U}` ? U
   : never;
+
+function fullyQualifyPropName(
```
ssanjay1, line 306: Same logic as before, just put in helper function

packages/client/src/internal/conversions/modernToLegacyWhereClause.test.ts
```diff
@@ -40,14 +49,32 @@ describe(modernToLegacyWhereClause, () => {
         } as const satisfies ObjectOrInterfaceDefinition;
 
         const r = modernToLegacyWhereClause({
-          prop: 5,
+          $and: [
+            { prop: { $eq: 5 } },
```
ssanjay1, line 53: Changed to a $ filter, because this broke the old code, so want to make sure we can handle that case

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
nihalbhatnagar: This makes sense, can we document this in code with a comment in case we forget

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -159,12 +175,16 @@ function handleWherePair(
...
   if (!hasDollarSign) {
-    // Future case for structs
-    throw new Error(
-      `Unsupported filter. Did you forget to use a $-prefixed filter? (${
-        JSON.stringify(filter)
-      })`,
+    const structFilter = Object.entries(filter);
+    invariant(
+      structFilter.length === 1,
+      "Cannot filter on more than one struct field in the same clause, need to use an and clause",
```
nihalbhatnagar, line 181: Is it possible to make this a type error the same way we did this for the rest of the where clause?

nihalbhatnagar: Also, this may not be in line with the rest of how we do where clauses, but is it tough to just support this?

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -119,29 +126,38 @@ function makeGeoFilterPolygon(
 }
 
 function handleWherePair(
-  [field, filter]: [string, any],
+  [fieldName, filter]: [string, any],
   objectOrInterface: ObjectOrInterfaceDefinition,
-): SearchJsonQueryV2 {
+  structFieldSelector?: { propertyApiName: string; structFieldApiName: string },
+): any {
```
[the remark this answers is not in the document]
ssanjay1: Ah good catch thats from before I had the new types, let me fix

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
ssanjay1: Sorry could you elaborate on what you mean by type error the same way as the rest? But to answer the second question, its less so tough, but more so just trying to keep in parity with how we format the rest of our where clauses not allowing more than one field, didn't want it to be confusing

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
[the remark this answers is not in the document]
ssanjay1: Fixed

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
ssanjay1: Done!

nihalbhatnagar (approved):

PR #1081, opened 2025-01-09
Add typed lint checks and fix promise related things

ericanderson:


packages/client/src/createClient.ts
```diff
@@ -103,8 +103,10 @@ export function createClientInternal(
       throw new Error("Invalid ontology RID");
     }
   } else {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
     ontologyRid.then((ontologyRid) => {
       if (!ontologyRid.startsWith("ri.")) {
+        // FIXME this promise is not await so this just shows up as an unhandled promise rejection
         throw new Error("Invalid ontology RID");
```
ericanderson, line 110: @nihalbhatnagar have a follow up for you

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -667,7 +672,7 @@ export class ObjectSetListenerWebsocket {
 /** @internal */
 export function constructWebsocketUrl(
   baseUrl: string,
-  ontologyRid: string | Promise<string>,
+  ontologyRid: string,
```
ericanderson, line 675: Caught an actual bug. The error was on line 679 because `...${ontologyRid}...` is a problem if its a promise!

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.test.ts
```diff
@@ -494,7 +494,8 @@ describe("generator", () => {
     const diagnostics = compileThis(helper.getFiles(), BASE_PATH);
     for (const q of diagnostics) {
       console.error(
-        `${q.file?.fileName}:${q.file?.getLineStarts()} ${q.messageText}`,
+        `${q.file?.fileName}:${q.file?.getLineStarts()}`,
+        q.messageText,
```
ericanderson, line 498: Caught bug. `.messageText` can be a string or an object (strangely).

packages/generator/src/v2.0/generateMetadata.ts
```diff
@@ -26,7 +26,7 @@ export async function generateOntologyMetadataFile(
   { fs, outDir, ontology, ontologyApiNamespace }: GenerateContext,
   userAgent: string,
 ) {
-  fs.writeFile(
+  await fs.writeFile(
```
ericanderson, line 29: Fixes race condition in some test scenarios and possibly a real bug

packages/generator/src/v2.0/generateRootIndexTsFile.ts
```diff
@@ -22,7 +22,7 @@ import { formatTs } from "../util/test/formatTs.js";
 export async function generateRootIndexTsFile(
   { fs, outDir, importExt, ontologyApiNamespace, ontology }: GenerateContext,
 ) {
-  fs.writeFile(
+  await fs.writeFile(
```
ericanderson, line 25: Fixes race condition/real bug

packages/maker/src/api/defineInterface.ts
```diff
@@ -62,7 +62,9 @@ export function defineInterface(
         ) {
           invariant(
             isPropertyTypeType(type),
-            `Invalid data type ${type} for property ${apiName} on InterfaceType ${apiName}`,
+            `Invalid data type ${
+              JSON.stringify(type)
```
ericanderson, line 66: `type` may be an object here...

packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -161,8 +161,10 @@ export function createPublicOauthClient(
 
   // as an arrow function, `useHistory` is known to be a boolean
   const go = async (x: string) => {
-    if (useHistory) return window.history.replaceState({}, "", x);
-    else window.location.assign(x);
+    if (useHistory) {
+      window.history.replaceState({}, "", x);
+      return;
```
ericanderson, line 166: Auto fixer did this when I saved a different change. I kinda like the `return void window...` syntax better but there are two different promise rules and that causes an issue with another one since `.assign` isn't returning a promise, it returns void and so the `void window.history...` is "redundant". But its not cause we need it for the async function. Sigh. So i left this auto fixer as is.

packages/tool.release/src/publishPackages.ts
```diff
@@ -55,7 +55,7 @@ export async function publishPackages(
   );
 
   // need to produce the output to save for future run of create tag/release
-  fs.writeFile(
+  await fs.writeFile(
```
ericanderson, line 58: More races/real bug

packages/widget.vite-plugin.unstable/src/plugin.ts
```diff
@@ -248,7 +248,7 @@ export function FoundryWidgetVitePlugin(_options: Options = {}): Plugin {
                 res.statusCode = settingsResponse.status;
                 res.statusMessage =
                   `Unable to set widget settings in Foundry: ${settingsResponse.statusText}`;
-                settingsResponse.text().then((err) => {
+                await settingsResponse.text().then((err) => {
```
ericanderson, line 251: Prevents unhandled rejection + improves stack trace

packages/widget.vite-plugin.unstable/src/plugin.ts
```diff
@@ -669,6 +669,7 @@ function getPluginTransformHook(
     typeof plugin.transformIndexHtml === "object"
     && "transform" in plugin.transformIndexHtml
   ) {
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
     return plugin.transformIndexHtml.transform;
```
ericanderson, line 673: I think in general its good for it to be an error when we use deprecated types but I dont want to deep research vite plugins to understand if there is a consequence to this right now so we will save it as is.

turbo.json
```diff
@@ -27,7 +27,7 @@
         "tsconfig.json"
       ],
       "outputs": [],
-      "dependsOn": ["//#global-eslint-config", "//#dprint"]
+      "dependsOn": ["//#global-eslint-config", "//#dprint", "typecheck"]
     },
```
ericanderson, line 32: Gotta be sure we have compiled all of our dependencies

packages/client/src/createClient.ts
nihalbhatnagar: Ah just need a .catch here?

ericanderson: I think you probably don't want to throw because it will just end up as an unhandled promise. Maybe we want to set `ontologyRid = ontologyRid.then(...` and that way when we do end up using it it will pop up through a call stack?

We could maybe add a console.error() for the first time its hit too?

nihalbhatnagar: I see what you mean, this makes sense to me

nihalbhatnagar (approved):
Looks good to me, the warnings at the bottom look relatively simple to fix as well when we clean up

PR #1082, opened 2025-01-09
Add's more descriptive error messages for 429 errors and other errors with no body

### LINKED ISSUE #1070: Add a more descriptive error for 429 Errors/errors with no body
When we receive an error response from API Gateway, we try and [parse the body of the response](https://github.com/palantir/osdk-ts/blob/adf3b122ae166556dd859925e2e514dd854289f3/packages/shared.net.fetch/src/createFetchOrThrow.ts#L45). For some errors(specifically a 429), there is no response body, so a call to response.json fails and logs an "Unexpected end of JSON input" error instead of passing on the response. We should ensure we don't swallow the error in this case.
image
image

nihalbhatnagar:
Fixes #1070

### DIFF
--- a/packages/shared.net.fetch/src/createFetchOrThrow.ts
+++ b/packages/shared.net.fetch/src/createFetchOrThrow.ts
@@ -43,12 +43,14 @@ export function createFetchOrThrow(fetchFn: typeof fetch = fetch) {
       }
 
       let body;
+
       try {
         body = await response.json();
       } catch (e) {
         throw convertError(
           e,
           "A network error occurred while reading response",
+          response,
         );
       }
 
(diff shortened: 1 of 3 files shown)


packages/shared.net.fetch/src/createFetchOrThrow.ts
```diff
@@ -69,7 +71,17 @@ export function createFetchOrThrow(fetchFn: typeof fetch = fetch) {
 function convertError(
   e: any,
   msgIfNotError: string = "An unknown error occurred",
+  response?: Response,
 ) {
+  if (response?.status && response.statusText) {
+    // Catches errors like 429 where there is no response body but a code is returned
+    return new UnknownError(
```
ericanderson, line 78: Hmm. I don't think a statusText is a requirement. I'd probably not do this separate if statement and instead just include the statusCode on the 2 throws below. If you think the status text is important you could add it to the first argument of the message down there.

The use of `UNKNOWN` aligns with the existing palantir api errors that are used by conjure and co and that populate through so using a custom message like the `statusText` for that feels off.

ericanderson (changes requested):

packages/shared.net.fetch/src/createFetchOrThrow.ts
```diff
@@ -46,9 +46,10 @@ export function createFetchOrThrow(fetchFn: typeof fetch = fetch) {
       try {
         body = await response.json();
       } catch (e) {
-        throw convertError(
-          e,
-          "A network error occurred while reading response",
+        throw new PalantirApiError(
```
nihalbhatnagar, line 49: This is fine since we already know we're in an error state. The JSON body being unparseable isn't an error here

PR #1083, opened 2025-01-10
fix ToDo sample to allow build on lesson level

ewitkon:
ToDo tutorial on version 2.0 fails to build when you complete only lesson 1 out of the 2-3 lessons. This is due to type mismatch with the mock objects. 
When users are using VS code in platform and check in the code the CI will fail as it run `npm run build`, which is a poor experience. 
This PR extract the dependency of most of the code from the Mock objects so you can implement each method at a time and code will still compile. This PR will go together with a [forge pr](https://github.palantir.build/foundry/forge/pull/153270) to fix the suggested solutions to the lessons.

### DIFF
--- a/examples/example-tutorial-todo-aip-app-sdk-2.x/src/useProjects.ts
+++ b/examples/example-tutorial-todo-aip-app-sdk-2.x/src/useProjects.ts
@@ -1,19 +1,37 @@
 import { useCallback } from "react";
 import useSWR from "swr";
-import type { MockProject } from "./mocks";
 import Mocks from "./mocks";
 
+export interface IProject {
+  $apiName: string;
+  $primaryKey: string;
+  id: string;
+  name: string;
+  description: string;
+}
+
+export interface ITask {
+  $apiName: string;
+  $primaryKey: string;
+  id: string;
+  title: string;
+  description: string;
+  projectId: string;
+}
+
 function useProjects() {
-  const { data, isLoading, isValidating, error, mutate } = useSWR<
-    MockProject[]
-  >("projects", async () => {
+  const { data, isLoading, isValidating, error, mutate } = useSWR<IProject[]>("projects", async () => {
     // Try to implement this with the Ontology SDK!
-    return Mocks.getProjects();
-  });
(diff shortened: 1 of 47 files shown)


examples/example-tutorial-todo-aip-app-sdk-2.x/src/useProjects.ts
```diff
@@ -1,19 +1,41 @@
...
+    async () => {
+      // Try to implement this with the Ontology SDK!
+      const projectsList: IProject[] = (await Mocks.getProjects()).map((
+        project,
+      ) => ({
+        ...project,
+      }));
+      return projectsList;
+    },
+  );
```
zeyadkhaled, line 34: This feels like its redundant and could be done like this?

```suggestion
  const { data, isLoading, isValidating, error, mutate } = useSWR<IProject[]>(
    "projects",
    async () => {
      // Try to implement this with the Ontology SDK!
      const projectsList: IProject[] = await Mocks.getProjects();
      return projectsList;
    },
  );
```

examples/example-tutorial-todo-aip-app-sdk-2.x/src/useProjects.ts
```diff
@@ -1,19 +1,41 @@
...
 
+export interface IProject {
+  $apiName: string;
+  $primaryKey: string;
+  id: string;
+  name: string;
+  description: string;
+}
+
+export interface ITask {
```
zeyadkhaled, line 13: This should live in useProjectTasks

zeyadkhaled (approved):
Just minor comments, but those apply to all templates

examples/example-tutorial-todo-aip-app-sdk-2.x/src/useProjects.ts
ewitkon: Right, but I thought it would be good to keep the mapping code here as when you implement the lesson you need a similar thing to map from your OSDK object to IProject.

zeyadkhaled (approved):
Thanks!

zeyadkhaled (approved):

PR #1084, opened 2025-01-10
Creates new internal export 

nihalbhatnagar:
Creates new internal export for helper methods we want to hide aimed at internal developers.

Currently deciding to export: 
`createAttachmentFromRid`
`createTemporaryObjectSet`

These methods have been renamed accordingly.

A follow up PR will also export `preexistingObjectSet`


packages/client/src/public-utils/createAndFetchTempObjectSetRid.ts
```diff
@@ -0,0 +1,48 @@
...
+} from "@osdk/api";
+import { OntologyObjectSets } from "@osdk/internal.foundry.ontologiesv2";
+import { additionalContext, type Client } from "../Client.js";
+import { getWireObjectSet } from "../objectSet/createObjectSet.js";
+
+/**
+ * Fetches a temporary object set RID from the Foundry stack for the given object set.
+ *
+ * @param client - An OSDK client.
+ * @param objectSet - The object set to fetch a RID for.
```
nihalbhatnagar, line 30: Thought this sounded weird, I always thought it was based on the actual letter but its actually based on [sound] of the beggining.(https://www.scribbr.com/commonly-confused-words/a-vs-an/#:~:text=a%20useful%20one.-,A%20or%20an%20before%20acronyms,-The%20same%20rule) I prodounce it as "R I D" so it sounds weird to me(should be an), but if Foundry used "rid" as a words then its "a"

ssanjay1 (approved):

PR #1085, opened 2025-01-10
Interface support in actions

ssanjay1:
Actions can now take interface params


packages/api/src/actions/Actions.ts
```diff
@@ -55,6 +56,14 @@ export namespace ActionParam {
    * Helper type to convert action definition parameter object sets to typescript types
    */
   export type ObjectSetType<T extends ObjectTypeDefinition> = ObjectSet<T>;
+
+  /**
+   * Helper type to convert action definition parameter interface types to typescript types
+   */
+  export type InterfaceType<T extends InterfaceDefinition> = {
+    $objectType: string;
```
ssanjay1, line 64: So, we now get a list of objects that implement an interface that we theoretically could use to limit the object types you can pass in here. But, there's a chance that someone implements this interface with a new object type we won't know about after codegen

ericanderson (dismissed):

PR #1086, opened 2025-01-10
Prep work for splitting clis

ericanderson:
The ultimate goal is to move a bunch of packages to `osdk-ts-clis`, specifically:

- `create-app`
- `create-app.*`
- `create-widget`
- `create-widget.*`
- `cli`
- `cli.common`
- `example-generator`

NOTE: Intent is to keep `cli.cmd.typescript` in this repo still.

### DIFF
--- a/babel.config.mjs
+++ b/babel.config.mjs
@@ -18,6 +18,7 @@ import { findUp } from "find-up";
 import { readFile } from "fs/promises";
 import * as path from "node:path";
 
+process.env.STABLE_PACKAGE_CLIENT_VERSION = "~2.0.11";
 process.env.PACKAGE_VERSION = await readPackageVersion(process.cwd());
 process.env.PACKAGE_API_VERSION = await readPackageVersion("packages/api");
 process.env.PACKAGE_CLIENT_VERSION = await readPackageVersion(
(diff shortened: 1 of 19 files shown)


packages/cli.cmd.typescript/package.json
```diff
@@ -30,8 +30,9 @@
   },
   "dependencies": {
     "@arethetypeswrong/cli": "^0.15.2",
-    "@osdk/cli.common": "workspace:~",
-    "@osdk/generator": "workspace:~",
+    "@osdk/cli.common": "workspace:*",
+    "@osdk/generator": "workspace:*",
+    "@osdk/generator-utils": "workspace:*",
```
ericanderson, line 35: For these we don't want surprises so we are locking exactly in.

packages/create-app/src/cli.test.ts
```diff
@@ -47,6 +55,14 @@ beforeEach(() => {
 describe.each(TEMPLATES)("template $id", (template) => {
   const supportedVersions = Object.keys(template.files);
   describe.each(supportedVersions)("For SDK version %s", (sdkVersion) => {
+    beforeAll(() => {
+      vi.stubEnv("STABLE_PACKAGE_CLIENT_VERSION", "2.1.0-beta.20");
```
ericanderson, line 59: We don't run babel or anything for tests

babel.config.mjs
```diff
@@ -18,6 +18,7 @@ import { findUp } from "find-up";
 import { readFile } from "fs/promises";
 import * as path from "node:path";
 
+process.env.STABLE_PACKAGE_CLIENT_VERSION = "~2.0.11";
```
ericanderson, line 21: After we move this to the other repo, we can put a devDependency on `@osdk/client` in the packages that need this and can just read from there instead of doing the hard coding.

PR #1087, opened 2025-01-10
Export hydrateObjectSetFromRid from internal

nihalbhatnagar:
I originally wanted to add a check for the number of object types in the object set is only one(which is why this is a separate PR), but that endpoint is not exported from gateway yet. Now that I think about this though we likely want to keep this a synchronous operation, so it may not be necessary. I might consider a custom error on a standard fetch somehow that can catch this when an actual request is made in order to have a clear error


packages/client/src/public-utils/hydrateObjectSetFromRid.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param definition An OSDK object or interface definition.
+ * @param rid The RID of the temporary object set.
+ * @returns
+ */
+export function hydrateObjectSetFromRid<T extends ObjectOrInterfaceDefinition>(
+  client: Client,
+  definition: T,
+  rid: string,
+) {
+  OntologyObjectSets.get;
```
ssanjay1, line 35: This intentional?

nihalbhatnagar: No, thank you good catch!

ssanjay1 (approved):

PR #1089, opened 2025-01-10
Fixes infinite redirect issue on some errors in the OAuth flow

### LINKED ISSUE #953: Prevent oauth infinite loop in createPublicOauthClient
Tim did similar to the older version here: https://github.com/palantir/osdk-ts/pull/389/files

Assigning to you @ssanjay1 but feel free to move it elsewhere

nihalbhatnagar:
Fixes #953. Implements the approach used  in #389 for legacy oauth which moves the codeVerifier to session storage to avoid a bad state.

The error is caught on the authRedirect page and can be displayed to the user. The caveat here is that the "back" button is not going to redirect you to where the sign in began as the state is replaced. The expected flow here as discussed in the previous PR/offline discussions is that the user incorporates their own handling when something goes wrong here.

image


packages/oauth/src/common.ts
```diff
@@ -103,6 +99,31 @@ export function readLocal(client: Client): LocalStorageState {
   );
 }
 
+export function saveSession(client: Client, x: SessionStorageState) {
```
nihalbhatnagar, line 102: I left the above localStorage in a helper just to endure we still `localStorage?` due to the node.js restriction

packages/oauth/src/publicOauth.test.ts
```diff
@@ -250,7 +270,10 @@ describe(createPublicOauthClient, () => {
         );
       });
 
-      if (Object.keys(initialLocalState).length === 0) {
+      if (
+        Object.keys(initialState.localStorage).length === 0
+        && Object.keys(initialState.sessionStorage).length === 0
```
nihalbhatnagar, line 275: This is the equivalent as above, if we don't have the refresh token or the pkce code verifier, we need the whole flow

packages/oauth/src/publicOauth.test.ts
```diff
@@ -319,7 +342,7 @@ describe(createPublicOauthClient, () => {
         }
       }
 
-      if (initialLocalState.codeVerifier) {
+      if (initialState.sessionStorage.codeVerifier) {
```
nihalbhatnagar, line 345: Then we check if we have the code verifier

packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -269,14 +274,15 @@ export function createPublicOauthClient(
       && window.location.href !== loginPage
       && window.location.pathname !== loginPage
     ) {
-      saveLocal(client, { oldUrl: postLoginPage });
+      saveSession(client, { oldUrl: postLoginPage });
```
ericanderson, line 277: This code previously ensured that we were no longer storing `refreshToken` as it overwrote state but this line no longer does that. Is that intentional or should we be saving an empty object here

nihalbhatnagar: Good catch. I don't think this has any behavioral effects as `initiateLoginRedirect` seems to only be called when there is no refresh token in local storage here. However, we should ensure we are always reflecting what the state should be and similarly follow what we do immediately below, so I'll add the fix here.

ericanderson (dismissed):

ericanderson (approved):

PR #1091, opened 2025-01-13
Cleanup MRL / standardize more

ericanderson:
# Cleanup MRL / standardize more

In commit order:

## Wonky Codegen for `e2e.test.foundry-sdk-generator`

The existing code generated into the `src` which then required us to special case not including it in the tsc output. This was unnecessary and caused special casing in mrl that we could cleanup. We now expect codegen will go into either or both: `./build/codegen` and `./src/generatedNoCheck`. 

## Codegen

Just regenerating ` packages/e2e.sandbox.todoapp/src/generatedNoCheck2` with extensions

## Proper extensions for e2e.sandbox.todoapp

Manually renaming remaining extensions in imports

## Refactor `widget.vite-plugin` to use standard paths

Drops a special snow flake tsconfig file. Produces output in a standardized location for turbo.

## MRL applied boilerplate

I committed this prior to the MRL changes so that there would be no errors in lint-staged. This is just MRL applied changes for the next commit

## Standardize vite apps

Break out 5 projects from `nonStandardPackages` into `viteApp` and `usesReact`.  Updates turbo.json with better defaults for the inputs and outputs of the `build` task and adds comments explaining why things are there.

## Remove foundry-sdk-generator from snowflakes

Following the first commit and other prep work, this commit makes all consumer facing cli apps act the same. Notably this means must use exact version matches (this helps prevent surprises for us) just like foundry-sdk-generator used to do. We also had special casing for `consola` in our tests but it didn't wire up to anything so was pointeless

## Empty changeset

There isn't really anything to report here.

## Fix lint issues

Semantic merge issues fixed here

## Big mrl refactor and improvements

Introduced a new feature into MRL for archetypes to make it easier to keep track of changes. This is part of the step towards being able to split out parts of our code base. It also importantly ensures we don't apply changes multiple times nor do we accidentally forget something.

From this we also get a set of improvements to the code base:
* No need to include `files: []` on packages that are private.
* Reducting in packages that need to check-attw, build cjs, and build browser, which should help with build speed
* Fixing a bug where we intended to bundle `client.unstable` and `client.unstable.tpsa` but we typo'd.
* We don't need to transpile types for templates
* We don't need to check-attw on the e2e.generated packages because we only use them internally
* the vite plugin might actually want to have the types so you get some type checking when you put it into a vite config


.monorepolint.config.mjs
```diff
@@ -76,13 +71,23 @@ const privatePackages = [
...
+// Package architypes (mutually exclusive)
+
 const consumerCliPackages = [
   "@osdk/cli",
   "@osdk/create-app",
   "@osdk/foundry-sdk-generator",
 ];
 
-const testWithHappyDomPackages = [
+const viteApp = [
```
nihalbhatnagar, line 81: When we move widgets out or want to make more sample apps in other repo's, we'd have to update these first? When we move the MRL config out, is it somehow possible to make these inputs of some sort so these names come from the repos that the packages live in themselves?

nihalbhatnagar: For both this var and others

nihalbhatnagar (dismissed):

.monorepolint.config.mjs
ericanderson: This file is part of a temporary state. The goal will be to refactor this whole file to be a "library" where individual repos would "configure" it with the packages in their own repo that fit the various archetypes.

.monorepolint.config.mjs
```diff
@@ -774,17 +789,45 @@ NOTE: DO NOT EDIT THIS README BY HAND. It is generated by monorepolint.
...
+    // }),
+
     ...standardPackageRules({
-      includePackages: ["@osdk/widget.vite-plugin.unstable"],
+      includePackages: [
+        ...viteApp,
+      ],
     }, {
-      tsVersion: LATEST_TYPESCRIPT_DEP,
+      react: true,
```
ssanjay1: Do we also need to set react explicitly to true for the vite apps?

.monorepolint.config.mjs
```diff
@@ -35,77 +36,219 @@ const LATEST_TYPESCRIPT_DEP = "~5.5.4";
...
+    "consumerCliPackages",
+    [
+      "@osdk/cli",
+      "@osdk/create-app",
+      "@osdk/create-widget",
+      "@osdk/foundry-sdk-generator",
+    ],
+    {
+      ...LIBRARY_RULES,
+      output: {
```
ssanjay1, line 137: You think its worth making a separate enum for bundling ESM, or do you think this will be the only case?

ssanjay1 (approved):
Overall seems fine to me, left a couple of questions but approving to unblock

.monorepolint.config.mjs
ssanjay1: Was going commit by commit, realize this is now irrelevant

.monorepolint.config.mjs
ericanderson: I figured this was the only case.

PR #1095, opened 2025-01-14
Enforce action params do not allow more parameters than in the specified type

### LINKED ISSUE #1093: Arguments passed to `applyAction` allows additional props which are not accepted by Actions
Calling `applyAction` allows additional args in the typescript SDK. ideally this would be a type error, since the actions API does not accept extra args. I can get around this by using satisfies:

```typescript
$(applySchedulingGridChanges).applyAction({
        department: departmentId,
        installationConfig: '1',
        jsonPayload: selectPayload(gridRef.current.shifts, editedShifts),
} satisfies applySchedulingGridChanges.Params)
```

nihalbhatnagar:
Fixes #104 
Before:
image
Generic expands type def only when fully complete:
image

After:
image

image


packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
```diff
@@ -83,7 +83,7 @@ export async function generatePackage(
     beta: options.beta,
   });
 
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+
```
ssanjay1, line 86: was this intentional?

packages/api/src/OsdkBase.ts
```diff
@@ -35,6 +35,18 @@ export type OsdkBase<
   readonly $objectSpecifier: ObjectSpecifier<Q>;
 };
 
+export type OsdkWithPrimaryKey<
```
ssanjay1, line 38: nit: maybe should be `OsdkBaseWithPrimaryKey`

packages/client/src/util/isOsdkObject.ts
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-import type { OsdkBase } from "@osdk/api";
-
-export function isOsdkBaseObject(o: any): o is OsdkBase<any> {
-  return o && typeof o === "object" && typeof o.$apiName === "string"
+export function isActionOrQueryObjectParameter(
+  o: any,
+): o is { $primaryKey: any } {
```
ssanjay1, line 19: Assuming you intentionally didn't use `OsdkBaseWithPrimaryKey` here because you didn't want to add it to our contract yet?

packages/e2e.sandbox.catchall/src/runActionsTest.ts
```diff
@@ -0,0 +1,64 @@
...
+await client($Actions.createOsdkTestObject).applyAction({
+  description: "test",
+  osdk_object_name: "OsdkTestObject",
+  string_property: "test",
+});
+
+const objectArray = await client(OsdkTestObject).fetchPage();
+
+// Testing Object action with OSDK instance
+await client($Actions.editOsdkTestObject).applyAction({
```
ssanjay1, line 30: Can we add type tests in actions.test.ts to make sure our actions accept osdkInstance+primaryKey+osdkBase+osdkBaseWithPrimaryKey?

packages/client/src/util/isOsdkObject.ts
nihalbhatnagar: Yep, i don't want to export this from API yet if ever

packages/e2e.sandbox.catchall/src/runActionsTest.ts
nihalbhatnagar: Sure, I will add them there as well. I should have commented this but I added the tests in our e2e so they'll be run during our backcompat tests as well. I think for the future its probably best we have coverage in both places as well

ssanjay1 (dismissed):

ssanjay1 (approved):

PR #1101, opened 2025-01-16
Fix up JSDocs for public utils

nihalbhatnagar:


packages/client/src/public-utils/hydrateObjectSetFromRid.ts
```diff
@@ -20,11 +20,11 @@ import { additionalContext } from "../Client.js";
 import { createObjectSet } from "../objectSet/createObjectSet.js";
 
 /**
- * Creates an OSDK object set from a temporary object set RID.
- * @param client An OSDK client.
- * @param definition An OSDK object or interface definition.
- * @param rid The RID of the temporary object set.
- * @returns
+ * Creates an OSDK object set from an object set RID.
+ * @param client - An OSDK client.
```
ssanjay1, line 24: Don't think you need to add the dash actually, pretty sure it'll auto render like that but nbd.

ssanjay1 (dismissed):

PR #1109, opened 2025-01-22
Upgrade custom widgets Vite plugin to V6

rosscourt:
Upgrades the custom widgets Vite plugin to V6 of Vite, which was released in November last year.

The only breaking change to the API was in how HTML transformations from other plugins are made available to us. Now instead of looking at each plugin and having to look for a potential `transformIndexHtml` implementation, we just have access to a top level `transformIndexHtml` method on the `ViteDevServer` instance. The result of this is slightly different, being rendered script tags as a HTML string instead of the JSON `HtmlTagDescriptor` format.

I'm using [parse5](https://www.npmjs.com/package/parse5) to parse this HTML content and extract the script information. This library is one of the smaller options that achieves what we need.

### DIFF
--- a/packages/widget.vite-plugin.unstable/src/extractInjectedScripts.ts
+++ b/packages/widget.vite-plugin.unstable/src/extractInjectedScripts.ts
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { DefaultTreeAdapterTypes } from "parse5";
+import { parse as parse5 } from "parse5";
+import type { ViteDevServer } from "vite";
+
+export interface InjectedScripts {
+  scriptSources: string[];
+  inlineScripts: string[];
+}
+
+/**
+ * Extracts inline scripts usually injected by Vite plugins during HTML transformation.
+ *
+ * Vite plugins can inject scripts into the HTML entrypoint. This function captures
+ * those injections, specifically inline scripts, which are needed for our server-side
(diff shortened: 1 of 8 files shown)


packages/widget.vite-plugin.unstable/src/plugin.ts
```diff
@@ -575,11 +566,12 @@ function setWidgetSettings(
   localhostUrl: string,
   entrypointToJsSourceFileMap: Record<string, Set<string>>,
   entrypointFileName: string,
+  injectedScriptSources: string[],
 ) {
   const widgetDevModeSettings = {
     entrypointJs: [
       `/${VITE_INJECTIONS}`,
-      "/@vite/client",
```
tzyl, line 582: What is `injectedScriptSources` expected to contain now? Is it still `@vite/client`?

pnpm-lock.yaml
```diff
@@ -5173,6 +5185,12 @@ packages:
     cpu: [arm64]
     os: [android]
 
+  '@esbuild/android-arm64@0.24.2':
```
tzyl, line 5188: If we can get all of the other templates to v6 that would be great then these esbuild dependencies should collapse

packages/widget.vite-plugin.unstable/src/extractInjectedScripts.ts
```diff
@@ -0,0 +1,88 @@
...
+ * those injections, specifically inline scripts, which are needed for our server-side
+ * rendered pages. It calls the `transformIndexHtml` hook on each plugin, collects
+ * the script descriptors, and returns the concatenated inline script contents.
+ *
+ * See documentation: https://vite.dev/guide/api-plugin#transformindexhtml
+ */
+export async function extractInjectedScripts(
+  server: Pick<ViteDevServer, "transformIndexHtml">,
+): Promise<InjectedScripts> {
+  const result = await server.transformIndexHtml("", "");
```
tzyl, line 39: Do you see this as more stable than before having to go through all the plugins, even though the AST has to be parsed separately now?

tzyl (approved):

packages/widget.vite-plugin.unstable/src/plugin.ts
rosscourt: Yea, Vite uses the same mechanism to inject this script

packages/widget.vite-plugin.unstable/src/extractInjectedScripts.ts
rosscourt: Less likely to run into issues with the variations of return types that the previous API could return. That said I'd still like to consider using webpack instead as it doesn't do any HTML transformation which is much more consistent with our approach.

PR #1110, opened 2025-01-22
Support media upload 

mnayan:
Added support to upload media through media reference action parameters.


packages/client/src/object/mediaUpload.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function isMediaReference(o: any): o is MediaReference {
+  return typeof o === `object`
+    && typeof o.mimeType === "string"
+    && "reference" in o
+    && o.reference.type === "mediaSetViewItem";
+}
+
+export function createMediaUpload(args: {
```
mnayan, line 20: Wondering if we need this at all or have the users directly construct a mediaUpload type.

packages/client/src/util/toDataValue.ts
```diff
@@ -87,13 +104,17 @@ export async function toDataValue(
     return getWireObjectSet(value);
   }
 
+  if (isMediaReference(value)) {
+    return value;
```
ssanjay1, line 108: Just double checking, the action param requires the mimtype and the media ref rids?

packages/client/src/util/toDataValue.ts
```diff
@@ -87,13 +104,17 @@ export async function toDataValue(
...
+  }
+
   // TODO (during queries implementation)
   // two dimensional aggregation
   // three dimensional aggregation
 
   // struct
   if (typeof value === "object") {
-    return Object.entries(value).reduce(
+    return await Object.entries(value).reduce(
```
ssanjay1, line 117: Oh was this incorrect before?

packages/client/src/object/mediaUpload.ts
ssanjay1: Mmm kinda agree with you on this since you're not doing any extra processing. I don't think we need to keep in strict parity with the attachments workflow in this aspect, how about we leave this out for now and not export it as part of our API. And based on usage, we can add helpers if needed for some reason?

packages/api/src/object/Media.ts
```diff
@@ -25,6 +25,13 @@ export interface Media {
   fetchContents(): Promise<Response>;
 }
 
+export interface MediaUpload {
```
ssanjay1, line 28: So, just clarifying the workflow, a user will need to know what object type and property name they want to use right? Like the action param def doesn't tell you what media set that param should be looking at based on the underlying implementation? (E.g. if its a create action for `Foo` object type, we still need the user to tell us the `objectTypeApiName` right?)

packages/api/src/mapping/DataValueMapping.ts
```diff
@@ -34,6 +35,7 @@ export interface DataValueWireToClient {
   short: number;
   string: string;
   timestamp: string;
+  mediaReference: MediaUpload;
```
ssanjay1, line 38: If someone already has a MediaReference, we should probably let them pass that directly in too

ssanjay1 (commented):
Overall looks good! Left some conceptual questions/comments!

packages/api/src/mapping/DataValueMapping.ts
mnayan: Good point!

packages/api/src/object/Media.ts
mnayan: That's correct.

packages/client/src/util/toDataValue.ts
mnayan: Yup, it does.

packages/client/src/util/toDataValue.ts
mnayan: Need to double check. Maybe this got left out from testing.

packages/api/src/object/Media.ts
nihalbhatnagar: Will the underlying action/function throw an error if we pass in a different api name/property name than they expect? Basically, is the action tied to a specific objectType/propertyApiName and will throw if they don't get the one they expect or will they execute if you specify an objectType/propertyApiName that is a media set you have access to

packages/client/src/object/mediaUpload.ts
nihalbhatnagar: Agreed here as well

nihalbhatnagar (commented):
Overall looks good to me! Will defer to @ssanjay1 for approval

packages/api/src/object/Media.ts
mnayan: The action does not use the object/prop name but the media upload endpoint does - the new media needs to be uploaded to the media set which backs the media reference property on the object type. If the media is not in the correct backing media set, it will error out. 
We should do something similar to actions API has done in Foundry where the action definition has information about the affected objecttype and propertyId. However, this still wouldn't work for function backed actions which doesn't have information of the corresponding object type/prop name.

packages/api/src/mapping/DataValueMapping.ts
mnayan: discussed offline - we're going to do this in a follow-up PR.

packages/client/src/object/mediaUpload.ts
```diff
@@ -0,0 +1,34 @@
...
+    && "fileName" in o
+    && "objectTypeApiName" in o
+    && "propertyApiName" in o
+    && o.data instanceof Blob;
+}
+
+export function isMediaReference(o: any): o is MediaReference {
+  return typeof o === `object`
+    && typeof o.mimeType === "string"
+    && "reference" in o
```
ssanjay1, line 32: Should we also check that `typeof o.reference` is object before accessing?

ssanjay1: Also, related to the thread offline, since the wire types are broken right now, we should probably hold on merging this since we rely on this type internally from the return of upload

mnayan: Done

packages/client/src/object/mediaUpload.ts
```diff
@@ -0,0 +1,38 @@
...
+    && typeof o.reference === "object"
+    && o.reference.type === "mediaSetViewItem"
+    && typeof o.reference.mediaSetRid === "string"
```
ssanjay1, line 35: This is wrong now right? Because we now have the internal mediasetviewitem wrapper?

mnayan: Not until the new types flow through right?

packages/generator-converters/src/wireActionTypeV2ToSdkActionMetadata.ts
```diff
@@ -103,7 +104,7 @@ function actionPropertyToSdkPropertyDefinition(
       };
     default:
       throw new Error(
-        `Unsupported action parameter type: ${parameterType.type}`,
+        `Unsupported action parameter type: ${parameterType}`,
```
ssanjay1, line 107: Why did we remove `.type` here?

packages/api/src/experimental/createMediaReference.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @returns media reference of the uploaded media item.
+ */
+type createMediaReference = <
+  Q extends ObjectTypeDefinition,
+  const L extends PropertyKeys<Q>,
+>(
+  data: Blob,
+  fileName: string,
+  objectType: Q,
```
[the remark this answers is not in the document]
mnayan: Yeah, it's possible they could want to use action on another OT/Prop not in their SDK. Having the affected OT and prop in their SDK makes it less error prone - we can have stricter type checking.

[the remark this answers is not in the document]
mnayan: Ended up changing this to string.

packages/generator-converters/src/wireActionTypeV2ToSdkActionMetadata.ts
ssanjay1: Ok can we `JSON.stringify()` it then please? That way when people look at the logs it'll be more clear than `[Object object]`

ssanjay1 (approved):

PR #1113, opened 2025-01-23
Rework RDP types

ericanderson:
# Summary

* Reworked the types a bit to make it even easier to read/write by hand
* Wrote some type only tests in @osdk/api

# Effect of changes
## Return result from `.withProperties()`: 
image

## Return result from cascading `.withProperties()`:
image

## Cleaner "builder" name
image

# Things I would still change

## Final instance object

The end return type for a returned object is unmanageable right now:
image

Maybe we could get it to:
```ts
const a: Osdk.Instance<Employee, never, PropertyKeys<Employee>, {
    a: "integer" | undefined;
}>
```
or

Since the Rdp properties are implied selected (weird not to include them) maybe this?
```
const a: Osdk.Instance<WithRdp<Employee, {
    a: "integer" | undefined;
}>, never, PropertyKeys<Employee>>
```


packages/api/src/derivedProperties/Rdp.ts
```diff
@@ -0,0 +1,59 @@
...
+  BaseWithPropertyObjectSet,
+  WithPropertyObjectSet,
+} from "./WithPropertyObjectSet.js";
+
+export namespace Rdp {
+  export type SelectorResult<
+    T extends SimplePropertyDef,
+  > = {
+    type: T;
+  };
```
ericanderson, line 31: Was 
```ts
export type WithPropertyDefinition<
> = {
  type: T;
};
```

packages/api/src/derivedProperties/Rdp.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  export type Selector<
+    Q extends ObjectOrInterfaceDefinition,
+    T extends SimplePropertyDef,
+  > = (
+    baseObjectSet: Rdp.Builder<Q>,
+  ) => SelectorResult<T>;
+
+  export interface Builder<Q extends ObjectOrInterfaceDefinition> extends
+    // Fixme, just move the contents of BaseWithPropertyObjectSet in place
```
ericanderson, line 47: Didn't do this yet to keep the PR small.

packages/api/src/derivedProperties/Rdp.ts
```diff
@@ -0,0 +1,59 @@
...
+    T extends SimplePropertyDef,
+  > = {
+    type: T;
+  };
+
+  export type Clause<
+    Q extends ObjectOrInterfaceDefinition,
+  > = {
+    [key: string]: Selector<Q, SimplePropertyDef>;
+  };
```
ericanderson, line 37: Was:

```ts
export type WithPropertiesClause<
> = {
  [key: string]: (
    baseObjectSet: BaseWithPropertyObjectSet<Q>,
  ) => WithPropertyDefinition<ObjectMetadata.Property>;
};
```

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -142,9 +144,48 @@ export interface InterfaceObjectSet<
...
+export type ExtractRdp<
+  D extends ObjectSet<any, any> | Record<string, SimplePropertyDef>,
+> = D extends Record<string, SimplePropertyDef> ? D : {};
+
+export namespace ObjectSet {
+  export namespace Fns {
+    export interface WithProperties<
+      Q extends ObjectOrInterfaceDefinition = any,
+      D extends Record<string, SimplePropertyDef> = {},
+    > {
```
ericanderson, line 156: This needs to be split out so we can recombine multiple calls to `withProperties` without insane hoop jumping

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -142,9 +144,48 @@ export interface InterfaceObjectSet<
...
+      readonly withProperties: <
+        R extends Record<string, SimplePropertyDef>,
+      >(
+        clause: { [K in keyof R]: Rdp.Selector<Q, R[K]> },
+      ) => ObjectSet<
+        Q,
+        {
+          [NN in keyof R | keyof D]: NN extends keyof R ? R[NN]
```
ericanderson, line 164: Combine inline so we don't get a resolved type of
```ts
{ 
  foo: "string"
} & {
  bar: "string"
}
```

And instead get:
```ts
{ 
  foo: "string"
  bar: "string"
}
```

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -142,9 +144,48 @@ export interface InterfaceObjectSet<
...
+    any
+  >,
+> extends
+  ObjectSetPrime<MergeObjectSet<Q, D>>,
+  ObjectSet.Fns.WithProperties<Q, ExtractRdp<D>>
+{
+}
+
+interface ObjectSetPrime<
+  Q extends ObjectOrInterfaceDefinition = any,
```
ericanderson, line 188: Splitting meat out so we can run `MergeObjectSet<Q,D>` once instead of every place there was a `Q`.

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -261,45 +305,4 @@ export interface ObjectSet<
...
-        (baseObjectSet: any) => WithPropertyDefinition<infer P> ? PropertyDef<
-          P["type"],
-          "nullable",
-          P["multiplicity"] extends true ? "array" : "single"
-        >
-        : never;
-    }
-  >;
 }
-
```
ericanderson, line 282: These extra types aren't needed now that we can repurpose `ObjectSet`

packages/api/src/ontology/ObjectOrInterface.ts
```diff
@@ -16,11 +16,28 @@
...
+  > = {
+    __DefinitionMetadata: {
+      properties: {
+        [T in keyof D]: SimplePropertyDef.ToPropertyDef<D[T]>;
+      };
+      props: {
+        [T in keyof D]: SimplePropertyDef.ToRuntimeProperty<D[T]>;
+      };
+    };
+  } & K;
```
ericanderson, line 38: The intersection is fine here because this type never shows up anywhere

packages/api/src/ontology/ObjectOrInterface.ts
```diff
@@ -16,11 +16,28 @@
...
 import type { InterfaceDefinition } from "./InterfaceDefinition.js";
 import type { ObjectTypeDefinition } from "./ObjectTypeDefinition.js";
+import type { SimplePropertyDef } from "./SimplePropertyDef.js";
 
 export type ObjectOrInterfaceDefinition =
   | ObjectTypeDefinition
   | InterfaceDefinition;
 
+export namespace ObjectOrInterfaceDefinition {
+  export type WithRdp<
```
ericanderson, line 26: Given how long `ObjectOrInterfaceDefinition.WithRdp` is we should probably just shorten this special one to `WithRdp`

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -671,9 +665,9 @@ describe("ObjectSet", () => {
         // @ts-expect-error
         "derivedPropertyName": { "$eq": [1, 2] },
       });
-      expectTypeOf(setAggregationObjectSet).branded.toEqualTypeOf<
-        ObjectSetWithProperties<Employee, {
-          derivedPropertyName: PropertyDef<"string", "nullable", "array">;
+      expectTypeOf(setAggregationObjectSet).toEqualTypeOf<
+        ObjectSet<Employee, {
+          derivedPropertyName: "string"[] | undefined;
```
ericanderson, line 670: Can you hear the chorus of the angels? I can.

packages/api/src/derivedProperties/Rdp.ts
```diff
@@ -0,0 +1,59 @@
...
+import type {
+  ObjectOrInterfaceDefinition,
+} from "../ontology/ObjectOrInterface.js";
+import type { SimplePropertyDef } from "../ontology/SimplePropertyDef.js";
+import type {
+  BaseWithPropertyObjectSet,
+  WithPropertyObjectSet,
+} from "./WithPropertyObjectSet.js";
+
+export namespace Rdp {
```
nihalbhatnagar, line 26: Related to my other comment, but I think we should rename this away from "Rdp" if the types can be accessed on hover? I wish there was more consistent naming in the platform(doesn't seem like derived properties are it). I think WithProperty___ anything may be a bit wordy, but can't think of any meaningful alternatives

PR #1130, opened 2025-01-27
Rename create-widget widgetSetRid arg

tzyl:
`--widgetSetRid` -> `--widgetSet` for consistency with create-app `--application`. Also update the unstable client deps to `^2.0.0` which is the range they are being published at now matching the other public packages

### DIFF
--- a/packages/create-widget/src/templates.ts
+++ b/packages/create-widget/src/templates.ts
@@ -42,8 +42,5 @@ export interface Template {
 
 export interface TemplateContext {
   project: string;
-  foundryUrl: string;
   osdkPackage: string;
-  widgetSetRid: string;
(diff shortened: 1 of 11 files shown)


packages/create-widget/src/templates.ts
```diff
@@ -42,8 +42,5 @@ export interface Template {
 
 export interface TemplateContext {
   project: string;
-  foundryUrl: string;
   osdkPackage: string;
-  widgetSetRid: string;
-  clientVersion: string;
```
tzyl, line 48: These aren't currently being used in any of the template handlebars but we could add them back in if we do use them later

rosscourt (approved):

PR #1135, opened 2025-01-27
Remove where clause from aggregate

ssanjay1:
Removing where clause from aggregation options because we don't even pass this through to the backend now.

### DIFF
--- a/.changeset/wet-parents-check.md
+++ b/.changeset/wet-parents-check.md
@@ -3,4 +3,4 @@
 "@osdk/api": patch
 ---
 
-Remove where clause from aggregate options.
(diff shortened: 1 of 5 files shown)


.changeset/wet-parents-check.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/client": patch
+"@osdk/api": patch
+---
+
+Remove where clause from aggregate options.
```
ericanderson, line 6: Can we say more here? like "as it was never actually doing anything" or whatever sounds good?

ericanderson (changes requested):

ericanderson (approved):

PR #1138, opened 2025-01-28
Generator emits cjs

ericanderson:
PR is broken into 3 commits and I recommend you review them by commit.

1. Actually update the generator to produce valid cjs
2. All generated code to just use `.js` extensions everywhere
3. Tests that node10, node16-esm, and node16-cjs can compile


packages/foundry-sdk-generator/src/generate/betaClient/generatePackageJson.ts
```diff
@@ -42,21 +42,29 @@ export async function generatePackageJson(options: {
   const packageJson = {
     name: options.packageName,
     version: options.packageVersion,
-    main: "./index.js",
-    types: "./index.d.ts",
+    main: "./cjs/index.js",
```
ssanjay1, line 45: So are these two fields cjs now because node10 default looks at the main field?

.github/workflows/ci.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
   build:
     name: Build and Test
-    timeout-minutes: 15
+    timeout-minutes: 20
```
ssanjay1, line 71: Dang our tests taking long now

ssanjay1 (approved):
Left some questions above, but approving to unblock

packages/foundry-sdk-generator/src/generate/betaClient/compileInMemory.ts
```diff
@@ -39,9 +39,9 @@ export function compileInMemory(
 } {
   const inMemoryOutputFileSystem: { [fileName: string]: string } = {};
   const compilerOptions: CompilerOptions = {
-    module: ModuleKind.NodeNext,
+    module: type === "cjs" ? ModuleKind.CommonJS : ModuleKind.ES2022,
     target: ScriptTarget.ES2020,
-    moduleResolution: ModuleResolutionKind.NodeNext,
+    resolvePackageJsonExports: true,
```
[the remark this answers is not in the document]
ericanderson: Correct. So I want to be sure we are generating in a way that will work with Node16 so I turn this on.

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.test.ts
```diff
@@ -521,567 +521,567 @@ describe("generator", () => {
       expect(
         tweakedFilesForSnapshotConsistency(helper.getFiles()),
       ).toMatchInlineSnapshot(`
-      {
```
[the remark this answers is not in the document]
ericanderson: Yes, it just simplified things and I wrote a test project to be sure it didn't matter

packages/foundry-sdk-generator/src/generate/betaClient/generatePackageJson.ts
ericanderson: Yes

PR #1141, opened 2025-01-28
Fix fetch function 

ssanjay1:


.changeset/twenty-trees-warn.md
```diff
@@ -0,0 +1,38 @@
+---
+"@osdk/foundry.thirdpartyapplications": patch
```
ssanjay1, line 2: Hmm just saw this, shoudl I just downselect to the actually changed packages? Confused why changesets did this

ericanderson: it compares by default your current to main if I am not mistaken and so this has a lot of files that are different than main?

ericanderson (approved):
Approving but we shoudl fix the changest

.changeset/twenty-trees-warn.md
ssanjay1: Ah I see, yea just fixed

PR #1146, opened 2025-01-30
Adds cloneAndUpdate to Osdk Objects

nihalbhatnagar:
This method will be helpful for updating objects received back from subscriptions.

### DIFF
--- a/packages/api/src/OsdkObjectFrom.ts
+++ b/packages/api/src/OsdkObjectFrom.ts
@@ -204,15 +204,15 @@ export namespace Osdk {
         ConvertProps<Q, NEW_Q, P>
       >;
 
-      readonly $clone: (
+      readonly $clone: <NEW_PROPS extends PropertyKeys<Q>>(
         updatedObject?:
-          | Osdk.Instance<Q, any, any>
+          | Osdk.Instance<Q, any, NEW_PROPS>
           | {
-            [K in PropertyKeys<Q>]?: CompileTimeMetadata<
+            [K in NEW_PROPS]?: CompileTimeMetadata<
               Q
             >["props"][K];
           },
-      ) => Osdk.Instance<Q, OPTIONS, P>;
+      ) => Osdk.Instance<Q, OPTIONS, P | NEW_PROPS>;
     }
     // We are hiding the $rid field if it wasn't requested as we want to discourage its use
     & (IsNever<OPTIONS> extends true ? {}
--- a/packages/api/src/OsdkObjectFrom.ts
+++ b/packages/api/src/OsdkObjectFrom.ts
@@ -203,6 +203,16 @@ export namespace Osdk {
         OPTIONS,
         ConvertProps<Q, NEW_Q, P>
       >;
+
+      readonly $cloneAndUpdate: (
+        updatedObject:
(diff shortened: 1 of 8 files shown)


packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -61,6 +61,31 @@ const basePropDefs = {
...
   },
+  "$cloneAndUpdate": {
+    value: function(
+      this: InternalOsdkInstance & ObjectHolder<any>,
+      update: Record<string, any>,
+    ) {
+      const rawObj = this[UnderlyingOsdkObject];
+      const def = this[ObjectDefRef];
+
+      if (
```
nihalbhatnagar, line 72: I opted to throw here because I didn't want a mistake like this to go unnoticed, but I realize this may not be what we want to happen in production. I could instead throw only in develop and console.warn in develop? Also, understandably we can expose a more powerful cloneAndUpdate that will replace your primary key for you and hope you know what you're doing, but I'm slightly against that since I'd like to know upfront in develop if I do something wrong

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -61,6 +61,31 @@ const basePropDefs = {
...
+        throw new Error(
+          `Cannot update ${def.apiName} object with differing primary key values `,
+        );
+      }
+
+      if (def.titleProperty in update && !("$title" in update)) {
+        update.$title = update[def.titleProperty];
+      }
+
+      const newObject = { ...this[UnderlyingOsdkObject], ...update };
```
nihalbhatnagar, line 85: I don't filter undefineds, subscribe isn't going to give undefineds for properties it didn't recieve updates for but rather omit the property entirely, and if you get a new object with an undefined via a load I think it should update with that.

packages/api/src/OsdkObjectFrom.ts
```diff
@@ -203,6 +203,16 @@ export namespace Osdk {
...
+
+      readonly $cloneAndUpdate: (
+        updatedObject:
+          | Osdk.Instance<Q, any, any>
+          | {
+            [K in PropertyKeys<Q>]?: CompileTimeMetadata<
+              Q
+            >["props"][K];
+          },
+      ) => Osdk.Instance<Q, OPTIONS, P>;
```
nihalbhatnagar, line 215: We maintain the scope of the object we call cloneAndUpdate on always in the type, though the raw object may have more properties as well

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -61,6 +61,31 @@ const basePropDefs = {
...
+      return createOsdkObject(this[ClientRef], this[ObjectDefRef], newObject);
```
nihalbhatnagar, line 86: This also allows random properties to end up as attached to the object. I don't have strong feelings here but opted to expose this functionality to users. It would be easy to filter by known properties only though

ericanderson (dismissed):

PR #1149, opened 2025-01-30
fix oauth and update fetch packages

ssanjay1:
This PR does two things:
- Updates our oauth client creation so the URL is agnostic to having a trailing slash or not 
- Updates our platform SDK dependencies so we can use the new fetch logic


packages/oauth/src/common.ts
```diff
@@ -266,7 +266,7 @@ export function createAuthorizationServer(
     | "revocation_endpoint"
   >
 > {
-  const issuer = `${new URL(ctxPath, url + "/")}`;
```
ericanderson, line 269: was this not working?

ericanderson (approved):

packages/oauth/src/common.ts
ssanjay1: Yea because if we somehow had a url that already had a trailing slash, it added two trailing slashes

PR #1156, opened 2025-02-03
Bump dependency versions

ericanderson:
Upgrades a handful of libraries and standardizes on `@types/node` to reduce node_modules bloat. PR looks big but its almost entirely lock file


packages/e2e.sandbox.todoapp/src/CreateTodoForm.tsx
```diff
@@ -14,7 +14,7 @@ export default function CreateTodoForm({
   const [error, setError] = useState<string>();
 
   return (
-    <div className="flex flex-shrink mb-6">
```
ericanderson, line 17: done automatically by the tailwind upgrade tool

nihalbhatnagar (approved):

package.json
```diff
@@ -60,7 +60,9 @@
   },
   "pnpm": {
     "overrides": {
-      "trim@0.0.1": "0.0.3"
+      "trim@0.0.1": "0.0.3",
+      "cookie@0.6.0": "0.7.2",
```
nihalbhatnagar, line 64: I'm a bit confused about what this does

ericanderson: it means when any dep (including transitive) wants `cookie: "0.6.0"` give them `0.7.2` instead

PR #1161, opened 2025-02-04
Export interfaces for function signatures

tmisirpash:
Export interfaces for the following:

- Integer
- Long
- Float
- Double
- Date (represented as string)
- Timestamp (represented as string)
- TwoDimensionalAggregation
- ThreeDimensionalAggregation

Would a new package `@osdk/functions` be a better place to export these types?

### DIFF
--- a/packages/api/src/queries/Aggregations.ts
+++ b/packages/api/src/queries/Aggregations.ts
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type Range<T extends AllowedBucketTypes> = {
+  startValue?: T;
+  endValue: T;
+} | {
+  startValue: T;
+  endValue?: T;
+};
+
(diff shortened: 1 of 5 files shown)


packages/api/src/queries/Aggregations.ts
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+export type Range<T extends AllowedBucketTypes> = {
+  startValue?: T;
+  endValue: T;
+} | {
+  startValue: T;
+  endValue?: T;
+};
```
tmisirpash, line 23: Previously `startValue` and `endValue` were both typed as being required, but this isn't strictly the case -- either the start value or the end value can be omitted from the range.

packages/api/src/queries/Aggregations.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export type AllowedBucketTypes = string | number | boolean;
```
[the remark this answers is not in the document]
ericanderson: If I am understanding the question, the generated types reflect the reality? e.g. https://github.com/palantir/osdk-ts/blob/main/packages/e2e.generated.catchall/src/generatedNoCheck/ontology/queries/queryTakesAllParameterTypes.ts

packages/api/src/queries/Primitives.ts
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type Integer<T extends number = number> = T;
+export type Float<T extends number = number> = T;
+export type Double<T extends number = number> = T;
+export type Long<T extends string = string> = T;
```
ericanderson, line 20: I wonder if we should brand these? (Truly a question, Im wondering without a brand do the aliases get erased making the static analysis harder?)

tmisirpash: For analysis purpose branding these strings would definitely be nice to prevent the compiler from interning the types.

PR #1169, opened 2025-02-04
Modifying query agg return types

ssanjay1:
Fleshing out aggregation types on a query so you can explicitly set the key and value types

### DIFF
--- a/packages/api/src/mapping/DataValueMapping.ts
+++ b/packages/api/src/mapping/DataValueMapping.ts
@@ -25,6 +25,7 @@ export interface DataValueWireToClient {
   boolean: boolean;
   byte: number;
   datetime: string;
+  date: string;
   decimal: string;
   float: number;
   double: number;
@@ -57,6 +58,7 @@ export interface DataValueClientToWire {
   boolean: boolean;
   byte: number;
   datetime: string;
+  date: string;
   decimal: string | number;
   float: number;
   double: number;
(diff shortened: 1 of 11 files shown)


packages/client/src/queries/queries.test.ts
```diff
@@ -139,13 +139,25 @@ describe("queries", () => {
   it("two dimensional aggregation response works", async () => {
     const result = await client(twoDimensionalAggregationFunction)
       .executeFunction();
+    expectTypeOf<typeof result>().toEqualTypeOf<
```
ssanjay1, line 142: Added type tests now that we more explicity type agg requests/results

packages/api/src/mapping/DataValueMapping.ts
```diff
@@ -25,6 +25,7 @@ export interface DataValueWireToClient {
   boolean: boolean;
   byte: number;
   datetime: string;
+  date: string;
```
ssanjay1, line 28: Added this because the internal subtypes for aggs use date and was easier to just remap here rather than remap date to datetime like actions did

packages/api/src/queries/Aggregations.ts
```diff
@@ -0,0 +1,73 @@
...
+  startValue: T;
+  endValue?: T;
+};
+
+export type AllowedBucketTypes = string | number | boolean;
+export type AllowedBucketKeyTypes =
+  | AllowedBucketTypes
+  | Range<AllowedBucketTypes>;
+
+export type TwoDimensionalAggregation<
```
tmisirpash, line 40: Will `TwoDimensionalAggregation` and `ThreeDimensionalAggregation` be exported from the api package's `index.ts`? Or will the functions package use a submodule import to re-export these?

ssanjay1: We wil reexport these in the functions package

PR #1170, opened 2025-02-04
Release versions 1.1.0 of Custom Widget client libraries and align package names

rosscourt:
Manually bringing us back down from the beta release range that the rest of the repo is using so that we can get another `^1` range release for our repo templates to consume. Also aligns the package name prefix to `@osdk/widget.*` where we were using `@osdk/widget-*` sometimes


packages/example-generator/src/run.ts
```diff
@@ -334,12 +334,12 @@ const UPDATE_PACKAGE_JSON: Mutator = {
       .replace(
         // Use locally generated SDK in the monorepo
         /"@osdk\/widget-client-react.unstable": "\^.*?"/,
```
tzyl, line 336:
```suggestion
        /"@osdk\/widget.client-react.unstable": "\^.*?"/,
```

packages/example-generator/src/run.ts
```diff
@@ -334,12 +334,12 @@ const UPDATE_PACKAGE_JSON: Mutator = {
...
-        `"@osdk/widget-client-react.unstable": "workspace:*"`,
+        `"@osdk/widget.client-react.unstable": "workspace:*"`,
       )
       .replace(
         // Use locally generated SDK in the monorepo
         /"@osdk\/widget-client.unstable": "\^.*?"/,
```
tzyl, line 341:
```suggestion
        /"@osdk\/widget.client.unstable": "\^.*?"/,
```

tzyl (commented):
Just checking if the example-generator check might fail that the contents are not up to date if the regex isn't working correctly

tzyl (dismissed):

tzyl (approved):

PR #1182, opened 2025-02-06
Add nearest neighbors support

mvanschellebeeck:
Add `nearestNeighbor` query support. 

See [runNearestNeighborsTest.ts](https://github.com/palantir/osdk-ts/pull/1182/files#diff-31d042ffcf1fa89634f2b4ae860575e9eef86faed622968d1972eee7af61e501) for example usage


packages/api/src/object/FetchPageArgs.ts
```diff
@@ -39,9 +39,11 @@ export interface OrderByArg<
...
   L extends PropertyKeys<Q> = PropertyKeys<Q>,
 > {
-  $orderBy?: {
-    [K in L]?: "asc" | "desc";
-  };
+  $orderBy?:
+    | {
+      [K in L]?: "asc" | "desc";
+    }
+    | "relevance";
```
ssanjay1, line 46: Can we not sort by ascending or descending relevance as well?

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -254,4 +255,23 @@ export interface ObjectSet<
...
     opts?: ObjectSetListenerOptions<Q, P>,
   ) => { unsubscribe: () => void };
+
+  /**
+   * Finds the nearest neighbors for a given text or vector within the object set.
+   *
+   * @param query - Queries support either a vector matching the embedding model defined on the property, or text that is
+        automatically embedded.
+   * @param numNeighbors - The number of objects to return. If the number of documents in the objectType is less than the provided
+            value, all objects will be returned. This value is limited to 1 &le; numNeighbors &ge; 500.
```
ssanjay1, line 265: nit: I'm not actually sure how `&le `will render, does throwing in < not work?

packages/foundry-sdk-generator/src/ontologyMetadata/__snapshots__/metadataResolver.test.ts.snap
```diff
@@ -235,318 +235,28 @@ exports[`Load Ontologies Metadata > Loads action types with media refs, interfac
  This is some more text.",
               "rid": "rid",
             },
-            "startDate": {
```
ssanjay1, line 238: Hmm why this large diff in the snapshots?

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -147,6 +147,65 @@ describe("ObjectSet", () => {
     expect(iter).toEqual(2);
   });
 
+  it("nearest neighbors object set", async () => {
+    const numNeighbors = 3;
+    const nearestNeighborsObjectSet = client(Employee).nearestNeighbors(
```
ssanjay1, line 152: Can we add some type tests in here to make sure the type of the call signature is what we expect?

packages/api/src/objectSet/ObjectSet.ts
mvanschellebeeck: Yeah, when I used `<` the markdown would show:
```
 // Warning: (tsdoc-malformed-html-name) Invalid HTML element: Expecting an HTML name
```

packages/api/src/object/FetchPageArgs.ts
mvanschellebeeck: Nope.
By default, nearest neighbors returns objects with the highest scores first, so sorting by relevance is equivalent to sorting by score in descending order.

PR #1183, opened 2025-02-07
Ea.cache

ericanderson:


packages/api/src/OsdkObjectFrom.ts
```diff
@@ -206,7 +206,7 @@ export namespace Osdk {
 
       readonly $clone: <NEW_PROPS extends PropertyKeys<Q>>(
         updatedObject?:
-          | Osdk.Instance<Q, any, NEW_PROPS>
```
ssanjay1, line 209: Why did this change?

packages/client/src/public/unstable-do-not-use.ts
```diff
@@ -15,3 +15,11 @@
  */
 
 export { augment } from "../object/fetchPage.js";
+
+// THIS IS NOT THE FINAL NAME DO NOT SHIP LIKE THIS
```
ssanjay1, line 19: It's fine to bikeshed names while in unstable though right?

packages/api/src/OsdkObjectFrom.ts
ericanderson: this is the fix to make intellisense work right

ericanderson: Will take it out for safety

packages/client/src/public/unstable-do-not-use.ts
ericanderson: yeah

nihalbhatnagar (approved):

PR #1193, opened 2025-02-11
Add minimal React widget template without OSDK

rosscourt:
Adds a new widget template that doesn't use an OSDK. We'd like to add an option to skip OSDK generation in the widget creation process, and will likely restrict initial users to this while are a working out the security model for SDK based data loading.

I'm using Blueprint in this example as for embedded widgets I do think we should be encouraging design consistency with the parent application. I've also adapted the layout of this example so that its more flexible for varied widget container sizes.

Screenshot of the UI:
Screenshot 2025-02-11 at 10 48 23

PR #1200, opened 2025-02-12
FLUPs for minimal react widget template

rosscourt:
Addressing 
- https://github.com/palantir/osdk-ts/pull/1193#discussion_r1952509860
- https://github.com/palantir/osdk-ts/pull/1193#discussion_r1952520141

### DIFF
--- a/packages/create-widget/src/cli.test.ts
+++ b/packages/create-widget/src/cli.test.ts
@@ -51,6 +51,7 @@ for (const template of TEMPLATES) {
         project: `expected-${template.id}`,
         template,
         sdkVersion,
+        requiresOsdk: template.requiresOsdk,
       });
     });
   });
@@ -60,12 +61,14 @@ async function runTest({
   project,
   template,
   sdkVersion,
+  requiresOsdk,
 }: {
   project: string;
   template: Template;
   sdkVersion: string;
+  requiresOsdk: boolean;
 }): Promise<void> {
-  await cli([
+  let options = [
     "npx",
     "@osdk/create-widget",
     project,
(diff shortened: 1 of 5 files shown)


packages/create-widget/src/cli.test.ts
```diff
@@ -93,8 +103,4 @@ async function runTest({
   expect(fs.existsSync(path.join(process.cwd(), project, "README.md"))).toBe(
     true,
   );
-
-  const packageJson = JSON.parse(
-    fs.readFileSync(path.join(process.cwd(), project, "package.json"), "utf-8"),
-  );
```
rosscourt, line 99: This was unused

packages/create-widget/src/cli.test.ts
```diff
@@ -93,8 +103,9 @@ async function runTest({
...
     true,
   );
-
-  const packageJson = JSON.parse(
-    fs.readFileSync(path.join(process.cwd(), project, "package.json"), "utf-8"),
+  const packageJsonContents = fs.readFileSync(
+    path.join(process.cwd(), project, "package.json"),
+    "utf-8",
   );
+  expect(() => JSON.parse(packageJsonContents)).not.toThrow();
```
rosscourt, line 110: I believe this was the intention of this code, so this just makes that clearer

tzyl (dismissed):
Also remembering a minor comment that we can remove emitEvent from the deps here not sure if the eslint in the template sets up that rule:
https://github.com/palantir/osdk-ts/blob/255a8f10c2d6eae85d6a15bcf8675acb37c61e6c/examples/example-widget-minimal-react-sdk-2.x/src/Widget.tsx#L33C18-L33C33

packages/create-widget/src/cli.test.ts
tzyl: The create-app original copy was testing something with the client version:
https://github.com/palantir/osdk-ts/blob/255a8f10c2d6eae85d6a15bcf8675acb37c61e6c/packages/create-app/src/cli.test.ts#L114-L128

Otherwise the file existing was probably enough but no harm to check it's JSON

tzyl (approved):

PR #1204, opened 2025-02-13
Replace usage of internal.foundry packages

ssanjay1:


packages/e2e.sandbox.catchall/src/runFoundrySdkClientVerificationTest.ts
```diff
@@ -45,13 +45,13 @@ export async function runFoundrySdkClientVerificationTest(
     );
   }
 
-  if (!branchesResult.data.find(b => b.branchId === "master")) {
+  if (!branchesResult.data.find(b => b.name === "master")) {
```
ssanjay1, line 48: There was a break somewhere in the new datasets external api that we never had to deal with because we kept using internal, just changed branchId to name

packages/shared.test/src/handlers/loadObjectsEndpoints.ts
```diff
@@ -103,6 +103,30 @@ export const loadObjectsEndpoints: Array<RequestHandler> = [
     },
   ),
 
+  /**
+   * List ontologies
+   */
+  handleOpenApiCall(
```
ssanjay1, line 109: Needed these handlers since switched over to new endpoint

nihalbhatnagar (approved):

packages/e2e.sandbox.catchall/src/runFoundrySdkClientVerificationTest.ts
ssanjay1: Yea I think that should be fine

PR #1208, opened 2025-02-14
Use options overload for scopes parameter in create-app

nadprana:
Fixing scopes templating in create-app templates, we now use the options overload instead of defining defaults for other params

Also fixed the scopes prompt CLI as there was a faulty line causing the prompt to never show up unless the value was already supplied


examples/example-tutorial-todo-app-sdk-1.x/src/client.ts
```diff
@@ -6,6 +6,10 @@ const redirectUrl = import.meta.env.VITE_FOUNDRY_REDIRECT_URL;
 checkEnv(url, "VITE_FOUNDRY_API_URL");
 checkEnv(clientId, "VITE_FOUNDRY_CLIENT_ID");
 checkEnv(redirectUrl, "VITE_FOUNDRY_REDIRECT_URL");
+const scopes = [
+    "api:ontologies-read",
+    "api:ontologies-write",
```
tzyl, line 11: Is this the expected indentation to match the existing template? Should it be two spaces?

packages/create-app/src/prompts/promptScopes.ts
```diff
@@ -21,26 +21,27 @@ const scopeNameRegex = /^[a-zA-Z-_:]+$/;
 export async function promptScopes(
   { scopes }: { scopes?: string[] },
 ): Promise<string[] | undefined> {
-  if (scopes == null) {
-    return undefined;
-  }
-
   while (true) {
```
tzyl, line 24: Can you add a test for this following the other prompts?

packages/create-app/src/prompts/promptScopes.ts
```diff
@@ -21,26 +21,27 @@ const scopeNameRegex = /^[a-zA-Z-_:]+$/;
...
+      consola.fail(
+        `Scopes [ ${joinedInvalidScopes} ] are invalid. Scope names can only contain letters, hyphens, underscores, and colons`,
+      );
+    }
+    const stringScopes = await consola.prompt(
+      "Enter the scopes to request during OAuth:",
+      {
+        type: "text",
+        placeholder: "api:read-data api:write-data",
+        default: "api:read-data api:write-data",
```
tzyl, line 42: Just double checking are these the preferred defaults still?

nadprana: hm good point, i guess we can switch them over now actually!

examples/example-tutorial-todo-app-sdk-1.x/src/client.ts
nadprana: oh nice catch, looks like the older scopes templating had incorrect indentation as well so i went ahead and fixed all of them

packages/create-app.template.react.beta/templates/src/createClientAndAuth.ts.hbs
```diff
@@ -19,9 +19,9 @@ export default function createClientAndAuth() {
   const redirectUrl = import.meta.env.VITE_FOUNDRY_REDIRECT_URL;
   {{#if scopes}}
   const scopes = [
-    {{#each scopes}}
-      "{{this}}",
-    {{/each}}
+   {{#each scopes}}
+    "{{this}}",
+   {{/each}}
```
tzyl, line 24: I think this indentation is confusing to have the `#each` helper indented with one space in the context of this file (three spaces total), can we either keep this inline with the original indentation level (remove one space) or match the content (add one space)

```suggestion
  {{#each scopes}}
    "{{this}}",
  {{/each}}
```

Same applies for the all the rest of the templates!

examples/example-tutorial-todo-app-sdk-1.x/src/client.ts
tzyl: It's not really wrong as if it should be consistent then a formatter should be applied to it but may as well try to make this consistent as it's low lift

tzyl (approved):

ssanjay1 (approved):
Rubberstamping

PR #1209, opened 2025-02-14
Adding script to simulate minor release

ssanjay1:
This PR is split into 2 commits:
- Adding the script
- Running the script

**Details**

So, added a quick script that simulates a minor release, what it does is:
- Removes all changesets from pre.json file
- Deletes all old changesets
- Bump the initial versions in pre.json to whatever would have been released in the minor release
- Bump all package.json current versions to the next incremented beta version
- Add a new changeset for each package that simulates the next beta release

Once we actually cut the 2.1 release, will need to port over all the changelog.md changes to this branch as well, though that's not pressing.


packages/tool.release/src/simulateMinorBump.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import fs from "node:fs";
+import path from "node:path";
+import * as semver from "semver";
+
+export function simulateMinorBump(): void {
+  const cwd = process.cwd();
+
+  const packagesDir = path.join(cwd, "./packages");
+  const changesetsDir = path.join(cwd, "./.changeset");
```
ericanderson, line 25: fyi most tools are fine with just `./packages` if you want but I do like being explicit cause it makes the code easier to refactor and use for other things.

packages/tool.release/src/simulateMinorBump.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      // Increment minor version and add a beta tag, write back
+      const newVersion = semver.inc(version, "minor") + "-beta.1";
+      packageJson.version = newVersion;
+      fs.writeFileSync(
+        packageJsonPath,
+        JSON.stringify(packageJson, null, 2) + "\n",
+      );
+
+      // Add a changeset file that indicates a minor bump happened, write back
```
ericanderson, line 66: doesnt matter now but this could have been cleaner with a single changeset file.

ericanderson (approved):

packages/tool.release/src/simulateMinorBump.ts
ssanjay1: Oh, yea so true

PR #1213, opened 2025-02-18
bump a few versions

ericanderson:
`pnpm up` with a few additions where changes should be back compat.


packages/maker/src/cli/main.ts
```diff
@@ -107,22 +107,6 @@ export default async function main(
   }
 }
 
-async function loadOntologyViaJiti(input: string) {
```
ssanjay1, line 110: Is this just not being used anymore?

ssanjay1 (approved):

PR #1214, opened 2025-02-19
Add fetchAsBase64Url to media reference props

mnayan:
Added fetchAsBase64Url method to media reference properties that encodes the response as a base64 url.

### DIFF
--- a/packages/client/src/createMediaReferenceProperty.ts
+++ b/packages/client/src/createMediaReferenceProperty.ts
@@ -59,4 +59,16 @@ export class MediaReferencePropertyImpl implements Media {
       mediaType: r.mediaType,
     };
   }
+
+  public async fetchAsBase64Url(): Promise<string> {
+    const response = await this.fetchContents();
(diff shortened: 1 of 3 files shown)


packages/client/src/createMediaReferenceProperty.ts
```diff
@@ -59,4 +59,16 @@ export class MediaReferencePropertyImpl implements Media {
...
+  public async fetchAsBase64Url(): Promise<string> {
+    const response = await this.fetchContents();
+    if (!response.ok) {
+      throw new Error(`Error fetching media contents: ${response.status}`);
+    }
+    const arrayBuffer = await response.arrayBuffer();
+    const buffer = Buffer.from(arrayBuffer);
+    const base64String = buffer.toString("base64");
+    const mimeType = response.headers.get("content-type") || "image/jpeg";
+    return `data:${mimeType};base64,${base64String}`;
```
mnayan, line 72: We can also just return a the base64 string instead of url here.

PR #1219, opened 2025-02-19
Capture transitive CSS files in widget Vite builds

rosscourt:
Restores this logic from the original implementation of this plugin: https://github.com/palantir/osdk-ts/blob/121475a91919afc8365e7301cee97aa04e0beda8/packages/widget.vite-plugin.unstable/src/widgetSetManifest.ts#L76-L102

Some of the CSS files that should be included for a widget are not surfaced directly on the entrypoint chunk, but are instead present on the chunks imported by this entrypoint. Fortunately this appears to only require this single level of depth when extracting transitive CSS file references.

E.g. My working project has roughly this file setup:
```
src/
  - main.js
  - main.config.js
  - main.css
  - Widget.js
  - components/
    - Component.js
    - Component.css
    - Nested.js
    - Nested.css
```

And the build outputs are roughly:
```
main-[hash].js
main-[hash].css
component-[hash].js
component-[hash].css
```

Where any CSS files below the `main.js` file in the dependency tree is bundled up into the `component-[hash].css` file (i.e. `Component.css` + `Nested.css` and their imported NPM library CSS files). Currently the Vite plugin only captures the `main-[hash].css`, but with this change also captures `component-[hash].css`.

Took this opportunity to add some unit tests for this specific function which includes an example of the Vite chunks output in the above case.

### DIFF
--- a/packages/widget.vite-plugin.unstable/src/build-plugin/buildWidgetSetManifest.ts
+++ b/packages/widget.vite-plugin.unstable/src/build-plugin/buildWidgetSetManifest.ts
@@ -30,26 +30,15 @@ export function buildWidgetSetManifest(
   configFiles: Record<string, WidgetConfig<ParameterConfig>>,
   bundle: Rollup.OutputBundle,
 ): WidgetSetManifest {
-  const chunks = Object.values(bundle).filter(
-    (chunk): chunk is Rollup.OutputChunk => chunk.type === "chunk",
+  const entrypointChunks = Object.values(bundle).filter(
+    (chunk): chunk is Rollup.OutputChunk =>
+      chunk.type === "chunk" && chunk.isEntry && chunk.facadeModuleId != null,
   );
-
-  // Capture CSS imports for non-entrypoint chunks, to be matched against entrypoint chunks
-  const nonEntrypointChunkCssImports = chunks.filter(
-    chunk => !chunk.isEntry && (chunk.viteMetadata?.importedCss.size ?? 0) > 0,
-  ).reduce<Record<string, string[]>>((acc, chunk) => {
-    acc[chunk.fileName] = Array.from(chunk.viteMetadata?.importedCss ?? []);
-    return acc;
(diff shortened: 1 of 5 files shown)


packages/widget.vite-plugin.unstable/src/build-plugin/buildWidgetSetManifest.ts
```diff
@@ -84,8 +96,15 @@ function getChunkConfigFile(
 function widgetConfig(
   chunk: Rollup.OutputChunk,
   widgetConfig: WidgetConfig<ParameterConfig>,
+  nonEntrypointChunkCssImports: Record<string, string[]>,
 ): WidgetManifestConfig {
-  const cssFiles = Array.from(chunk.viteMetadata?.importedCss ?? []);
+  const directCssFiles = Array.from(chunk.viteMetadata?.importedCss ?? []);
+  const transitiveCssFiles = chunk.imports.flatMap((chunk) =>
+    nonEntrypointChunkCssImports[chunk] ?? []
+  );
```
tzyl, line 104: > Some of the CSS files that should be included for a widget are not surfaced directly on the entrypoint chunk, but are instead present on the chunks imported by this entrypoint. Fortunately this appears to only require this single level of depth when extracting transitive CSS file references.

Do you know why this happens? What's the reason for having the entrypoint css file and all other CSS gets bundled into one other file? Is it to do with if multiple entrypoints share common chunks but if that's the case I would have thought it would be possible to be more than one other file depending on how it decides to share chunks across the entrypoints.

Also, if you dynamically import a component which gets split into a different chunk then that's completely separate from this right as it falls under a separate non-entrypoint tree and the CSS will get correctly added to the page at import time?

I'm wondering if this isn't explicitly documented anywhere then we could modify to do an actual DFS traversal from the entrypoints to be exhaustive in case this changes or it's required in more complicated setups

rosscourt:
Closing in favor of #1230

PR #1222, opened 2025-02-20
Make createMediaReference typesafe

mnayan:
Added typesafety for objectType and propertyType params.

### DIFF
--- a/packages/api/src/experimental/createMediaReference.ts
+++ b/packages/api/src/experimental/createMediaReference.ts
@@ -15,7 +15,7 @@
  */
 
 import type { MediaReference } from "../object/Media.js";
-import type { MediaPropertyKeys } from "../ontology/MediaPropertyKeys.js";
+import type { FilteredPropertyKeys } from "../ontology/FilteredPropertyKeys.js";
 import type { ObjectTypeDefinition } from "../ontology/ObjectTypeDefinition.js";
 import type { Experiment } from "./Experiment.js";
 
@@ -33,7 +33,7 @@ import type { Experiment } from "./Experiment.js";
  */
 type createMediaReference = <
   Q extends ObjectTypeDefinition,
-  const L extends MediaPropertyKeys<Q>,
+  const L extends FilteredPropertyKeys<Q, "mediaReference">,
 >(
   args: {
     data: Blob;
--- a/packages/api/src/experimental/createMediaReference.ts
+++ b/packages/api/src/experimental/createMediaReference.ts
@@ -15,7 +15,7 @@
  */
 
 import type { MediaReference } from "../object/Media.js";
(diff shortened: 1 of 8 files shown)


packages/api/src/experimental/createMediaReference.ts
```diff
@@ -24,17 +26,20 @@ import type { Experiment } from "./Experiment.js";
...
- * @param propertyTypeApi - Api name of the media reference property of the corresponding object type to which the media item will be uploaded.
+ * @param objectType - Object type to which the media item will be uploaded.
+ * @param propertyType - Media reference property of the corresponding object type to which the media item will be uploaded.
  *
  * @returns media reference of the uploaded media item.
  */
-type createMediaReference = (
+type createMediaReference = <
+  Q extends ObjectTypeDefinition,
+  const L extends PropertyKeys<Q>,
```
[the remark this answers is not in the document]
mnayan: @ssanjay1 - updated to use the above type.

packages/e2e.sandbox.catchall/src/runMediaTest.ts
```diff
@@ -40,9 +40,9 @@ export async function runMediaTest(): Promise<void> {
         )
           .createMediaReference({
             data: await response.blob(),
-            fileName: "test13.png",
-            objectTypeApi: MnayanOsdkMediaObject.apiName,
-            propertyTypeApi: "mediaReference",
```
mnayan, line 45: Should we do a minor version bump since these have changed? Or should we keep the prop names same?

ssanjay1: Yea we hid this behind the experimental flag for this reason, fine to break. Let's just keep as a patch

packages/api/src/ontology/MediaPropertyKeys.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+import type { ObjectOrInterfaceDefinition } from "./ObjectOrInterface.js";
+
+export type MediaPropertyKeys<
+  O extends ObjectOrInterfaceDefinition,
+> = {
+  [K in keyof NonNullable<O["__DefinitionMetadata"]>["properties"]]:
+    NonNullable<O["__DefinitionMetadata"]>["properties"][K]["type"] extends
+      "mediaReference" ? K : never;
```
ssanjay1, line 24: 1. Can we add some type tests to make sure this type is working as intended?
2. I think it may be good to actually generalize this type potentially to accept any wirepropertytype as another generic to then limit the property keys to, maybe like `FilteredPropertyKeys<O extends ObjectOrInt, T extends WirePropertyTypes>`

nihalbhatnagar (approved):

PR #1224, opened 2025-02-20
Allow selecting derived properties

nihalbhatnagar:


packages/api/src/derivedProperties/WithPropertiesAggregationOptions.ts
```diff
@@ -30,5 +29,6 @@ export type NumericWithPropAggregateOption =
   | "max"
   | "sum"
   | "avg"
+  | "approximatePercentile"
```
nihalbhatnagar, line 32: This was wrong before

ericanderson (approved):

PR #1230, opened 2025-02-21
Alternative approach to Vite build processing

rosscourt:
An alternative approach to #1219 which should be more robust and less reliant on trying to match Vite internals.

The approach in my other PR isn't actually complete, as Vite does a DFS from the entrypoint in order to find all CSS references, whereas my approach and the previous implementation of this only traverses one layer of depth. There is also another code path in Vite when the `cssCodeSplit‎` setting is set to false, and the way we'd need to detect CSS files in that case is very different.

- Vite's DFS for CSS files: https://github.com/vitejs/vite/blob/f6926caa1f2c9433ca544172378412795722d8e1/packages/vite/src/node/plugins/html.ts#L797
- Vite's handling of disabled CSS code splitting: https://github.com/vitejs/vite/blob/f6926caa1f2c9433ca544172378412795722d8e1/packages/vite/src/node/plugins/html.ts#L931

Instead this approach lets Vite do its usual thing to the entrypoint HTML file, and then we extract the information we need from this. I expect this to be a much more reliable way of matching the build output.


packages/widget.vite-plugin.unstable/src/build-plugin/FoundryWidgetBuildPlugin.ts
```diff
@@ -54,27 +75,33 @@ export function FoundryWidgetBuildPlugin(): Plugin {
      *
      * Write the manifest to the expected location in the dist directory.
      */
-    async generateBundle(_, bundle) {
+    async writeBundle(_, bundle) {
```
rosscourt, line 78: This step happens after the HTML files have been written (we don't have a hook to access them post-transformation)

packages/widget.vite-plugin.unstable/src/build-plugin/FoundryWidgetBuildPlugin.ts
```diff
@@ -54,27 +75,33 @@ export function FoundryWidgetBuildPlugin(): Plugin {
...
-      });
+      const manifestPath = path.join(
+        config.build.outDir,
+        MANIFEST_FILE_LOCATION,
+      );
+      fs.mkdirSync(path.dirname(manifestPath), { recursive: true });
+      fs.writeFileSync(
+        manifestPath,
+        JSON.stringify(widgetSetManifest, null, 2),
+      );
```
rosscourt, line 104: The `emitFile` method isn't available anymore here

tzyl: Does this have any effects on the build? Would it no longer show up in the vite stats about the bundle for example as it's written without any way for vite to track it

packages/widget.vite-plugin.unstable/src/build-plugin/extractBuildOutputs.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (node.nodeName === "link" && getAttribute(node, "rel") === "stylesheet") {
+    return [parseStylesheetNode(node as DefaultTreeAdapterTypes.Element)];
+  }
+  if ("childNodes" in node) {
+    return node.childNodes.flatMap((childNode) => visitNode(childNode));
+  }
+  return [];
+}
+
+function parseScriptNode(
```
tzyl, line 75: Should we be stricter on this to not allow other script type attribute values and other script attributes? I'm not sure if it's bad to break on the assumption of what tag structure vite will inject into the html but also at the moment we would let through a lot of stuff perhaps unintentionally:

- Only allow script tag if type attribute is module, javascript mime type or empty??
- Only allow src attribute and no others
- Disallow any text content inside the tag

packages/widget.vite-plugin.unstable/src/build-plugin/extractBuildOutputs.ts
```diff
@@ -0,0 +1,108 @@
...
+    throw new Error("Inline scripts are not supported");
+  }
+  return {
+    type: "script",
+    scriptType: typeAttribute ?? "text/javascript",
+    src: srcAttribute,
+  };
+}
+
+function parseStylesheetNode(
```
tzyl, line 92: Same question on here on whether it would be intentional to extract stylesheets with extra attributes:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#conditionally_loading_resources_with_media_queries

tzyl (dismissed):

packages/widget.vite-plugin.unstable/src/build-plugin/extractBuildOutputs.ts
```diff
@@ -75,12 +75,14 @@ function visitNode(node: DefaultTreeAdapterTypes.Node): BuildOutput[] {
...
-  const srcAttribute = getAttribute(node, "src");
-  const typeAttribute = getAttribute(node, "type") as
-    | ScriptType
-    | undefined;
-  if (srcAttribute == null) {
-    throw new Error("Inline scripts are not supported");
+  assertEmptyNode(node);
+  assertAttributes(node, ["src", "type"]);
+  const srcAttribute = getAttribute(node, "src")!;
+  const typeAttribute = getAttribute(node, "type")!;
```
tzyl, line 81: We could also handle empty string or undefined for "classic" scripts:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type#attribute_is_not_set_default_an_empty_string_or_a_javascript_mime_type

packages/widget.vite-plugin.unstable/src/build-plugin/__tests__/extractBuildOutputs.test.ts
```diff
@@ -66,7 +66,33 @@ test("extractBuildOutputs throws error for inline scripts", () => {
...
+        <script src="/script.js" type="text/unknown"></script>
+      </head>
+    </html>
+  `);
+  expect(() => extractBuildOutputs("dummy/path.html")).toThrow(
+    "Invalid script type attribute found in Vite HTML output: text/unknown",
+  );
+});
+
+test("extractBuildOutputs throws error for unexpected attributes", () => {
```
tzyl, line 86: Add a case for the stylesheet unexpected attributes too?

tzyl (dismissed):

ssanjay1 (commented):
Rubber stamping

PR #1232, opened 2025-02-21
Ontology edits

marshallavail:
Adds a set of types and a helper for creating ontology edits (similar to the Python SDK) for use in Functions.


palantirtech:
Thanks for your interest in [palantir/osdk-ts](https://github.com/palantir/osdk-ts), @marshallavail! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/client/src/edits/createEditBatch.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  public getEdits(): X[] {
+    return this.edits;
+  }
+}
+
+export function createEditBatch<T extends AnyEdit>(
+  _client: Client,
+  _editScope?: T[],
```
marshallavail, line 92: I made this optional. That will allow users to (if they want to) create their edit batch the way I did in the unit test:
```
const editBatch: EditBatch<MyEditScope> = createEditBatch(client);
```

packages/client/src/edits/createEditBatch.ts
```diff
@@ -0,0 +1,95 @@
...
+  AddLinkEdits,
+  AnyEdit,
+  CreateObjectEdits,
+  DeleteObjectEdits,
+  RemoveLinkEdits,
+  UpdateObjectEdits,
+} from "./types.js";
+
+class InMemoryEditBatch<X extends AnyEdit = never> implements EditBatch<X> {
+  private edits: X[] = [];
```
marshallavail, line 29: According to Actions folks, we don't need to compress edits or detect invalid edit sequences ourselves like we do in the legacy TypeScript SDK and the Python OSDK. Will link a thread to this discussion via DM.

packages/client/src/edits/createEditBatch.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  | RemoveLink<Task, "RP">;
+
+describe(createEditBatch, () => {
+  let editBatch: EditBatch<TestEditScope>;
+
+  beforeEach(() => {
+    editBatch = createEditBatch(createClientMockHelper().client);
+  });
+
+  it("collects all edits", () => {
```
marshallavail, line 44: Note that for all of the operations below, you can also just pass in an `Osdk.Instance<...>`
For example:

```
editBatch.create(Task, { id: 99, ...someExistingInstance });
```
Lets you create a copy of some existing object instance.

```
editBatch.delete(someExistingInstance);
```
Lets you delete some existing object instance.

```
editBatch.update(instanceA, instanceB);
```
Allows you to update instance A with the (non-primary key) properties of instance B

packages/client/src/edits/createEditBatch.ts
marshallavail: I added separate unit tests for `buildEditScope` which I feel is sufficient now
That said, I do think we are leaning towards a UX where people use the edit scope builder.

packages/functions/src/edits/buildEditScope.ts
```diff
@@ -0,0 +1,55 @@
...
+  ) => EditScopeBuilder<T | DeleteObject<S>>;
+
+  update: <S extends ObjectTypeDefinition>(
+    s: S,
+  ) => EditScopeBuilder<T | UpdateObject<S>>;
+
+  build: () => T[];
+}
+
+declare function buildEditScope(): EditScopeBuilder;
```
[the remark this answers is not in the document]
marshallavail: Yeah, that's MB. I added a no-op implementation. Not sure if that's the cleanest way to do this though.

packages/functions/src/edits/createEditBatch.test.ts
```diff
@@ -0,0 +1,108 @@
...
+describe(createEditBatch, () => {
+  let client: Client;
+  let editBatch: EditBatch<TestEditScope>;
+
+  beforeEach(() => {
+    client = vi.fn() as any as Client;
+    editBatch = createEditBatch(client);
+  });
+
+  it("collects all edits", () => {
```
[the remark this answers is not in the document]
marshallavail: Done!

packages/functions/package.json
```diff
@@ -35,13 +35,16 @@
...
+    "test": "vitest run --pool=forks",
     "transpileBrowser": "monorepo.tool.transpile -f esm -m normal -t browser",
     "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
     "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
     "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
+    "@osdk/api": "workspace:~",
+    "@osdk/client": "workspace:~",
```
tmisirpash, line 47: Probably a question for @ssanjay1 , but do we need these additional dependencies here?

tmisirpash: I'm referring to `@types/geojson`, `fetch-retry`, etc.

packages/functions/src/edits/createEditBatch.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  | AddLink<Task, "RP">
+  | RemoveLink<Task, "RP">;
+
+describe(createEditBatch, () => {
+  let client: Client;
+  let editBatch: EditBatch<TestEditScope>;
+
+  beforeEach(() => {
+    client = vi.fn() as any as Client;
+    editBatch = createEditBatch(client);
```
tmisirpash, line 43: Should we also be passing the edits scope produced by the builder into this client if that's how we expect people to use it?

packages/functions/src/edits/buildEditScope.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public update<S extends ObjectTypeDefinition>(
+    s: S,
+  ): EditScopeBuilder<T | UpdateObject<S>> {
+    return this;
+  }
+
+  public build(): T[] {
+    return [];
```
tmisirpash, line 64: Hm, it does feel slightly odd to me that we're returning an empty list here, it seems generally useful to have the edit scope as a concrete value e.g. for logging purposes. Deferring to @ericanderson and @ssanjay1 on this, though.

packages/functions/src/edits/createEditBatch.ts
```diff
@@ -0,0 +1,94 @@
...
+  public getEdits(): X[] {
+    return [...this.edits];
+  }
+}
+
+export function createEditBatch<T extends AnyEdit>(
+  _client: Client,
+): EditBatch<T> {
+  return new InMemoryEditBatch<T>();
+}
```
marshallavail, line 94: Example usage in Functions:
```
type AssignOnboardingTaskEdit =
  | Edits.Object<Task>
  | Edits.Link<Task, "RP">;

function assignOnboardingTask(
  client: Client,
  emp: Osdk.Instance<Employee>,
): AssignOnboardingTaskEdit[] {
  const editBatch = createEditBatch<AssignOnboardingTaskEdit>(client);
  const newTask = {...}
  editBatch.create(Task, newTask);
  editBatch.link(newTask, "RP", emp);
  return editBatch.getEdits();
}
```

tmisirpash (dismissed):
Happy to start out with this! CC @ericanderson and @ssanjay1 for their sign-off.

packages/functions/src/edits/createEditBatch.ts
tmisirpash: I'm personally not against having an API that looks like

```typescript
const obj = editBatch.create(Task, { id: 0 });
editBatch.update(obj, { name: "New Name" });
```

My preference would probably be to make it as clear as possible that we're just populating a store and not modifying objects directly. If the API becomes more oriented around working with concrete object instances, my concern is that we muddy things a bit, but I don't think it's too bad as long as it's still clear we're updating the batch, not the object. (For example I'd be more concerned about `obj.name = "New Name";`.)

packages/functions/package.json
marshallavail: Removed in 8df151f9a2ef6c8ac27fe2f7f58658655bea32a8

packages/functions/src/edits/createEditBatch.ts
marshallavail: I'm not even sure we can return a fully-fledged `Osdk.Instance` since I would need to populate fields like `$objectType` which I don't think we get from inputs to `create` like `create(Task, { id: 0 })`

We can revisit this later, but I'm also slightly concerned about muddying up the true functionality of the edit batch (i.e., it doesn't actually create any object instances itself, it just stores a proposal for an object creation).

packages/functions/src/edits/EditBatch.ts
```diff
@@ -0,0 +1,54 @@
...
+    obj: O["obj"],
+  ) => void;
+
+  update: <O extends UpdateObjectEdits<X>>(
+    obj: O["obj"],
+    properties: O["properties"],
+  ) => void;
+
+  getEdits: () => X[];
+}
```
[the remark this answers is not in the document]
marshallavail: Ack!

packages/functions/src/edits/createEditBatch.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  beforeEach(() => {
+    client = vi.fn() as any as Client;
+    editBatch = createEditBatch<TestEditScope>(client);
+  });
+
+  it("collects all edits", () => {
+    editBatch.create(Task, { id: 0, name: "My Task Name" });
+    editBatch.create(Task, { id: 1, name: "My Other Task Name" });
+    editBatch.delete({ apiName: "Task", id: 0 });
+    editBatch.update({ apiName: "Task", id: 0 }, { name: "My New Task Name" });
```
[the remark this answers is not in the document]
marshallavail: Ack!

examples/example-next-static-export-sdk-1.x/.next/cache/.rscinfo
```diff
@@ -0,0 +1 @@
+{"encryption.key":"r4a0qb5yIowxjOO7rNMsomCCUp037/b5HJgaTDqx7Ic=","encryption.expire_at":1741367325745}
```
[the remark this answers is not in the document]
marshallavail: From a bad merge. I rebased and force-pushed!

ericanderson (approved):

PR #1240, opened 2025-02-25
Add Object Specifier and Function Map Type Generation to the OSDK

nihalbhatnagar:
This PR adds the ObjectSpecifier concept to the OSDK. This was done in order to support Function Map type generation to the OSDK.

It's probably best reviewable commit by commit to isolate some of the changes


packages/client/src/object/createObjectSpecifierFromPrimaryKey.ts
```diff
@@ -0,0 +1,27 @@
...
+import type {
+  ObjectOrInterfaceDefinition,
+  ObjectSpecifier,
+  PrimaryKeyType,
+} from "@osdk/api";
+
+export function createObjectSpecifierFromPrimaryKey<
+  Q extends ObjectOrInterfaceDefinition,
+>(def: Q, primaryKey: PrimaryKeyType<Q>): ObjectSpecifier<Q> {
+  return `${def.apiName}:${primaryKey}` as ObjectSpecifier<Q>;
```
nihalbhatnagar, line 26: Maybe we should codify the shape of this somewhere? I want to make creating these manually not suggested so that we can experiment with the shape of this.

nihalbhatnagar: Also are we okay with the "apiName:primaryKey" shape? Do we want brackets around it?

packages/api/src/ontology/ObjectSpecifier.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import type {
+  InterfaceDefinition,
+  InterfaceMetadata,
+} from "./InterfaceDefinition.js";
+import type { ObjectOrInterfaceDefinition } from "./ObjectOrInterface.js";
+
+export type ObjectSpecifier<Q extends ObjectOrInterfaceDefinition> = string & {
+  __apiName?:
```
ssanjay1, line 24: Talked offline, but think it would be good to make this non-optional so people aren't just passing in random strings and are forced to use our type handler.

packages/client/src/object/createObjectSpecifierFromPrimaryKey.ts
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+import type {
+  ObjectOrInterfaceDefinition,
+  ObjectSpecifier,
+  PrimaryKeyType,
+} from "@osdk/api";
+
+export function createObjectSpecifierFromPrimaryKey<
```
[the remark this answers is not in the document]
nihalbhatnagar: I thought about doing this, but I kind of wanted to see if we got any feedback from users that actually want this. Although I can see if you get a object specifier back as a key, you'd potentially want to get the PK from it? Any objections to feeling out the usage?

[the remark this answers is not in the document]
tmisirpash: Functions would definitely benefit from having a helper in the runtime that takes an `ObjectSpecifier` and allows us to extract both the API name and primary key from the specifier, but we could also hard-code it to split by `:` or something similar.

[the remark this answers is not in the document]
nihalbhatnagar: Hmm ok, can I FLUP this? I think for now the shape will remain stable and we can do the `:` split, and I'll work on getting these right. I want to think about if these should be made available a bit more

packages/api/src/ontology/ObjectSpecifier.test.ts
```diff
@@ -0,0 +1,96 @@
...
+describe("ObjectSpecifier", () => {
+  const fauxObjectSet = {
+    fetchOne: vi.fn(() => Promise.resolve(() => "")),
+  } as any as Employee.ObjectSet;
+
+  it("should only work with same ObjectID", async () => {
+    type EmployeeObjectId = ObjectSpecifier<Employee>;
+
+    expectTypeOf<EmployeeObjectId>().toMatchTypeOf<
+      string & { __apiName?: "Employee" }
```
ssanjay1, line 31: The brand shouldn't be optional right?

ssanjay1 (approved):

ssanjay1 (approved):

PR #1244, opened 2025-02-25
Nihalb/object set to string

### LINKED ISSUE #1143: Provide a mechanism to get the representation of an ObjectSet
Either use `toString()` or something similar so that you can:

```
function foo() {
  const objectSet = buildObjectSet();
  console.log(objectSet.toString())
}
```

nihalbhatnagar:
Fixes #1143 . I opted to just stringify the underlying wire object set instead of doing any custom formatting here.

### DIFF
--- a/.changeset/spotty-dogs-jog.md
+++ b/.changeset/spotty-dogs-jog.md
@@ -1,0 +1,6 @@
+---
+"@osdk/client": patch
+"@osdk/api": patch
+---
+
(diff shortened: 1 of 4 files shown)


ericanderson (changes requested):
I think we should probably get this to our cleaned up version. Once we put this out I feel like people will expect the string format. 

We probably should just rework the internals a bit so we store our own format instead of the final.

PR #1248, opened 2025-02-27
Add nearest neighbors support

mvanschellebeeck:
Add nearestNeighbor query support.

See [runNearestNeighborsTest.ts](https://github.com/palantir/osdk-ts/blob/8a05287267be41091c26ab00b8696cfcb1f52566/packages/e2e.sandbox.catchall/src/runNearestNeighborsTest.ts) for example usage


packages/api/src/OsdkBase.ts
```diff
@@ -30,6 +30,8 @@ export type OsdkBase<
   readonly $primaryKey: PrimaryKeyType<Q>;
 
   readonly $title: string | undefined;
+
+  readonly $score?: number | undefined;
```
nihalbhatnagar, line 34: I don't think we should have this belong in OSDK base, since I imagine this will be undefined 90% of the time. What if instead we passed through a generic through the ObjectSet type that determined if we were making a nearest neighbors query, and conditionally added it to the Osdk.Instance type?

nihalbhatnagar: Or actually looking further, it seems we only want score to be returned when ordering by relevance? Which makes it easier I think, we don't need to pass through a generic we can just add it to the type from a fetchPage result

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -39,9 +39,11 @@ export interface OrderByArg<
...
   L extends PropertyKeys<Q> = PropertyKeys<Q>,
 > {
-  $orderBy?: {
-    [K in L]?: "asc" | "desc";
-  };
+  $orderBy?:
+    | {
+      [K in L]?: "asc" | "desc";
+    }
+    | "relevance";
```
nihalbhatnagar, line 46: If we order by relevance but we haven't called nearestNeighbors yet, do we throw?

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -209,6 +210,27 @@ interface WithProperties<
   >;
 }
 
+interface NearestNeighbors<Q extends ObjectOrInterfaceDefinition> {
+  /**
+   * Finds the nearest neighbors for a given text or vector within the object set.
+   *
+   * @param query - Queries support either a vector matching the embedding model defined on the property, or text that is
+        automatically embedded.
```
nihalbhatnagar, line 218: This is cool

.changeset/rotten-scissors-wonder.md
```diff
@@ -0,0 +1,9 @@
+---
+"@osdk/foundry-sdk-generator": minor
```
ssanjay1, line 2: Can we make these all patch

packages/api/src/ontology/ObjectOrInterface.ts
```diff
@@ -38,9 +39,21 @@ export namespace DerivedObjectOrInterfaceDefinition {
...
 export type PropertyKeys<
   O extends ObjectOrInterfaceDefinition,
   RDPs extends Record<string, SimplePropertyDef> = {},
 > =
-  & (keyof NonNullable<O["__DefinitionMetadata"]>["properties"] | keyof RDPs)
+  & (keyof Properties<O> | keyof RDPs)
   & string;
+
+export type VectorType = Extract<BaseWirePropertyTypes, "vector">;
+export type VectorPropertyKeys<O extends ObjectOrInterfaceDefinition> = keyof {
```
ssanjay1, line 54: Can you use our `FilteredPropertyKeys` type?

packages/api/src/object/FetchPageArgs.ts
ssanjay1: Also, do we imagine that there will ever be any extra modifiers for relevance ordering? Trying to think if this should be a string or object option

mvanschellebeeck: > Also, do we imagine that there will ever be any extra modifiers for relevance ordering? Trying to think if this should be a string or object option

This is how its modeled in the internal APIs so any added modifiers there would be a breaking change

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -209,6 +210,27 @@ interface WithProperties<
...
+   * @param numNeighbors - The number of objects to return. If the number of documents in the objectType is less than the provided
+            value, all objects will be returned. This value is limited to 1 &le; numNeighbors &ge; 500.
+   * @param property - The property key with a defined embedding model to search over.
+   *
+   * @returns An object set containing the `numNeighbors` nearest neighbors. To return the objects ordered by relevance and each
```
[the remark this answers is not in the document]
mvanschellebeeck: It operates the same way as nested unions/intersects in that it executes from inner to outer and returns the outer result. Added an example in https://github.com/palantir/osdk-ts/pull/1248/commits/25380158f9daae24e3b44d14ef5a2298aa8adc20

packages/api/src/OsdkBase.ts
mvanschellebeeck: I'm having trouble implementing the types here - would it be the same approach taken with `$rid` in https://github.com/palantir/osdk-ts/blob/main/packages/api/src/OsdkObjectFrom.ts#L224-L228?

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -35,13 +35,21 @@ export interface SelectArg<
   $includeRid?: R;
 }
 
+export type OrderByType<
```
ssanjay1, line 38: nit: can we rename this `OrderByOptions?`

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -174,6 +174,125 @@ describe("ObjectSet", () => {
...
+    const nearestNeighborsObjectSet = client(Employee).nearestNeighbors(
+      "python3",
+      numNeighbors,
+      "skillSetEmbedding",
+    );
+    const { data: employees } = await nearestNeighborsObjectSet.fetchPage();
+    expect(employees).toHaveLength(numNeighbors);
+    // Check that no score is returned when not ordered by relevance
+    // @ts-ignore
+    employees.forEach(e => expect(e.$score).toBeUndefined());
```
ssanjay1, line 188: What did you have to ts-ignore here?

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -246,6 +248,51 @@ describe("ObjectSet", () => {
...
   });
 
+  it("type checking nearestNeighbor return type", async () => {
+    const nearestNeighbors = client(Employee).nearestNeighbors(
+      "python3",
+      3,
+      "skillSetEmbedding",
+    );
+
+    const { data: unOrdered } = await nearestNeighbors.fetchPage();
```
nihalbhatnagar, line 258: Do you mind doing a sanity check here and trying to access .$score on one of the resulting objects (and adding //ts-expect-error) or actually for most of these cases

nihalbhatnagar: I don't think we have type tests for OsdkObjectFrom but if we do that might be a better place to put it

packages/client/src/objectSet/ObjectSet.test.ts
ssanjay1: Oh is it because $score isn't even a property on the object? If so, could we actually modify by adding some type tests to make sure that $score is not an available option?

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -174,6 +174,125 @@ describe("ObjectSet", () => {
...
+      Osdk.Instance<Employee, never, PropertyKeys<Employee>, {}, {
+        employeeId: "desc";
+      }>[]
+    >;
+
+    const { data: orderedByRelevance } = await nearestNeighbors.fetchPage({
+      $orderBy: "relevance",
+    });
+    expectTypeOf<typeof orderedByRelevance>().toMatchTypeOf<
+      Osdk.Instance<Employee, never, PropertyKeys<Employee>, {}, "relevance">[]
```
ssanjay1, line 276: Hmm, I'm not a big fan of `relevance` being in the type when in reality it's `$score` that pops up. I think what we want to do is something similar to `ExtractOptions` so instead of passing in `Z` directly to `Osdk.Instance` you pass in `ExtractOrderByOptions<Z>` which will give you either `never` or`$score` that way if someone looks at their return type, its clear what they're getting back on their object. 

Like `Osdk.Instance<quickAndDirty, "$rid", "name" | "foo">` makes it clear to me I'm getting name, foo, and $rid back, so would be nice to keep it consistent and also throw in `$score` rather than relevance

mvanschellebeeck: yep agreed - fixed in
https://github.com/palantir/osdk-ts/pull/1248/commits/0b6b35a9cb10ad476a722f431af02058277bcb26

packages/client/src/objectSet/ObjectSet.test.ts
mvanschellebeeck: update in https://github.com/palantir/osdk-ts/pull/1248/commits/0b6b35a9cb10ad476a722f431af02058277bcb26

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -35,13 +35,21 @@ export interface SelectArg<
...
+  Q extends ObjectOrInterfaceDefinition,
+  L extends PropertyKeys<Q> = PropertyKeys<Q>,
+> =
+  | {
+    [K in L]?: "asc" | "desc";
+  }
+  | "relevance";
+
 export interface OrderByArg<
   Q extends ObjectOrInterfaceDefinition,
```
nihalbhatnagar, line 48: Noting for myself that this isn't exported, so this arg reorder is fine

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -212,6 +224,27 @@ interface WithProperties<
...
+   * objects associated score, specify "relevance" in the orderBy.
+ */
+
+  readonly nearestNeighbors: (
+    query: string | number[],
+    numNeighbors: number,
+    property: VectorPropertyKeys<Q>,
```
nihalbhatnagar, line 244: Can we just use FilteredPropertyKeys here instead?

mvanschellebeeck: nice, [f6eb11b](https://github.com/palantir/osdk-ts/pull/1248/commits/f6eb11b52e88a3a3634d987e19ffa5bccfc36d68)

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -672,7 +794,7 @@ describe("ObjectSet", () => {
         },
       });
 
-      expectTypeOf(objectSet).branded.toEqualTypeOf<
+      expectTypeOf(objectSet).toEqualTypeOf<
```
ssanjay1, line 797: Why'd we remove branded here?

mvanschellebeeck: ah I was seeing `Type instantiation is excessively deep and possibly infinite.` Appears to be a failure mode for this branded check according to the docs:

```
branded:
     * // ❌ The following line doesn't compile, even though the types are arguably the same.
     * expectTypeOf<{ a: 1 } & { b: 2 }>().toEqualTypeOf<{ a: 1; b: 2 }>()
     * ```
     * This helper works around this problem by using
     * a more permissive but less performant check.
     *
     * __Note__: This comes at a performance cost, and can cause the compiler
     * to 'give up' if used with excessively deep types, so use sparingly.
 ```

PR #1249, opened 2025-02-27
add all interface prop support

ssanjay1:
We now fix a bug where properties from parent interfaces were not populated

### DIFF
--- a/packages/e2e.generated.catchall/ontology.json
+++ b/packages/e2e.generated.catchall/ontology.json
@@ -1857,6 +1857,35 @@
           "description": "Description of Description"
         }
       },
+      "allProperties": {
+        "name": {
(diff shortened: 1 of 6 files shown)


packages/e2e.generated.catchall/ontology.json
```diff
@@ -1857,6 +1857,35 @@
           "description": "Description of Description"
         }
       },
+      "allProperties": {
```
nihalbhatnagar, line 1860: Assuming this is the new field we get that has inherited properties?

nihalbhatnagar (approved):

packages/e2e.generated.catchall/ontology.json
ssanjay1: Yup

PR #1253, opened 2025-02-28
Update where clause error to say use and clause

### LINKED ISSUE #710: Improve error message when using mixed $ entries
https://github.com/palantir/osdk-ts/blob/main/packages/client/src/internal/conversions/modernToLegacyWhereClause.ts#L130-L137

Report was need to change:

```ts
$(Thing).where({
    age: {
        $gte: start;
        $lte: end;
    }
})
```

to

```ts
$(Thing).where({
  $and: [
    { age: { $gte: start } },
    { age: { $lte: end } }
  ]
}) 
```

Error message currently says:
`Uncaught Error: Invariant failed: WhereClause Filter with multiple clauses isn't allowed`

nihalbhatnagar:
This was done in order to fix #710. We also have added type guards for this now, so this would only be if someone was brute forcing this through

### DIFF
--- a/packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
+++ b/packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
@@ -176,7 +176,7 @@ function handleWherePair(
   invariant(
     !hasDollarSign
       || keysOfFilter.length === 1,
-    "A WhereClause Filter with multiple clauses/fields is not allowed. Instead, use an 'and' clause to combine multiple filters.",
+    "A WhereClause Filter with multiple clauses/fields is not allowed. Instead, use an 'or'/'and' clause to combine multiple filters.",
   );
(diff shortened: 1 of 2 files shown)


packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -176,7 +176,7 @@ function handleWherePair(
   invariant(
     !hasDollarSign
       || keysOfFilter.length === 1,
-    "WhereClause Filter with multiple clauses isn't allowed",
+    "A WhereClause Filter with multiple clauses/fields is not allowed. Instead, use an 'and' clause to combine multiple filters.",
```
ssanjay1, line 179: Should we specify and/or

ssanjay1 (approved):

PR #1264, opened 2025-03-03
modify script

ssanjay1:


packages/tool.release/src/runVersion.ts
```diff
@@ -117,7 +117,7 @@ export async function runVersion({
   const isReleaseBranch = context.branch.startsWith("release/")
     || process.env.PRETEND_BRANCH?.startsWith("release/");
 
-  const runGitCommands = !process.env.PRETEND_BRANCH;
+  const runGitCommands = true;
```
ssanjay1, line 120: This lets us use the pretend_branch main to cut release like we're on main and still generate the version PR automatically.

nihalbhatnagar (approved):

PR #1271, opened 2025-03-04
Add deleted objects and links to the Apply Action endpoint

nihalbhatnagar:


packages/api/src/actions/ActionResults.ts
```diff
@@ -27,15 +27,19 @@ interface ObjectEdits {
   type: "edits";
   addedObjects: Array<ObjectReference>;
   modifiedObjects: Array<ObjectReference>;
+  deletedObjects?: Array<ObjectReference>;
   addedLinks: Array<LinkReference>;
+  deletedLinks?: Array<LinkReference>;
```
nihalbhatnagar, line 32: There are a couple of decisions I made to arrive at this point. For context, batch actions will never return deleted object or edits.

BatchAction and ApplyAction return the [ActionReturnTypeForOptions](https://github.com/palantir/osdk-ts/blob/nihalb/new-apply-action/packages/api/src/actions/ActionReturnTypeForOptions.ts#L26-L27). We also export the [ActionEditResponse](https://github.com/palantir/osdk-ts/blob/nihalb/new-apply-action/packages/api/src/actions/Actions.ts#L76-L77) type. I did not split out ActionEditResponse into a union of BatchActionResults and RegularActionResults since we don't really get any type guarantees there and adding BatchActionEditResponse would be a break all together.

Secondly, I made the deletedLinks and deletedObjects properties optional to avoid a break as well. I think this is required if you were typing against ActionResults where you would suddenly see a break.

ericanderson: Would this only be a problem if you were creating your own?

ericanderson (dismissed):

ericanderson (approved):

PR #1273, opened 2025-03-05
Allow cloning interface objects

nihalbhatnagar:


packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -80,4 +84,37 @@ export function createOsdkInterface<
...
+        x != null
+      ),
+    );
+
+    return underlying.$clone(remappedProps).$as(interfaceDef);
+  }
+  function mapProperty(propertyName: string, value: any) {
+    const objDef = underlying[ObjectDefRef];
+    const targetPropName =
+      objDef.interfaceMap![interfaceDef.apiName][propertyName];
```
nihalbhatnagar, line 111: This isn't the map we get from the interface load, its from the object metadata. Presumably should be the same

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -197,11 +197,11 @@ export async function convertWireToOsdkObjects2(
  * Utility function that lets us take down selected property names from an interface
  * and convert them to an array of property names on an object.
  */
-function convertInterfacePropNamesToObjectPropNames(
-  objectDef: FetchedObjectTypeDefinition & { interfaceMap: {} },
+export function convertInterfacePropNamesToObjectPropNames(
```
ericanderson, line 200: Since this wasn't exported before, can we mark it as `@internal` so we don't accidentally leak it?

ericanderson (changes requested):
Looks good. One thing to fix

ericanderson (approved):

PR #1279, opened 2025-03-10
Load all underlying properties for interface

ssanjay1:
We can now load all the underlying properties of an interface by adding a flag like this:
```  
const athletes = await dsClient(Athlete).where({
    name22: { $eq: "Michael Jordan" },
  }).fetchPage({ $includeAllBaseObjectProperties: true });
```

The type we get back will now include the flag in the options section of `Osdk.Instance`. Any sort of selects will override this flag and should only give back the properties you selected.
image


packages/api/src/OsdkObjectFrom.ts
```diff
@@ -71,10 +71,17 @@ export type MapPropNamesToObjectType<
   FROM extends ObjectOrInterfaceDefinition,
   TO extends ObjectTypeDefinition,
   P extends ValidOsdkPropParams<FROM>,
-> = PropMapToObject<
-  FROM,
-  TO
->[JustProps<FROM, P> & keyof PropMapToObject<FROM, TO>];
+  OPTIONS extends never | "$rid" | "$allBaseProperties" = never,
+> = "$allBaseProperties" extends OPTIONS
+  ? PropertyKeys<FROM> extends P ? PropertyKeys<TO>
```
nihalbhatnagar, line 76: Nice find, its cool we already had the building blocks here already

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -883,8 +883,19 @@ describe("ObjectSet", () => {
         >()
           .toEqualTypeOf<"fooSpt">();
 
-        expectTypeOf<ConvertProps<FooInterface, Employee, "fooSpt">>()
```
nihalbhatnagar, line 886: Does it make sense to keep this test as well?

ssanjay1: Honestly yea probably should, lemme add back

ssanjay1: Cool, added back

nihalbhatnagar (approved):

PR #1283, opened 2025-03-11
Framework for accepting special types for RDP aggregations

nihalbhatnagar:
OSS allows collectList/Set on almost every type, which means special types weren't being properly parsed. This PR allows us to store the definition when defined, and pass it through down to the conversion. Additionally, I updated the types to allow booleans and GeoJSON types as well.

Alternatives I considered:
Parsing the Object Set definition to extract these definitions instead of passing it around. This is was possible but was a bit messy to me and I think passing this around could allow us to use it if needed in other places. Passing it around itself is a bit messy however, I would be open to refactoring this to parse the definition instead

FLUPs from this PR
1. I need to check into collect list/set on array types and if they're supported/how to handle them
2. When OSS adds support for the other special types, should be a fairly easy commit to start supporting these


packages/api/src/aggregate/AggregatableKeys.ts
```diff
@@ -36,24 +39,39 @@ export type NumericAggregateOption =
...
+type WITH_PROPERTIES_AGG_FOR_TYPE<WIRE_TYPE extends WirePropertyTypes> =
+  number extends GetWirePropertyValueFromClient<WIRE_TYPE>
+    ? NumericWithPropAggregateOption
+    : string extends GetWirePropertyValueFromClient<WIRE_TYPE>
+      ? StringWithPropAggregateOption
+    : WITH_PROPERTIES_AGG_FOR_SPECIAL_WIRE_TYPE<WIRE_TYPE>;
+
+type WITH_PROPERTIES_AGG_FOR_SPECIAL_WIRE_TYPE<
+  WIRE_TYPE extends WirePropertyTypes,
+> = WIRE_TYPE extends "attachment" | "geopoint" | "geoshape" | "boolean"
```
ssanjay1, line 56: Can we pull these out into a helper type, maybe `CollectListPropertyTypes`? Maybe even in propertyValueMapping you could have a new type that has all the possible wireTypes mapped to true or false for whether it supports collectSet

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -124,6 +125,27 @@ export function createOsdkObject<
...
+      && typeof (derivedPropertyTypeByName[propKey].type) === "string"
+      && specialPropertyTypes.has(derivedPropertyTypeByName[propKey].type)
+    ) {
+      if (derivedPropertyTypeByName[propKey].type === "attachment") {
+        if (Array.isArray(rawObj[propKey])) {
+          rawObj[propKey] = rawObj[propKey].map(a =>
+            hydrateAttachmentFromRidInternal(client, a.rid)
+          );
+        }
+        rawObj[propKey] = hydrateAttachmentFromRidInternal(
```
ssanjay1, line 139: Won't this still run in the array case?

nihalbhatnagar: yep, fixed

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -243,29 +273,41 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
...
     withProperties: (clause) => {
-      const definitionMap = new Map<any, DerivedPropertyDefinition>();
+      const definitionMap: DerivedPropertyDefinitionMap = new Map();
 
-      const derivedProperties: Record<string, DerivedPropertyDefinition> = {};
+      const derivedPropertyDefinitionByName: Record<
+        string,
+        DerivedPropertyDefinition
+      > = {};
+      const derivedPropertyTypeByName: Record<
```
ericanderson, line 282: This shadows the argument passed into `createObjectSet`. Is that intended?

nihalbhatnagar: Ah an oversight on my part, probably should use different names + extend. This would overwrite. Thank you for the catch

PR #1297, opened 2025-03-18
Add new TimeSeriesQueryV2 type with nested operations

saakshaat-singh:
The current `TimeSeriesQuery` type has a flat structure, making it difficult to add and maintain any new operations or args.

So we're adding a new `TimeSeriesQueryV2` type with nested operations, which will make it easy for users to express their time series operations.

### Handling Unions
To avoid breaking existing `TimeSeriesQuery` usages, I've added a union of `TimeSeriesQuery` and `TimeSeriesQueryV2`.

To handle unions, I've added time-series specific type-guards.

Alternatively I considered using discriminated unions however they would require users to pass the discriminator which, although explicit and clean, I'm not confident would be the best UX or be consistent with the rest of the SDK. I'm open to alternatives for handling unions.

### DIFF
--- a/packages/api/src/timeseries/timeSeriesGuards.ts
+++ b/packages/api/src/timeseries/timeSeriesGuards.ts
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type {
+  AbsoluteTimeRange,
+  RelativeTimeRange,
+  TimeSeriesQuery,
+  TimeSeriesQueryV2,
+  TimeSeriesQueryWrapper,
+  TimeSeriesRange,
+} from "./timeseries.js";
+
+/**
+ * Type Guards to handle union time series types.
+ */
+
+/**
+ * Checks if the query is a TimeSeriesQueryV2 (new nested version)
+ * by verifying that it has any of the available operations:
+ * - `range`
+ */
+export function isTimeSeriesQueryV2(
+  query: TimeSeriesQueryWrapper,
+): query is TimeSeriesQueryV2 {
+  return "range" in query;
+}
+
+/**
+ * Checks if the query is a legacy TimeSeriesQuery type.
+ * Since legacy queries have a flat structure, they do not include nested operations
+ * such as the `range` operation used in TimeSeriesQueryV2.
+ */
+export function isLegacyTimeSeriesQuery(
+  query: TimeSeriesQueryWrapper,
+): query is TimeSeriesQuery {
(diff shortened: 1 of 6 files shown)


packages/api/src/timeseries/timeseries.ts
```diff
@@ -46,6 +46,17 @@ export type TimeSeriesQuery =
     $unit?: never;
   };
 
+export type TimeSeriesQueryV2 = {
+  range: TimeSeriesRange;
+};
+
+export type TimeSeriesQueryWrapper = TimeSeriesQuery | TimeSeriesQueryV2;
+
+export type TimeSeriesRange = {
```
ssanjay1, line 55: What happens if users pass in an empty range to the backend?

packages/api/src/timeseries/timeSeriesGuards.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Checks if the query is a legacy TimeSeriesQuery type.
+ * Since legacy queries have a flat structure, they do not include nested operations
+ * such as the `range` operation used in TimeSeriesQueryV2.
+ */
+export function isLegacyTimeSeriesQuery(
+  query: TimeSeriesQueryWrapper,
+): query is TimeSeriesQuery {
+  return !("range" in query);
```
ssanjay1, line 46: Could we just check that `!isTimeSeriesQueryV2`, so we don't need to maintain changes in both functions?

packages/client/src/util/timeseriesUtils.ts
```diff
@@ -46,6 +54,32 @@ export function getTimeRange(body: TimeSeriesQuery): TimeRange {
     };
 }
 
+export const parseTimeSeriesRangeV2 = (
```
ssanjay1, line 57: nit: can we use `export function` syntax to keep parity with the rest of the codebase?

packages/api/src/timeseries/timeseries.ts
ssanjay1: Wondering if we should require atleast one value is passed in?

saakshaat-singh: For no range/empty range, the backend (api-gateway layer) will return the full range of the time-series. It's a valid request to get all the points in the time-series.

packages/api/src/timeseries/timeSeriesGuards.ts
saakshaat-singh: good point, done

.changeset/yummy-beds-bet.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/client": minor
+"@osdk/api": minor
```
saakshaat-singh, line 3: doing a minor release since we're not breaking the existing type

ssanjay1: Can we actually make this a patch? For reference, we usually use patch for most of our updates since we usually push out beta releases until we're ready to GA the next minor.(Right now, 2.1 is GA, and we have 2.2-betas)

saakshaat-singh: thanks for SA, updated

packages/api/src/timeseries/timeseries.ts
```diff
@@ -46,6 +46,17 @@ export type TimeSeriesQuery =
     $unit?: never;
   };
 
+export type TimeSeriesQueryV2 = {
+  range: TimeSeriesRange;
```
ssanjay1, line 50: Can you prepend these with $ to be in line with our other apis? We use `$` to indicate anything that is an OSDK api field, and not some field from the object

saakshaat-singh: ah good to know, updated

PR #1303, opened 2025-03-18
test framework media

ericanderson:


packages/shared.test/src/FauxFoundry/FauxDataStore.ts
```diff
@@ -249,6 +263,85 @@ export class FauxDataStore {
...
+    property: OntologiesV2.PropertyApiName,
+  ): MediaMetadataAndContent {
+    const obj = this.getObjectOrThrow(objectType, primaryKey);
+    const propertyDef =
+      this.ontology.getObjectTypeFullMetadataOrThrow(objectType)
+        .objectType.properties[property];
+
+    if (!propertyDef) {
+      // FIXME: what would the backend do here?
+      throw new OpenApiCallError(
```
ssanjay1, line 289: If the property is not found it throws 404 `Api:PropertiesNotFound"` with the objectType and properties as unsafe args

packages/shared.test/src/FauxFoundry/FauxDataStore.ts
```diff
@@ -249,6 +263,85 @@ export class FauxDataStore {
...
+            propertyBaseType: "unknown",
+          },
+        } satisfies PropertyApiNameNotFound,
+      );
+    }
+
+    if (propertyDef.dataType.type !== "mediaReference") {
+      // FIXME: what would the backend do here?
+      throw new OpenApiCallError(
+        400,
```
ssanjay1, line 306: You just get a 400 invalid argument here, doesn't log out the property type

packages/shared.test/src/handlers/loadObjectsEndpoints.ts
```diff
@@ -506,33 +484,43 @@ export const loadObjectsEndpoints: Array<RequestHandler> = [
...
-
-        if (!areArrayBuffersEqual(body, expectedBodyArray)) {
-          throw new OpenApiCallError(400, InvalidContentTypeError);
-        }
+      const content = await request.arrayBuffer();
+      const mediaType = request.headers.get("Content-Type")
+        ?? "application/octet-stream";
+
+      // FIXME is this what a rid for this looks like?
+      const mediaItemRid = `ri.media.main.${randomUUID()}`;
```
ssanjay1, line 500: It's: `ri.mio.main.media-item`

ssanjay1 (approved):
Approving to unblock, but left comments to address your FIX-MEs

packages/shared.test/src/FauxFoundry/FauxDataStore.ts
ericanderson: This is an INVALID_ARGUMENT. We don't have a generic one in the types.

PR #1306, opened 2025-03-19
Links as code

aep000:


packages/maker/src/api/types.ts
```diff
@@ -203,6 +210,50 @@ export interface StructPropertyType extends
...
+export type ManyToManyLinkTypeDefinition = {
+  id: LinkTypeId;
+  many: ManyToManyObjectLinkReference;
+  toMany: ManyToManyObjectLinkReference;
+  editsEnabled?: boolean;
+  status?: OntologyIrLinkTypeStatus;
+  redacted?: boolean;
+};
+
+export type ManyToManyObjectLinkReference = {
```
ericanderson, line 245: Where possible, if you use `interface` instead of `type` you do the compiler a big favor.

aep000: Fixed

ericanderson (approved):

PR #1308, opened 2025-03-19
Ea.redo tests.3

ericanderson:
Each PR can be reviewed independently.


packages/shared.test/src/handlers/MockOntologiesV2.ts
```diff
@@ -0,0 +1,403 @@
...
+    OntologiesV2.OntologyObjectSets.aggregate,
+    ["ontologyApiName"],
+  ),
+  loadMultipleObjectTypes: handleOpenApiCallFactory(
+    OntologiesV2.OntologyObjectSets.loadMultipleObjectTypes,
+    ["ontologyApiName"],
+  ),
+};
+
+const MOntologiesV2: {
```
ssanjay1: typo?

ericanderson: Was from a refactor to split these out. Couldnt have a MockOntologies raw object that had a MockOntologies child object without shenanigans

PR #1319, opened 2025-03-21
Check requested scopes are equal before exchanging refresh token

nadprana:
We don't check whether the requested scopes have changed before using the refresh token in local storage if there is one. This has led to an issue flagged by a couple different users where, even after they've requested more scopes in their code, they can't get a token back with those permissions because we continue to exchange their old refresh token.

The scopes granted from exchanging a refresh token are taken from the intersection of 1) the requested scopes during the initial auth code grant (not to be confused with the scopes that actually get granted and returned in the auth response) and 2) the allowed client operations at this current moment in time.

So there's no risk of having a refresh token based off stale client allowed operations, as those get re-evaluated every time we refresh our token. But we do need to check whether the requested scopes have changed since we kicked off the auth code grant last, which this PR does by storing those scopes in local storage and checking if they've changed before attempting to exchange the refresh token.

### DIFF
--- a/packages/oauth/src/createConfidentialOauthClient.ts
+++ b/packages/oauth/src/createConfidentialOauthClient.ts
@@ -44,6 +44,7 @@ export function createConfidentialOauthClient(
   const client: Client = { client_id, client_secret };
   const authServer = createAuthorizationServer(ctxPath, url);
   const oauthHttpOptions: HttpRequestOptions = { [customFetch]: fetchFn };
+  const joinedScopes = scopes.join(" ");
 
   const { getToken, makeTokenAndSaveRefresh } = common(
     client,
@@ -52,7 +53,7 @@ export function createConfidentialOauthClient(
     oauthHttpOptions,
     undefined,
     undefined,
-    scopes,
+    joinedScopes,
   );
 
   async function _signIn() {
@@ -64,7 +65,7 @@ export function createConfidentialOauthClient(
           await clientCredentialsGrantRequest(
             authServer,
             client,
-            new URLSearchParams({ scope: scopes.join(" ") }),
+            new URLSearchParams({ scope: joinedScopes }),
             oauthHttpOptions,
           ),
         ),
--- a/packages/oauth/src/createConfidentialOauthClient.ts
+++ b/packages/oauth/src/createConfidentialOauthClient.ts
@@ -52,6 +52,7 @@ export function createConfidentialOauthClient(
     oauthHttpOptions,
     undefined,
     undefined,
+    scopes,
   );
 
   async function _signIn() {
--- a/packages/oauth/src/createPublicOauthClient.ts
+++ b/packages/oauth/src/createPublicOauthClient.ts
@@ -138,11 +138,12 @@ export function createPublicOauthClient(
   ctxPath?: string,
 ): PublicOauthClient {
   let refreshTokenMarker: string | undefined;
+  let joinedScopes: string;
   ({
     useHistory,
     loginPage,
     postLoginPage,
-    scopes,
+    joinedScopes,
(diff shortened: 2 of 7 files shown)


packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -187,10 +188,20 @@ export function createPublicOauthClient(
...
-    const { refresh_token, refreshTokenMarker: lastRefreshTokenMarker } =
-      readLocal(client);
+    const {
+      refresh_token,
+      refreshTokenMarker: lastRefreshTokenMarker,
+      requestedScopes: initialRequestedScopes,
+    } = readLocal(client);
+
+    const areScopesEqual = initialRequestedScopes != null
+      && scopes!.every(scope => initialRequestedScopes.includes(scope))
```
nadprana, line 198: i don't like using the `!` but the types here are a bit funky -- this would technically only be called after the defaults have been assigned i.e. non-null right? alternatively i can extract the default scopes from `processOptionsAndAssignDefaults` and over here do `scopes ?? DEFAULT_SCOPES`

ericanderson: I don't mind the `!`, as long as you just write a comment about the assumption above it. 

That said, why not just string sort these and then you can just `===` the stored value with the one you have? At the end of the day we are only passing a string to MP anyway:
```ts
  ...
  scope: ["offline_access", ...scopes].join(" "),
 ...
```

Instead of joining them in `initiateLoginRedirect`, you can sort and join them in `createPublicOauthClient` up front and use in both places.

nadprana: good point! i knew there had to be a better a way to better compare list equality here, something felt off 😭

packages/oauth/src/utils.ts
```diff
@@ -62,7 +65,8 @@ export function processOptionsAndAssignDefaults(
     useHistory: options.useHistory ?? true,
     loginPage: options.loginPage,
     postLoginPage: options.postLoginPage || window.location.toString(),
-    scopes: options.scopes ?? ["api:read-data", "api:write-data"],
+    joinedScopes: [...options.scopes ?? ["api:read-data", "api:write-data"]]
```
nadprana, line 68: made a shallow copy here to avoid modifying the original array but not sure if it really matters

packages/oauth/src/createConfidentialOauthClient.ts
```diff
@@ -44,6 +44,7 @@ export function createConfidentialOauthClient(
   const client: Client = { client_id, client_secret };
   const authServer = createAuthorizationServer(ctxPath, url);
   const oauthHttpOptions: HttpRequestOptions = { [customFetch]: fetchFn };
+  const joinedScopes = scopes.join(" ");
```
ericanderson, line 47: I was going to comment on these being sorted but actually, it doesn't really matter that much. Unless the developer consumer is randomizing their array per page load, this would only change per app update in the worst case and re-signing in is not that big of a deal in that case.

packages/oauth/src/utils.ts
ericanderson: Appreciate the extra care. The 1 time this gets called per page load can handle the temporary duplication of the array.

ericanderson (approved):

packages/oauth/src/createConfidentialOauthClient.ts
nadprana: ooh i wasn't sure if this applied for confidential clients, my understanding was they don't use refresh tokens so these never get stored? 🤔 could be wrong though

ericanderson (approved):

PR #1326, opened 2025-03-25
Support widget dev server on non-root paths

rosscourt:
Support running the widget dev-server plugin's setup UI on non-root paths by making all asset and API paths relative. When running within Foundry on a container the set up UI is accessed using a proxy route rather than directly via it's port.

I'm also redirecting from `.palantir/setup` to `.palantir/setup/` in order to ensure a single consistent route is used to hit the setup page.

### DIFF
--- a/packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
+++ b/packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
@@ -26,7 +26,7 @@ import type { Plugin, ViteDevServer } from "vite";
 import {
   CONFIG_FILE_SUFFIX,
   ENTRYPOINTS_PATH,
-  PALANTIR_PATH,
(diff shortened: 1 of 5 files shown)


packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -205,9 +229,13 @@ function getFullSourcePath(source: string, importer: string): string {
   return path.resolve(path.dirname(importer), source);
 }
 
+function serverPath(server: ViteDevServer, subPath: string): string {
+  return path.resolve(server.config.base, subPath);
```
rosscourt, line 233: Restores this from the original implementation https://github.com/palantir/osdk-ts/blob/121475a91919afc8365e7301cee97aa04e0beda8/packages/widget.vite-plugin.unstable/src/plugin.ts#L106

tzyl (approved):

PR #1328, opened 2025-03-25
Framework for accepting special types for RDP aggregations pt 2

nihalbhatnagar:
A rework of #1283.

OSS allows collectList/Set on almost every type, which means special types weren't being properly parsed. This PR allows us to store the definition when defined, and pass it through down to the conversion. Additionally, I updated the types to allow booleans and GeoJSON types as well.

I chose to parse the object set for RDP definitions instead of the previous method of passing it through because I feel this may come in handy for expression derived properties. Additionally, piping the definitions through was tedious and could have led to obvious bugs if the definitions weren't piped through specific codepaths.


packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -477,6 +486,7 @@ describe("convertWireToOsdkObjects", () => {
           [object],
           undefined,
           undefined,
+          {},
```
ericanderson, line 489: Curious why you opted for inserting a param in the middle vs adding one to the end with a default value.

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -127,6 +129,29 @@ export function createOsdkObject(
...
       );
+    } else if (
+      propKey in derivedPropertyTypeByName
+      && typeof (derivedPropertyTypeByName[propKey].type) === "string"
+      && specialPropertyTypes.has(derivedPropertyTypeByName[propKey].type)
+    ) {
+      const rawValue = rawObj[propKey as any];
+      if (derivedPropertyTypeByName[propKey].type === "attachment") {
+        if (Array.isArray(rawValue)) {
+          rawObj[propKey] = rawValue.map(a =>
```
ericanderson, line 140: Should we extract a method shared by this and `createSpecialProperty` ?

packages/client/src/object/fetchPage.ts
```diff
@@ -142,6 +143,7 @@ async function fetchInterfacePage<
       result.data as OntologyObjectV2[], // drop readonly
       interfaceType.apiName,
       !args.$includeRid,
+      (await extractRdpDefinition(client, objectSet, interfaceType.apiName))[0],
```
ericanderson, line 146: Why the `[0]`th one? If intentional, it should have a comment in the code above it to explain this.

packages/client/src/object/fetchPage.ts
```diff
@@ -329,6 +332,7 @@ export async function fetchObjectPage<
       r.data as OntologyObjectV2[],
       undefined,
       undefined,
+      (await extractRdpDefinition(client, objectSet, objectType.apiName))[0],
```
ericanderson, line 335: ditto

packages/shared.test/src/stubs/objects.ts
```diff
@@ -267,3 +267,16 @@ export const objectWithAllPropertyTypesEmptyEntries = {
...
+  ...objectWithAllPropertyTypes1,
+  __rid:
+    "ri.phonograph2-objects.main.object.401ac022-89eb-4591-8b7e-0a912b9efb44",
+  __primaryKey: 5,
+  __apiName: "objectTypeWithAllPropertyTypes",
+  __title: "5",
+  id: 5,
+} as {
+  __primaryKey: number;
+  __apiName: string;
```
ericanderson, line 281: Why this as? Seems like we are stripping a lot of information.

packages/client/src/object/convertWireToOsdkObjects.test.ts
nihalbhatnagar: I wanted to make sure if we start using convertWireToOsdkObjects elsewhere we don't forget to pipe through RDPs which would lead to a bug in that codepath.

packages/client/src/object/fetchPage.ts
nihalbhatnagar: The 0th since `extractRdpDefinition` returns a tuple. Since we don't need the second externally i'll just wrap the function

packages/shared.test/src/stubs/objects.ts
nihalbhatnagar: --isolatedDefinitions doesn't like the `...` expansion of objectWithAllPropertyTypes1, and this was the minimum cast needed for LegacyFauxFoundry. I'll just add some mock data instead and remove the cast

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
nihalbhatnagar: I don't think it's worth it since one's a return and ones an assignment, also just one branch

ericanderson (approved):

PR #1338, opened 2025-03-26
[WIP][1/?] Test back-compatibility across minor versions

nihalbhatnagar:
This PR adds code to test back-compatibility across versions by transpiling the client test code with SDKs from older generator versions. This PR may be followed up by additional tests

### DIFF
--- a/packages/client.test.ontology/generateMockOntology.js
+++ b/packages/client.test.ontology/generateMockOntology.js
@@ -16,15 +16,49 @@
 
 // @ts-check
 
-import { generateClientSdkVersionTwoPointZero } from "@osdk/generator";
 import { LegacyFauxFoundry } from "@osdk/shared.test";
-import { rmSync } from "node:fs";
+import { execa } from "execa";
+import { promises as fs, rmSync } from "node:fs";
 import { mkdir, readdir, writeFile } from "node:fs/promises";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "url";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
-const outDir = join(__dirname, "src", "generatedNoCheck");
+
+const version = process.argv[2];
+const packageName = "@osdk/generator";
+
+async function installAndImport(packageName, version) {
+  const tmpDir = join(__dirname, "tmp");
+  await mkdir(tmpDir, { recursive: true });
+
+  await execa("pnpm", [
+    "install",
+    `${packageName}@${version}`,
+    "--prefix=./tmp",
+    "--lockfile=false",
+  ], {
+    cwd: __dirname,
+  });
+
+  const packageJsonPath = join(
+    tmpDir,
+    "node_modules",
+    packageName,
+    "package.json",
+  );
+  const packageJsonContent = await fs.readFile(packageJsonPath, "utf8");
+  const packageJson = JSON.parse(packageJsonContent);
+  const mainEntry = packageJson.main;
+
+  const packageMainFile = join(tmpDir, "node_modules", packageName, mainEntry);
+
+  const importedModule = await import(packageMainFile);
+
+  return importedModule;
+}
+
+const outDir = join(__dirname, "src", `generatedNoCheck`);
 
 try {
   rmSync(outDir, { recursive: true, force: true });
@@ -43,13 +77,20 @@ const ontologyWithoutUnsupportedAction = {
   },
 };
 
-// the generator does not correctly handle actions that point to object types outside of the ontology
-// this step is typically handled by code upstream of the actual generator
 delete ontologyWithoutUnsupportedAction.actionTypes["unsupportedAction"];
+
+let generateClientSdkVersionTwoPointZero;
+if (version) {
+  const generatorModule = await installAndImport(packageName, version);
+  generateClientSdkVersionTwoPointZero =
+    generatorModule.generateClientSdkVersionTwoPointZero;
+} else {
+  ({ generateClientSdkVersionTwoPointZero } = await import("@osdk/generator"));
+}
 
 await generateClientSdkVersionTwoPointZero(
   ontologyWithoutUnsupportedAction,
-  "typescript-sdk/dev osdk-cli/dev",
+  `typescript-sdk/dev osdk-cli/dev`,
   {
     writeFile: (path, contents) => {
       return writeFile(path, contents, "utf-8");
--- a/packages/test.minor-backcompat/package.json
+++ b/packages/test.minor-backcompat/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "@osdk/client.test.ontology",
+  "name": "@osdk/tests.minor-backcompat",
   "private": true,
-  "version": "2.2.0-beta.11",
+  "version": "0.7.0-beta.2",
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
@@ -26,31 +26,25 @@
   "scripts": {
     "check-spelling": "cspell --quiet .",
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
-    "codegen": "node ./generateMockOntology.js",
     "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
     "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
-    "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
-    "@osdk/api": "workspace:*"
+    "chalk": "^5.4.1",
+    "consola": "^3.4.0",
+    "execa": "^9.5.1",
+    "path": "0.12.7",
+    "semver": "^7.6.3"
   },
-  "peerDependencies": {},
   "devDependencies": {
-    "@osdk/generator": "workspace:~",
-    "@osdk/monorepo.api-extractor": "workspace:~",
-    "@osdk/monorepo.tsconfig": "workspace:~",
-    "@osdk/shared.test": "workspace:~",
-    "execa": "9.5.2",
+    "@types/node": "^18.0.0",
+    "@types/semver": "^7.5.8",
     "typescript": "~5.5.4"
   },
-  "publishConfig": {
-    "access": "public"
-  },
   "files": [],
-  "main": "./build/cjs/index.cjs",
   "module": "./build/esm/index.js",
-  "types": "./build/cjs/index.d.cts",
+  "types": "./build/esm/index.d.ts",
   "type": "module"
 }
--- a/packages/test.minor-backcompat/src/index.ts
+++ b/packages/test.minor-backcompat/src/index.ts
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import chalk from "chalk";
+import consola from "consola";
+import { execa } from "execa";
+import { promises as fs } from "fs";
+import * as path from "node:path";
+import { exit } from "node:process";
+import { fileURLToPath } from "node:url";
+import semver from "semver";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const packagesPath = path.join(
+  __dirname,
+  "..",
+  "..",
+  "..",
+);
+
+async function codegenAtVersion(version: string) {
+  const dirPath = path.join(
+    packagesPath,
+    "client.test.ontology",
+  );
+
+  consola.info(`Running codegen for version ${chalk.blue(version)}`);
+  const { failed } = await execa("pnpm", ["codegen", version], {
+    cwd: dirPath,
+  });
+  if (failed) {
+    consola.error(`Codegen failed for version ${chalk.red(version)}`);
+    exit(1);
+  }
+}
+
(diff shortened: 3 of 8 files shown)


packages/test.minor-backcompat/src/index.ts
```diff
@@ -0,0 +1,128 @@
...
+for (const version of versions.reverse()) {
+  consola.info(`Processing version ${chalk.blue(version)}`);
+
+  try {
+    await codegenAtVersion(version);
+  } catch (e) {
+    consola.warn(
+      `Failed to codegen at version ${
+        chalk.yellow(version)
+      } due to likely incompatible generator features. Defaulting to standard tests.`,
```
nihalbhatnagar, line 90: When we add new features, like when we added the Function Map type to the ontology.json, we can't generate and compare with these older versions now. Will probably add some fallbacks gere

packages/client.test.ontology/generateMockOntology.js
```diff
@@ -16,15 +16,49 @@
...
+
+  const packageJsonPath = join(
+    tmpDir,
+    "node_modules",
+    packageName,
+    "package.json",
+  );
+  const packageJsonContent = await fs.readFile(packageJsonPath, "utf8");
+  const packageJson = JSON.parse(packageJsonContent);
+  const mainEntry = packageJson.main;
```
nihalbhatnagar, line 52: should probably make this module

packages/client.test.ontology/generateMockOntology.js
```diff
@@ -43,13 +77,20 @@ const ontologyWithoutUnsupportedAction = {
...
-// this step is typically handled by code upstream of the actual generator
 delete ontologyWithoutUnsupportedAction.actionTypes["unsupportedAction"];
 
+let generateClientSdkVersionTwoPointZero;
+if (version) {
+  const generatorModule = await installAndImport(packageName, version);
+  generateClientSdkVersionTwoPointZero =
+    generatorModule.generateClientSdkVersionTwoPointZero;
+} else {
+  ({ generateClientSdkVersionTwoPointZero } = await import("@osdk/generator"));
```
nihalbhatnagar, line 88: This is the default case (I called pnpm turbo transpile on the whole workspace)

packages/test.minor-backcompat/src/index.ts
```diff
@@ -0,0 +1,128 @@
...
+async function fetchMatchingVersions(range: string): Promise<string[]> {
+  consola.info(`Fetching matching versions for range ${chalk.blue(range)}`);
+  const { stdout } = await execa("npm", [
+    "view",
+    "@osdk/client",
+    "versions",
+    "--json",
+  ]);
+  const versions = JSON.parse(stdout);
+  return versions.filter((version: string) => semver.satisfies(version, range));
```
nihalbhatnagar, line 59: Seems redundant but was running into weird issues trying to specify versions in the npm view, will take another look

packages/client.test.generate/bin/generateMockOntology.mjs
```diff
@@ -0,0 +1,150 @@
...
+    outDir,
+    "module",
+    undefined,
+    undefined,
+    undefined,
+    true,
+  );
+}
+
+async function installAndImport(packageName, version) {
```
nihalbhatnagar, line 116: This function is the main change from this diff, besides the branching on version

packages/client.test.generate/bin/generateMockOntology.mjs
```diff
@@ -0,0 +1,150 @@
...
+  const __dirname = dirname(fileURLToPath(import.meta.url));
+  const tmpDir = join(__dirname, "..", "build", "fauxInstall");
+  await mkdir(tmpDir, { recursive: true });
+
+  await execa("pnpm", [
```
nihalbhatnagar, line 121: Installs package from NPM and stores it away

packages/test.minor-backcompat/src/index.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  await modifyPackageJsonToUseDirectPaths(path.join(outPath, "package.json"));
+
+  const pkgPath = path.join(outPath, "package.json");
+  const pkgData = JSON.parse(await fs.readFile(pkgPath, "utf8"));
+
+  /**
+   * Babel core is expecting these modules to be present in the workspace.
+   */
+  pkgData.dependencies = {
```
nihalbhatnagar, line 90: I don't like this but I can't seem to think of any workarounds since monorepo.tool.transpile is always called from this cwd. The only optimization I can think of is making a global workspace

packages/test.minor-backcompat/src/index.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const packagesPath = path.join(
+  __dirname,
+  "..",
+  "..",
+  "..",
+);
+
+async function codegenAtVersion(version: string) {
```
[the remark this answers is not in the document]
nihalbhatnagar: I will need to look into a way to run specifically when the underlying ontology in shared.test changes somehow. Otherwise, we would run into issues when we update this but the matching versions haven't changed so we don't codegen

packages/test.minor-backcompat/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@osdk/tests.minor-backcompat",
```
nihalbhatnagar, line 2: Bad naming on my part, will fix before merge but not changing rn to avoid losing comments

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,306 @@
...
+    path.join(outPath, "pnpm-workspace.yaml"),
+    "",
+  );
+
+  await modifyPackageJsonToUseDirectPaths(path.join(outPath, "package.json"));
+
+  const pkgPath = path.join(outPath, "package.json");
+  const pkgData = JSON.parse(await fs.readFile(pkgPath, "utf8"));
+
+  /**
```
nihalbhatnagar, line 87: Lost some comments so repeating a few, but since we're in a different workspace I need to reinstall these deps. I'm not sure theres a better workaround, any ideas?

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  await execa("pnpm", ["install"], { cwd: outPath });
+
+  /**
+   * monorepo.tool.transpile uses find-up on the pnpm-workspace.yaml file to find the root.
+   * Failing to remove it will cause the transpile command to fail.
+   */
+  await fs.unlink(path.join(outPath, "pnpm-workspace.yaml"));
+
+  await execa("pnpm", ["transpileEsm"], { cwd: outPath });
```
nihalbhatnagar, line 106: I need to transpile the generated files to be usable as a package. The alternatives is that I add a shortcut in `foundry-sdk-generator` to allow me to pass in the ontology.json and then bundle the package, but seems like a lot of work when I can workaround this way

nihalbhatnagar: Maybe I don't have to use our transpile tool? I could call babel myself but if this works and easily works with the copied package.json why not

packages/client/src/intellisense.test.ts
```diff
@@ -65,7 +66,7 @@ describe("intellisense", () => {
 
   beforeAll(() => {
     const clientsPackageJson = findUpSync("package.json", {
-      cwd: import.meta.dirname,
+      cwd: path.dirname(fileURLToPath(import.meta.url)),
```
nihalbhatnagar, line 69: I started running into TSC issues when doing the client copy. Were we testing this specific behavior/is this okay as a stopgap? Will continue to debug why this is happening, possibly because my node version isn't specified?

packages/tests.backcompat/src/e2e-backcompat.ts
```diff
@@ -0,0 +1,122 @@
...
+    `--prefix=${versionedDir}`,
+    "--lockfile=false",
+  ], {
+    cwd: rootPath,
+  });
+
+  return path.join(versionedDir, "node_modules", "@osdk");
+}
+
+export async function runE2eBackcompatTests(): Promise<void> {
```
nihalbhatnagar, line 68: WIP but this is it generally aside from version traversal

packages/client.test.generate/bin/generateMockOntology.mjs
```diff
@@ -0,0 +1,150 @@
...
+    alias: "v",
+    describe: "Version of the @osdk/generator package to use",
+    type: "string",
+  })
+  .help()
+  .argv;
+
+(async () => {
+  await generateMockOntologyAtDir(argv.outDir, argv.generatorVersion);
+})();
```
ericanderson, line 62: FWIW, since this is an `mjs` file you can just do top level await, you dont have to make the IIfe

packages/client/src/intellisense.test.ts
ericanderson: I thought vitest was supposed to handle this but this is a perfectly fine stop gap.

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,309 @@
...
+import { promises as fs } from "fs";
+import * as path from "node:path";
+import { fileURLToPath } from "node:url";
+import semver from "semver";
+
+const THIS_FILE_DIR = path.dirname(fileURLToPath(import.meta.url));
+const pnpmLockFile = findUpSync("pnpm-lock.yaml");
+
+if (!pnpmLockFile) {
+  throw new Error("pnpm-lock.yaml not found. :(");
```
ericanderson, line 30: Based on feedback I got for some of my errors before I went on vacation, we should explain in the error why this is important. Like `"Expected to find pnpm-lock.yaml higher up in our path to detect the workspace root. This suggests that this tool is being run outside of the code checkout, which is not its intent."` or IDK what.

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,309 @@
...
+
+const rootPath = path.dirname(pnpmLockFile);
+
+const monorepoTranspilePath = path.join(
+  rootPath,
+  "packages",
+  "monorepo.tool.transpile",
+  "bin",
+  "transpile2.mjs",
+);
```
ericanderson, line 41: This feels fragile to me on a few levels. 

1. Hard coded paths are always fun to debug later and now if we ever change the script path we have to change this too. So if we have to keep it this way we should at least throw some tiny-invariant's in with some file existence checks that explain the assumption so its faster for us to debug in the future. (So much easier for me to say this as the reviewer than as the author of so many of our CLIs, I'm sorry for my sins).
2. This creates the implication that we are forever going to maintain back compat in our compile script (or I guess fork it and change this code when that happens).
3. The exact dependency tree for turbo is probably also no a part of our back compat story which is a bit weird too.

For the first issue, I feel like we could just add a dependency on `monorepo.tool.transpile` in the workspace to simplify this a bit and then invoke the tool directly (and not change the package scripts if we do this).

The second issue is less clear what we should do. In theory we should be able to just publish the transpile tool that goes with that set of versions. This would enable us to continue to use the exact same transpilation assumptions for our back compat checks as we had when we built the file in the first place. To really pull this off though we are likely going to need to publish everything that is currently private. This is probably fine.

The third issue makes improving and upgrading turbo a problem. For example, the exact task ordering dependencies we had before I did the last optimization round don't likely exist any more.

The last alternative is to actually not use the build infrastructure at all for compiling this stuff. When we pull in some old code we COULD just use TSC directly and manually invoke it instead of relying on turbo to do it. The thing we truly care is that code that compiled before can still compile in the simple case, as our users will likely be living in that world.  So doing a transpile and a transpileTypes and a typecheck as a separate task doesn't really matter.

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,309 @@
...
+
+const builtSdksPath = path.join(
+  THIS_FILE_DIR,
+  "codegen",
+  "sdks",
+);
```
ericanderson, line 47: So when this runs we are codegening into `build/esm/codegen/sdks`? I think you mean to go up a directory to hit the `build/codegen` thing I mentioned?

packages/tests.backcompat/src/client-backcompat.ts
ericanderson: I find this strange as these should be in the full tree right? If the package was installed at `/Volumes/git/osdk-ts/packages/test.backcompat/build/codegen/sdks/{whatever}` then when doing module resolution would include the node_modules directory in the workspace root. 

What am I missing in my knowledge/context?

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,309 @@
...
+  };
+
+  await fs.writeFile(pkgPath, JSON.stringify(pkgData, null, 2));
+
+  await execa("pnpm", ["install"], { cwd: outPath });
+
+  /**
+   * monorepo.tool.transpile uses find-up on the pnpm-workspace.yaml file to find the root.
+   * Failing to remove it will cause the transpile command to fail.
+   */
```
ericanderson, line 103: I think I'm confused by this too. What does the tool do when this file doesn't exist?

packages/tests.backcompat/src/client-backcompat.ts
ericanderson: Yeah or pure tsc.

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,309 @@
...
+  await fs.unlink(path.join(outPath, "pnpm-workspace.yaml"));
+
+  await execa("pnpm", ["transpileEsm"], { cwd: outPath });
+
+  await execa("pnpm", ["transpileTypes"], { cwd: outPath });
+}
+
+async function copyClientPackage() {
+  const srcPath = path.join(rootPath, "packages", "client");
+  const destPath = path.join(THIS_FILE_DIR, "clientCopy");
```
ericanderson, line 113: I'm guessing this is wrong too.

packages/tests.backcompat/src/client-backcompat.ts
```diff
@@ -0,0 +1,309 @@
...
+
+  consola.info(`Copying client into ${destPath} for testing`);
```
ericanderson, line 115: Why copy instead of just using `pnpm link` or a symlink ourselves?

packages/tests.backcompat/src/client-backcompat.ts
ericanderson: Looking at things below I think this is related to how the client gets copied into place?

PR #1340, opened 2025-03-27
Standardize client.ts across templates

zeyadkhaled:


packages/create-app.template.react.beta/templates/src/client.ts.hbs
tzyl, line 1: nit: this is somewhat unrelated to the specific change of this PR and should really be enforced by some kind of automatic check instead but while we're here can we make all versions of this file across the templates the same?

- https://github.com/palantir/osdk-ts/blob/929b2a44431f76d5fb348880a4207c267ce2c89e/packages/create-app.template.react.beta/templates/src/client.ts.hbs
- https://github.com/palantir/osdk-ts/blob/929b2a44431f76d5fb348880a4207c267ce2c89e/packages/create-app.template.tutorial-todo-app.beta/templates/src/client.ts.hbs
- https://github.com/palantir/osdk-ts/blob/929b2a44431f76d5fb348880a4207c267ce2c89e/packages/create-app.template.vue.v2/templates/src/client.ts.hbs

Some of them export client + auth, other export auth and default export client, the vue has slightly different ordering of the checks in the file

tzyl (approved):

PR #1357, opened 2025-04-01
Fix RDP aggregation types and add tests

nihalbhatnagar:
Fixes regression in 2.2.0-beta.13


packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -111,6 +114,8 @@ type Aggregatable<
     V extends `${infer N}:${infer P}`
       ? P extends CollectWithPropAggregations
         ? Array<CompileTimeMetadata<Q>["properties"][N]["type"]> | undefined
+      : P extends MinMaxWithPropAggregateOption
```
nihalbhatnagar, line 117: Ensures that Min and Max take on the type of the property they're applied to

ssanjay1 (approved):

PR #1358, opened 2025-04-01
Fixes RDP types

ericanderson:
Rough gist, if you assigned the result of loading RDPs, the typechecker might infer the wrong thing. If you had let it assign to a typeless const and then compared it was fine though! Very strange.

Basically this should have worked in `ObjectSet.test.ts` and didn't:

```ts
    const shouldBeAssignablePage: PageResult<
          Osdk.Instance<
            EmployeeApiTest,
            never,
            PropertyKeys<EmployeeApiTest>,
            {
              mom: "integer";
              dad: "string" | undefined;
              sister: "string"[] | undefined;
            }
          >
        > = await withFamily.fetchPage();

        const shouldBeAssignableSingle: Osdk.Instance<
          EmployeeApiTest,
          never,
          PropertyKeys<EmployeeApiTest>,
          {
            mom: "integer";
            dad: "string" | undefined;
            sister: "string"[] | undefined;
          }
        > = await withFamily.fetchOne(1);
```

### DIFF
--- a/etc/api.report.api.md
+++ b/etc/api.report.api.md
@@ -232,7 +232,7 @@ export interface AsyncIterArgs<
 	S extends NullabilityAdherence = NullabilityAdherence.Default,
 	T extends boolean = false,
 	RDP_KEYS extends string = never
-> extends SelectArg<Q, K, R, S, RDP_KEYS>, OrderByArg<Q, PropertyKeys<Q>> {
+> extends SelectArg<Q, K, R, S>, OrderByArg<Q, PropertyKeys<Q>> {
     	// (undocumented)
     $__UNSTABLE_useOldInterfaceApis?: boolean;
     	// (undocumented)
@@ -791,6 +791,23 @@ export interface ObjectQueryDataType<T_Target extends ObjectTypeDefinition = nev
     object: string;
 }
 
+// @public (undocumented)
+export namespace ObjectSet {
+    	// (undocumented)
+    export interface FetchPageSignature<
+    		Q extends ObjectOrInterfaceDefinition,
+    		RDPs extends Record<string, SimplePropertyDef> = {}
+    	> {
+        		// Warning: (tsdoc-malformed-inline-tag) Expecting a TSDoc tag starting with "{@"
+        // Warning: (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
+        // Warning: (ae-forgotten-export) The symbol "ValidFetchPageArgs" needs to be exported by the entry point index.d.ts
+        // Warning: (ae-forgotten-export) The symbol "ExtractOptions2" needs to be exported by the entry point index.d.ts
+        // Warning: (ae-forgotten-export) The symbol "SubSelectKeys" needs to be exported by the entry point index.d.ts
+        // Warning: (ae-forgotten-export) The symbol "SubSelectRDPs" needs to be exported by the entry point index.d.ts
+        <const X extends ValidFetchPageArgs<Q, RDPs> = never>(args?: X): Promise<PageResult<Osdk.Instance<Q, ExtractOptions2<X>, SubSelectKeys<Q, X>, SubSelectRDPs<RDPs, X>>>>;
+        	}
+}
+
 // Warning: (ae-forgotten-export) The symbol "ObjectSetCleanedTypes" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "ExtractRdp" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "MergeObjectSet" needs to be exported by the entry point index.d.ts
@@ -1144,13 +1161,12 @@ export interface SelectArg<
 	Q extends ObjectOrInterfaceDefinition,
 	L extends PropertyKeys<Q> = PropertyKeys<Q>,
 	R extends boolean = false,
-	S extends NullabilityAdherence = NullabilityAdherence.Default,
-	RDP_KEYS extends string = never
+	S extends NullabilityAdherence = NullabilityAdherence.Default
 > {
     	// (undocumented)
     $includeRid?: R;
     	// (undocumented)
-    $select?: readonly (L | RDP_KEYS)[];
+    $select?: readonly L[];
 }
 
 // @public (undocumented)
--- a/etc/api.report.api.md
+++ b/etc/api.report.api.md
@@ -791,23 +791,6 @@ export interface ObjectQueryDataType<T_Target extends ObjectTypeDefinition = nev
     object: string;
 }
 
-// @public (undocumented)
-export namespace ObjectSet {
-    	// (undocumented)
-    export interface FetchPageSignature<
-    		Q extends ObjectOrInterfaceDefinition,
-    		RDPs extends Record<string, SimplePropertyDef> = {}
-    	> {
-        		// Warning: (tsdoc-malformed-inline-tag) Expecting a TSDoc tag starting with "{@"
-        // Warning: (tsdoc-escape-right-brace) The "}" character should be escaped using a backslash to avoid confusion with a TSDoc inline tag
-        // Warning: (ae-forgotten-export) The symbol "ValidFetchPageArgs" needs to be exported by the entry point index.d.ts
-        // Warning: (ae-forgotten-export) The symbol "ExtractOptions2" needs to be exported by the entry point index.d.ts
-        // Warning: (ae-forgotten-export) The symbol "SubSelectKeys" needs to be exported by the entry point index.d.ts
-        // Warning: (ae-forgotten-export) The symbol "SubSelectRDPs" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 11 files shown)


etc/api.report.api.md
```diff
@@ -967,10 +986,7 @@ export interface PropertyDef<
 }
 
 // @public (undocumented)
-export type PropertyKeys<
-	O extends ObjectOrInterfaceDefinition,
-	RDPs extends Record<string, SimplePropertyDef> = {}
-> = (keyof NonNullable<O["__DefinitionMetadata"]>["properties"] | keyof RDPs) & string;
+export type PropertyKeys<O extends ObjectOrInterfaceDefinition> = (keyof NonNullable<O["__DefinitionMetadata"]>["properties"]) & string;
```
ericanderson, line 989: Changing this back to have `PropertyKeys` continue to mean explicitly the ones defined on the object.

packages/api/src/OsdkObjectFrom.test.ts
```diff
@@ -343,21 +342,12 @@ describe("ExtractOptions", () => {
...
-    ): Awaited<ReturnType<ObjectSet<O>["fetchPage"]>>["data"] {
-      return {} as any;
-    }
-
-    class Helper<O extends ObjectOrInterfaceDefinition> {
-      constructor(private objectSet: ObjectSet<O>) {}
-      public go() {
-        return createObjectSetChannel(this.objectSet);
-      }
-    }
```
ericanderson, line 357: These helpers were fragile. Don't need them

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -34,13 +43,32 @@ describe("ObjectSet", () => {
...
+  const fauxObjectSet = createMockObjectSet<EmployeeApiTest>();
 
   describe("normal", () => {
     test("select none", async () => {
       const result = await fauxObjectSet.fetchPage();
       expectTypeOf<typeof result.data[0]>().toEqualTypeOf<
-        Osdk.Instance<EmployeeApiTest, never>
+        Osdk.Instance<EmployeeApiTest, never, PropertyKeys<EmployeeApiTest>>
+      >();
+
```
ericanderson, line 63: Below I am testing if you had previously used explicit generics on `fetchPage` that it continues to work.

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -52,13 +80,127 @@ describe("ObjectSet", () => {
...
+            "attachment",
+            "class",
+            "employeeId",
+            "fullName",
+            "geopoint",
+            "geotimeSeriesReference",
+            "isActive",
+            "mediaReference",
+            "timeseries",
+          ],
```
ericanderson, line 164: I have a version of this that just says you can have `$select` or $includeAllBaseObjectProperties` which we should consider.

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -52,13 +80,127 @@ describe("ObjectSet", () => {
...
+  test("includeRid", async () => {
+    const x = await fauxObjectSet
+      .where({ class: "idk" })
+      .fetchPage({ $includeRid: true });
+
+    expectTypeOf(x).toEqualTypeOf<
+      PageResult<
+        Osdk.Instance<EmployeeApiTest, "$rid">
+      >
+    >();
```
ericanderson, line 178: Just want to be sure we didn't break this either.

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -52,13 +80,127 @@ describe("ObjectSet", () => {
...
   describe(".withProperties", () => {
     test("single property", async () => {
       const withA = fauxObjectSet.withProperties({
         "a": (base) => {
           return base.pivotTo("lead").aggregate("class:exactDistinct");
         },
       });
+      withA.pivotTo("lead");
+
+      const isWithAAssignable: $ObjectSet<EmployeeApiTest, {}> = withA;
```
ericanderson, line 203: One of the weird things we saw from the bug report was the way the inference happened if it was assigned right away from a return result or if it was assigned after the fact. Just bolstering things here.

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -78,6 +220,15 @@ describe("ObjectSet", () => {
...
     });
 
+    it("can be sub-selected", () => {
+      const objectWithUndefinedRdp = fauxObjectSet.withProperties({
+        "derivedPropertyName": (base) =>
+          base.pivotTo("lead").selectProperty("employeeId"),
+      }).fetchOne(3, {
+        $select: ["derivedPropertyName"],
+      });
+    });
```
ericanderson, line 230: A version of this test broke for me in @osdk/client so I wrote one in @osdk/api to iterate faster.

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -385,13 +607,13 @@ interface Subscribe<
...
   D extends Record<string, SimplePropertyDef>,
-  MERGED extends ObjectOrInterfaceDefinition,
+  MERGED extends ObjectOrInterfaceDefinition & Q,
 > extends
   MinimalObjectSet<Q, D>,
   WithProperties<Q, D>,
   Aggregate<MERGED>,
   SetArithmetic<MERGED>,
-  PivotTo<MERGED>,
+  PivotTo<Q>,
```
ericanderson, line 616: I don't think it makes sense to retain the "merged" object when you pivot since those properties are contextually gone now.

packages/client/src/object/convertWireToOsdkObjects/getDollarLink.ts
```diff
@@ -49,21 +50,25 @@ export function get$link(
 
         const value = !linkDef.multiplicity
           ? {
-            fetchOne: <A extends SelectArg<any>>(options?: A) =>
+            fetchOne: <A extends SelectArg<any, any, any, any>>(
+              options?: A,
+            ) =>
```
ericanderson, line 55: Overloads required more specificity

packages/e2e.sandbox.catchall/src/runInterfacesTest2.ts
```diff
@@ -62,6 +62,7 @@ export async function runInterfacesTest2(): Promise<void> {
     >
   >(true);
 
+  // You cannot specify both $select and $includeAllBaseObjectProperties
```
ericanderson, line 65: When this failed for me during dev I as confused as to why, so extra comment for future us

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -220,6 +402,34 @@ describe("ObjectSet", () => {
               }
             >
           >();
+
```
ericanderson, line 405: This is the test that confirms the bug is gone.

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -67,11 +108,93 @@ export interface MinimalObjectSet<
...
 > {
+  readonly fetchPage: FetchPageSignature<Q, RDPs>;
+  readonly fetchPageWithErrors: FetchPageWithErrorsSignature<Q, RDPs>;
+}
+
+type ValidFetchPageArgs<
+  Q extends ObjectOrInterfaceDefinition,
+  RDPs extends Record<string, SimplePropertyDef>,
+> =
+  | ObjectSetArgs.FetchPage<
```
nihalbhatnagar, line 136: I am a bit confused by this, why do we need to separate these two cases out?

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -39,22 +39,63 @@ import type {
...
+
+type MaybeSimplifyPropertyKeys<
+  Q extends ObjectOrInterfaceDefinition,
+  L extends PropertyKeys<Q>,
+> = PropertyKeys<Q> extends L ? PropertyKeys<Q> : L;
+
+type SubSelectKeysHelper<
+  Q extends ObjectOrInterfaceDefinition,
+  L extends string,
+> = [L] extends [never] ? PropertyKeys<Q>
```
nihalbhatnagar, line 75: I love subselect 😢

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -39,22 +39,63 @@ import type {
...
+  : T extends abstract new(...args: any[]) => any ? T
+  : { [K in keyof T]: T[K] };
+
+type SubSelectRDPsHelper<
+  X extends ValidFetchPageArgs<any, any> | ValidAsyncIterArgs<any, any>,
+  DEFAULT extends string,
+> = [X] extends [never] ? DEFAULT
+  : (X["$select"] & string[])[number] & DEFAULT;
+
+type SubSelectRDPs<
```
nihalbhatnagar, line 94: Could you explain how this approach is circumventing the problem of both inferring and expecting the generic to be provided that caused the previous bug?

nihalbhatnagar (approved):

PR #1352, opened 2025-03-31
Update templates to use getOsdkConfig utility

zeyadkhaled:
This PR applies changes to supported 2.x templates to make them deployable across stacks by using a deployment.config.file. This only affects Foundry-hosted applications.

1. Update client.ts to use getOsdkConfig
2. Add utilities for generating deployment.config.json
3. Update examples to use the new config file
4. Stop generating .env.production and remove existing test/introduce new test for config file
5. Update the templates README and introduce new test

PR #1363, opened 2025-04-02
Update templates with meta tags and use it for client creation

zeyadkhaled:
This PR is in place of https://github.com/palantir/osdk-ts/pull/1352 and it is meant to be a simplification from the original one. 

We leverage [VITE constant replacement](https://vite.dev/guide/env-and-mode#html-constant-replacement) to inject the .env variables in the index.html. We then read those injected meta tags while intializing the client.

The PR adds support for providing ontologyRid in create-app args and uses that to write the FOUNDRY_ONTOLOGY_RID env variable in the generated .env files so they could be injected in the index.html.

This allows the template to remain usable by both foundry-hosted apps and externally hosted apps.

### DIFF
--- a/.changeset/calm-ducks-bake.md
+++ b/.changeset/calm-ducks-bake.md
@@ -1,14 +1,14 @@
 ---
-"@osdk/create-app.template.tutorial-todo-aip-app.beta": minor
-"@osdk/create-app.template.tutorial-todo-app.beta": minor
-"@osdk/create-app.template.tutorial-todo-aip-app": minor
-"@osdk/create-app.template.tutorial-todo-app": minor
-"@osdk/create-app.template.react.beta": minor
-"@osdk/create-app.template.vue.v2": minor
-"@osdk/create-app.template.react": minor
-"@osdk/create-app.template.vue": minor
-"@osdk/example-generator": minor
-"@osdk/create-app": minor
+"@osdk/create-app.template.tutorial-todo-aip-app.beta": patch
+"@osdk/create-app.template.tutorial-todo-app.beta": patch
+"@osdk/create-app.template.tutorial-todo-aip-app": patch
+"@osdk/create-app.template.tutorial-todo-app": patch
+"@osdk/create-app.template.react.beta": patch
+"@osdk/create-app.template.vue.v2": patch
+"@osdk/create-app.template.react": patch
+"@osdk/create-app.template.vue": patch
+"@osdk/example-generator": patch
+"@osdk/create-app": patch
 ---
 
-Update templates with meta tags
+Read environment variables from index.html meta tags
--- a/.changeset/calm-ducks-bake.md
+++ b/.changeset/calm-ducks-bake.md
@@ -1,0 +1,14 @@
+---
+"@osdk/create-app.template.tutorial-todo-aip-app.beta": minor
+"@osdk/create-app.template.tutorial-todo-app.beta": minor
+"@osdk/create-app.template.tutorial-todo-aip-app": minor
+"@osdk/create-app.template.tutorial-todo-app": minor
+"@osdk/create-app.template.react.beta": minor
+"@osdk/create-app.template.vue.v2": minor
+"@osdk/create-app.template.react": minor
+"@osdk/create-app.template.vue": minor
+"@osdk/example-generator": minor
+"@osdk/create-app": minor
+---
+
+Update templates with meta tags
--- a/packages/create-app.template.react.beta/templates/src/client.ts.hbs
+++ b/packages/create-app.template.react.beta/templates/src/client.ts.hbs
@@ -5,7 +5,7 @@ function getMetaTagContent(tagName: string): string {
   const elements = document.querySelectorAll(`meta[name="${tagName}"]`);
   const element = elements.item(elements.length - 1);
   const value = element ? element.getAttribute("content") : null;
-  if (value == null) {
+  if (value == null || value === "") {
     throw new Error(`Meta tag ${tagName} not found`);
   }
   if (value.match(/%.+%/)) {
--- a/packages/create-app.template.react.beta/templates/src/client.ts.hbs
+++ b/packages/create-app.template.react.beta/templates/src/client.ts.hbs
@@ -1,11 +1,20 @@
-import type { Client } from "@osdk/client";
-import { createClient } from "@osdk/client";
-import { $ontologyRid } from "{{osdkPackage}}";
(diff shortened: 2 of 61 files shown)


packages/create-app.template.react.beta/templates/src/client.ts.hbs
```diff
@@ -40,8 +36,8 @@ export const auth = createPublicOauthClient(
  * Initialize the client to interact with the Ontology SDK
  */
 const client: Client = createClient(
-  url,
-  $ontologyRid,
```
tzyl, line 44: It's a bit weird in the single stack case that we add duplication for the ontology rid when this is already part of the SDK but I don't think it's a big deal when we prefill this correctly most of the time

packages/create-app.template.react.beta/templates/src/client.ts.hbs
```diff
@@ -1,11 +1,20 @@
...
-import { createPublicOauthClient } from "@osdk/oauth";
+import { createClient, type Client } from "@osdk/client";
+import { createPublicOauthClient, type PublicOauthClient } from "@osdk/oauth";
 
-const url = import.meta.env.VITE_FOUNDRY_API_URL;
-const clientId = import.meta.env.VITE_FOUNDRY_CLIENT_ID;
-const redirectUrl = import.meta.env.VITE_FOUNDRY_REDIRECT_URL;
+function getMetaTagContent(tagName: string): string {
+  const elements = document.querySelectorAll(`meta[name="${tagName}"]`);
+  const element = elements.item(elements.length - 1);
```
tzyl, line 6: I know you have temporary reasons for this but maybe if we choose from scratch then following the behaviour of `document.querySelector` and picking the first rather than last tag which is what workspace does would follow the fact that the browser also executes the page in order (earlier scripts run before later scripts) but I guess you can make the argument that the later one overrides the first one

packages/example-generator/src/run.ts
```diff
@@ -91,6 +91,7 @@ async function generateExamples(tmpDir: tmp.DirResult): Promise<void> {
         "https://fake.palantirfoundry.com/artifacts/api/repositories/ri.artifacts.main.repository.fake/contents/release/npm",
       corsProxy: false,
       scopes: ["api:ontologies-read", "api:ontologies-write"],
+      ontologyRid: "ri.ontologies.main.ontology.fake",
```
tzyl, line 94:
```suggestion
      ontologyRid: "ri.ontology.main.ontology.fake",
```

packages/create-app/src/cli.ts
```diff
@@ -141,6 +147,7 @@ export async function cli(args: string[] = process.argv): Promise<void> {
   const osdkRegistryUrl: string = await promptOsdkRegistryUrl(parsed);
   const corsProxy: boolean = await promptCorsProxy(parsed);
   const scopes: string[] | undefined = await promptScopes(parsed);
+  const ontologyRid: string = await promptOntologyRid(parsed);
```
tzyl, line 150: Does it make sense conceptually to prompt for this last? We expect to help prefill this most of the time but should it be grouped after the application rid?

packages/create-app/src/cli.ts
```diff
@@ -122,6 +124,10 @@ export async function cli(args: string[] = process.argv): Promise<void> {
             array: true,
             describe:
               "List of client-side scopes to be used when creating a client",
+          })
+          .option("ontologyRid", {
+            type: "string",
+            describe: "RID of the Ontology your application is using",
```
tzyl, line 130: If we do change the prompt order then let's also change this option definition order. For the consistency of the terminology can you follow how application is defined? (call the option ontology and describe it as `Ontology resource identifier (rid)``

```
          .option("application", {
            type: "string",
            describe: "Application resource identifier (rid)",
          })
```

.changeset/calm-ducks-bake.md
```diff
@@ -0,0 +1,14 @@
...
+"@osdk/create-app.template.tutorial-todo-app": minor
+"@osdk/create-app.template.react.beta": minor
+"@osdk/create-app.template.vue.v2": minor
+"@osdk/create-app.template.react": minor
+"@osdk/create-app.template.vue": minor
+"@osdk/example-generator": minor
+"@osdk/create-app": minor
+---
+
+Update templates with meta tags
```
tzyl, line 14: Can you make this describe a little more what this means? e.g. configuration environment variables are now read from index.html meta tags

.changeset/calm-ducks-bake.md
```diff
@@ -0,0 +1,14 @@
...
+"@osdk/create-app.template.tutorial-todo-aip-app.beta": minor
+"@osdk/create-app.template.tutorial-todo-app.beta": minor
+"@osdk/create-app.template.tutorial-todo-aip-app": minor
+"@osdk/create-app.template.tutorial-todo-app": minor
+"@osdk/create-app.template.react.beta": minor
+"@osdk/create-app.template.vue.v2": minor
+"@osdk/create-app.template.react": minor
+"@osdk/create-app.template.vue": minor
+"@osdk/example-generator": minor
+"@osdk/create-app": minor
```
tzyl, line 11: Depending what you need to do with this change (check with team) it may need to be a patch if it has to be cherry picked

packages/create-app/src/cli.ts
tzyl: Also if you don't mind keeping the README up to date (looks like we should put scopes in there too)
https://github.com/palantir/osdk-ts/tree/main/packages/create-app#options

packages/create-app.template.react.beta/templates/src/client.ts.hbs
zeyadkhaled: Yeah we had already merged the change to replace instead of append. But I think the argument of choosing last added makes more sense than first.

packages/create-app/src/cli.ts
zeyadkhaled: Good flag. I was taking the path of just stacking the options at the end. But grouping it with application makes sense too.

packages/create-app.template.react.beta/templates/src/client.ts.hbs
```diff
@@ -1,11 +1,23 @@
...
+  const value = element ? element.getAttribute("content") : null;
+  if (value == null) {
```
tzyl, line 8: Should we do another check for empty string in case value is set but empty?

tzyl (approved):

tzyl (approved):

PR #1365, opened 2025-04-02
Runtime Derived Property Expressions

nihalbhatnagar:
This PR lays the groundwork for the planned full capabilities of calculated RDP's in the near future. The backend does not support some features of this PR:

- Inlined property definitions within another definition (`base.selectProperty("hello").add(base.pivotTo("a).selectProperty("bye")`)
- Literal expressions
(`base.constant.double(5)`)

Additionally, we are still finalizing the API for extractDate. These should be commented out to be added incrementally as features are added.

All supported syntax can be viewed in the test bed in `ObjectSet.test.ts`.

Below is common syntax as well as their hover definitions

Storing the builder in a variable:
image

Hover for a nested derived property definition on a calculation:
image

Hover for a property reference on a calculation:
image


packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -23,45 +23,78 @@ import type {
...
+import type {
+  DatetimeExpressions,
+  DefinitionForType,
+  NumericExpressions,
+} from "./Expressions.js";
 import type { CollectWithPropAggregations } from "./WithPropertiesAggregationOptions.js";
 
 export namespace DerivedProperty {
-  export type SelectorResult<
+  export interface Definition<
```
nihalbhatnagar, line 34: I chose the terminology of `Definition` to extend throughout all the types because I felt like it conveyed the fact that this was returnable at any time. Every "Definition" is valid and doesn't need to be extended by a calculation

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -23,45 +23,78 @@ import type {
...
 
   export type Clause<
     Q extends ObjectOrInterfaceDefinition,
   > = {
-    [key: string]: Selector<Q, SimplePropertyDef>;
+    [key: string]: Creator<Q, SimplePropertyDef>;
   };
 
-  export type Selector<
+  export type Creator<
```
nihalbhatnagar, line 57: Creator seemed most relevant to Definition. Additionally, these aren't just used for "selecting" anymore

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -23,45 +23,78 @@ import type {
...
-    baseObjectSet: DerivedProperty.Builder<Q, false>,
-  ) => SelectorResult<T>;
+    baseObjectSet: Builder<Q, false>,
+  ) =>
+    | Definition<T, Q>
+    | NumericPropertyDefinition<T, Q>
+    | DatetimePropertyDefinition<T, Q>;
 
-  export interface Builder<
+  interface BaseBuilder<
```
nihalbhatnagar, line 67: Not sure of a better name for this. This was abstracted out to make sure that the types it inherits can be easily extended by its children and maintains the readability of this syntax. If we hate it, easy enough to remove and have it's children inherit Filterable<Q, CONSTRAINED>, Pivotable<Q, CONSTRAINED>

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -23,45 +23,78 @@ import type {
...
+  > extends BaseBuilder<Q, CONSTRAINED>, Aggregatable<Q> {
   }
 
   export interface SelectPropertyBuilder<
     Q extends ObjectOrInterfaceDefinition,
     CONSTRAINED extends boolean,
   > extends AggregateBuilder<Q, CONSTRAINED>, Selectable<Q> {
   }
+
+  export type ValidParts =
```
nihalbhatnagar, line 91: Will probably remove when if we merge for a beta

packages/api/src/derivedProperties/Expressions.ts
```diff
@@ -0,0 +1,197 @@
...
+  : SimplePropertyDef.ExtractWirePropertyType<T> extends WithPropertiesDatetime
+    ? DerivedProperty.DatetimePropertyDefinition<T, Q>
+  : DerivedProperty.Definition<T, Q>;
+
+type NumericExpressionArg<Q extends ObjectOrInterfaceDefinition> =
+  | number
+  | FilteredPropertyKeys<Q, WithPropertiesNumerics>
+  | DerivedProperty.NumericPropertyDefinition<any, Q>;
+
+type ReturnTypeForNumericMethod<
```
nihalbhatnagar, line 51: This handles the type coercion logic for numerics.

packages/api/src/derivedProperties/Expressions.ts
```diff
@@ -0,0 +1,197 @@
...
+  : "long" extends (LEFT | RIGHT) ? DerivedProperty.NumericPropertyDefinition<
+      SimplePropertyDef.Make<"long", false, false>,
+      Q
+    >
+  : DerivedProperty.NumericPropertyDefinition<
+    SimplePropertyDef.Make<"integer", false, false>,
+    Q
+  >;
+
+type ReturnTypeForDatetimeMethod<
```
nihalbhatnagar, line 72: Abstracting this doesn't affect the hover readability 😄

packages/client/src/derivedProperties/derivedPropertyDefinitionFactory.ts
```diff
@@ -0,0 +1,121 @@
...
+import type { DerivedProperty } from "@osdk/api";
+import type { DerivedPropertyDefinition } from "@osdk/foundry.ontologies";
+import invariant from "tiny-invariant";
+
+/** @internal */
+export function derivedPropertyDefinitionFactory(
+  wireDefinition: DerivedPropertyDefinition,
+  definitionMap: Map<any, DerivedPropertyDefinition>,
+): DerivedProperty.NumericPropertyDefinition<any, any> {
+  const definition: DerivedProperty.NumericPropertyDefinition<any, any> = {
```
nihalbhatnagar, line 26: Missing extractPart for now

packages/api/src/derivedProperties/Expressions.ts
```diff
@@ -0,0 +1,195 @@
...
+  : ARG extends FilteredPropertyKeys<Q, WithPropertiesNumerics>
+    ? NonNullable<CompileTimeMetadata<Q>["properties"][ARG]["type"]>
+  : never;
+
+export type NumericExpressions<
+  Q extends ObjectOrInterfaceDefinition,
+  LEFT_PROPERTY_TYPE extends SimplePropertyDef,
+> = {
+  readonly add: <A extends NumericExpressionArg<Q>>(
+    value: A,
```
nihalbhatnagar, line 102: Opted for a single property rather than many args in order to keep syntax simple

packages/api/src/derivedProperties/DerivedProperty.ts
ssanjay1: Yea lets remove if we're not using

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -95,6 +128,43 @@ type Pivotable<
     : DerivedProperty.SelectPropertyBuilder<LinkedType<Q, L>, false>;
 };
 
+type Constant<Q extends ObjectOrInterfaceDefinition> = {
```
ssanjay1, line 131: So the constants aren't supported yet right? If so, we can we leave a comment on anything in this PR that isn't officially supported?

packages/client/src/derivedProperties/createWithPropertiesObjectSet.test.ts
```diff
@@ -114,4 +114,103 @@ describe(createWithPropertiesObjectSet, () => {
...
+  describe("expressions", () => {
+    it("can handle expressions referencing other property keys", () => {
+      const map = new Map<any, DerivedPropertyDefinition>();
+      const deriveObjectSet = createWithPropertiesObjectSet(Employee, {
+        type: "methodInput",
+      }, map);
+
+      const clause: DerivedProperty.Clause<Employee> = {
+        "derivedPropertyName": (base) =>
+          base.pivotTo("lead").selectProperty("employeeId").add("employeeId"),
```
ssanjay1, line 127: So in this example I'm adding the same property to itself?

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
```diff
@@ -96,27 +101,47 @@ export function createWithPropertiesObjectSet<
...
           type: "get",
           selectedPropertyApiName: name,
         },
-      });
-      return selectorResult;
+      };
+      const selectorResult: DerivedProperty.Definition<any, any> =
+        derivedPropertyDefinitionFactory(wrappedObjectSet, definitionMap);
+      definitionMap.set(selectorResult, wrappedObjectSet);
+      return selectorResult as any;
```
ssanjay1, line 126: Do we need the any here?

packages/api/src/derivedProperties/DerivedProperty.ts
ssanjay1: Hmm, maybe we can rename the original `Builder` to something else?

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -95,6 +128,43 @@ type Pivotable<
...
+      value: number,
+    ) => DerivedProperty.NumericPropertyDefinition<
+      SimplePropertyDef.Make<"double", false, false>,
+      Q
+    >;
+
+    readonly integer: (
+      value: number,
+    ) => DerivedProperty.NumericPropertyDefinition<
+      SimplePropertyDef.Make<"integer", false, false>,
```
ssanjay1, line 143: Not blocking, but we use this Make type a lot now with both of the generic options, maybe we should consider using consts to denote what each option really is for better readability here

packages/api/src/derivedProperties/Expressions.ts
```diff
@@ -0,0 +1,195 @@
...
+  : DerivedProperty.Definition<T, Q>;
+
+type NumericExpressionArg<Q extends ObjectOrInterfaceDefinition> =
+  | number
+  | FilteredPropertyKeys<Q, WithPropertiesNumerics>
+  | DerivedProperty.NumericPropertyDefinition<any, Q>;
+
+type ReturnTypeForNumericMethod<
+  Q extends ObjectOrInterfaceDefinition,
+  LEFT extends WirePropertyTypes,
```
ssanjay1, line 53: Can we use something besides left right to make it more clear what we're talking about here?

ssanjay1: Also, is there ever going to be a case where the type coercion will be different basedon if the type is in left or right? As of now, its just looking at either arg

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
nihalbhatnagar: Yeah we start running into type weirdness by conditionally returning datetime or numeric definitions here, similar in aggregate

packages/client/src/derivedProperties/createWithPropertiesObjectSet.test.ts
nihalbhatnagar: Yeah, it's a bit odd but still highlights the functionality we need. It was the only numeric on the object type

nihalbhatnagar: I'll probably add more extensive client tests once we get backend support for the things we need

packages/api/src/derivedProperties/DerivedProperty.ts
nihalbhatnagar: Yeah I wanted this PR to be what we'd have if everything worked perfectly and if we want some basic features out in a beta we can just comment these out. So TLDR yeah we will comment this out and explain why, and when the time comes we will uncomment

packages/api/src/derivedProperties/DerivedProperty.ts
nihalbhatnagar: cool will remove

packages/api/src/derivedProperties/Expressions.ts
nihalbhatnagar: Do you have any ideas? The best I can think of is LEFT_PROPERTY_TYPE and RIGHT_PROPERTY_TYPE, or are you saying the concept of left and right doesn't make sense?

nihalbhatnagar: And as of right now no, and I don't think so in the future as I imagine that would lead to a but of a weird API. I will confirm with OSS but this in a scenario where it does matter It's not too hard to modify this type in that case

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
```diff
@@ -32,8 +37,8 @@ export function createWithPropertiesObjectSet<
   objectType: Q,
   objectSet: WireObjectSet,
   definitionMap: Map<any, DerivedPropertyDefinition>,
-): DerivedProperty.SelectPropertyBuilder<Q, false> {
-  const base: DerivedProperty.SelectPropertyBuilder<Q, false> = {
+): WithConstSelect<Q> {
+  const base: WithConstSelect<Q> = {
```
nihalbhatnagar, line 41: Just doing some pattern matching with create object set which is why Ive typed both here and saved it in a type, but I think this is probably not needed

packages/client/src/derivedProperties/derivedPropertyDefinitionFactory.ts
```diff
@@ -0,0 +1,138 @@
...
+    divide(
+      value,
+    ) {
+      return derivedPropertyDefinitionFactory({
+        "type": "subtract",
+        "left": wireDefinition,
+        "right": getDefinitionFromMap(value, definitionMap),
+      }, definitionMap);
+    },
+    type: 1,
```
ssanjay1, line 107: What is this?

packages/client/src/derivedProperties/derivedPropertyDefinitionFactory.ts
```diff
@@ -0,0 +1,138 @@
...
+    any
+  >["extractPart"];
+} {
+  const definition: DerivedProperty.NumericPropertyDefinition<any, any> & {
+    extractPart: DerivedProperty.DatetimePropertyDefinition<
+      any,
+      any
+    >["extractPart"];
+  } = {
+    abs() {
```
ericanderson, line 37: We may want to come back through here later and do something clever with classes or prototypes but this is really just a future thought to myself more than anything else so you can pretty much disregard.

ericanderson (commented):
Aligned on changes and strategies. Just a few questions and we're good on my end.

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
```diff
@@ -22,8 +22,13 @@ import type {
 } from "@osdk/foundry.ontologies";
 import invariant from "tiny-invariant";
 import { modernToLegacyWhereClause } from "../internal/conversions/modernToLegacyWhereClause.js";
+import { derivedPropertyDefinitionFactory } from "./derivedPropertyDefinitionFactory.js";
 
-const idCounter = 0;
+type WithConstSelect<Q extends ObjectOrInterfaceDefinition> =
```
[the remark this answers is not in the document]
nihalbhatnagar: Yeah I can make this an interface. I'm using this type to make sure that we can get type safety on our methods and what we return. We don't have a good interface to type our return that has every method on it (selectProperty, aggregate, pivotTo, etc.), so I'm using SelectPropertyBuilder which has all of them except constant which is only exposed off the `Builder` api.

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -110,27 +174,52 @@ type Aggregatable<
...
     SimplePropertyDef.Make<
       CompileTimeMetadata<Q>["properties"][R]["type"],
-      CompileTimeMetadata<Q>["properties"][R]["nullable"],
-      CompileTimeMetadata<Q>["properties"][R]["multiplicity"]
+      CompileTimeMetadata<Q>["properties"][R]["nullable"] extends true
+        ? "nullable"
+        : "non-nullable",
+      CompileTimeMetadata<Q>["properties"][R]["multiplicity"] extends true
+        ? "array"
+        : "single"
```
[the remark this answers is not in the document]
nihalbhatnagar: Do you mean doing this instead: 
image
The hover stays the same 🤷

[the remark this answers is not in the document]
nihalbhatnagar: Do you mean like why we made it "string"[] rather than `string[]`? I think we did this so the readability of our hover was exactly our wire types 
image

packages/api/src/derivedProperties/DerivedProperty.ts
nihalbhatnagar: Mm maybe, tho wondering if this aligns with standard java factories which are usually classes with the factory methods exposed off of them

packages/client/src/derivedProperties/createWithPropertiesObjectSet.test.ts
nihalbhatnagar: This was wrong before but the updated version is more clear as to whats happening

ericanderson (approved):

PR #1370, opened 2025-04-03
Add geotime and mediareference datasources in OAC

davidpaquin:
Some decisions that were made:
- Right now it's impossible to use the same geotime/mediaset datasource for multiple properties. We can add that later if necessary though
- ~~If all properties are exotic, then we ignore the object-level datasource, since it would have no properties sourced from it~~
- Timeseries was not included yet, since it's a special case and includes some extra metadata that's unrelated to these sources

### DIFF
--- a/packages/maker/src/api/defineOntology.ts
+++ b/packages/maker/src/api/defineOntology.ts
@@ -179,14 +179,20 @@ function convertToWireOntologyIr(
 function convertObject(
   objectType: ObjectType,
 ): OntologyIrObjectTypeBlockDataV2 {
-  const datasource: OntologyIrObjectTypeDatasource = {
-    rid: "ri.ontology.main.datasource.".concat(objectType.apiName),
-    datasource: convertDatasourceDefinition(objectType),
-    editsConfiguration: {
-      onlyAllowPrivilegedEdits: false,
-    },
-    redacted: false,
-  };
+  const propertyDatasources: OntologyIrObjectTypeDatasource[] =
+    (objectType.properties ?? [])
+      .flatMap(prop => extractPropertyDatasource(prop, objectType.apiName));
+
+  const objectDatasource =
+    propertyDatasources.length < (objectType.properties ?? []).length
+      ? [buildDatasource(
+        objectType.apiName,
+        convertDatasourceDefinition(
+          objectType,
+          objectType.properties!.filter(p => !isExotic(p)),
+        ),
+      )]
+      : [];
 
   const implementations = objectType.implementsInterfaces ?? [];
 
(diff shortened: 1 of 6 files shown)


packages/maker/src/api/defineOntology.ts
```diff
@@ -273,6 +327,12 @@ function convertDatasourceDefinition(
   }
 }
 
+function isExotic(property: ObjectPropertyType): boolean {
```
aep000, line 330: This function is a good use case for type narrowing predicates: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

This way you don't need to do the "as" casting in the extract method. You can also use a [typeof typeguard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#typeof-type-guards) to narrow in this method.

packages/maker/src/api/overall.test.ts
```diff
@@ -2945,5 +2945,269 @@ describe("Ontology Defining", () => {
...
       );
     });
+
+    it("Property-level datasources are properly defined", () => {
+      const objectDatasourceIgnored = defineObject({
+        titlePropertyApiName: "bar",
+        displayName: "objectDatasourceIgnored",
+        pluralDisplayName: "objectDatasourceIgnored",
+        apiName: "foo",
+        primaryKeys: ["bar"],
```
aep000, line 2955: FLUP: We should validate types for primary keys because geotime and media can't be primary keys.

packages/maker/src/api/overall.test.ts
```diff
@@ -2945,5 +2945,269 @@ describe("Ontology Defining", () => {
         `[Error: Invariant failed: Retention period "bad retention period string" on object "buzz" is not a valid ISO 8601 duration string]`,
       );
     });
+
+    it("Property-level datasources are properly defined", () => {
+      const objectDatasourceIgnored = defineObject({
+        titlePropertyApiName: "bar",
```
aep000, line 2951: FLUP: We should also look into whether titles can be anything but primitive types.

packages/maker/src/api/defineOntology.ts
```diff
@@ -179,14 +179,20 @@ function convertToWireOntologyIr(
...
-      onlyAllowPrivilegedEdits: false,
-    },
-    redacted: false,
-  };
+  const propertyDatasources: OntologyIrObjectTypeDatasource[] =
+    (objectType.properties ?? [])
+      .flatMap(prop => extractPropertyDatasource(prop, objectType.apiName));
+
+  const objectDatasource =
+    propertyDatasources.length < (objectType.properties ?? []).length
```
aep000, line 187: We should check what OMA does here if you have an object with exclusively a media property or geotime. It may also be that they are mapped here as well if they aren't edit only properties.

aep000 (commented):
Couple thoughts and FLUPs

packages/maker/src/api/defineOntology.ts
aep000: And we are currently assuming that no properties are edit only. (IE they all have a backing datasource)

davidpaquin: It looks like they do get mapped to both

packages/maker/src/api/overall.test.ts
davidpaquin: Checked and it seems like titles are restricted to primitives

davidpaquin: Added a verification to check that title/PKs are primitives

PR #1377, opened 2025-04-07
Fix up extractRdpDefinition for intersect object sets

nihalbhatnagar:
Users started encountering errors of the OSDK fetching the metadata of an empty string object type when using intersection, subtraction, or union object set types. This was due to improperly handled edge cases. This PR cleans up the extractRdpDefinition method and removes as many runtime errors as possible besides those that we know will throw in OSS

### DIFF
--- a/packages/client/src/util/extractRdpDefinition.ts
+++ b/packages/client/src/util/extractRdpDefinition.ts
@@ -55,7 +55,7 @@ async function extractRdpDefinitionInternal(
         );
 
       if (childObjectType === undefined || childObjectType === "") {
-        return { definitions: {}, childObjectType: "" };
+        return { definitions: {} };
       }
       const objDef = await clientCtx.ontologyProvider.getObjectDefinition(
         childObjectType,
@@ -145,6 +145,7 @@ async function extractRdpDefinitionInternal(
           )
         ),
       );
+
       const definitions = objectSetTypes.reduce(
         (acc, { definitions }) => ({ ...acc, ...definitions }),
         {},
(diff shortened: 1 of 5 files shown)


packages/client/src/util/extractRdpDefinition.ts
```diff
@@ -39,36 +38,46 @@ export async function extractRdpDefinition(
 async function extractRdpDefinitionInternal(
   clientCtx: MinimalClient,
   objectSet: ObjectSet,
-  methodInputObjectType: string,
+  methodInputObjectType: string | undefined,
```
nihalbhatnagar, line 41: Not optional so we remember to pass this in

packages/client/src/util/extractRdpDefinition.ts
```diff
@@ -118,12 +133,49 @@ async function extractRdpDefinitionInternal(
...
+        objectSets.map((os) =>
+          extractRdpDefinitionInternal(
+            clientCtx,
+            os,
+            methodInputObjectType,
+          )
+        ),
+      );
+
+      const definitions = objectSetTypes.reduce(
```
nihalbhatnagar, line 149: Building out this logic since this since OSS will likely start just dropping RDPs instead of throwing

packages/client/src/util/extractRdpDefinition.ts
```diff
@@ -39,36 +38,46 @@ export async function extractRdpDefinition(
...
-        methodInputObjectType,
-      );
+      const { definitions, childObjectType } =
+        await extractRdpDefinitionInternal(
+          clientCtx,
+          objectSet.objectSet,
+          methodInputObjectType,
+        );
+
+      if (childObjectType === undefined || childObjectType === "") {
```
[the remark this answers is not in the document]
ericanderson: Prefer to fail hard fast and early for now

ericanderson (approved):

PR #1378, opened 2025-04-07
Publish e2e sandboxes and generated code for minor version testing

nihalbhatnagar:
To be used in #1338

### DIFF
--- a/packages/e2e.sandbox.catchall/package.json
+++ b/packages/e2e.sandbox.catchall/package.json
@@ -1,6 +1,5 @@
 {
   "name": "@osdk/e2e.sandbox.catchall",
-  "private": true,
   "version": "0.4.0-beta.1",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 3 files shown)


packages/e2e.sandbox.catchall/package.json
```diff
@@ -59,7 +58,13 @@
   "publishConfig": {
     "access": "public"
   },
-  "files": [],
+  "files": [
+    "src/",
```
nihalbhatnagar, line 62: Not including build to make sure that if anyone stumbles on this package it's unusable. LMK if this is overkill

ericanderson: This makes sense. I wonder if we should put in a README.md

ericanderson: Yeah okay I'm thinking we have MRL write a README and we add some comments into monorepolint. My thought is in 1 or 2 years we will have no recollection of why we did these things

ericanderson (approved):
Approving to unblock but lets follow up?

PR #1380, opened 2025-04-08
Make release depend on build and e2e succeeding

zeyadkhaled:


.github/workflows/release.yml
```diff
@@ -17,6 +17,9 @@ jobs:
     name: Release
     runs-on: ubuntu-latest
     if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/') }}
+    needs:
```
ericanderson, line 20: Does this work? Can you do this across workflows?

ericanderson: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

PR #1386, opened 2025-04-08
Send baseHref when setting widget set settings

rosscourt:
Send the local dev server's base URL as the `baseHref` value when setting widget set dev mode settings (e.g. `http://localhost:8080/`). This value is used as the `href` value of the `<base />` element for custom widgets when their HTML is returned from the server. This fixes issues with loading image assets from localhost locations when running a widget in dev mode.

### DIFF
--- a/packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
+++ b/packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
@@ -143,7 +143,12 @@ export function FoundryWidgetDevPlugin(): Plugin {
             configFiles,
             getLocalhostUrl(server),
           );
-          await publishDevModeSettings(server, widgetIdToOverrides, res);
+          await publishDevModeSettings(
+            server,
+            widgetIdToOverrides,
+            getBaseHref(server),
+            res,
+          );
         },
       );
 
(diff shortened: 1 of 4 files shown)


packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -222,6 +227,10 @@ function getLocalhostUrl(server: ViteDevServer): string {
   }://localhost:${server.config.server.port}`;
 }
 
+function getBaseHref(server: ViteDevServer): string {
+  return `${getLocalhostUrl(server)}${server.config.base}`;
+}
```
rosscourt, line 232: This value doesn't work for Code Workspaces, but I'll hit that in a single pass for Code Workspaces awareness. There are environment variables that contain the values we need.

tzyl: Do we know if this can handle all the different variations of trailing / leading slashes? I guess this one is controlled by us for the internal vite config used for the dev server so it doesn't need to potentially handle all possible weird cases:
https://github.com/palantir/osdk-ts/blob/7700076bedb06b239b7ba566b9e33a27347d59ec/packages/widget.vite-plugin.unstable/vite.config.ts

tzyl (approved):

packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
rosscourt: Yea I tested with a few cases and this always seems to be a full path with a trailing slash, which is what we want

PR #1394, opened 2025-04-14
Add geoshape action support

ssanjay1:
Actions now support geoshape and geohash parameters


packages/e2e.sandbox.catchall/src/runGeoQueriesTest.ts
```diff
@@ -137,7 +137,16 @@ export async function runGeoQueriesTest(): Promise<void> {
     geometry10M: { $isNull: false },
   }).fetchPage();
 
-  console.log(nonNullGeoProps.data.map(data => data.usState));
+  // console.log(nonNullGeoProps.data.map(data => data.usState));
+
+  // const huh = await client(createTestGeoAction).applyAction({
```
ssanjay1, line 142: Commenting out action so we don't accidentally keep calling it

nihalbhatnagar: Can we uncomment and potentially just do a delete call right after it? Or place the code in a function that's never called? Mostly just thinking from the backcompat testing standpoint I'd rather have the code be compile checked rather than ignored. If actions is always going to need this however maybe we start looking at aleternatives for the backcompat tests

ssanjay1: Yea I just gated off the call for now

nihalbhatnagar (approved):

PR #1396, opened 2025-04-14
Remove ObjectSpecifier from OSDK base

nihalbhatnagar:
Having `ObjectSpecifier` on OSDK Base caused issues since we use that type for Functions and Queries object parameters. Leaving `ObjectSpecifier` on OSDK Base and modifying the accepted type of Functions and Queries caused issues since object literals cannot have more than the specifier properties. Additionally, edited the return type of queries that returned objects since they should include the object specifier

### DIFF
--- a/packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
+++ b/packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
@@ -14,11 +14,10 @@
  * limitations under the License.
  */
 
-import type { ObjectSetSubscription, PropertyKeys } from "@osdk/api";
+import type { ObjectSetSubscription, Osdk, PropertyKeys } from "@osdk/api";
 import { $ontologyRid, Employee } from "@osdk/client.test.ontology";
 import type {
   ObjectSetStreamSubscribeRequests,
-  ObjectUpdate,
   StreamMessage,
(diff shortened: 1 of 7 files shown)


packages/client/src/queries/applyQuery.ts
```diff
@@ -382,7 +383,7 @@ export function createQueryObjectResponse<
 >(
   primaryKey: PrimaryKeyType<Q>,
   objectDef: Q,
-): OsdkBase<Q> {
+): OsdkBase<Q> & { $objectSpecifier: ObjectSpecifier<Q> } {
```
nihalbhatnagar, line 386: Did not want the type in the contract

packages/client/src/objectSet/ObjectSetListenerWebsocket.test.ts
```diff
@@ -125,7 +124,10 @@ describe("ObjectSetListenerWebsocket", async () => {
     >;
     let oslwInst = 0;
 
-    let updateReceived: ObjectUpdate | undefined = undefined;
+    let updateReceived: {
+      object: Osdk.Instance<Employee>;
```
ssanjay1, line 128: This is fine, but why did this change happen?

ssanjay1 (approved):

PR #1397, opened 2025-04-14
Add configurable string types

ethanai22:


packages/maker/src/api/defineOntology.ts
```diff
@@ -602,6 +602,19 @@ function convertType(
         struct: { structFields },
       };
 
+    case (typeof type === "object" && "isLongText" in type):
+      return {
+        "type": "string",
+        string: {
```
aep000, line 608: nit: wrap this key in quotes. I think string is a keyword

packages/maker/src/api/types.ts
```diff
@@ -302,6 +303,13 @@ type PropertyTypeTypeStruct = {
   };
 };
 
+type PropertyTypeTypeString = {
+  type: "string";
+  isLongText: boolean;
```
aep000, line 308: Make these all optional (add a question mark before the colon). And if they aren't set make them the default from the simple string. This way someone can say `{type: "string", isLongText: true}` and not worry about the others.

aep000 (commented):
2 small comments, otherwise looks great 😄

aep000 (approved):

PR #1399, opened 2025-04-14
[Functions] Ontology edits API improvements

marshallavail:
Makes the following improvements:
1. Optional properties can be omitted from calls to `create` for creating objects. At runtime, an omitted property will be treated the same as explicitly providing `undefined` values.
2. All properties can be omitted from calls to `update` for updating objects. At runtime, an omitted property will be treated as a no-op (i.e., keep the property value the same).
3. Fix `createEditBatch` default typing to `never`

### DIFF
--- a/packages/functions/src/edits/createEditBatch.test.ts
+++ b/packages/functions/src/edits/createEditBatch.test.ts
@@ -48,12 +48,14 @@ describe(createEditBatch, () => {
 
     editBatch.create(Task, { id: 0, name: "My Task Name" });
     editBatch.create(Task, { id: 1, name: "My Other Task Name" });
+    editBatch.create(Task, { id: 3 });
     editBatch.delete({ $apiName: "Task", $primaryKey: 0 });
     editBatch.delete(taskInstance);
     editBatch.update({ $apiName: "Task", $primaryKey: 0 }, {
       name: "My New Task Name",
     });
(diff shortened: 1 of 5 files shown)


packages/functions/src/edits/createEditBatch.test.ts
```diff
@@ -81,6 +83,11 @@ describe(createEditBatch, () => {
         obj: Task,
         properties: { id: 1, name: "My Other Task Name" },
       },
+      {
+        type: "createObject",
+        obj: Task,
+        properties: { id: 3 },
+      },
```
marshallavail, line 90: I tested this with actions BE and it works fine

packages/functions/src/edits/createEditBatch.ts
```diff
@@ -87,7 +87,7 @@ class InMemoryEditBatch<X extends AnyEdit = never> implements EditBatch<X> {
   }
 }
 
-export function createEditBatch<T extends AnyEdit>(
+export function createEditBatch<T extends AnyEdit = never>(
```
marshallavail, line 90: Before, it defaulted to `AnyEdit` which gave the user zero type information and allowed them to call `create`, `updated`, etc. with whatever they wanted.

tmisirpash (approved):
LGTM

ssanjay1 (approved):

PR #1401, opened 2025-04-15
Simple create/modify actions on interfaces

davidpaquin:
Didn't add field overrides yet because I figured refactoring the namespace stuff is a higher priority


aep000 (commented):
Looks great lets 🚢 it. One comment to address and I will approve. 

I think we should collect feedback on these methods before we invest much more in them.

aep000:
👍

aep000 (approved):

PR #1404, opened 2025-04-16
Add peer dep on client for functions package

ssanjay1:
For internal users, we will assume they will already have the client dep loaded.


packages/functions/package.json
```diff
@@ -52,9 +52,11 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
-    "@osdk/api": "workspace:~",
-    "@osdk/client": "workspace:~"
+    "@osdk/api": "workspace:~"
```
ssanjay1, line 55: I'm thinking if we also just export whatever we need in api from client and then we just have that one peer dep honestly.

nihalbhatnagar (approved):

PR #1405, opened 2025-04-16
Fix Pk typings for interface actions

braeden:
Let's match what we actually get from the `OSDK.Instance` types, so things are compatible: https://github.com/palantir/osdk-ts/blob/25177505b516a421d6134af27753abb5f222f200/packages/client/src/object/SimpleOsdkProperties.ts#L29

### DIFF
--- a/packages/client/src/actions/actions.test.ts
+++ b/packages/client/src/actions/actions.test.ts
@@ -98,7 +98,7 @@ describe.each([
       },
     );
 
-    // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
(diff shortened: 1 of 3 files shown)


packages/client/src/actions/actions.test.ts
```diff
@@ -98,7 +98,7 @@ describe.each([
       },
     );
 
-    // eslint-disable-next-line @typescript-eslint/no-confusing-void-expression
```
braeden, line 101: Pre-commit hooks seem to have done this?

PR #1409, opened 2025-04-18
Code Workspaces awareness for the Widgets Vite plugin

rosscourt:
Support a `code-workspaces` mode, activated by running dev mode via `vite --mode=code-workspaces`. This uses the relevant environment variables present in the Code Workspaces container to communicate with the correct API and determine the correct base URL for dev mode assets.

The setup page won't receive a `redirectUrl` in this case and instead waits for the parent application to mount the preview components natively. I cleaned up the setup page a bit in this PR too.

### DIFF
--- a/packages/widget.vite-plugin.unstable/package.json
+++ b/packages/widget.vite-plugin.unstable/package.json
@@ -45,8 +45,7 @@
     "fs-extra": "^11.2.0",
     "parse5": "^7.2.1",
     "picocolors": "^1.1.1",
-    "sirv": "^3.0.0",
-    "yargs": "^17.7.2"
+    "sirv": "^3.0.0"
   },
   "peerDependencies": {
     "vite": "^6.2.4"
@@ -61,7 +60,6 @@
     "@types/fs-extra": "^11.0.4",
     "@types/react": "^18.3.12",
     "@types/react-dom": "^18.3.1",
-    "@types/yargs": "^17.0.29",
     "@vitejs/plugin-react": "^4.2.0",
     "react": "^18",
     "react-dom": "^18",
--- a/packages/widget.vite-plugin.unstable/package.json
+++ b/packages/widget.vite-plugin.unstable/package.json
@@ -45,7 +45,8 @@
     "fs-extra": "^11.2.0",
     "parse5": "^7.2.1",
     "picocolors": "^1.1.1",
-    "sirv": "^3.0.0"
+    "sirv": "^3.0.0",
+    "yargs": "^17.7.2"
   },
   "peerDependencies": {
     "vite": "^6.2.4"
@@ -60,6 +61,7 @@
     "@types/fs-extra": "^11.0.4",
     "@types/react": "^18.3.12",
     "@types/react-dom": "^18.3.1",
+    "@types/yargs": "^17.0.29",
     "@vitejs/plugin-react": "^4.2.0",
     "react": "^18",
     "react-dom": "^18",
--- a/packages/widget.vite-plugin.unstable/src/client/app.tsx
+++ b/packages/widget.vite-plugin.unstable/src/client/app.tsx
@@ -27,7 +27,7 @@ type PageState =
   }
   | {
     state: "success";
-    redirectUrl: string | null;
+    isRedirecting: boolean;
   };
 
 const POLLING_INTERVAL = 250;
@@ -52,12 +52,21 @@ export const App: React.FC = () => {
           if (result.status === "pending") {
             return;
           }
+
+          // On success or failure, we clear the poll and end the loading state
           window.clearInterval(poll);
           setPageState(
             result.status === "success"
-              ? { state: "success", redirectUrl: result.redirectUrl }
+              ? { state: "success", isRedirecting: result.redirectUrl != null }
               : { state: "failed", error: result.error },
           );
+
+          // When running in Code Workspaces the parent app will handle the redirect
+          if (result.status === "success" && result.redirectUrl != null) {
+            setTimeout(() => {
+              window.location.href = result.redirectUrl!;
+            }, REDIRECT_DELAY);
+          }
         })
         .catch((error: unknown) => {
           window.clearInterval(poll);
@@ -69,16 +78,6 @@ export const App: React.FC = () => {
     }, POLLING_INTERVAL);
     return () => window.clearInterval(poll);
   }, []);
-
-  // Redirect on reaching success state
-  useEffect(() => {
-    // When running in Code Workspaces the parent app will handle the redirect
-    if (pageState.state === "success" && pageState.redirectUrl != null) {
-      setTimeout(() => {
-        window.location.href = pageState.redirectUrl!;
-      }, REDIRECT_DELAY);
-    }
-  }, [pageState]);
 
   return (
     <div className="body">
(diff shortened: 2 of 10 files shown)


packages/widget.vite-plugin.unstable/package.json
```diff
@@ -45,7 +45,8 @@
     "fs-extra": "^11.2.0",
     "parse5": "^7.2.1",
     "picocolors": "^1.1.1",
-    "sirv": "^3.0.0"
+    "sirv": "^3.0.0",
+    "yargs": "^17.7.2"
```
rosscourt, line 49: Wanted to use a lightweight library that we already use in this repo

packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -61,6 +64,11 @@ export function FoundryWidgetDevPlugin(): Plugin {
      * setup page and trigger module parsing.
      */
     buildStart(options) {
+      safeGetEnvVar(
+        process.env,
+        "FOUNDRY_TOKEN",
+        "This value is required to run dev mode.",
+      );
```
rosscourt, line 71: Re-using this util to fail ASAP when starting dev mode

packages/widget.vite-plugin.unstable/package.json
tzyl: This is used for one thing only to grab the mode from the command right?

From digging around it looks like this should be accessible on `ResolvedConfig.mode` and in the build plugin we store the resolved config from https://vite.dev/guide/api-plugin#configresolved but in the dev plugin it looks like we might only care about it in configureServer from which do we already have it in `server.config.mode`?

packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -251,3 +286,8 @@ function printSetupPageUrl(server: ViteDevServer) {
     }: ${color.green(setupRoute)}`,
   );
 }
+
+async function getIsCodeWorkspacesMode(): Promise<boolean> {
```
tzyl, line 290: From the other comment, if it's possible should we not pass `isCodeWorkspacesMode: boolean` around and instead reuse the existing server parameter with `isCodeWorkspacesMode(server)`?

packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -64,6 +67,20 @@ export function FoundryWidgetDevPlugin(): Plugin {
       htmlEntrypoints = getInputHtmlEntrypoints(options);
     },
 
+    /**
+     * Check that the FOUNDRY_TOKEN environment variable is set when running in dev mode. This is
+     * required to publish the widget overrides to Foundry.
+     */
+    config(_, { command }) {
+      if (command === "serve" && process.env.VITEST == null) {
```
tzyl, line 75: What are the two checks here for? It sounds like "serve" means "dev" for backwards compatibility and is VITEST for this repo only during tests?

packages/widget.vite-plugin.unstable/src/client/app.tsx
```diff
@@ -41,53 +47,66 @@ export const App: React.FC = () => {
...
         <NonIdealState
           title="Started dev mode"
           icon="tick-circle"
           description={
             <div className="description">
               <Spinner intent="primary" size={SpinnerSize.SMALL} />{" "}
-              Redirecting you…
+              {pageState.redirectUrl != null
+                ? <span>Redirecting you…</span>
+                : <span>Loading preview…</span>}
```
tzyl, line 100: How is this state going to be detected? Is it based on whether the dev mode overrides have been set?

packages/widget.vite-plugin.unstable/src/dev-plugin/publishDevModeSettings.ts
```diff
@@ -94,3 +93,13 @@ export async function publishDevModeSettings(
     );
   }
 }
+
+function getCodeWorkspacesFoundryUrl(): string {
+  return `https://${
+    safeGetEnvVar(
+      process.env,
+      "FOUNDRY_PROXY_URL",
```
tzyl, line 101: Are all the environment variables we're treating as API and expecting to be present:
- FOUNDRY_PROXY_URL
- DEV_SERVER_DOMAIN
- DEV_SERVER_BASE_PATH

rosscourt: Plus `FOUNDRY_TOKEN`

packages/widget.vite-plugin.unstable/src/client/app.tsx
rosscourt: Yea the parent app checks for dev mode settings on the widget, because if present we would skip this step. That logic already polls because it also waits for the dev server to be running

tzyl: Got it, my question was going to be if the dev mode overrides are set but the dev server is not running (for example refreshing the page or restarting the workspace) what would happen but it sounds like we would check for both before trying to swap over to preview

packages/widget.vite-plugin.unstable/src/client/app.tsx
```diff
@@ -41,53 +47,66 @@ export const App: React.FC = () => {
...
 
+  // Redirect on reaching success state
+  useEffect(() => {
+    // When running in Code Workspaces the parent app will handle the redirect
+    if (pageState.state === "success" && pageState.redirectUrl != null) {
+      setTimeout(() => {
+        window.location.href = pageState.redirectUrl!;
+      }, REDIRECT_DELAY);
+    }
+  }, [pageState]);
```
tzyl, line 81: Should we still kick off the redirect once we set the page state rather than doing it separately in a useEffect? The rendering of the text will still work based on the set page state but it might be cleaner to do the redirect in the existing place rather than doing it afterwards in a side effect

packages/widget.vite-plugin.unstable/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -64,6 +68,25 @@ export function FoundryWidgetDevPlugin(): Plugin {
       htmlEntrypoints = getInputHtmlEntrypoints(options);
     },
 
+    /**
+     * Capture configuration options from the Vite config. This is used to determine if we are running in
+     * Code Workspaces mode and in dev mode.
+     */
+    config(resolvedConfig, { command }) {
+      // Check for the optional `--mode` argument when running `vite`
+      isCodeWorkspacesMode = resolvedConfig.mode === CODE_WORKSPACES;
```
tzyl, line 77: It looks like all the places we pass `isCodeWorkspacesMode` to we also pass `server: ViteDevServer`. Do you prefer to do it this way to capture this once during config resolved, or reuse the existing `server` parameter and compute it on the fly so it doesn't need to be passed around?
```
isCodeWorkspacesMode(server: ViteServer) => server.config.mode === CODE_WORKSPACES
```

tzyl (dismissed):

PR #1413, opened 2025-04-21
Add timeseries datasource in oac

ethanai22:


packages/client.unstable/src/generated/ontology-metadata/api/index.ts
```diff
@@ -639,10 +639,13 @@ export type {
   OntologyIrActionNotificationRecipients,
   OntologyIrActionsObjectSet,
   OntologyIrActionType,
+  OntologyIrActionTypeDisplayMetadata,
```
aep000, line 642: For future Reference: API Bump PRs should be done on their own PR. This way we don't have huge diffs of code-gen mixed with implementation.

packages/maker/src/api/overall.test.ts
```diff
@@ -1665,6 +1666,156 @@ describe("Ontology Defining", () => {
     `);
   });
 
+  it("TimeSeries fields properly set", () => {
+    const seriesValueMetadata: SeriesValueMetadata_numeric = {
+      type: "numeric",
+      numeric: {
+        defaultInternalInterpolation: {
+          type: "propertyType",
+          propertyType: "bar",
```
aep000, line 1675: Can we add a validation that this property exists?

aep000, line 1675: Additionally we need to consider name spacing here. So when this is configured on shared property types (Not regular object properties) these api names need to be namespaced under the hood. For object properties we don't need to do this. Shared property types are ontology scoped so they need api namespaces.

packages/maker/src/api/overall.test.ts
```diff
@@ -1665,6 +1666,156 @@ describe("Ontology Defining", () => {
     `);
   });
 
+  it("TimeSeries fields properly set", () => {
```
aep000, line 1669: Can we also get an object test?

aep000 (commented):
Implementation looks good, one note for the future and a few minor additional validations/functionality notes.

packages/maker/src/api/overall.test.ts
ethanai22: done

ethanai22: Added to defineSharedPropertyType

packages/maker/src/api/overall.test.ts
ethanai22: done

aep000 (approved):
I am approving but I don't care for some of the syntax. Mostly a vibes thing. i think we may want to re-think the property type type anyway so leaving for now.

ssanjay1:
@ethanai22 do you still need this PR, or can we close?

PR #1419, opened 2025-04-23
Backport changes to functions deps

nihalbhatnagar:
Backports #1404 and #1411


packages/client/src/index.ts
```diff
@@ -23,12 +23,15 @@ export type {
   ActionValidationResponse,
   ApplyActionOptions,
   ApplyBatchActionOptions,
+  Attachment,
+  CompileTimeMetadata,
```
nihalbhatnagar, line 27: Just highlighting that these reexports are being backported now. It seems fine to me

ericanderson (approved):

PR #1424, opened 2025-04-24
Query Version Pinning

nihalbhatnagar:
This PR adds the ability to pin versions of queries at generation time, and the ability to execute and load metadata of queries when they are generated with pinned versions.

Here is an outline of the changes for each package:

API: Adds the `version` and `pinned` fields to our Query Type definitions. These will provide the metadata necessary to know when and what versions to execute at. Note that we already generated our queries with the `version` field. I chose not to simple know if a function was pinned solely through the presence of `version` for 2 reasons. 1. To maintain backcompat when all functions were generated with version(the bigger reason). 2. Having it there even if it's not pinned is helpful for the user for knowing what version this shape is for. Theoretically, we can update the contents of version to contain an identifier but I think this makes it hard for a user to know immediately if their version is pinned

Client: Updates executeQuery and fetchMetadata to use pinned versions. Our internal metadata cache now also uses `apiName:version` as a key for the metadata. 

foundry-sdk-generator: Utilizes the new loadMetadata endpoint where all ontology properties can be specified, not requiring us to filter the metadata ourselves. This also starts producing a `pinnedQueries` array that is determined by whether there was a colon passed in the argument. If so, we will mark it as specified.

generator: Uses the new pinnedQueries param to determine whether to generate query metadata with the pinned field.


packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -215,9 +167,21 @@ export class OntologyMetadataResolver {
 
       const dataScope = sdk.inputs.dataScope.ontologyV2;
 
+      const objectTypes: ObjectTypeV2[] = [];
```
nihalbhatnagar, line 170: This is very silly, not intended to merge. Providing more context in slack

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -240,60 +217,44 @@ export class OntologyMetadataResolver {
...
+        queryTypes: [...queryTypes],
+        interfaceTypes: [...interfaceTypes],
+        linkTypes: Array.from(linkTypes.entries()).flatMap(
+          ([_, linkTypeApiNames]) => [...linkTypeApiNames],
+        ),
       },
-      extPackageInfo,
     );
 
     const validData: Result<{}, string[]> = this.validateLoadedOntologyMetadata(
```
nihalbhatnagar, line 238: Probably will remove, still iterating

packages/shared.test/src/FauxFoundry/FauxOntology.ts
```diff
@@ -305,14 +356,28 @@ export class FauxOntology {
...
     implementation?: FauxQueryImpl,
   ): void {
-    if (def.apiName in this.#ontology.queryTypes) {
+    if (`${def.apiName}:${def.version}` in this.#ontology.queryTypes) {
       throw new Error(
         `QueryType ${def.apiName} already registered`,
       );
     }
-    this.#ontology.queryTypes[def.apiName] = def;
+    this.#ontology
```
nihalbhatnagar, line 364: I thought that registering our queries by `apiName:version` for our query types worked best since we got to maintain the shape of our ontology internally (which was OntologyFullMetadata where query types are ` Record<VersionedQueryTypeApiName, QueryTypeV2>` this introduces complexity because on access and insertion we need to always access by the concatenated string, but I think this is an acceptable tradeoff

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.test.ts
```diff
@@ -2089,6 +2099,382 @@ describe("generator", () => {
     });
   });
 
+  describe("queries with versions", () => {
```
nihalbhatnagar, line 2102: This is here and not in `generateQueryPerDataFiles` to make sure that `pinnedQueries` is passed through correctly. It is a relatively simple code path though so if we really care I can move it

packages/shared.test/src/handlers/createQueryHandlers.ts
```diff
@@ -27,9 +28,20 @@ export const createQueryHandlers: FauxFoundryHandlersFactory = (
   OntologiesV2.Queries.execute(
     baseUrl,
     async ({ request, params: { ontologyApiName, queryApiName } }) => {
+      const queryParams = Object.fromEntries(
```
nihalbhatnagar, line 31: Could probably find a way to encode this in our handler definition types but will track as a FLUP/polish

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -192,17 +121,40 @@ export class OntologyMetadataResolver {
...
+      linkTypes.get(objectTypeApiName)?.add(linkTypeApiName);
+    }
+
+    const queryTypes = new Set<string>();
+    const pinnedQueryTypes = [];
+
+    for (const queryType of entities.queryTypesApiNamesToLoad ?? []) {
+      if (queryTypes.has(queryType)) {
+        return Result.err([
+          `Query type ${queryType} was specified multiple times.`,
```
nihalbhatnagar, line 148: This could be considered a behavior break, but I think it's acceptable. Possible to expand to other entities as well but restricting to queries for now

packages/api/src/ontology/QueryDefinition.ts
```diff
@@ -35,6 +35,8 @@ export interface QueryCompileTimeMetadata<T> {
 export interface QueryDefinition<T = any> {
   type: "query";
   apiName: string;
+  version: string;
+  pinned: boolean;
```
ssanjay1, line 39: should this be optional?

packages/client/src/queries/queries.test.ts
```diff
@@ -338,4 +338,31 @@ describe("queries", () => {
...
     expect(result).toEqual(["Pitt", "Clooney", "Reynolds"]);
   });
+
+  it("pinned queries call execute with the right version", async () => {
+    const result = await client($Queries.addOne).executeFunction({
+      n: 2,
+    });
+    expect(result).toEqual(3);
+
+    const result2 = await client({ ...$Queries.addOne, version: "0.0.1" })
```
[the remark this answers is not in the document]
ssanjay1: Talked offline, if we don't want to support this first class for now, that's fine

packages/api/src/ontology/QueryDefinition.ts
nihalbhatnagar: pinned should definitely be optional, I think we have always generated version so maybe not but can make it optional just in case

packages/client/src/queries/queries.test.ts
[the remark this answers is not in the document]
nihalbhatnagar: Yeah spitballing we can have `client(myQuery.atVersion())` or `client(myQuery).executeQuery({}, {version:})` or smth but not in scope for right now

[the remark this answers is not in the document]
nihalbhatnagar: Maybe we don't even do any of this and require you to specify at runtime? Idk I don't like that

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -192,6 +193,7 @@ export class OntologyMetadataResolver {
       ]);
     }
 
+    if (extPackageInfo.size > 0) {
     const ontologyFullMetadata = await OntologiesV2.getFullMetadata(
```
nihalbhatnagar, line 197: Intentionally committed with this formatting issue for diff

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -274,27 +287,105 @@ export class OntologyMetadataResolver {
...
+          queryTypes: [...queryTypes],
+          interfaceTypes: [...interfaceTypes],
+          linkTypes: Array.from(linkTypes.entries()).flatMap(
+            ([_, linkTypeApiNames]) => [...linkTypeApiNames],
+          ),
+        },
+      );
+
+      const validData: Result<{}, string[]> = this
+        .validateLoadedOntologyMetadata(
```
nihalbhatnagar, line 367: Endpoint won't return if we throw in unknown names, so we need to validate we received the expected valued

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
nihalbhatnagar: Will fix before merge

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -274,27 +287,105 @@ export class OntologyMetadataResolver {
...
+        ontology.rid as OntologyIdentifier,
+        {
+          actionTypes: [...actionTypes],
+          objectTypes: [...objectTypes],
+          queryTypes: [...queryTypes],
+          interfaceTypes: [...interfaceTypes],
+          linkTypes: Array.from(linkTypes.entries()).flatMap(
+            ([_, linkTypeApiNames]) => [...linkTypeApiNames],
+          ),
+        },
```
nihalbhatnagar, line 363: Need to wait for new platform SDK to pass in beta flag 😢

packages/api/src/ontology/QueryDefinition.ts
```diff
@@ -35,6 +35,8 @@ export interface QueryCompileTimeMetadata<T> {
 export interface QueryDefinition<T = any> {
   type: "query";
   apiName: string;
+  version?: string;
```
nihalbhatnagar, line 38: Made these optional as discussed. Note that `version` is also in our compile time query metadata (in DefinitionMetadata type) but need to add here for runtime

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -107,7 +108,7 @@ export class OntologyMetadataResolver {
       Object.entries(ontologyFullMetadata.actionTypes).filter(
         ([actionApiName]) => {
           if (
-            expectedEntities.actionTypes.has(this.camelize(actionApiName))
```
nihalbhatnagar, line 110: We were camelizing out action names here for no reason. We were basically camelizing, and then always camelizing whatever we needed to compare to. The API also returns as a key for the action whatever we pass to the generator. However, on actual SDK generation we never use the key, thather actions[actionType].apiName or an equivalent. I also smoketested and make sure our action API names are still camelized (so no change in behavior)

packages/client/src/fetchMetadata.ts
```diff
@@ -61,6 +61,7 @@ export const fetchMetadataInternal = async <
   } else if (definition.type === "query") {
     return client.ontologyProvider.getQueryDefinition(
       definition.apiName,
+      definition.isFixedVersion ? definition.version : undefined,
```
ssanjay1, line 64: So this tracks with our old behavior, but im questioning whether our old behavior was correct. Theoretically, even if I always wanted latest, if I hadn't regenerated my OSDK, my generated query code could be on version 0.2 but i could be fetching metadata for 0.3 which could have entirely different params. I feel like maybe we should always just use the version in the generated metadata

packages/client/src/queries/applyQuery.ts
```diff
@@ -48,7 +48,10 @@ export async function applyQuery<
 ): Promise<
   QueryReturnType<CompileTimeMetadata<QD>["output"]>
 > {
-  const qd = await client.ontologyProvider.getQueryDefinition(query.apiName);
+  const qd = await client.ontologyProvider.getQueryDefinition(
+    query.apiName,
+    query.isFixedVersion ? query.version : undefined,
```
ssanjay1, line 53: In contrast to my comment above, yea I feel like this makes sense, because if we changed it we'd break people who want the latest version of their function logic and they're never like breaking the contract by adding/removing params

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
ssanjay1: This is all good now right

packages/generator/src/GenerateContext/GenerateContext.ts
```diff
@@ -26,7 +26,7 @@ export interface GenerateContext {
   fs: MinimalFs;
 
   outDir: string;
-
+  pinnedQueryTypes: string[];
```
ssanjay1, line 29: maybe rename fixedVersionQueryTypes?

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -36,9 +36,10 @@ type PackageInfo = Map<string, {
 }>;
 
 export interface OntologyInfo {
-  filteredFullMetadata: OntologyFullMetadata;
+  requestedMetadata: OntologyFullMetadata;
   externalInterfaces: Map<string, string>;
   externalObjects: Map<string, string>;
+  pinnedQueryTypes: string[];
```
ssanjay1, line 42: nit: Save it for FLUP to not block this if you want, but maybe we should rename this fixedVersionQueryTypes to be consistent with everything else

ssanjay1 (dismissed):

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
nihalbhatnagar: Fixed

ssanjay1 (approved):

PR #1434, opened 2025-05-02
Delete logger from client and reexport from api

nihalbhatnagar:
In #1315, we duplicated the Logger file to export from API. This PR deletes the version from client, reexports the one from API in client, and fixes import statements


packages/api/src/index.ts
```diff
@@ -134,4 +134,4 @@ export type {
 } from "./timeseries/timeseries.js";
 export type { LinkedType, LinkNames } from "./util/LinkUtils.js";
 export {};
-export type { Logger } from "./Logger.js";
+export type { LogFn, Logger } from "./Logger.js";
```
ericanderson, line 137: Do we need to export LogFn? If so should we hide it in a `Logger` namespace?

nihalbhatnagar: I think we have enough usage in client where the export becomes valuable, converted to namespace export

ericanderson (approved):

PR #1436, opened 2025-05-05
Fix typing of params in EditBatch interface

marshallavail:


packages/functions/src/edits/EditBatch.ts
```diff
@@ -15,40 +15,44 @@
...
-
-  update: <O extends UpdateObjectEdits<X>>(
-    obj: O["obj"],
-    properties: O["properties"],
-  ) => void;
+  link: (...args: ExtractAddLinkArgs<X>) => void;
+
+  unlink: (...args: ExtractRemoveLinkArgs<X>) => void;
+
+  create: (...args: ExtractCreateObjectArgs<X>) => void;
```
marshallavail, line 51: This rest parameter syntax allows users to provide the arguments as `create(Person, { id: 2 })`
If I did `([obj, properties]: ExtractCreateObjectArgs<X>)`, then users would have to wrap their arguments in a tuple explicitly `create([Person, { id: 2}])`

packages/functions/src/edits/EditBatch.ts
```diff
@@ -15,40 +15,44 @@
...
+  AddLink<infer S, infer L> ? [ObjectLocator<S>, L, AddLink<S, L>["target"]]
+  : never;
+
+export type ExtractRemoveLinkArgs<X extends AnyEdit> = X extends
+  RemoveLink<infer S, infer L>
+  ? [ObjectLocator<S>, L, RemoveLink<S, L>["target"]]
+  : never;
+
+export type ExtractCreateObjectArgs<X extends AnyEdit> = X extends
+  CreateObject<infer O> ? [O, CreateObject<O>["properties"]] : never;
```
marshallavail, line 37: Typing the args this way makes it so that we end up with a union of `(obj, properties)`, where each member corresponds to exactly one object type (i.e., prevents users from calling `create(Animal, { officeId: 2 })` when `X = CreateObject<Animal> | CreateObject<Office>`)

PR #1443, opened 2025-05-06
Mv/fix edit batch params may sixth

marshallavail:


packages/functions/src/edits/EditBatch.ts
```diff
@@ -14,40 +14,109 @@
  * limitations under the License.
```
marshallavail, line 14: The git diff in this file is wonky... you'll probably want to clone locally to see the changes

.changeset/tired-bars-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/functions.unstable": patch
```
tmisirpash, line 2: Scope creep, but I think we agreed to rename this to `@osdk/functions`. Would it work to do that here?

marshallavail: Mind if I do that in a FLUP?

packages/functions/src/edits/createEditBatch.test.ts
```diff
@@ -15,16 +15,19 @@
...
+import { Office, Task } from "@osdk/client.test.ontology";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { createEditBatch } from "./createEditBatch.js";
 import type { EditBatch } from "./EditBatch.js";
 import type { Edits } from "./types.js";
 
 type TestEditScope =
   | Edits.Object<Task>
-  | Edits.Link<Task, "RP">;
+  | Edits.Object<Office>
```
ssanjay1, line 27: Can we add a quick test in here where the types do flag something wrong (trying to mix properties like in the PR description) and ts-expect error it?

marshallavail: Done!

ssanjay1 (approved):

PR #1451, opened 2025-05-07
Support for actions (create, modify, delete) on objects in OAC

ethanai22:
Ontology as code support for actions that create, modify, or delete objects

### DIFF
--- a/packages/maker/src/api/defineAction.ts
+++ b/packages/maker/src/api/defineAction.ts
@@ -23,11 +23,13 @@ import type {
   ActionParameterTypePrimitive,
   ActionType,
   InterfaceType,
+  ObjectPropertyType,
   ObjectType,
+  PropertyTypeType,
   SharedPropertyType,
 } from "./types.js";
 
-export function defineCreateAction(
+export function defineCreateInterfaceObjectAction(
   interfaceType: InterfaceType,
   objectType?: ObjectType,
 ): ActionType {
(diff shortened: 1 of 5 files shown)


packages/maker/src/api/defineAction.ts
```diff
@@ -339,6 +504,40 @@ function extractAllowedValuesFromSpt(
   }
 }
 
+function extractActionParameterTypeFromOpt(
+  opt: ObjectPropertyType,
+): ActionParameterType {
+  const typeType = opt.type;
```
aep000, line 510: Object property types and SPT Property types use the same PropertyTypeType field as well as the same array field, I think we can just union this method with that method.

packages/maker/src/api/overall.test.ts
```diff
@@ -4025,15 +4028,765 @@ describe("Ontology Defining", () => {
         `);
     });
 
+    it("Concrete actions are properly defined", () => {
+      const exampleAction = defineAction({
```
aep000, line 4032: Side note (Not to be actioned in this PR) I think we should remove defineAction all together. I eventually want us to support custom parameter -> property mappings, but I don't think it makes any sense to give users full control of actions.

aep000 (commented):
Looks good. One small re-usability nit.

packages/maker/src/api/defineAction.ts
ethanai22: Done

PR #1454, opened 2025-05-08
Fix query return types

### LINKED ISSUE #1442: [OSDK V1] fetchPage for queries that return lists of objects
For queries that return a list of objects, the TS v1 osdk calls `getObject` for each object in the list before returning the full list of objects.
[Link to code](https://github.com/palantir/osdk-ts/blob/c9a9050911aff643fd0c2302bc6bc5a13ef32c91/packages/legacy-client/src/client/net/executeQuery.ts#L157). They should instead take all the primary key's of the list of objects and call the [load objectSet endpoint](https://www.palantir.com/docs/foundry/api/v2/ontologies-v2-resources/ontology-object-sets/load-object-set/) or (some fetchPage method) with an `in` filter to make this more optimized.

nihalbhatnagar:
For arrays of objects returned from queries, we used to make several fetch one requests. Now, we dispatch a fetch page instead. Fixes #1442

### DIFF
--- a/packages/legacy-client/src/client/net/executeQuery.ts
+++ b/packages/legacy-client/src/client/net/executeQuery.ts
@@ -48,8 +48,9 @@ import type {
   QueryParameters,
   WrappedQueryReturnType,
 } from "../queries.js";
-import { isOk } from "../Result.js";
+import { isErr, isOk } from "../Result.js";
 import { getObject } from "./getObject.js";
+import { loadAllObjects } from "./loadObjects.js";
 import { getParameterValueMapping } from "./util/getParameterValueMapping.js";
 import { wrapResult } from "./util/wrapResult.js";
 import type { WireOntologyObjectV2 } from "./WireOntologyObjectV2.js";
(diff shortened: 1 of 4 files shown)


packages/legacy-client/src/client/net/executeQuery.ts
```diff
@@ -123,6 +124,27 @@ async function remapQueryResponseType<
...
   // handle arrays
   if (definition.multiplicity) {
+    if (definition.type === "object") {
+      const objects = await loadAllObjects(client, definition.object, {
+        type: "filter",
+        where: {
+          type: "in",
+          field: client.ontology.objects[definition.object].primaryKeyApiName,
+          value: responseValue,
+        } as any,
```
nihalbhatnagar, line 134: Having a typed in filter would require me to regenerate gateway package, which I would rather not do to save compile breaks. Ive tested this e2e so can confirm shape is right

packages/legacy-client/src/client/net/executeQuery.ts
```diff
@@ -150,152 +172,150 @@ async function remapQueryResponseType<
...
-              [],
-            );
 
-            if (isOk(result)) {
-              return result.value;
-            } else {
-              throw result.error;
-            }
-          }
+    case "object": {
```
nihalbhatnagar, line 176: Diff from here on down is messy but I made no changes other than moving this out of the switch

ssanjay1 (approved):

PR #1457, opened 2025-05-08
add example-advance-to-do-application

ewitkon:
This PR adds `example-advance-to-do-application` to the example folder. 
The `example-advance-to-do-application` is not generated and is based on the [sdk and ontology on hatch](https://hatch.palantirfoundry.com/workspace/developer-console/app/ri.third-party-applications.main.application.94401aab-d091-42e0-8d20-20f35ccd250d/overview)


examples/example-advance-to-do-application/.jemma/settings.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+export NODE_INSTALLATION_VERSION=18.18.0
+export REPOSITORY_RID=ri.stemma.main.repository.2ee1df2f-ef0c-434b-8d0a-f6948a4d648b
```
ericanderson, line 3: I am assuming this is specific to you? Should we take this stuff out or extract into a `.env` or `.envrc` file?

examples/example-advance-to-do-application/.jemma/download-node.py
```diff
@@ -0,0 +1,15 @@
...
+if __name__ == "__main__":
+    url = sys.argv[1]
+    filename = sys.argv[2]
+    token = sys.argv[3]
+    save_path = os.path.join(os.getcwd(), filename)
+
+    request = urllib.request.Request(url, headers={"Authorization": f"Bearer {token}"})
+    with urllib.request.urlopen(request) as response:
+        with open(save_path, "wb") as out_file:
+            out_file.write(response.read())
```
ericanderson, line 15: Is there no out of the box Jemma usage?

ewitkon: this all directory is coming from Authoring integration and is part of how they removed the need for gradle. 
If we remove this folder all together it will mean the code can't be synced to a stemma repo and work.

examples/example-advance-to-do-application/.jemma/settings.sh
ewitkon: this is a good question, as the next one is - I guess for now we can remove this and have this from here. 
I will add this to the marketplace product that build the OSDK template

ericanderson (approved):
Looks good, we may need to tweak over time.

PR #1458, opened 2025-05-08
Add refresh on scopes change to the Legacy Typescript OSDK

beloiual:
Previously, we added support for regenerating our Public OAuth token if the scopes on multipass change on the 2.x version in https://github.com/palantir/osdk-ts/pull/1319 

We need to port this change over to our legacy client to make sure users don't run into the same issue

### DIFF
--- a/packages/oauth/src/createPublicOauthClient.ts
+++ b/packages/oauth/src/createPublicOauthClient.ts
@@ -69,6 +69,8 @@ export function createPublicOauthClient(
   const client: Client = { client_id, token_endpoint_auth_method: "none" };
   const authServer = createAuthorizationServer(ctxPath, url);
   const oauthHttpOptions: HttpRequestOptions = { [customFetch]: fetchFn };
+  let joinedScopes: string = [...scopes ?? ["api:read-data", "api:write-data"]]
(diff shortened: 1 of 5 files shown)


packages/oauth/src/createPublicOauthClient.ts
```diff
@@ -89,12 +92,21 @@ export function createPublicOauthClient(
...
+      readLocal(client);
     if (!refresh_token) {
       if (expectRefreshToken) throw new Error("No refresh token found");
       return;
     }
 
+    const areScopesEqual = initialRequestedScopes != null
+      && joinedScopes === initialRequestedScopes;
+
+    if (!refresh_token || !areScopesEqual) {
```
nihalbhatnagar, line 105: Can you merge with/place inside the if statement above?

nihalbhatnagar (commented):
Looks mostly good to me, just confirming that you've tested login e2e with a sample react app? Also wondering if we need a test similar to `should not allow refresh if scopes in local storage is undefined` in nadiras PR but if you smoketests thats fine with me since this is legacy

nihalbhatnagar (approved):

PR #1459, opened 2025-05-08
Explicitly passing undefined for actions passes through nulls

### LINKED ISSUE #1416: Type checks prevent setting null values for optional OSDK fields
If a field is marked as optional in a Foundry action, the codegen for the field gets typed as either the primitive or undefined. However, this prevents the action from "resetting" the value as null. Ideally at runtime, the BE can explicitly check if the value is undefined so that the value no longer gets populated.

nihalbhatnagar:
Fixes #1416

### DIFF
--- a/packages/client/src/actions/actions.test.ts
+++ b/packages/client/src/actions/actions.test.ts
@@ -34,6 +34,7 @@ import {
   moveOffice,
 } from "@osdk/client.test.ontology";
 import type {
+  ApplyActionRequestV2,
   BatchApplyActionResponseV2,
   SyncApplyActionResponseV2,
 } from "@osdk/foundry.ontologies";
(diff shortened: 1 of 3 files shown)


packages/client/src/actions/actions.test.ts
```diff
@@ -558,6 +559,52 @@ describe.each([
...
+              submissionCriteria: [],
+              parameters: {},
+            },
+          };
+        }));
+
+        await client($Actions.createStructPerson).applyAction({
+          name: "testMan",
+        });
+        // @ts-ignore
```
ssanjay1, line 603: Why are we ts-ignoring here?

ssanjay1 (approved):

PR #1460, opened 2025-05-09
Discover widget set input specs

tzyl:
Discover OSDK packages from widget set project's package.json dependencies and include these in the built widget set manifest. The package.json format is to be determined.


packages/widget.vite-plugin.unstable/src/build-plugin/FoundryWidgetBuildPlugin.ts
```diff
@@ -86,10 +87,14 @@ export function FoundryWidgetBuildPlugin(): Plugin {
       const widgetBuilds = htmlEntrypoints.map((input) =>
         getWidgetBuildOutputs(bundle, input, config.build.outDir, configFiles)
       );
+      const widgetSetInputSpec = await getWidgetSetInputSpec(
+        path.resolve(process.cwd(), "package.json"),
```
tzyl, line 91: I baked in the assumption that the vite plugin is run from the same directory as the root package.json not sure if there's a vite plugin place which we can grab this from. Otherwise, when grabbing the package.json version to put in the manifest if using that strategy we use find-up https://github.com/palantir/osdk-ts/blob/a73744207ad5092a1b2aebc3f5c2b8656923f825/packages/foundry-config-json/src/autoVersion.ts#L66

PR #1463, opened 2025-05-12
Lint React 2.0 template

amishshah:


ericanderson (changes requested):
I'm not sure this is the best pattern to propagate. Lets discuss offline.

PR #1467, opened 2025-05-12
Write site url to the jemma metadata file

nfdiop:
We write the `siteUrl` to the job custom metadata file so we can later re-use it.

### DIFF
--- a/packages/cli/src/commands/site/deploy/siteDeployCommand.mts
+++ b/packages/cli/src/commands/site/deploy/siteDeployCommand.mts
@@ -124,7 +124,7 @@ export default async function siteDeployCommand(
     }
   }
 
-  // Write the site link to the custom metadata file if it exists
+  // Write the site link and version to the custom metadata file when the site link is present
   if (siteLink != null) {
     maybeUpdateJemmaCustomMetadata({ siteLink, siteVersion });
   }


packages/cli/src/util/maybeUpdateJemmaCustomMetadata.ts
```diff
@@ -0,0 +1,86 @@
...
+const JEMMA_JOB_CUSTOM_METADATA_PATH = "JEMMA_JOB_CUSTOM_METADATA_PATH";
+
+/**
+ * Writes the site link to the custom metadata file (if the path is provided by the JEMMA_JOB_CUSTOM_METADATA_PATH environment variable).
+ *
+ * The custom metadata file will be updated with the following format:
+ *
+ * {
+ *   "siteLink": "https://example.com"
+ * }
```
tzyl, line 31: Should we include the site version as well? Although you will generally infer it from the tag, including it here would be the source of truth for what actually got deployed

packages/cli/src/util/maybeUpdateJemmaCustomMetadata.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import * as fs from "node:fs";
+
+// Environment variable names
+const JEMMA_JOB_CUSTOM_METADATA_PATH = "JEMMA_JOB_CUSTOM_METADATA_PATH";
+
+/**
+ * Writes the site link to the custom metadata file (if the path is provided by the JEMMA_JOB_CUSTOM_METADATA_PATH environment variable).
+ *
+ * The custom metadata file will be updated with the following format:
```
tzyl, line 27: Should we only allow newly creating the metadata file as that seems like the expected case when running in CI? Updating the existing content feels like it should never happen (e.g. does functions also do this?) and I would vote for one of:
- Only allowing newly creating and throwing if the file already exists
- Always overwriting the existing file contents if it exists (and log if we detect we're overwriting if desired)

Those would have stronger guarantees to start with and shouldn't cause problems given that this is more of a hidden feature when detecting the special jemma environment

packages/cli/src/commands/site/deploy/siteDeployCommand.mts
```diff
@@ -96,6 +97,7 @@ export default async function siteDeployCommand(
     archive,
   );
 
+  let siteLink: string | undefined;
```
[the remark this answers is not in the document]
tzyl: Hmm, which do you prefer? My opinion is that, although it removes specific references to jemma in the CLI, it's less obvious what's going on

packages/cli/src/util/maybeUpdateJemmaCustomMetadata.ts
samialfhaily: I think it would be good to include it here as well as a source of truth.

packages/cli/src/util/maybeUpdateJemmaCustomMetadata.ts
nfdiop: So I am checking with the relevant teams now. What I noticed while testing was that the file wasn't present but it was OK to create it.

packages/cli/src/commands/site/deploy/siteDeployCommand.mts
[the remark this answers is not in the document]
nfdiop: It is less obvious indeed: totally agree! Happy to keep the current implementation

nfdiop:
Closing in favour of https://github.com/palantir/osdk-ts/pull/1492

PR #1468, opened 2025-05-12
Return metadata for all loaded objects

nihalbhatnagar:
Adds a `$metadata` field to `Osdk.Instance` objects. Uses the already fetched metadata so we don't execute another load and uses the stored reference so objects have no added size


packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -81,21 +82,17 @@ describe("convertWireToOsdkObjects", () => {
     }).fetchPage();
 
     // Should not have $title
-    expect(JSON.stringify(employee)).toMatchInlineSnapshot(
-      `"{"employeeId":50030,"fullName":"John Doe","office":"NYC","class":"Red","startDate":"2019-01-01","employeeStatus":{},"employeeSensor":{},"employeeLocation":{},"$apiName":"Employee","$objectType":"Employee","$primaryKey":50030,"$objectSpecifier":"Employee:50030"}"`,
-    );
+    expect(JSON.stringify(employee)).toMatchSnapshot();
```
nihalbhatnagar, line 85: I moved a lot of these to file snapshots as to avoid bloat from metadata. Can also exclude metadata from these tests if we prefer inline style

packages/client/src/object/convertWireToOsdkObjects/InterfaceHolder.ts
```diff
@@ -23,4 +23,6 @@ export interface InterfaceHolder<
   _Q extends Osdk.Instance<any> = never,
 > extends BaseHolder {
   [InterfaceDefRef]: InterfaceMetadata;
+
+  $metadata: InterfaceMetadata;
```
ericanderson, line 27: I think this will lead to confusion about state since the Object and Interface instance of this are both called `$metadata`. 

```ts
const employee = await loadEa();
employee.$metadata;
employee.$as(FullTimer).$metadata;
```

The pattern we used for `$apiName` was to set `$objectType` to always be the underlying and for `$apiName` to be contextual. 

Maybe we should do the same thing here? 

```
employee.$metadata.apiName == "Employee";
employee.$objectMetadata.apiName == "FullTimer";
employee.$as(FullTimer).$metadata.apiName == "FullTimer";
```

packages/client/src/object/convertWireToOsdkObjects.test.ts
ericanderson: I think the test result before might have been right. 

Should doing a `JSON.stringify()` include $metadata? I'm not sure that it should.

PR #1474, opened 2025-05-12
Fix ObjectSpecifier Break

nihalbhatnagar:
We saw a break in an internal library where using `typeof` on an OsdkInstance object resolved to 
`OsdkBaseWithObjectSpecifier & {} & {} & ...`. However, we did not export `OsdkBaseWithObjectSpecifier`, leading to an issue similar to [this](https://github.com/microsoft/TypeScript/issues/42873) one, which indicated the type was not resolving correctly.. To solve the above issue, we can either start exporting OsdkBaseWithObjectSpecifier or an equivalent orconvert OsdkBaseWithObjectSpecifier to a type instead of an interface. We need to figure out a test or standard on what the support should be


packages/api/src/OsdkBase.ts
```diff
@@ -21,21 +21,21 @@ import type { ObjectTypeDefinition } from "./ontology/ObjectTypeDefinition.js";
...
 
+  readonly $objectType: string;
+
   readonly $title: string | undefined;
-};
+}
 
-export interface OsdkBaseWithObjectSpecifier<
+export interface OsdkBase<
   Q extends ObjectOrInterfaceDefinition,
```
nihalbhatnagar, line 37: Interface? Type? Not sure if we should always defer to type now to circumvent these issues

packages/api/src/OsdkBase.ts
```diff
@@ -21,21 +21,21 @@ import type { ObjectTypeDefinition } from "./ontology/ObjectTypeDefinition.js";
...
-> = {
+> {
   readonly $apiName: Q["apiName"];
 
-  readonly $objectType: string;
-
   readonly $primaryKey: PrimaryKeyType<Q>;
 
+  readonly $objectType: string;
```
nihalbhatnagar, line 31: Alternatively we decouple from OsdkBase entirely and make objectType and title optional

nihalbhatnagar: We cant remove them entirely in decoupling due to [backcompat](https://dev.to/this-is-learning/understanding-excess-property-checking-in-typescript-ook)

packages/api/src/OsdkBase.ts
```diff
@@ -21,21 +21,21 @@ import type { ObjectTypeDefinition } from "./ontology/ObjectTypeDefinition.js";
 import type { PrimaryKeyTypes } from "./ontology/PrimaryKeyTypes.js";
 import type { OsdkObjectPrimaryKeyType } from "./OsdkObjectPrimaryKeyType.js";
 
-export type OsdkBase<
+export interface MinimalOsdkInstance<
```
nihalbhatnagar, line 24: So does this mean we need to start exporting this as well when someone does `Parameters<typeof fn>[0];`? Do we need to make this a type?

packages/api/src/OsdkBase.ts
```diff
@@ -21,21 +21,21 @@ import type { ObjectTypeDefinition } from "./ontology/ObjectTypeDefinition.js";
...
   Q extends ObjectOrInterfaceDefinition,
-> = {
+> {
   readonly $apiName: Q["apiName"];
 
-  readonly $objectType: string;
```
ericanderson, line 29: nit this didnt need to move

packages/api/src/OsdkBase.ts
ericanderson: Hrm, vscode deleted my entire writeup. That sucks. I think we need to fix this so that we have one type for "Accepts an OSDK Object" and one type for "We put these on all objects".

Calling functions/actions only requires `{ $apiName, $primaryKey }`. If we switch to that, we are not breaking back compat because we are loosening the constraint. Maybe this is called `ObjectRef<T>`

We also need the "Base" for objects but I am afraid of breaking `OsdkBase`, as silly as that might be. So we could instead create a new interface to represent that?

ericanderson: I could probably be convinced to break `OsdkBase` if we make the new type for queries/etc.

packages/api/src/OsdkBase.ts
```diff
@@ -21,22 +21,22 @@ import type { ObjectTypeDefinition } from "./ontology/ObjectTypeDefinition.js";
 import type { PrimaryKeyTypes } from "./ontology/PrimaryKeyTypes.js";
 import type { OsdkObjectPrimaryKeyType } from "./OsdkObjectPrimaryKeyType.js";
 
-export type OsdkBase<
+export interface ObjectIdentifiers<
   Q extends ObjectOrInterfaceDefinition,
-> = {
+> {
   readonly $apiName: Q["apiName"];
```
nihalbhatnagar, line 27: We don't really need $apiName at all. Not sure if we want to also add `$objectType` for consistency or remove both or keep this

packages/api/src/OsdkBase.ts
nihalbhatnagar: Let me know if keeping it is preferable than removing both

packages/api/src/OsdkBase.ts
ssanjay1: If we don't have apiName/objectType, while that is correct, doesn't that take away some of the type safeguards we gave people when passing us concrete objects? If it's just PK I could theoretically pass in another object type that has overlapping pk type right?

ssanjay1 (commented):
Should we also modify `toDataValue` and `toDataValueQueries` to speak our new types? Right now they're checking for osdkBase, which the type guard still technically passes with but isn't totally correct if we're really giving it `ObjectIdentifiers`

ssanjay1 (approved):

PR #1475, opened 2025-05-12
Fix marking/list validation

davidpaquin:
https://github.com/palantir/osdk-ts/pull/1462 was accidentally reverted in https://github.com/palantir/osdk-ts/pull/1466


aep000:
👍

aep000 (approved):

PR #1488, opened 2025-05-13
Fixes for custom widgets Vite plugin in Code Workspaces mode

rosscourt:
Consolidated some of the logic around usage of the `baseHref` value and started using this value in `getWidgetIdOverrideMap`, where previously there was still usage of a localhost prefix even if running in Code Workspaces mode. I've tested this in a workspace and added tests based on representative example data in order to validate the changes.

Also corrected the usage of the `FOUNDRY_PROXY_TOKEN` environment variable, which points to a file to be read instead of containing the token value directly.

### DIFF
--- a/packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
+++ b/packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
@@ -25,8 +25,6 @@ describe("getBaseHref", () => {
   });
 
   test("returns localhost URLs in local mode", () => {
-    vi.spyOn(codeWorkspacesMode, "isCodeWorkspacesMode").mockReturnValue(false);
-
     const server = {
       config: {
         mode: "",
@@ -40,7 +38,6 @@ describe("getBaseHref", () => {
   });
 
   test("returns code workspaces URLs in code workspaces mode", () => {
-    vi.spyOn(codeWorkspacesMode, "isCodeWorkspacesMode").mockReturnValue(true);
     // Representative value when running dev mode in Code Workspaces mode
     vi.spyOn(codeWorkspacesMode, "getCodeWorkspacesBaseHref").mockReturnValue(
       "https://workspace.stack.com/proxy/path",
--- a/packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
+++ b/packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as codeWorkspacesMode from "../codeWorkspacesMode.js";
+import { getBaseHref } from "../getBaseHref.js";
+
+describe("getBaseHref", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  test("returns localhost URLs in local mode", () => {
+    vi.spyOn(codeWorkspacesMode, "isCodeWorkspacesMode").mockReturnValue(false);
```
tzyl, line 28: Do we need to mock isCodeWorkspacesMode if we have server.config.mode to the correct value anyway?

packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    const result = getBaseHref(server);
+    expect(result).toEqual("http://localhost:5173/");
+  });
+
+  test("returns code workspaces URLs in code workspaces mode", () => {
+    vi.spyOn(codeWorkspacesMode, "isCodeWorkspacesMode").mockReturnValue(true);
+    // Representative value when running dev mode in Code Workspaces mode
+    vi.spyOn(codeWorkspacesMode, "getCodeWorkspacesBaseHref").mockReturnValue(
+      "https://workspace.stack.com/proxy/path",
+    );
```
tzyl, line 47: Maybe we can stub env variables to make this cover the whole getBaseHref code path?
https://vitest.dev/api/vi.html#vi-stubenv

Or we can test the codeWorkspacesMode level separately

packages/widget.vite-plugin.unstable/src/dev-plugin/getWidgetIdOverrideMap.ts
```diff
@@ -48,16 +48,22 @@ export async function getWidgetIdOverrideMap(
...
     },
     {},
   );
 
   return widgetIdToOverrides;
 }
+
+function removeBasePath(scriptPath: string, baseHref: string): string {
+  const baseHrefPath = new URL(baseHref).pathname;
+  return scriptPath.replace(baseHrefPath, "");
```
tzyl, line 68: Is there anything we want to do here to check that we actually performed this replace? We're expecting baseHrefPath to always be the prefix of scriptPath right?

tzyl (approved):

packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
rosscourt: They can be removed

packages/widget.vite-plugin.unstable/src/dev-plugin/__tests__/getBaseHref.test.ts
rosscourt: Added some separate tests for codeWorkspacesMode

packages/widget.vite-plugin.unstable/src/dev-plugin/getWidgetIdOverrideMap.ts
rosscourt: Are you imagining some kind of runtime check? I think the unit tests with representative examples should be enough for confidence

tzyl: Yeah I was thinking roughly something like:
```
if !scriptPath startsWith baseHrefPath {
  throw
}

return scriptPath stripPrefix baseHrefPath
```

but if we're pretty sure this should line up then can leave it

tzyl (approved):

PR #1508, opened 2025-05-16
Add doc snippets for contains + searchAround

mvanschellebeeck:


packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1839,7 +1850,7 @@ versions:
 
             // Object set operations can be chained. e.g. To find all objects in objectSetA 
             // that are present in objectSetB but do not exist in objectSetC:
```
mvanschellebeeck, line 1852: drive-by fixes: this type is incorrect. It should actually be `$Objects.{{objectType}}.ObjectSet` but I decided to omit it because this feels simpler

nihalbhatnagar: I think you should also be able to do the prior (assuming the object type name is the name we generate with) with the right import, but agreed this is a lot simpler

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -302,6 +302,17 @@ versions:
         - template: |-
             const {{objectType}}ObjectSet = client.ontology.objects.{{objectType}}
                 .where(query => query.{{property}}.containsAllTerms("foo bar"));
+      containsTemplate:
+        - template: |-
+            import { {{objectType}} } from "{{{packageName}}}";
+            // Edit this import if your client location differs
+            import { client } from "./client";
```
nihalbhatnagar, line 309: Can we add a newline here to make is consistent with the rest of the snippets I recall

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1880,9 +1891,19 @@ versions:
...
             // Return all objects common to objectSetA, objectSetB and objectSetC
-            const result: {{objectType}}.ObjectSet = objectSetA
+            const result = objectSetA
               .intersect(objectSetB)
               .intersect(objectSetC); // alternatively: objectSetA.intersect(objectSetB, objectSetC)
+      searchAround:
+        - template: |-
+            // Edit this import if your client location differs
+            import { client } from "./client";
+            import { {{sourceObjectType}} } from "{{{packageName}}}";
```
nihalbhatnagar, line 1901: Same here

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1880,9 +1891,19 @@ versions:
 
 
             // Return all objects common to objectSetA, objectSetB and objectSetC
-            const result: {{objectType}}.ObjectSet = objectSetA
+            const result = objectSetA
               .intersect(objectSetB)
               .intersect(objectSetC); // alternatively: objectSetA.intersect(objectSetB, objectSetC)
+      searchAround:
+        - template: |-
+            // Edit this import if your client location differs
```
nihalbhatnagar, line 1899: I like this client import, though note for us is that I think we should make sure this remains consistent with the templates

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1880,9 +1891,19 @@ versions:
...
+            import { client } from "./client";
+            import { {{sourceObjectType}} } from "{{{packageName}}}";
+            // Take 3 `{{sourceObjectType}}` objects
+            const objects = await client({{sourceObjectType}}).take(3);
```
nihalbhatnagar, line 1903: Hmm we don't have a `take`, additionally our pivotTo's are on the object set class. I think you're probably looking for await client({{sourceObjectType}}).pivotTo("{{linkApiName}}")

packages/typescript-sdk-docs/src/documentation.yml
nihalbhatnagar: We probably want to move these to the 2.0 section since this looks like 2.0 style

packages/typescript-sdk-docs/src/documentation.yml
mvanschellebeeck: oops seems like thats only supported in FoO

nihalbhatnagar (approved):

PR #1521, opened 2025-05-20
[old] Support datasetv2 struct properties

jscaii:
# Before
struct properties were not properly mapped for datasetv2; they were mapped in the same way as a normal column

# After
struct properties have their fields mapped as well


packages/maker/src/api/defineOntology.ts
jscaii, line 1: relevant changes are here

jscaii:
all changes that aren't in defineOntology.ts were from running turbo fix-lint

jscaii:
Closing in favor of #1526 
Running fix-lint affected a bunch of other packages that we don't want to touch

PR #1525, opened 2025-05-20
Enforce that struct, mediaReference, and GTSR type SPTs cannot have render hints

davidpaquin:
Changes the default for struct, mediaReference, and geotimeSeriesReference types and adds an invariant to make it a build time error


aep000:
👍

aep000 (approved):

PR #1538, opened 2025-05-22
Update OAC codegen

davidpaquin:
There were a couple of issues with the codegen 
- We didn't wrap with a proxy, so everything got imported -- this wasn't caught because we weren't actually importing from a package during testing (because we didn't publish a package) 
- Enums are weird -- strings couldn't be cast into the enum for typechecking purposes, so we use the actual enum
- The way conjure defines optionals, e.g. `failureMessage: string | undefined` instead of `failureMessage?: string` meant that something like `JSON.stringify(myValueType) as ValueType` would fail because `failureMessage` didn't exist in the stringified version
- Some other misc things

### DIFF
--- a/.changeset/light-eagles-allow.md
+++ b/.changeset/light-eagles-allow.md
@@ -1,0 +1,5 @@
+---
+"@osdk/maker": patch
+---
+
+Update OAC codegen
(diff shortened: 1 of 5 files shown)


aep000 (approved):
LGTM

ethanai22 (approved):

aep000 (approved):

PR #1548, opened 2025-05-23
Experimental Metadata property on OSDK Objects

nihalbhatnagar:


packages/client/src/ontology/StandardOntologyProvider.ts
```diff
@@ -59,14 +60,14 @@ export const createStandardOntologyProviderFactory: (
         [InterfaceDefinitions]: interfaceDefs,
       };
 
-      return fullObjectDef;
+      return deepFreeze(fullObjectDef);
```
nihalbhatnagar, line 63: This was also a bug before, but for our client.fetchMetadata that uses Standard Ontology Provider, we end up returning the raw cached value to the user. This is an issue because they can mess with the object we store in our caches and would then mess with internal OSDK client code. I opted to deep freeze the returned object but keep the reference the same to keep duplication to a minimum

packages/client/src/ontology/StandardOntologyProvider.ts
```diff
@@ -59,14 +60,14 @@ export const createStandardOntologyProviderFactory: (
...
 
     async function loadInterface(
       client: MinimalClient,
       key: string,
     ) {
-      return loadInterfaceMetadata(client, key);
+      return deepFreeze(await loadInterfaceMetadata(client, key));
     }
 
     async function loadQuery(
```
nihalbhatnagar, line 73: I probably need to add freezing to query and action metadata as well

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -121,7 +129,7 @@ export function createOsdkObject(
         enumerable: false,
         value: simpleOsdkProperties,
       },
-      [ObjectDefRef]: { value: objectDef, enumerable: false },
+      [ObjectDefRef]: { value: objectDef, enumerable: false }, // TODO: Potentially update when GA metadata field
```
nihalbhatnagar, line 132: Keeping internal references to avoid refactors if we remove

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -619,7 +619,7 @@ describe("ObjectSet", () => {
       type ObjectSetType = $ObjectSet<
         EmployeeApiTest,
         {
-          mom: "integer" | undefined;
```
nihalbhatnagar, line 622: I'm not going to lie, I don't know why my changes forced me to change this. The weird thing is, this was wrong before and this is now correct, $count returns a non-nullable option. Potentially my extra ternary sparked the compiler to do some deeper digging but I have no idea

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -68,6 +68,13 @@ export function createOsdkInterface<
         value: clone,
         enumerable: false,
       },
+      "$__experimental_metadata": {
```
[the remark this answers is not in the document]
ssanjay1: I think we should keep in parity with our other experimental features so like `__EXPERIMENTAL__NOT_SUPPORTED_YET__`

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -513,7 +513,7 @@ export class ObjectSetListenerWebsocket {
           o.object.__apiName,
           sub.interfaceApiName,
         ),
-      ) as Array<Osdk.Instance<any, never, any>>;
+      ) as Array<Osdk.Instance<any>>;
```
ssanjay1, line 516: Whats going on here?

nihalbhatnagar: I just cleaned this up, didnt need those generics

packages/api/src/OsdkObjectFrom.ts
```diff
@@ -213,6 +217,15 @@ export namespace Osdk {
             >["props"][K];
           },
       ) => Osdk.Instance<Q, OPTIONS, P | NEW_PROPS>;
+
+      readonly $__EXPERIMENTAL__NOT_SUPPORTED_YET__metadata: Q extends
+        ObjectTypeDefinition ? {
+          object: Q;
```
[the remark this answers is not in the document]
nihalbhatnagar: Sure, I just went this way because potentially `$metadata.objectMetadata` might be redundant but happy w this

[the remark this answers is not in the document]
ssanjay1: Yea I just think in the non interface case people will be confused why there's an extra object field a level down

ssanjay1 (dismissed):

ssanjay1 (approved):

PR #1549, opened 2025-05-26
Remove extra JSDoc spaces

MariaSolOs:
These extra spaces aren't needed and can be interpreted as invalid indentation (and lint rules like [`jsdoc/check-indentation`](https://github.com/gajus/eslint-plugin-jsdoc/blob/e26a11a39930ebd07f61e509d91e6cb87f017dde/docs/rules/check-indentation.md) will flag these.


MariaSolOs:
Is there a command to update the snapshots causing the CI test failures? They seem to differ by just whitespace so the diff looks expected.

PR #1552, opened 2025-05-27
Handle enums correctly so that they won't throw on a fetchMetadata call or generation

nihalbhatnagar:


packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -33,7 +33,7 @@ export interface ActionMetadata {
       modified: boolean;
     }>
   >;
-  status: ReleaseStatus;
+  status?: ReleaseStatus;
```
ssanjay1, line 36: Should this also be `| undefined` rather than optional

nihalbhatnagar: Ah I forgot to remove it, ReleaseStatus is already | undefined so just need to remove

PR #1557, opened 2025-05-27
Add property deprecation to JS docs

nihalbhatnagar:
image

### DIFF
--- a/examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Todo.ts
+++ b/examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Todo.ts
@@ -30,6 +30,7 @@ export namespace Todo {
     readonly isComplete: $PropType['boolean'] | undefined;
     /**
      *   display name: 'Title',
+     *
      *   description: The text of the todo
      */
     readonly title: $PropType['string'] | undefined;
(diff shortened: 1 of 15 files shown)


examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Todo.ts
```diff
@@ -84,6 +85,7 @@ export interface Todo extends $ObjectTypeDefinition {
       isComplete: $PropertyDef<'boolean', 'nullable', 'single'>;
       /**
        *   display name: 'Title',
+       *
```
ssanjay1, line 88: Why did all of these get added?

packages/generator/src/shared/propertyJsdoc.ts
```diff
@@ -15,32 +15,52 @@
...
+      ret.push(` *   description: ${property.description}\n`);
     }
   } else {
-    ret += ` * (no ontology metadata)\n`;
+    ret.push(` * (no ontology metadata)\n`);
   }
-
-  ret += ` */\n`;
-  return ret;
+  return `/**\n` + ret.join("*\n") + ` */\n`;
```
ssanjay1, line 65: Pretty sure it's this join adding the extra lines

ssanjay1: I don't feel super strongly against it, but would be nice if we didn't have to have super large diff for new codegen without the status stuff if not needed

examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Todo.ts
nihalbhatnagar: I added it intentionally, without it all of the metadata lines are stuffed onto one line in intellisense. I thought it looked cleaner this way but happy to remove or do a flup.

PR #1563, opened 2025-05-30
Add group permission configuration for OAC actions

ethanai22:
Relevant changes in maker pacakge.
Users can add a custom group with submission permissions to OAC generated actions on objects and interfaces. Can add organization permission support in follow up


packages/maker/src/api/defineOntology.ts
```diff
@@ -399,7 +400,21 @@ function convertToWireBlockData(
       }),
     ),
     blockPermissionInformation: {
-      actionTypes: {},
+      actionTypes: Object.fromEntries(
+        Object.entries(ontology[OntologyEntityTypeEnum.ACTION_TYPE])
+          .filter(([apiName, action]) => action.validation)
+          .map<
+            [string, ActionTypePermissionInformation]
+          >(([apiName, action]) => {
```
davidpaquin, line 408: I'm a little confused on how the type is a `ActionTypePermissionInformation` instead of `ActionType`, shouldn't you need to map the `ActionType` to `ActionTypePermissionInformation` somewhere?

davidpaquin: Oh you don't actually use the action so it doesn't matter

packages/maker/src/api/types.ts
```diff
@@ -196,6 +196,17 @@ export interface ActionTypeInner {
...
+export type ActionValidationRule = OntologyIrValidationRule;
+
+export type ActionValidationDefinition =
+  | GroupValidationRule
+  | OntologyIrValidationRule;
+
+export type GroupValidationRule = {
+  type: "group";
+  name: string;
+};
```
davidpaquin, line 208: I wonder if this would be better
```
export type GroupValidationRule = {
    type: "submitterInGroup";
    groupName: string;
}
```
I'm trying to think about the end user that doesn't know much, and I feel like `type: "group"` might be a little vague

davidpaquin (approved):
Looks good! just a note on the names of stuff for UX purposes

packages/maker/src/api/types.ts
ethanai22: Yea based off the configuration menus in OMA it seems that submitter permissions are the most common validation that gets attached and they only list "group", "organization", and "user" as options so I was basing it off that. Agreed that we can add more descriptive names in a follow up when organizations get added.

PR #1564, opened 2025-05-30
Fix count aggregation for RDPs

nihalbhatnagar:
Count aggregations return strings from the wire instead of numbers, We want to keep typing them as numbers, so we check the definition and convert the string to a number at runtime.

### DIFF
--- a/packages/client/src/derivedProperties/derivedPropertyRuntimeMetadata.ts
+++ b/packages/client/src/derivedProperties/derivedPropertyRuntimeMetadata.ts
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -142,11 +142,31 @@ export function createOsdkObject(
       );
     } else if (
       propKey in derivedPropertyTypeByName
```
ssanjay1, line 144: For ease of reading, want to nest the derived property checkers under one else branch?

packages/client/src/derivedProperties/derivedPropertyRuntimeMetadata.ts
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectMetadata } from "@osdk/api";
+import type { DerivedPropertyDefinition } from "@osdk/foundry.ontologies";
+
+export type DerivedPropertyRuntimeMetadata = Record<string, {
+  definition: DerivedPropertyDefinition;
+  propertyType: ObjectMetadata.Property | undefined;
```
ssanjay1, line 22: Why is this typed as `| undefined ` now?

nihalbhatnagar: We only use propertyType for collectSet, collectList, or selectProperty operations to know the underlying data type of the properties we selected. Prior to this, I only created propagated the RDP metadata for the above three cases. Because I now want the definition for all, I made this or undefined(not optional so we don't miss it). I'll update the name to be more clear

ssanjay1 (approved):

PR #1565, opened 2025-06-02
Update property default TypeClasses to match SPT default TypeClasses 

jscaii:
Update property default TypeClasses to match SPT default TypeClasses


davidpaquin (approved):
Looks good! Weird that the diff in the test file was so big since only type classes were added

PR #1572, opened 2025-06-03
[widgets] Use window.parent instead of window.top

amishshah:
`window.top` isn't actually the desired window for the widgets workflow (e.g. if we iframe a workshop containing a widget into some other page, we don't want the other page to start receiving the messages!)


tzyl (dismissed):
Also highlighting that this still is limited by the server controlled meta tag origin so messages will not be sent to unexpected origins and the frame-ancestors used at the moment would also add another guard

tzyl (approved):

PR #1588, opened 2025-06-06
Add value type support for concrete type properties

jscaii:


packages/maker/src/api/types.ts
```diff
@@ -291,7 +291,7 @@ export interface ObjectPropertyTypeInner extends
 {
   type: PropertyTypeType;
   array?: boolean;
-  valueType: string | ValueTypeDefinitionVersion;
```
jscaii, line 294: don't think that valueType is ever a string rather than the actual definition; removing

packages/maker/src/api/overall.test.ts
```diff
@@ -65,6 +65,74 @@ describe("Ontology Defining", () => {
...
+        ?.propertyTypeConstraints[0];
+      expect(constraint?.constraints.type).toBe("string");
+      expect(constraint?.constraints.string?.length?.minSize).toBe(5);
+      expect(constraint?.constraints.string?.length?.maxSize).toBe(20);
+      expect(constraint?.failureMessage?.message).toBe(
+        "String must be between 5 and 20 characters",
+      );
+    });
+
+    // I had originally added a test for struct conversions too, but value type structs are converted to structV2 which are not supported
```
jscaii, line 134: will delete this comment before merging, but I originally had added a test for struct conversions which failed since value type structs are converted to structV2, which is not supported in property types yet (there's no PropertyTypeDataConstraint for structv2)

packages/maker/src/api/overall.test.ts
```diff
@@ -65,6 +65,74 @@ describe("Ontology Defining", () => {
...
+      });
+
+      expect(objectPropertyType.dataConstraints).toBeDefined();
+      expect(objectPropertyType.dataConstraints?.propertyTypeConstraints)
+        .toHaveLength(1);
+
+      const constraint = objectPropertyType.dataConstraints
+        ?.propertyTypeConstraints[0];
+      expect(constraint?.constraints.type).toBe("string");
+      expect(constraint?.constraints.string?.length?.minSize).toBe(5);
```
jscaii, line 127: passes locally but circle is unhappy about constraint.string-- will fix

packages/maker/src/api/defineOntology.ts
```diff
@@ -652,13 +670,130 @@ function convertProperty(property: ObjectPropertyType): OntologyIrPropertyType {
...
+        failureMessage: constraint.constraint.failureMessage,
+      }),
+    );
+
+  return {
+    propertyTypeConstraints,
+  };
+}
+
+function dataConstraintToPropertyTypeDataConstraint(
```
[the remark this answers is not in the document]
jscaii: makes sense! extracted some methods into a utils file, lmk what you think

aep000 (approved):

PR #1597, opened 2025-06-10
Add automation definitions

anik545:


packages/client.unstable/src/blockDataIr.ts
```diff
@@ -62,3 +64,18 @@ export type OntologyIrValueTypeBlockDataEntry = {
...
+export type SingleAutomationIrBlockData = {
+  automation: MarketplaceMonitor;
+  // All the object sets that are used by this automation
+  objectSets: ObjectSetBlockDataEntry[];
+};
+
+export type ObjectSetBlockDataEntry = {
+  objectSet: TemplatedObjectSet;
+  templateId: string;
+};
```
davidpaquin, line 81: This stuff will have to change to match the IR, but if it doesn't break things, we can probably still push it through to start wiring things up

davidpaquin (dismissed):

davidpaquin (approved):

aep000 (dismissed):

PR #1620, opened 2025-06-14
improve media upload interface (beta)

wxyangf:


packages/client/src/util/toDataValue.ts
```diff
@@ -38,6 +38,7 @@ import { isWireObjectSet } from "./WireObjectSet.js";
 export async function toDataValue(
   value: unknown,
   client: MinimalClient,
+  actionApiName: string,
```
ssanjay1, line 41: Can you have this take actionMetadata instead? @nihalbhatnagar is doing a separate change that kinda overlaps which will need the entire actionMetadata, so may as well just do that here to avoid conflicts

packages/api/src/mapping/PropertyValueMapping.ts
```diff
@@ -75,7 +75,7 @@ export interface PropertyValueClientToWire {
   short: number;
   string: string;
   timestamp: string;
-  mediaReference: Media;
+  mediaReference: Media | MediaUpload;
```
ssanjay1, line 78: You don't need to add this

ssanjay1: It's ignored for attachments too, we just have kept it there for consistency from previously

wxyangf: will remove

packages/client/src/util/extractRdpDefinition.ts
```diff
@@ -188,7 +188,7 @@ async function extractRdpDefinitionInternal(
       return { definitions: {} };
     // We don't have to worry about new object sets being added and doing a runtime break and breaking people since the OSDK is always constructing these.
     default:
-      const _: never = objectSet;
+      const _ = objectSet;
```
nihalbhatnagar, line 191: Can we instead keep the never and add a "interfaceLinkObjectSet" right above this? This will make sure we handle this case properly when we add this support

wxyangf: fixed!

packages/client/src/util/toDataValue.test.ts
```diff
@@ -161,10 +201,26 @@ describe(toDataValue, () => {
       { name: "file1.txt" },
     );
 
-    const converted = await toDataValue(file, clientCtx);
+    const converted = await toDataValue(file, clientCtx, mockActionMetadata);
     expect(converted).toMatch(/ri\.attachments.main.attachment\.[a-z0-9\-]+/i);
   });
 
+  /*
+  it("converts media uploads correctly", async () => {
```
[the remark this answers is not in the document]
nihalbhatnagar: This codepath won't run on e2e tests. This seems to be the core codepath that should be tested, why is this failing? If its related to @mnayan s work then that should be the test below right?

packages/client/src/util/toDataValue.ts
```diff
@@ -85,22 +88,40 @@ export async function toDataValue(
         filename: value.name as string,
       },
     );
-    return await toDataValue(attachment.rid, client);
+    return await toDataValue(attachment.rid, client, actionMetadata);
+  }
+
+  // new media item upload interface, very similar to how attachments work above
```
nihalbhatnagar, line 94: Nit: should remove comment

packages/api/src/mapping/DataValueMapping.ts
```diff
@@ -72,7 +72,7 @@ export interface DataValueClientToWire {
   string: string;
   timestamp: string;
   set: Set<any>;
-  mediaReference: MediaReference;
+  mediaReference: MediaReference | MediaUpload;
```
nihalbhatnagar, line 75: Note: Intellisense is messed up. Will figure out after dev con

nihalbhatnagar (dismissed):

packages/shared.test/src/stubs/objectTypeV2.ts
```diff
@@ -609,30 +609,35 @@ export const BGaoNflPlayerObjectType: ObjectTypeV2 = {
             dataType: {
               type: "string",
             },
+            rid: "rid.a.b",
```
tdeitch, line 612: Nit: RIDs start with `ri.`, not `rid.`, and then have four components, so `ri.a.b.c.d`. Even better would be to use realistic ones like `ri.ontology.main.property.ab2f1665-7652-4932-bf60-0f36bf9eb55e`.

tdeitch (commented):
Approving with one nit

tdeitch:
👍

packages/shared.test/src/stubs/objectTypeV2.ts
wxyangf: changed to `ri.a.b.c.d` for now

PR #1629, opened 2025-06-17
Split Faux into its own package

ericanderson:


packages/faux/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@osdk/faux",
+  "private": true,
+  "version": "2.3.0-beta.5",
```
ssanjay1, line 4: Are you just keeping in line with shared.test? Does it matter enough to make them fixed together like client/api

packages/faux/CHANGELOG.md
```diff
@@ -0,0 +1,1047 @@
+# @osdk/shared.test
```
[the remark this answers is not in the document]
ericanderson: just did a copy paste of the directories as step one. will fix

packages/faux/package.json
ssanjay1: Ah I see, I think its fine keeping separate for now I don't feel super strong about it yet

ssanjay1 (approved):

PR #1631, opened 2025-06-18
Remove unstable suffix from widget libraries

amishshah:
Intending to kick off the new stable packages at `1.0.0`, let me know if this is the wrong way to do it!

### DIFF
--- a/.changeset/puny-rabbits-walk.md
+++ b/.changeset/puny-rabbits-walk.md
@@ -1,0 +1,5 @@
+---
+"@osdk/widget.client": patch
+---
+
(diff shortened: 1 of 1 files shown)


packages/create-widget.template.minimal-react.v2/templates/package.json.hbs
```diff
@@ -11,10 +11,10 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@osdk/widget.client-react.unstable": "^1.1.0",
-    "@osdk/widget.client.unstable": "^1.1.0"
+    "@osdk/widget.client-react": "workspace:*",
```
ssanjay1, line 14: Was this intentional?

PR #1651, opened 2025-06-23
Use root id rather than body for widgets react-dom

tzyl:
Updates the following:

- Widget minimal template
- Widget OSDK template
- Widget vite plugin

To use the root id element rather than body for react-dom. This cleans up some console errors and may help prevent interference if other libraries directly inject more elements into the body element

### DIFF
--- a/packages/create-widget.template.minimal-react.v2/templates/src/main.tsx
+++ b/packages/create-widget.template.minimal-react.v2/templates/src/main.tsx
@@ -1,14 +1,15 @@
 import "@blueprintjs/core/lib/css/blueprint.css";
+import "./main.css";
 
 import { FoundryWidget } from "@osdk/widget.client-react";
 import { createRoot } from "react-dom/client";
 import MainConfig from "./main.config.js";
 import { Widget } from "./Widget.js";
 
-const root = document.querySelector("body")!;
+const root = document.getElementById("root")!;
 
 createRoot(root).render(
   <FoundryWidget config={MainConfig}>
     <Widget />
-  </FoundryWidget>,
+  </FoundryWidget>
 );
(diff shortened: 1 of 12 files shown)


packages/create-widget.template.minimal-react.v2/templates/src/main.tsx
```diff
@@ -1,14 +1,15 @@
 import "@blueprintjs/core/lib/css/blueprint.css";
+import "./main.css";
```
tzyl, line 2: This file was never being included and it resets the body margin to zero:
https://github.com/palantir/osdk-ts/blob/main/packages/create-widget.template.minimal-react.v2/templates/src/main.css

I updated the widget's container CSS to put the margin back to keep the visuals the same as before

packages/create-widget.template.react.v2/templates/index.html
```diff
@@ -6,6 +6,7 @@
     <title>Widget: Ontology SDK + React</title>
   </head>
   <body>
+    <div id="root"></div>
```
tzyl, line 9: This is a bit confusing but this index.html for the widgets don't actually get used for rendering, it's only used by the vite plugin to extract out the entrypoints. For rendering we provide the index.html but I updated these as well for consistency

packages/widget.vite-plugin/index.html
```diff
@@ -6,9 +6,7 @@
     <title>Palantir: Widget dev mode setup</title>
   </head>
   <body>
-    <div id="root-container">
-      <div id="root"></div>
-    </div>
+    <div id="root"></div>
```
tzyl, line 9: This is used for showing the loading spinner before redirect on the dev setup page. The "root-container" part was copied over from the create-app templates which sets up some styles for putting the preview banner on the page but not relevant here so I removed the extra level

PR #1684, opened 2025-07-02
Add hydrateObjectSetFromObjectRidsUtil

asaliling:
This util will provide an easy way for developers to create a static OSDK object set from a list of object rids without any network calls involved. I modeled this after the `hydrateObjectSetFromRid` util.

### DIFF
--- a/packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
+++ b/packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
@@ -20,10 +20,10 @@ import { additionalContext } from "../Client.js";
 import { createObjectSet } from "../objectSet/createObjectSet.js";
 
 /**
- * Creates an OSDK object set from an object set RID.
+ * Creates an OSDK object set from object RIDs.
  * @param client - An OSDK client.
  * @param definition - An OSDK object or interface definition.
- * @param rids - An array of object rids
+ * @param rids - An array of object RIDs.
  * @returns An OSDK object set.
  */
 export function hydrateObjectSetFromObjectRids<


packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { ObjectOrInterfaceDefinition } from "@osdk/api";
+import type { Client } from "../Client.js";
+import { additionalContext } from "../Client.js";
+import { createObjectSet } from "../objectSet/createObjectSet.js";
+
+/**
+ * Creates an OSDK object set from an object set RID.
+ * @param client - An OSDK client.
+ * @param definition - An OSDK object or interface definition.
+ * @param rids - An array of object rids
```
asaliling, line 26: this argument could also be `objects` to be more consistent with the wire object set shape, but I named it `rids` to make it more clear to consumers that these are object rids and not objects.

packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @returns An OSDK object set.
+ */
+export function hydrateObjectSetFromObjectRids<
+  T extends ObjectOrInterfaceDefinition,
+>(client: Client, definition: T, rids: string[]) {
+  return createObjectSet(definition, client[additionalContext], {
+    type: "intersect",
+    objectSets: [
+      {
+        type: "base",
```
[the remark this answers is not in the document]
ssanjay1: If you're referring to the type strings nah we don't, that should be fine

packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+import type { ObjectOrInterfaceDefinition, ObjectSet } from "@osdk/api";
+import type { Client } from "../Client.js";
+import { additionalContext } from "../Client.js";
+import { createObjectSet } from "../objectSet/createObjectSet.js";
+
+/**
+ * Creates an OSDK object set from an object set RID.
```
ssanjay1, line 23: Can we update this comment please?

packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
ssanjay1: Works for me! could also call it `objectRids`

ssanjay1 (approved):

PR #1686, opened 2025-07-07
Async processing on function definition metadata

iw-an:
When executing a function with either a `map` or `struct` return type the resolution of each object's metadata happens synchronously.

If you have many types in the struct this can balloon the time to execution on the first run.

Making this async keeps execution time roughly the same regardless of the amount of types in the struct/map


ericanderson:
Closing and reopening to trigger builds

PR #1689, opened 2025-07-07
Fix spt object equality check

aep000:


packages/maker/src/api/defineObject.ts
```diff
@@ -214,7 +214,7 @@ function validateInterfaceImplProperty(
         `Object property mapped to interface does not exist. Object Property Mapped: ${mappedObjectProp}`,
     };
   }
-  if (spt.type !== objProp?.type) {
+  if (JSON.stringify(spt.type) !== JSON.stringify(objProp?.type)) {
```
davidpaquin, line 217: technically I think `stringify` isn't stable, but it's probably fine in whatever modern environment this is running in

davidpaquin (approved):

PR #1696, opened 2025-07-08
[Functions] For `multiplicity: true` links, allow lists of target objects in EditBatch::link and EditBatch::unlink

marshallavail:
When users want to add multiple links from the same source to many targets (multiplicity is true on the link), they currently have to write a for-each loop. Same is true for unlinking. For example,
```
targetsToLink.forEach(target => editBatch.link(source, "myLink", target));
```

This PR updates the `link` and `unlink` to allow lists when multiplicity is true on the relevant link. For example,
```
editBatch.link(source, "myLink", targetsToLink);
```


packages/functions/src/edits/EditBatch.ts
```diff
@@ -43,7 +43,12 @@ export type AddLinkTargets<
   X extends AnyEdit,
   SOL extends ObjectLocator<any>,
   A extends string,
-> = X extends AddLink<ObjectTypeDefinitionForLocator<SOL>, A> ? X["target"]
+> = X extends AddLink<ObjectTypeDefinitionForLocator<SOL>, A>
+  ? (X extends AddLink<infer OTD, A>
+    ? (CompileTimeMetadata<OTD>["links"][A]["multiplicity"] extends true
+      ? Array<X["target"]> | X["target"]
```
marshallavail, line 49: Accepts list or singleton target for backcompat

ssanjay1 (approved):

PR #1699, opened 2025-07-09
Conditional overrides + condition usability improvements

ethanai22:


packages/maker/src/api/ontologyUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+        type: "hidden",
+        hidden: {},
+      };
+    default:
+      return {
+        type: "editable",
+        editable: {},
+      };
+  }
+}
```
davidpaquin, line 54: You should be able to make this generic, e.g.
```
const vis = visibility ?? "editable";
return {
  type: vis.
  [vis]: {}
};
```

packages/maker/src/api/overall.test.ts
```diff
@@ -6597,35 +6881,39 @@ describe("Ontology Defining", () => {
...
-            },
-            right: {
-              type: "staticValue",
-              staticValue: {
-                type: "stringList",
-                stringList: {
-                  strings: [
-                    "87ef507e-f954-457e-ad68-e0df71ef7567",
-                  ],
+          type: "and",
```
aep000, line 6884: Curious about the and being applied here? Its technically a no-op but just curious.

packages/maker/src/api/overall.test.ts
```diff
@@ -6199,8 +6481,10 @@ describe("Ontology Defining", () => {
         }],
       });
       const createAction = defineCreateObjectAction(sampleObject, {
-        type: "group",
-        name: "testGroup",
+        condition: {
```
aep000, line 6484: note for FLUP rename this to describe what level of a condition this is.

packages/maker/src/api/ontologyUtils.ts
ethanai22: good point

packages/maker/src/api/ontologyUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+  OntologyIrParameterValidationBlockOverride,
+} from "@osdk/client.unstable";
+
+import type {
+  ActionParameterConditionalOverride,
+  ActionParameterValidation,
+  ConditionDefinition,
+} from "./types.js";
+
+export function convertVisibility(
```
aep000, line 29: can you rename to "convertActionVisibility"?

packages/maker/src/api/ontologyUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+  visibility: "editable" | "disabled" | "hidden" | undefined,
+): any {
```
aep000, line 31: Why does this return any?

packages/maker/src/api/types.ts
```diff
@@ -143,8 +145,41 @@ export interface ActionParameter {
 export interface ActionParameterValidation {
   allowedValues: ActionParameterAllowedValues;
   required: ActionParameterRequirementConstraint;
+  defaultVisibility?: "editable" | "disabled" | "hidden";
+  conditionalOverrides?: Array<ActionParameterConditionalOverride>;
 }
 
+// TODO(ethana): add more commonly used conditions - parameter matching, organizations, etc.
```
aep000, line 152: We should follow up on this ASAP because Cacti wants to use some more complex conditions.

aep000 (commented):
Some questions and some nits but looks good.

packages/maker/src/api/overall.test.ts
ethanai22: Yea this can be cleaned up. I originally enforced top level conditions to be union types, but I changed it later and forgot to update this test

packages/maker/src/api/overall.test.ts
ethanai22: this is technically the condition inside the action level validation. The field will eventually look like 
```
actionLevelValidation: {
   condition: ...
   displayMetadata: ... // optional
}
```

packages/maker/src/api/types.ts
ethanai22: Agreed I'll sync with them to see what common conditions they need

packages/maker/src/api/ontologyUtils.ts
ethanai22: actually had trouble with this when declaring the return type as ParameterVisibility instead of any, so keeping it like this for now

davidpaquin: I think if you do `return { ... } as ParameterVisibility` it should work?

packages/maker/src/api/ontologyUtils.ts
ethanai22: fixed

aep000 (approved):

PR #1701, opened 2025-07-09
[hydrateObjectSetFromObjectRids] Take in a readonly rids argument

asaliling:
Take in a readonly rids argument.


packages/client/src/public-utils/hydrateObjectSetFromObjectRids.ts
```diff
@@ -38,8 +38,12 @@ export function hydrateObjectSetFromObjectRids<
...
       {
         type: "static",
-        objects: rids,
+        objects: asMutableArray(rids),
       },
     ],
   });
 }
+
+function asMutableArray<T>(array: readonly T[]): T[] {
```
asaliling, line 47: didn't see a similar util in the repo -- let me know if I missed it or if I should define this utils elsewhere.

ssanjay1 (approved):

PR #1702, opened 2025-07-10
Object properties are defined as maps

ethanai22:
When users define objects, supply properties as map of apiName -> property instead of list of object properties


aep000 (approved):
🔥

PR #1704, opened 2025-07-10
Remove features from stable release

nihalbhatnagar:
Removes

- Media
- RDP Expressions and Selection


packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -727,132 +727,136 @@ describe("ObjectSet", () => {
...
+            "fetchMetadata": [Function],
+            "rid": "ri.attachments.main.attachment.86016861-707f-4292-b258-6a7108915a75",
+          },
+        ]
       `,
       );
 
       expectTypeOf(objectWithRdp.geoSelectDp).toEqualTypeOf<
-        GeoJSON.GeoJSON | undefined
+        GeoJSON.GeoJSON[] | undefined
```
ssanjay1, line 828: Is this an array now because you can't do select?

PR #1705, opened 2025-07-10
Better syntax for parameter matching

ethanai22:
Goal is to provide a more intuitive syntax for the commonly used parameter matching condition:
Screenshot 2025-07-10 at 4 23 32 PM

```
{
  type: "parameter",
  parameterId: "param",
  matches: {
    type: "staticValue",
    staticValue: {
      type: "string",
      string: "foobar",
    }
  }
}
```

### DIFF
--- a/packages/maker/src/api/defineAction.ts
+++ b/packages/maker/src/api/defineAction.ts
@@ -25,12 +25,14 @@ import {
 import { convertConditionDefinition } from "./ontologyUtils.js";
 import {
   type ActionLevelValidationDefinition,
+  type ActionParameter,
   type ActionParameterAllowedValues,
   type ActionParameterType,
   type ActionParameterTypePrimitive,
   type ActionType,
   type ActionTypeDefinition,
   type ActionValidationRule,
+  type ConditionDefinition,
   type InterfaceType,
   type ObjectPropertyType,
   type ObjectType,
@@ -421,6 +423,7 @@ export function defineAction(actionDef: ActionTypeDefinition): ActionType {
     apiName: apiName,
     __type: OntologyEntityTypeEnum.ACTION_TYPE,
   } as ActionType;
+  validateActionValidation(fullAction);
   updateOntology(fullAction);
   return fullAction;
 }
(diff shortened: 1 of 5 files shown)


packages/maker/src/api/defineAction.ts
```diff
@@ -670,3 +673,70 @@ function convertValidationRule(
...
+      const overrideParamId = condition.parameterId;
+      invariant(
+        parameters?.some(p => p.id === overrideParamId),
+        `Parameter condition on ${currentParameterId} is referencing unknown parameter ${overrideParamId}`,
+      );
+      invariant(
+        overrideParamId !== currentParameterId,
+        `Parameter condition on ${currentParameterId} is referencing itself`,
+      );
+      invariant(
```
ethanai22, line 709: Could also validate this through the `parameterOrdering` field, but I think it feels more intuitive for users to manage the ordering in the parameters list itself

packages/maker/src/api/defineAction.ts
```diff
@@ -670,3 +673,70 @@ function convertValidationRule(
...
+        seenParameterIds.has(overrideParamId),
+        `Parameter condition on ${currentParameterId} is referencing later parameter ${overrideParamId}`,
+      );
```
davidpaquin, line 712: Is this actually an error? Like you need to manage the dependencies here?

packages/maker/src/api/defineAction.ts
```diff
@@ -670,3 +673,70 @@ function convertValidationRule(
...
+            c,
+            currentParameterId,
+            seenParameterIds,
+            parameters,
+          )
+        );
+      }
+      break;
+    default:
+      break;
```
davidpaquin, line 740: we should probably throw an error if the condition type is something we don't account for

ethanai22: this would be caught at compile time right? we wouldn't want to throw on other valid comparison types that don't require validation in this method

davidpaquin: If there are only a couple other types, then I'd explicitly list them and `break`, but if there's a bunch, then this is probably fine. Because we don't support TS files for ontology definitions yet, I just want to be careful about validating unknown inputs

packages/maker/src/api/defineAction.ts
davidpaquin: oh weird

davidpaquin (approved):

PR #1719, opened 2025-07-14
Support Ontology interface edits

tmisirpash:
This PR adds support for Ontology interface edits to `@osdk/functions`.

To edit an interface, a user needs to add the `Edits.Interface` type to the edits scope:

```typescript
type TestEditScope =
  | Edits.Object<Task>
  | Edits.Interface<FooInterface>;
```

## Create
```typescript
import { MyInterface } from "@ontology/sdk";

batch.create(MyInterface, { $objectType: "myObjectType", interfaceProperty: 3 });
```

There are some caveats to point out with interface creates:

- We need to provide the interface as the first argument to provide strong typing for filling in the values of the non-nullable interface properties. We also need to declare the specific object type (which won't necessarily be known at compile time) as a required `$objectType` field.
- It isn't guaranteed that interface properties currently map to the primary key properties of the implementing object type. 
- If using `create` with an interface does not assign a value to the underlying object's primary key, the action will fail at runtime. This limitation exists for non-function-backed interface actions.

## Update
```typescript
batch.update(obj, { interfaceProperty: 3 });
```

Note that it is possible to attempt to edit the underlying primary key property if that's what the interface property is mapped to, which will cause a runtime failure. This limitation exists in non-function-backed interface actions.

## Delete
```typescript
batch.delete(obj);
```

### DIFF
--- a/etc/functions.report.api.md
+++ b/etc/functions.report.api.md
@@ -75,12 +75,12 @@ export interface EditBatch<X extends AnyEdit = never> {
 
 // @public (undocumented)
 export namespace Edits {
-    	// Warning: (ae-forgotten-export) The symbol "CreateInterface" needs to be exported by the entry point index.d.ts
-    // Warning: (ae-forgotten-export) The symbol "UpdateInterface" needs to be exported by the entry point index.d.ts
-    // Warning: (ae-forgotten-export) The symbol "DeleteInterface" needs to be exported by the entry point index.d.ts
+    	// Warning: (ae-forgotten-export) The symbol "CreateObjectForInterface" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "UpdateObjectForInterface" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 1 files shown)


packages/functions/src/edits/createEditBatch.ts
```diff
@@ -93,25 +97,70 @@ class InMemoryEditBatch<X extends AnyEdit = never> implements EditBatch<X> {
...
+    objOrInterfaceType: T,
+    propertiesOrObjectType: any,
+    maybeProperties?: any,
   ): void {
-    this.edits.push({
-      type: "createObject",
-      obj,
-      properties,
-    } as unknown as X);
+    if (arguments.length === 3) {
```
tmisirpash, line 114: @RTiwary I think instead of checking the number of arguments, we can check whether `objOrInterfaceType.type === "interface"`

RTiwary: changed

ssanjay1 (approved):

PR #1732, opened 2025-07-15
Fix attachment type for creating objects

ssanjay1:
We want attachments to also take a string. Made a new mapping type so we can have whatever separate mappings we want for the internal edits workflows


marshallavail (approved):
Works for me

marshallavail (approved):

PR #1745, opened 2025-07-17
Typeclasses and list validation for object actions

jscaii:


packages/maker/src/api/defineAction.ts
```diff
@@ -168,8 +168,11 @@ export function defineCreateObjectAction(
         id: prop.apiName,
         displayName: prop.displayName,
         type: extractActionParameterType(prop),
+        typeClasses: prop.typeClasses ?? [],
         validation: {
-          required: true,
+          required: (prop.array ?? false)
+            ? { listLength: {} }
+            : true,
```
davidpaquin, line 175: I think we'll have to use `prop.nullability` to get the right info here

PR #1772, opened 2025-07-25
Writeable Client

nihalbhatnagar:
In progress while awaiting Gateway shape

### DIFF
--- a/packages/functions/src/transactions/WriteableClient.ts
+++ b/packages/functions/src/transactions/WriteableClient.ts
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


ssanjay1:
To fix ATTW, try adding a file like this to the root of functions to export from .cjs directly https://github.com/palantir/osdk-ts/blob/main/packages/client/unstable-do-not-use.d.ts

packages/functions/src/transactions/WriteableClient.ts
```diff
@@ -0,0 +1,71 @@
...
+  UpdatableObjectOrInterfaceLocators,
+} from "../edits/EditBatch.js";
+import type { AnyEdit } from "../edits/types.js";
+
+export interface WriteableClient<X extends AnyEdit = never>
+  extends Client, WriteMethods<X>
+{
+}
+
+export interface WriteMethods<X extends AnyEdit = never> {
```
nihalbhatnagar, line 39: I'm extending AnyEdit, which means that someone could pass in an interface scope to this. However, my create/modify/delete is scoped to object type because the v0 will not support interfaces OOTB. You will see a compile error when passing in an interface object. I will track scoping this down from AnyEdit as a flup

PR #1775, opened 2025-07-28
feat: Add link observation support to Observable Client and React hooks

ericanderson:
## Summary

Adds support for observing object links (relationships) in the Observable Client. Links now automatically update when related objects change.

### Observable Client
```typescript
observableClient.observeLinks(office, "occupants", 
  { where: { department: "Engineering" }, pageSize: 20 },
  observer
)
```

### React Hooks
```typescript
const { links, isLoading, fetchMore } = useLinks(office, "occupants", {
  where: { department: "Engineering" },
  orderBy: { name: "asc" },
  pageSize: 20
})
```

Currently exported from `@osdk/react/experimental`.

## Implementation

- **SpecificLinkQuery**: New query type for link-specific operations
- **BaseCollectionQuery**: Shared logic between lists and links
- **Reference-based storage**: Links store object references instead of duplicating data
- **Type-safe cache keys**: Replaced generic `CacheKey` with `KnownCacheKey` union type

## Known Limitations

Optimistic updates don't work for many-to-many link modifications. When you modify a many-to-many link, the UI waits for the server response before updating.

## Testing

- Added tests for cross-type links and foreign key updates
- Cache invalidation and selective revalidation tests
- Example app with Office-Employee relationships in `packages/e2e.sandbox.peopleapp`

## Notes

- No breaking changes - existing `observeList()` and `useOsdkObjects()` continue working
- Includes refactoring to extract common collection query patterns
- Added CLAUDE.md documentation files


KuberSethi (commented):
Looks mostly good! Main questions are around typing

packages/monorepo.cspell/dict.npm-packages.txt
```diff
@@ -8,7 +8,10 @@ immer
 jiti
 luxon
 manypkg
+maplibre
```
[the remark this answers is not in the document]
ericanderson: Was used in my sandbox

packages/client/src/observable/internal/links/SpecificLinkQuery.ts
```diff
@@ -221,6 +216,41 @@ export class SpecificLinkQuery extends BaseListQuery<
...
+
+    // For case 1 - direct source object type match
+    if (sourceObjectType === objectType) {
+      changes?.modified.add(this.cacheKey);
+      return this.revalidate(true);
+    } else {
+      // For case 2 - check if the link's target type matches the invalidated type
+      // We need to use the ontology provider to get the link metadata
+      // Since this is async, we'll collect all the metadata check promises
+      const linkName = this.cacheKey.otherKeys[2];
```
KuberSethi, line 238: Maybe I'm just new to the shape of this data but it's a little odd to be accessing `otherKeys[2]` and `otherKeys[1]` below in `ListQuery.ts`. Thoughts on some utils for accessing the values here to make it more readable as to whats being accessed? 

Also this just feels a little brittle but I'm less opinionated on how that might be improved barring retyping things.

packages/client/src/observable/internal/ListQuery.ts
```diff
@@ -381,7 +381,7 @@ export abstract class BaseListQuery<
     this.pendingFetch = this.fetchPageAndUpdate(
       "loaded",
       this.abortController?.signal,
-    ).finally(() => {
+    ).then(() => void 0).finally(() => {
```
KuberSethi, line 384: Is this commonly used in this repo? I'm more familiar with `.then(() => {})` for no-ops (assuming that's the goal here?)

packages/client/src/observable/ObservableClient/ObserveLink.ts
```diff
@@ -38,7 +39,7 @@ export namespace ObserveLink {
     where?: WhereClause<Q>;
     pageSize?: number;
     orderBy?: OrderBy<Q>;
-    invalidationMode?: "in-place" | "wait" | "reset";
+    invalidationMode?: InvalidationMode;
     expectedLength?: number;
     // streamUpdates?: boolean;
```
KuberSethi, line 44: nit: likewise if we want to remove this and the other instances (not sure of the history of this)

packages/client/src/observable/internal/ListQuery.ts
ericanderson: this is needed to type this as Promise<void>

packages/client/src/observable/internal/links/SpecificLinkQuery.ts
KuberSethi: That's helpful context, thanks. I think I was imagining something as simple as a wrapper helper function to just return `this.cacheKey.otherKeys[2]` as a means to improve readability since it's not immediately obvious what the shape/structure of `otherKeys` is without context like this. 

I think this is fine though, thanks for explaining!

packages/client/src/observable/internal/ListQuery.ts
KuberSethi: Yeah, was just confused at choice of `void 0` vs `.then(() => {})` which I've seen more of, but doesn't really matter though!

PR #1794, opened 2025-08-01
add interface support in queries

ssanjay1:
Adding support for input/output interfaces in queries

### DIFF
--- a/packages/api/src/queries/Queries.ts
+++ b/packages/api/src/queries/Queries.ts
@@ -19,7 +19,9 @@ import type {
   DataValueWireToClient,
 } from "../mapping/DataValueMapping.js";
 import type { ObjectSet } from "../objectSet/ObjectSet.js";
-import type { ObjectTypeDefinition } from "../ontology/ObjectTypeDefinition.js";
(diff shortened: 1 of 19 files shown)


packages/api/src/queries/Queries.ts
```diff
@@ -49,14 +51,33 @@ export namespace QueryParam {
...
-  export type ObjectType<T extends ObjectTypeDefinition> =
+  export type ObjectType<T extends ObjectOrInterfaceDefinition> =
     | ObjectIdentifiers<T>
     | OsdkObjectPrimaryKeyType<T>;
 
+  /**
+   * Helper type to convert action definition parameter interface types to typescript types
+   */
+  export type InterfaceType<T extends InterfaceDefinition> = {
+    $objectType: CompileTimeMetadata<T> extends { implementedBy: infer U }
```
ssanjay1, line 62: Either we let users pass in an osdk.instance (second part of type) or we let them pass in a concrete type of objectType and PK where we try and give them helper hints on objectType (first part of type)

nihalbhatnagar (approved):

PR #1795, opened 2025-08-02
Replace AST parsing with Vite module evaluation for widget config extraction

rosscourt:
The previous implementation used manual AST parsing with `escodegen` to extract widget configs from `.config.ts` files, which applied a naive transformation to convert the AST into JSON. We've already seen issues with this approach in simple cases, but it also prevents the use of executable JS/TS within the config source files.

Vite provides the ability to evaluate modules through its SSR functionality via `ViteDevServer.ssrLoadModule` (https://vitejs.dev/guide/ssr.html). This method can evaluate TypeScript/JavaScript modules, giving us access to the exported config object without the need for manual extraction logic.

I tested this against some example widget set builds by linking the package locally. Tested build and dev commands with more complex config source files, e.g. importing the widget name string from another file, and against events which update multiple parameters.

-----

#### Dev mode plugin

There is no longer the need for AST parsing within the `transform` hook. Instead the `getWidgetIdOverrideMap` function now calls `extractWidgetConfig` directly using the existing dev server instance.

#### Build mode plugin

During the `generateBundle` hook the build mode plugin now creates a Vite server instance for evaluating widget config modules. Instead of storing parsed configs during the `moduleParsed` hook, they are now extracted on-demand within `getWidgetBuildOutputs`, which calls `extractWidgetConfig` with the shared server instance. 

#### Shared extraction logic

The `extractWidgetConfig` function uses `ViteDevServer.ssrLoadModule` to load and evaluate config modules. This approach handles JS/TS syntax correctly, resolves imports, and evaluates any dynamic logic within the config module. Error handling associated with reading the module is contained here.

-----

### Why use module evaluation?

An alternative might be to simply require these configurations to be written in JSON, however I believe that the evaluation approach will be helpful in the future when we build support for object set/instance parameter types.

For example a widget config operating on a given object type could look like:

```typescript
import { MyObjectType } from "@custom-widget/sdk";
import { defineConfig } from "@osdk/widget.client";

export default defineConfig({
  // ...
  parameters: {
    myObjectSet: {
      displayName: "Object Set of MyObjectType",
      type: "object-set",
      objectType: MyObjectType,
    },
    myObjectInstance: {
      displayName: "Object Instance of MyObjectType",
      type: "object-instance",
      objectType: MyObjectType,
    },
  },
});
```

An approach like this would allow us to extract the RID of `MyObjectType` when building the widget set manifest, without requiring the user to manually work with RIDs. This object type RID can then be used in producing the search index, powering preview config UIs, and even runtime validation of parameter data.

Using `MyObjectType` should also allow us to infer the correct types for the parameter values as they are available for use within the widget code. i.e `myObjectSet` should be typed as `MyObjectType.ObjectSet` and `myObjectInstance` should be `MyObjectType.OsdkInstance`.


examples/example-widget-minimal-react-sdk-2.x/src/main.config.ts
```diff
@@ -20,5 +20,9 @@ export default defineConfig({
       displayName: "Set counter value",
       parameterUpdateIds: ["counterValue"],
     },
+    updateAllParameters: {
+      displayName: "Update All Parameters",
+      parameterUpdateIds: ["greetingName", "counterValue"],
+    },
```
rosscourt, line 26: Added this so that we have a test case with multiple `parameterUpdateIds`

packages/widget.vite-plugin/src/build-plugin/FoundryWidgetBuildPlugin.ts
```diff
@@ -114,3 +101,15 @@ async function computeWidgetSetVersion(
       ?? { "type": "package-json" },
   );
 }
+
+function writeManifest(
```
rosscourt, line 105: Just pulled this out to keep the `writeBundle` hook a bit more readable

tzyl (approved):
Nice, what coverage do we have that this works properly? We build the generated examples packages in this repo which means we're not throwing when running the build plugin at least but we don't know the manifest was created correctly or the dev plugin works. I wonder if we can do more integration style tests and run with a real vite dev server rather than a mocked one

tzyl (approved):

PR #1818, opened 2025-08-12
[2/2] Add nearestNeighbor order by relevance support 

mvanschellebeeck:
Follow up to: https://github.com/palantir/osdk-ts/pull/1666

### Summary
Adds the ability to order by relevance for nearest neighbors queries. When `$orderBy: "relevance"` is specified, objects are returned with an additional property: `$score`
```typescript
const { data } = await client(MyObjectType)
  .nearestNeighbors(
    "coffee",
    10,
    "myEmbeddingProperty",
  )
  .fetchPage({ $orderBy: "relevance" });

for (const obj of data) {
  console.log(obj.id, obj.$score);
} 
```

### Implementation

`OrderBy` is extended to `OrderByOptions` where `"relevance"` is introduced:
```typescript

  export type OrderByOptions<L extends string> =
  | {
    [K in L]?: "asc" | "desc";
  }
  | "relevance";

  export interface OrderBy<
    ORDER_BY_OPTIONS extends OrderByOptions<L>,
    L extends string = never,
  > {
    $orderBy?: ORDER_BY_OPTIONS
  }
```
`MaybeScore` wraps the `Osdk.Instance` type when `$orderBy: "relevance"` is specified in any of `asyncIter`, `fetchPage` or `fetchPageWithErrors`:

```typescript
export type MaybeScore<
	T extends Osdk.Instance<any>,
	ORDER_BY_OPTIONS extends ObjectSetArgs.OrderByOptions<any>
> = ORDER_BY_OPTIONS extends "relevance" ? T & {
    	$score: number
} : T;

// usage
MaybeScore<
  Osdk.Instance<
    Q,
    ....
  >,
  ORDER_BY_OPTIONS
>
```


packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -355,6 +357,13 @@ function fixObjectPropertiesInPlace(
       delete obj.__rid;
     }
 
+    if (obj.__score) {
+      if (includeScore) {
+        obj.$score = obj.__score;
+      }
+      delete obj.__score;
```
nihalbhatnagar, line 364: So this case is basically I added the nearest neighbor object set but I didnt order by? Is there ever a case where I just want the score and don't order by?

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -1349,5 +1350,104 @@ describe("ObjectSet", () => {
...
       );
     });
+
+    it("appropriately types fetchPage ordered by relevance", async () => {
+      const nearestNeighborsObjectSet = await fauxObjectSet.nearestNeighbors(
+        "textQuery",
+        3,
+        "skillSetEmbedding",
+      ).fetchPage({ $orderBy: "relevance" });
+
```
nihalbhatnagar, line 1360: Tried getting this test working but no luck 
```
const knnQuery = knn.fetchPage;
      
      type a = NonNullable<Parameters<typeof knnQuery>[0]>["$orderBy"];
      expectTypeOf<undefined | "relevance" | {}>()
        .toEqualTypeOf<a>();
```

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -57,7 +63,8 @@ export namespace ObjectSetArgs {
     K extends PropertyKeys<Q> = never,
     T extends boolean = false,
     RDP_KEYS extends string = never,
-  > extends AsyncIter<Q, K, T, RDP_KEYS> {
+    ORDER_BY_OPTIONS extends ObjectSetArgs.OrderByOptions<K> = {},
```
nihalbhatnagar, line 66: `= never` maybe to match the rest of the cases where we do this?

nihalbhatnagar: It seems to transpile locally

packages/client/src/object/convertWireToOsdkObjects.ts
mvanschellebeeck: ah right, we don't need it. Updated in https://github.com/palantir/osdk-ts/pull/1818/commits/ad6482abe8428f7cf9b12193892756e4bd9632a2

PR #1822, opened 2025-08-13
Include OSDK usage example for widgets

amishshah:
Include an example of how to create the osdk client in the non-minimal widget template

### DIFF
--- a/packages/create-widget.template.react.v2/templates/src/client.ts.hbs
+++ b/packages/create-widget.template.react.v2/templates/src/client.ts.hbs
@@ -2,7 +2,7 @@ import { $ontologyRid } from "{{osdkPackage}}";
 import { createClient } from "@osdk/client";
 
 export const client = createClient(
-    window.location.origin,
-    $ontologyRid,
-    () => Promise.resolve("widgets-auth"),
+  window.location.origin,
+  $ontologyRid,
+  () => Promise.resolve("widgets-auth"),
 );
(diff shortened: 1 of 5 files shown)


packages/create-widget.template.react.v2/templates/src/client.ts.hbs
```diff
@@ -0,0 +1,8 @@
+import { $ontologyRid } from "{{osdkPackage}}";
+import { createClient } from "@osdk/client";
+
+export const client = createClient(
+    window.location.origin,
+    $ontologyRid,
+    () => Promise.resolve("widgets-auth"),
+);
```
tzyl, line 8: nit: this is kind of hard to keep up without enforcing it as I don't think the template includes a formatter out of the box but we can keep the indentation consistent with the current template files:

https://github.com/palantir/osdk-ts/blob/5553fadd006a3364535b6d0a0add2cd07d1999ef/packages/create-widget.template.react.v2/templates/src/context.ts

```suggestion
export const client = createClient(
  window.location.origin,
  $ontologyRid,
  () => Promise.resolve("widgets-auth"),
);
```

packages/create-widget.template.react.v2/templates/src/client.ts.hbs
tzyl, line 1: Not relevant for this PR but we should remember to update the internal manifest file to include this new file as it will need replacing

tzyl (approved):

PR #1823, opened 2025-08-13
OAC actions support custom parameters

ethanai22:
We currently only support action parameters that are mapped to object properties. This PR allows you to add custom parameters to the `parameterConfiguration` map. This is mostly useful for adding objectReference parameters where you can then use the objectReference's properties to add default values or conditional overrides on other parts of the action (see the long test I added).

### DIFF
--- a/packages/maker/src/api/overall.test.ts
+++ b/packages/maker/src/api/overall.test.ts
@@ -8833,6 +8833,10 @@ describe("Ontology Defining", () => {
     });
 
     it("OAC defined object references as parameters are properly defined", () => {
+      // This test defines an action that creates an OAC-defined employee object type and also adds an additional parameter
+      // for an objectReference to a manager object type that is also defined in OAC. The action shows how you can use this
+      // objectReference parameter elsewhere in the action - in this case we are prefilling the managedBy property to be the
+      // id of the selected manager object type.
       const employeeObject = defineObject({
         apiName: "employee",
         displayName: "Employee",
@@ -8865,7 +8869,7 @@ describe("Ontology Defining", () => {
             customParameterType: {
               type: "objectReference",
               objectReference: {
-                objectTypeId: "manager",
+                objectTypeId: managerObject.apiName,
               },
             },
           },


packages/maker/src/api/defineAction.ts
```diff
@@ -832,7 +836,67 @@ function referencedParameterIds(
   return parameterIds;
 }
 
-function extractAllowedValuesFromType(
+function extractAllowedValuesFromActionParameterType(
```
ethanai22, line 839: At some point this should be moved to defineOntology. I think defineAction should just be returning an ActionParameterType and the ActionParameterAllowedValues that the user possibly defined (not the one we extract in this file), and then we can do the full conversion in defineOntology. In general the contract between what we return in defineAction and what we expect in convertToWireBlockData should be cleaner

ethanai22: ^ leaving this task for a more broad refactor PR

davidpaquin: I've been thinking about if it's possible to not need any of the conversions in`defineOntology` at all -- like if we just have each `defineFoo` function return exactly the object that gets passed into the IR. I agree with your comment though

packages/maker/src/api/overall.test.ts
```diff
@@ -8778,6 +8832,543 @@ describe("Ontology Defining", () => {
...
+
+      defineCreateObjectAction({
+        objectType: employeeObject,
+        parameterOrdering: ["myManager", "id", "managedBy"],
+        parameterConfiguration: {
+          "myManager": {
+            customParameterType: {
+              type: "objectReference",
+              objectReference: {
+                objectTypeId: "manager",
```
davidpaquin, line 8868: Does this have to be the api name of the object type? I forget how `objectTypeId` gets resolved. If so, can we make this line `objectTypeId: managerObject.apiName` just so that's more clear

packages/maker/src/api/overall.test.ts
```diff
@@ -8778,6 +8832,543 @@ describe("Ontology Defining", () => {
       );
     });
 
+    it("OAC defined object references as parameters are properly defined", () => {
```
davidpaquin, line 8835: Can you write a comment about what this test actually produces in Foundry?

PR #1845, opened 2025-08-20
Perf fix

ssanjay1:
When doing simple object fetches, we should be able to parallelize network requests to load object definition beforehand because we know the object type the user is asking for up front.

Since we cache metadata results on the client, if we fire it off first and it finishes, it should be ready to use in the cache.

### DIFF
--- a/packages/client/src/createClient.test.ts
+++ b/packages/client/src/createClient.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { Task } from "@osdk/client.test.ontology";
+import { BarInterface } from "@osdk/client.test.ontology";
 import * as SharedClientContext from "@osdk/shared.client.impl";
 import type { MockedFunction } from "vitest";
 import { beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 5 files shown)


packages/client/src/object/fetchPage.ts
```diff
@@ -321,6 +321,7 @@ export async function fetchObjectPage<
   objectSet: ObjectSet,
   useSnapshot: boolean = false,
 ): Promise<FetchPageResult<Q, L, R, S, T>> {
+  void client.ontologyProvider.getObjectDefinition(objectType.apiName);
```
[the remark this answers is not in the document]
nihalbhatnagar: Agreed

packages/client/src/createClient.test.ts
```diff
@@ -71,12 +71,12 @@ describe(createClient, () => {
     }
 
     it("works for objects", async () => {
-      await client(Task).fetchPage();
+      await client(BarInterface).fetchPage();
```
ssanjay1, line 74: The purpose of this test was to just check the user agent is passed correctly, this doesn't change that. Just changed it to interface so I don't need to mock and handle the whole metadata call we fire and forget now in the object case. 
We can't do it for interface loads because we don't know what the underlying object type will be

packages/client/src/ontology/StandardOntologyProvider.test.ts
```diff
@@ -49,8 +49,8 @@ describe(createStandardOntologyProviderFactory, () => {
 
     // first load should lookup employee and its link types
     expect(loads).toEqual([
-      "/api/v2/ontologies/ri.ontology.main.ontology.698267cc-6b48-4d98-beff-29beb24e9361/objectSets/loadObjects",
       "/api/v2/ontologies/ri.ontology.main.ontology.698267cc-6b48-4d98-beff-29beb24e9361/objectTypes/Employee/fullMetadata",
```
ssanjay1, line 52: We now fire off the fullMetadata call first technically (its parallel though since not awaited)

ericanderson (approved):

PR #1846, opened 2025-08-20
Existing object types as action parameters in OAC (maker-experimental)

ethanai22:
The goal is to support existing OTs as action parameters in OAC. I created an importOntologyEntity wrapper that takes in the minimal information needed to produce input shapes. We need a more complete definition to actually call importOntologyEntity, so I produced that with some inferred fields (primary key, title key, etc.). The actual type returned by this wrapper only contains fields that are safe to use and not inferred (it's just a OntologyIrImportedObjectType).
The test shows how this feature is supposed to be used.

This is pretty hacky overall, but that was the expectation going into this FR. We've made a maker-experimental package to reflect this

### DIFF
--- a/packages/maker-experimental/src/api/types.ts
+++ b/packages/maker-experimental/src/api/types.ts
@@ -17,14 +17,14 @@
 import type { PropertyTypeType } from "@osdk/maker";
 
 // basically a user friendly OntologyIrImportedActionType wrapper
-export type importObjectDefinition = {
+export type ImportObjectDefinition = {
   apiName: string;
-  properties: Record<string, importPropertyTypeDefinition>;
+  properties: Record<string, ImportPropertyTypeDefinition>;
   displayName?: string;
   description?: string;
 };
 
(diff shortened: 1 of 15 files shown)


packages/maker-experimental/src/api/types.ts
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { PropertyTypeType } from "@osdk/maker";
+
+// basically a user friendly OntologyIrImportedActionType wrapper
+export type importObjectDefinition = {
```
aep000, line 20: nit: Upper case ImportObjectDefinition

aep000 (commented):
One nit but looks good

PR #1856, opened 2025-08-22
Add RDP Expressions Docs

nihalbhatnagar:


packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -2035,3 +2035,44 @@ versions:
...
+        derivedPropertyNumericExpression:
+          - template: |-
+              import { {{objectType}} } from "{{{packageName}}}";
+              // Edit this import if your client location differs
+              import { client } from "./client";
+
+              const {{objectType}}WithExpression = await client({{objectType}})
+                  .withProperties({
+                    "newPropertyName": (baseObjectSet) =>
+                        {{^isUnary}}
```
bryantpark04, line 2049: nit: maybe switch the order of ^isUnary and #isUnary so it's consistent with #isExtractPart if there isn't a reason you put it first?

PR #1872, opened 2025-08-26
Add Principal Type

p-kona:
Add a principal type exported from @osdk/functions to allow for runtime differentiation between userid and groupid types (string types imported from PSDK)

### DIFF
--- a/packages/functions/src/Principal.ts
+++ b/packages/functions/src/Principal.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


packages/functions/src/Principal.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type Principal = {
+  type: "user";
```
ssanjay1, line 18: Will there ever be additional fields in these types, just wondering? Or is the only differentiation here the type

Issue #1536, opened 2025-05-22
Orderby,select args for fully qualified api names break


ssanjay1:
We do not remap fully qualified api names under the hood for orderby and select arguments.

PR #1878, opened 2025-08-27
fix(client): remap simple api names for orderby and select arguments

KuberSethi:
Fixes https://github.com/palantir/osdk-ts/issues/1536

There's currently a mismatch between what's expected for orderby and select args and what we pass, namely that we pass in simple API names instead of the expected fully qualified names (includes the namespace). This PR fixes this by using a similar approach to `modernToLegacyWhereClause` to remap names by extracting the namespace and combining names.


packages/client/src/object/fetchPage.ts
```diff
@@ -28,15 +28,15 @@ import type {
...
-  LoadObjectSetRequestV2,
-  LoadObjectSetV2MultipleObjectTypesRequest,
   ObjectSet,
   OntologyObjectV2,
   SearchJsonQueryV2,
   SearchObjectsForInterfaceRequest,
   SearchOrderByV2,
 } from "@osdk/foundry.ontologies";
 import * as OntologiesV2 from "@osdk/foundry.ontologies";
+import { metadataCacheClient } from "../__unstable/ConjureSupport.js";
```
KuberSethi, line 38: let me know if we don't use unstable dependencies in mainline code, the property mappings already existed there so figured to start I'd use it.

ssanjay1 (approved):

PR #1886, opened 2025-08-28
Fix casting support for object -> interface

ssanjay1:
We had a bug when using `$as` in our types.

Let's say we had an interface called: `apiName: "com.my.obscure.namespace.FooBarInterface";`
And it had properties:
```
 id: string;
 "com.my.even.more.obscure.namespace.originDate": string | undefined;
 fooInterface: number;
```

Because of the way OSDK strips namespaces on properties if they match the namespace of the interface, `id` and `fooInterface` are technically `com.my.obscure.namespace.id` and `com.my.obscure.namespace.fooInterface`. This is fine for ergonomics, but becomes tricky with casting.

For casting, we rely on the inverse interface map OSS gives us which maps interface props to object props, and these props all have things fully qualified. E.g.:
```
inverseInterfaceMap: {
        "com.my.obscure.namespace.FooBarInterface": {
          "foo": "com.my.obscure.namespace.fooInterface";
          "name": "com.my.obscure.namespace.id";
          "birthday": "com.my.even.more.obscure.namespace.originDate";
        };
      };
```

Before this fix, our `$as` type would not strip the namespace so we'd think there was a type mismatch because we get back `"com.my.obscure.namespace.id"` when we really need `id` . 

We now fix this by stripping the namespace from the key of the map if it matches the namespace of the overall interface!

### DIFF
--- a/.changeset/yellow-moments-open.md
+++ b/.changeset/yellow-moments-open.md
@@ -1,0 +1,5 @@
+---
+"@osdk/api": patch
+---
+
+Fix casting types with fully qualified property names.
--- a/packages/api/src/OsdkObjectFrom.test.ts
+++ b/packages/api/src/OsdkObjectFrom.test.ts
@@ -472,7 +472,7 @@ describe("ExtractOptions", () => {
   });
 
   describe("Interface casting works as intended", () => {
-    it("mapping as works", async () => {
+    it("mapping as with fqn property names works", async () => {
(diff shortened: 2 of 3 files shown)


KuberSethi:
Change makes sense, out of curiosity how was this issue surfaced?

nihalbhatnagar (approved):

PR #1892, opened 2025-08-29
fix: Rename ObserveLink namespace to ObserveLinks for consistency

ericanderson:
## Summary
- Rename ObserveLink namespace to ObserveLinks in ObserveLink.ts for naming consistency
- Update type imports and references across Observable Client implementation
- Update useLinks hook to use ObserveLinks types and resolvedList property
- Improve type safety in makeExternalStore with proper Snapshot typing

## Test plan
- [x] Verify existing tests still pass
- [x] Check TypeScript compilation
- [x] Confirm Observable Client functionality remains unchanged


packages/react/src/new/makeExternalStore.ts
```diff
@@ -31,7 +31,7 @@ export function makeExternalStore<X>(
   subscribe: (notifyUpdate: () => void) => () => void;
   getSnapShot: () => Snapshot<X>;
 } {
-  let lastResult: Snapshot<X>;
+  let lastResult: Snapshot<X> | undefined;
```
KuberSethi, line 34: is this needed? don't have full context but the idea of a `result` being potentially undefined is a typing smell

packages/react/src/new/useLinks.ts
```diff
@@ -143,19 +143,13 @@ export function useLinks<
...
-    hasMore?: boolean;
-  } | undefined;
+  const payload = React.useSyncExternalStore(
+    subscribe,
+    getSnapShot,
+  );
 
   return {
-    links: payload?.resolvedLinks,
+    links: payload?.resolvedList,
```
KuberSethi, line 152: is this intentional?

packages/react/src/new/makeExternalStore.ts
```diff
@@ -40,7 +40,7 @@ export function makeExternalStore<X>(
   function subscribe(notifyUpdate: () => void) {
     const obs = createObservation({
       next: (payload) => {
-        lastResult = payload;
+        lastResult = payload as Snapshot<X>;
```
KuberSethi, line 43: Do we need this? Tested removing it and types seem okay?

packages/react/src/new/makeExternalStore.ts
ericanderson: it doesn't need it because its redundant (the `Snapshot` is `| undefined). In the initial case though it can be undefined.

packages/react/src/new/makeExternalStore.ts
ericanderson: Doesnt work on my machine without this.

KuberSethi (dismissed):
LGTM, might need to restamp after you pull in latest + resolve merge conflicts

PR #1896, opened 2025-08-29
add new useAbortController hook

KuberSethi:
Adds a new `useAbortController` to centralize abort controller lifecycle code. As we expand the number of exposed hooks requiring abort controller usage (I'm working on `useObjectSet` for instance) this will prove helpful and cut down on repeated code + complexity


ericanderson:
Is there anything like this out in the wild? Seems like its not very osdk related.

PR #1932, opened 2025-09-08
all 'safe' upgrades

ericanderson:


examples-extra/docs_example/package.json
```diff
@@ -17,33 +17,33 @@
...
-    "@vitejs/plugin-react": "^4.2.1",
-    "autoprefixer": "^10.4.19",
-    "eslint": "^9.19.0",
+    "@types/react": "^18.3.24",
+    "@types/react-dom": "^18.3.7",
+    "@typescript-eslint/eslint-plugin": "^8.43.0",
+    "@typescript-eslint/parser": "^8.43.0",
+    "@vitejs/plugin-react": "^4.7.0",
+    "autoprefixer": "^10.4.21",
+    "eslint": "^9.35.0",
```
KuberSethi, line 39: Just a thought but if these linter dep bumps are resulting in new / updated rules being brought in that cause errors it might be easier to compile a list of what's been added/changed and ignoring them for now to unblock this PR?

packages/faux/src/handlers/util/handleOpenApiCall.ts
```diff
@@ -164,7 +163,7 @@ export function handleOpenApiCall<
             );
 
             if (result instanceof Response) {
-              return new HttpResponse(result.body) as StrictResponse<
+              return new HttpResponse(result.body) as HttpResponse<
```
KuberSethi, line 166: Is this change needed / will losing response type safety cause problems downstream?

KuberSethi (dismissed):
Guessing this might get invalidated as you fix linter errors but looks good!

examples-extra/docs_example/package.json
ericanderson: they made one rule catch a bit more than it used to. it wasnt too bad.

PR #1938, opened 2025-09-09
Fix applyQuery for marketplace installations

marshallavail:


packages/client/src/queries/applyQuery.ts
```diff
@@ -73,7 +73,7 @@ export async function applyQuery<
         )
         : {},
     },
-    { version: qd.version },
+    { version: query.isFixedVersion ? query.version : undefined },
```
marshallavail, line 76: Slight downside of this change is that there is a small race condition when `query.isFixedVersion` is false. Namely, if a new version is published between the call to `getQueryDefinition` above and the call to `execute` here.

In practice, I don't _think_ this should matter since everything should execute fine as long as the signatures are backward compatible (and if they aren't then we would have run into issues remapping the query parameters above).

ssanjay1: Could we instead just use `qd.apiName` as well so everything is remapped in our request?

ssanjay1: Defer to @nihalbhatnagar

nihalbhatnagar (approved):

PR #1950, opened 2025-09-10
More refactors and flups

ericanderson:


packages/client/src/observable/internal/Store.ts
```diff
@@ -18,99 +18,41 @@ import type {
...
-import type { Query } from "./Query.js";
-import { RefCounts } from "./RefCounts.js";
-import type { SimpleWhereClause } from "./SimpleWhereClause.js";
-import { tombstone } from "./tombstone.js";
+import { Queries } from "./Queries.js";
+import type { Subjects } from "./Subjects.js";
 import { WhereClauseCanonicalizer } from "./WhereClauseCanonicalizer.js";
 
-/*
-    Work still to do:
```
KuberSethi, line 55: looks a little out of date but guessing the later checkboxes here are still relevant?

packages/client/src/observable/internal/Layers.ts
```diff
@@ -0,0 +1,208 @@
...
+import { type Changes } from "./Changes.js";
+import { createInitEntry } from "./createInitEntry.js";
+import type { KnownCacheKey } from "./KnownCacheKey.js";
+import type { Entry } from "./Layer.js";
+import { Layer } from "./Layer.js";
+import type { OptimisticId } from "./OptimisticId.js";
+import { Subjects } from "./Subjects.js";
+import { tombstone } from "./tombstone.js";
+
+export class Layers {
```
KuberSethi, line 29: Class makes sense by itself and appears correct but it just feels a little odd to have a `Layers` class totally disjoint from where it's used yet having on opinions on how it should be used and in what context (caching structure) it should be used. 

Not sure of what might be better but it seems like this would be better if it was colocated with the `Store` - maybe `Store` itself can be broken up into a `BaseStore` and `LayeredStore` or something similar? Might be more immediately obvious that way, just a thought though.

packages/client/src/observable/internal/Store.ts
```diff
@@ -525,11 +229,7 @@ export class Store {
       await Promise.all(promises);
     } finally {
       if (process.env.NODE_ENV !== "production") {
-        // todo
-        this.logger?.child({ methodName: "maybeRevalidateQueries" }).debug(
-          "in finally",
-          DEBUG_ONLY__changesToString(changes),
-        );
+        logger?.debug("in finally", DEBUG_ONLY__changesToString(changes));
```
KuberSethi, line 232: hypernit: more useful log message

KuberSethi (commented):
Looks good, only real thought is on if `Layers` is the right level of abstraction. Not blocking since I think there's definitely multiple ways to tackle breaking up `Store` though.

KuberSethi (approved):

PR #1956, opened 2025-09-12
Add $primaryKey to object where clauses

bryantpark04:
One of two PRs that explore different approaches to resolving $primaryKey. This approach involves changing the WireObjectSets we pass through chained function calls to a Promise<WireObjectSet> and rearranging them as necessary. Promise<WireObjectSet>s are resolved at terminal commands like aggregate and fetch, reducing the need to color large parts of the call tree with async.

IntelliSense before: 
image

After:
image


packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -176,52 +176,56 @@ type Aggregatable<
...
-    Q,
-    V extends `${infer N}:${infer P}`
-      ? P extends CollectWithPropAggregations ? SimplePropertyDef.Make<
-          CompileTimeMetadata<Q>["properties"][N]["type"],
-          "nullable",
-          "array"
-        >
-      : P extends MinMaxWithPropAggregateOption ? SimplePropertyDef.Make<
-          CompileTimeMetadata<Q>["properties"][N]["type"],
+  ) => Promise<
```
nihalbhatnagar, line 179: This would be a break. We would probably instead need to wrap the entire derived property construction in another promise and resolve this at fetch page time

packages/client/src/derivedProperties/createWithPropertiesObjectSet.ts
```diff
@@ -125,7 +147,7 @@ export function createWithPropertiesObjectSet<
       }
       const wrappedObjectSet: DerivedPropertyDefinition = {
         type: "selection",
-        objectSet: objectSet,
+        objectSet: await objectSet,
```
nihalbhatnagar, line 150: Yeah so we shouldnt await here and instead have the entire definition be awaited at fetchPage time

packages/api/src/objectSet/ObjectSet.ts
```diff
@@ -549,7 +549,7 @@ interface Subscribe<
   >(
     listener: ObjectSetSubscription.Listener<Q, P>,
     opts?: ObjectSetSubscription.Options<Q, P>,
-  ) => { unsubscribe: () => void };
+  ) => Promise<{ unsubscribe: () => void }>;
```
nihalbhatnagar, line 552: This is also a break and I'm not sure why I grok why this needs to be a promise.

packages/api/src/aggregate/WhereClause.ts
```diff
@@ -189,11 +190,23 @@ export interface NotWhereClause<
 }
 
 export type PropertyWhereClause<T extends ObjectOrInterfaceDefinition> = {
-  [P in keyof CompileTimeMetadata<T>["properties"]]?: FilterFor<
+  [
```
nihalbhatnagar, line 193: I think its nice to have before/after pics of intellisense for these type changes generally

packages/api/src/derivedProperties/DerivedProperty.ts
bryantpark04: I think this would require us to roll our own internal DerivedPropertyDefinition type instead of using the generated version

bryantpark04: Fixed break

packages/api/src/aggregate/WhereClause.ts
bryantpark04: Added to the PR description!

packages/api/src/aggregate/WhereClause.ts
```diff
@@ -189,11 +190,23 @@ export interface NotWhereClause<
...
+  ]?: FilterFor<
     CompileTimeMetadata<T>["properties"][P]
   >;
 };
 
+type PropertyAsPrimaryKey<
+  T extends ObjectOrInterfaceDefinition,
+  P extends keyof CompileTimeMetadata<T>["properties"],
+> = T extends ObjectTypeDefinition
+  ? CompileTimeMetadata<T>["primaryKeyApiName"] extends P ? "$primaryKey"
```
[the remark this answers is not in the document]
bryantpark04: I finally got this working, but I'm not a huge fan of how the IntelliSense works. 
image

When you already have $primaryKey, it still suggests the actual primary key (employeeId in this case), and then the error it gives you when you have both is not too clear.
image

I also had to introduce a few `as any`s into the code because the types weren't inferring correctly.

packages/client/src/internal/conversions/modernToLegacyWhereClause.test.ts
```diff
@@ -913,34 +1074,80 @@ describe(modernToLegacyWhereClause, () => {
...
             }, {
-              integer: 10,
+              $primaryKey: 10,
             }],
           },
           objectTypeWithAllPropertyTypes,
-        )).toMatchInlineSnapshot(`
+          {
+            ...clientCtx,
+            ontologyProvider: {
```
bryantpark04, line 1125: poor man's mock lol

packages/client/src/observable/internal/list/ListQuery.ts
```diff
@@ -125,7 +125,7 @@ export class ListQuery extends BaseListQuery<
       type: this.#type,
       apiName: this.#apiName,
     } as ObjectTypeDefinition)
-      .where(this.#whereClause);
+      .where(this.#whereClause as any);
```
nihalbhatnagar, line 128: Why these as anys? Is it the same issue you described above?

packages/api/src/objectSet/ObjectSet.test.ts
```diff
@@ -104,6 +104,17 @@ describe("ObjectSet", () => {
...
+
+    it("has the right $primaryKey filters", async () => {
+      const filteredObject = await fauxObjectSet.where({
+        $primaryKey: { $lte: 400 },
+      }).fetchPage();
+      expectTypeOf<typeof filteredObject>().toEqualTypeOf<
+        PageResult<Osdk.Instance<EmployeeApiTest>>
+      >();
+      // @ts-expect-error
+      fauxObjectSet.where({ $primaryKey: "foo" });
```
nihalbhatnagar, line 116: Shouldn't this test $primaryKey and the actual primary key property fails?

packages/client/src/observable/internal/list/ListQuery.ts
bryantpark04: yeah it happened when I disallowed $primaryKey and primary key in the same where clause

Issue #1274, opened 2025-03-05
Allow `$primaryKey` as a valid key for Where Clauses


nihalbhatnagar:
`$primaryKey: { $in : ...`

if my PK property is pilot we probably want to prevent someone from going like:
.where({pilot: {$in: [..]}}, $primaryKey: {$in: [...]]]

PR #1959, opened 2025-09-12
Add $primaryKey to object where clauses by adding primaryKeyApiName to codegen

bryantpark04:
Rough alternative implementation of #1956 to address #1274. Modifies the codegen, so will require users to regenerate SDKs to use $primaryKey in where clauses.

Testing will be done after we decide on an approach.

### DIFF
--- a/examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Employee.ts
+++ b/examples-extra/docs_example/src/generatedNoCheck/ontology/objects/Employee.ts
@@ -68,7 +68,7 @@ export interface Employee extends $ObjectTypeDefinition {
   osdkMetadata: typeof $osdkMetadata;
   type: 'object';
   apiName: 'Employee';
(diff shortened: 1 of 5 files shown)


packages/api/src/ontology/ObjectTypeDefinition.ts
```diff
@@ -114,6 +114,7 @@ export interface ObjectTypeDefinition {
   type: "object";
   apiName: string;
   osdkMetadata?: OsdkMetadata;
+  primaryKeyApiName?: string;
   __DefinitionMetadata?:
```
nihalbhatnagar, line 118: I would think we should populate __DefinitionMetadata with this data. We have quite a few nucleus developers who manually create their osdk definitions `client(type, apiName)` and this might be confusing if present, even though it optional? On the other hand, always actually generating the __DefinitionMetadata may be ugly

PR #1962, opened 2025-09-15
add useObjectSet hook

KuberSethi:
Adds a new `useObjectSet` to unblock toolkit usage with object sets. Contains minimal perf. considerations considering this is a stopgap until `useOsdkObjects` reaches object set feature parity. If this is undesirable I can add some basic caching and request cancellation for this hook specifically

Also updates people app with nominal example usage.


packages/client/src/index.ts
```diff
@@ -71,3 +69,5 @@ export {
 } from "./util/datetimeConverters.js";
 
 export { createObjectSpecifierFromPrimaryKey } from "./util/objectSpecifierUtils.js";
+
+export { getWireObjectSet } from "./objectSet/createObjectSet.js";
```
ericanderson, line 73: Im not sure we want this in our public api.

packages/api/src/index.ts
```diff
@@ -46,14 +46,16 @@ export type {
   OsdkObjectLinksObject,
   SingleLinkAccessor,
 } from "./definitions/LinkDefinitions.js";
-export type { DerivedProperty } from "./derivedProperties/DerivedProperty.js";
+export type {
+  DerivedProperty,
+  DerivedPropertyCreator,
```
ericanderson, line 51: We have been trying to avoid making too many types on the root. Perhaps we should do this as `DervivedProperty.Creator`?

packages/api/src/index.ts
```diff
@@ -122,6 +124,7 @@ export type {
   ThreeDimensionalQueryAggregationDefinition,
   TwoDimensionalQueryAggregationDefinition,
 } from "./ontology/QueryDefinition.js";
+export type { SimplePropertyDef } from "./ontology/SimplePropertyDef.js";
```
ericanderson, line 127: Im not sure if we want this exported or how.

packages/react/src/new/useObjectSet.tsx
```diff
@@ -0,0 +1,458 @@
...
+  }, [nextPageToken, isLoading, data, serializedKey]);
+
+  return {
+    data,
+    isLoading,
+    error,
+    fetchMore: nextPageToken ? fetchMore : undefined,
+    objectSet: finalObjectSet,
+    revalidate,
+  };
```
ericanderson, line 457: I think we should do this slightly more formally as a `ObjectSetQuery` and `ObjectSetCacheKey` etc.

The way this stands right now, if an action invalidates an object type, this won't be refreshed.

ericanderson (changes requested):
Left some notes about exporting types but more importantly, I think we should make this slightly more in tune with the rest of the `ObservableClient`. Then we get better cache invalidation and we arent duplicating as much code.

packages/api/src/index.ts
KuberSethi: Why is this? The contents of the union are exported so i just declared inline types that effectively do the same thing, but it seems we should probably align on the right type to use for RDP values and also export.

packages/react/src/new/useObjectSet.tsx
KuberSethi: Took a stab at this - let me know if this is in-line with what you were thinking

PR #1964, opened 2025-09-15
make dedupeIntervals dynamic

KuberSethi:
Fixes static dedupeIntervals fixme to ensure dedupeIntervals are dynamic and the minimum gets used.


ericanderson (approved):
2 notes:
1. Failing builds so youll need to push up another pr for me to approve
2. I have optimization thoughts about the subscription id (e.g why not a number?) but they are encapsulated so I wouldn't block on it

PR #1966, opened 2025-09-15
Add endswith

ssanjay1:
We can now do `client(myObject).where({stringProp: {$endsWith: "suffix"}}).fetchPage()`! Which will allow you to do filters for leading wildcard searches on an object property!

### DIFF
--- a/.changeset/smooth-times-clap.md
+++ b/.changeset/smooth-times-clap.md
@@ -1,0 +1,6 @@
+---
+"@osdk/client": patch
+"@osdk/api": patch
+---
+
+Add ends with functionality.


packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -296,6 +296,15 @@ function handleWherePair(
     };
   }
 
+  if (firstKey === "$endsWith") {
+    return {
```
ssanjay1, line 300: May end up exposing an endsWith filter through gateway as well, but it will just be this under the hood

PR #1992, opened 2025-09-19
Export extractPrimaryKeyFromObjectSpecifier handler

nihalbhatnagar:
Users want to be able to use the object specifier to filter their object sets. Theres no good way to go from object specifier to object reference without this, unless they manually split by themselves.


KuberSethi (approved):
Agree on potentially renaming but overall looks good

PR #1993, opened 2025-09-19
Fix websocket subscription management

KuberSethi:
Fixes websocket subscription management issue where multiple subscriptions were being created and causing us to handle that downstream. We should arguably (perhaps in a FLUP) clean that up, but for now being defensive isn't bad.

### DIFF
--- a/packages/client/src/observable/internal/list/ListQuery.ts
+++ b/packages/client/src/observable/internal/list/ListQuery.ts
@@ -46,6 +46,7 @@ import type { Entry } from "../Layer.js";
 import { type ObjectCacheKey } from "../object/ObjectCacheKey.js";
 import { objectSortaMatchesWhereClause as objectMatchesWhereClause } from "../objectMatchesWhereClause.js";
 import type { OptimisticId } from "../OptimisticId.js";
+import { RefCounts } from "../RefCounts.js";
 import type { SimpleWhereClause } from "../SimpleWhereClause.js";
 import { OrderBySortingStrategy } from "../sorting/SortingStrategy.js";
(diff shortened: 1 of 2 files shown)


packages/client/src/observable/internal/list/ListQuery.ts
```diff
@@ -46,6 +46,7 @@ import type { Entry } from "../Layer.js";
 import { type ObjectCacheKey } from "../object/ObjectCacheKey.js";
 import { objectSortaMatchesWhereClause as objectMatchesWhereClause } from "../objectMatchesWhereClause.js";
 import type { OptimisticId } from "../OptimisticId.js";
+import { RefCounts } from "../RefCounts.js";
```
KuberSethi, line 49: Seemed like a good candidate for this problem but let me know if there's a better approach or if there's any issues with the class

ericanderson: Can you double check the websocket subscription code. I think there is a chance it already dedupes subscriptions?

KuberSethi: It dedupes but I don't think that's an ideal state to be in - ListQuery.registerStreamUpdates() was creating a new websocket subscription every time it was called, leading to multiple subscriptions for the same ListQuery instance. We might dedupe (I think we're just replacing all subscriptions?) but we're still in a nonideal state before then

KuberSethi:
> Thinking more, the existing ref count stuff will close out the list query after its no longer needed.

Fair point, but shouldn't we still prevent multiple subscriptions from being created in the first place?

PR #1994, opened 2025-09-20
Generate compiled version of OSDK docs

ewitkon:
Transforms Handlebars templates from 
  `@osdk/typescript-sdk-docs` into runnable TypeScript examples. Automatically generates variations for conditional blocks to test 
  different code paths and validates template syntax at build time.


packages/typescript-sdk-docs-examples/src/typescriptOsdkExamples.ts
```diff
@@ -0,0 +1,465 @@
...
+ * WARNING: This file is generated automatically by the generateExamples.ts script.
+ * DO NOT MODIFY this file directly as your changes will be overwritten.
+ */
+
+/**
+ * Generated examples hierarchy for SDK documentation
+ * This provides a mapping of example names to their file paths
+ * similar to how TYPESCRIPT_OSDK_SNIPPETS works for templates
+ */
+export const TYPESCRIPT_OSDK_EXAMPLES = {
```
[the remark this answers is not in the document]
ewitkon: We are this is part of what we export as JSON to other packages

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -951,118 +951,95 @@ versions:
...
-            // To fetch a page without a result wrapper, use fetchPage instead
-            const responseNoErrorWrapper: PageResult<Osdk.Instance<{{objectType}}>>
-                = await client({{objectType}}).fetchPage({ $pageSize: 30 });
+            import type { Osdk, PageResult } from "@osdk/client";
+            try {
+                const responseNoErrorWrapper: PageResult<Osdk.Instance<{{objectType}}>>
+                    = await client({{objectType}}).fetchPage({ $pageSize: 30 });
+            } catch (e) {
+                throw e;
+            }
```
ewitkon, line 1042: same - only try/catch

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1078,40 +1055,26 @@ versions:
...
-
-            // You can also get a linked object without the result wrapper
-            function getLinkedNoWrapper{{linkedObjectType}}(source: Osdk.Instance<{{sourceObjectType}}>, linkedObjectPrimaryKey: {{linkedPrimaryKeyPropertyV2.type}}): Osdk.Instance<{{linkedObjectType}}> {
-                return source.$link.{{linkApiName}}.fetchOne(linkedObjectPrimaryKey);
+            async function getLinked{{linkedObjectType}}(source: Osdk.Instance<{{sourceObjectType}}>) {
+                try {
+                    return await source.$link.{{linkApiName}}.fetchPage();
+                } catch (error) {
+                    return { error };
+                }
```
ewitkon, line 1066: keep only try/catch and make sure we have the right types imported

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1078,40 +1055,26 @@ versions:
...
-                return source.$link.{{linkApiName}}.fetchOneWithErrors();
-                {{/isLinkManySided}}
-            }
-
-            // You can also get a linked object by doing an object set searcharound
-            function getLinkedWithPivot{{linkedObjectType}}(){
-                return client({{sourceObjectType}}).pivotTo({{linkApiName}}).fetchPage();
+            async function getLinkedWithPivot{{linkedObjectType}}(){
+                return await client({{sourceObjectType}}).pivotTo("{{linkApiName}}").fetchPage();
             }
```
ewitkon, line 1078: same - only try/catch + simplify example 

Not sure if we need to extend this to include both one side and many side

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1120,10 +1083,10 @@ versions:
             import { client } from "./client";
 
             const num{{objectType}} = await client({{objectType}})
-                .where({{property}}: { $isNull : false }})
+                .where({ {{property}}: { $isNull : false }})
                 .aggregate({
                     $select: { $count: "unordered" },
-                    $groupBy: { name: "exact" },
+                    $groupBy: { {{property}}: "exact" },
                 });
```
ewitkon, line 1090: fix handlebar example

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1214,98 +1177,172 @@ versions:
             - propertyValueIncrementedV2
       applyAction:
```
ewitkon, line 1178: This required a larger re-write to take into account cases where we have attachment and not and if the attachment is already loaded or not.

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1365,12 +1426,12 @@ versions:
 
             const {{objectType}}ObjectSet = client({{objectType}})
                 .where({
-                    {{#structSubPropertyApiName}}
+                    {{#hasStructSubProperty}}
```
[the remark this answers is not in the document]
ewitkon: We used the same variable for both the block and the property which is wrong. we will need to validate this in Forge

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1623,18 +1683,19 @@ versions:
...
+                const page: PageResult<Osdk<{{interfaceApiName}}>> = await client({{interfaceApiName}})
+                    .fetchPage({
+                        $orderBy: { {{property}}: "asc"},
+                        $pageSize: 30
+                    });
+            } catch (e) {
+                throw e;
             }
+
+
```
ewitkon, line 1705: same - only try/catch

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1915,133 +1986,93 @@ versions:
...
                         if (update.state === "ADDED_OR_UPDATED") {
                             // An object has received an update or an object was added to the object set
-                            const currentObject = objects[update.object.$primaryKey];
-                            if (currentObject !== undefined) {
-                                currentObject["<propertyName>"] = update.object["<propertyName>"] ?? currentObject["<propertyName>"];
-                            }
+                            // Get the object using the $primaryKey from your cache
+                            // const currentObject = objects[update.object.$primaryKey];
+                            // use the update.object["<propertyName>"] to update your cache 
+                            //currentObject["<propertyName>"] = update.object["<propertyName>"] ?? currentObject["<propertyName>"];
```
ewitkon, line 2059: simplify the example to make it work and not assume how you store your cache

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1915,133 +1986,93 @@ versions:
...
                         // We could not keep track of all changes. Please reload the objects in your set.
                     },
                 },
-                { properties: [ {{#propertyNames}}"{{.}}", {{/propertyNames}}] }
+                { properties: [ {{{propertyNames}}} ] }
             );
 
+            // On dismount unsubscribe
             subscription.unsubscribe();
-      uploadMedia:
```
ewitkon, line 1997: Discussed - not shoing experimental not suported

.changeset/@osdk-api-simulatedRelease.md
```diff
@@ -1,6 +1,5 @@
 ---
 "@osdk/api": patch
 ---
-
```
nihalbhatnagar, line 4: Could you revert all of these linted files back to main? You can specify the whole directory with `git reset --hard origin/main -- path/to/your/folder/`

ewitkon: Done

packages/typescript-docs-example-generator/src/utils/generateFileHeader.ts
```diff
@@ -0,0 +1,54 @@
...
+export function generateFileHeader(
+  snippetKey?: string,
+  description?: string,
+): string {
+  const commentStyle = snippetKey ? "/**" : "/*";
+  const commentEnd = snippetKey ? " */" : " */";
+  const descriptionText = description ? ` (${description})` : "";
+
+  return `${commentStyle}
+ * Copyright 2023 Palantir Technologies, Inc. All rights reserved.
```
KuberSethi, line 33: nit: 2025

packages/typescript-docs-example-generator/src/validation/template-validator.ts
```diff
@@ -0,0 +1,302 @@
...
+            expected: variable.type,
+            actual: typeof value,
+          });
+        }
+      }
+    }
+
+    // Validate block variables
+    for (const block of template.blocks) {
+      const varName = block.name.substring(1); // Remove # or ^ prefix
```
KuberSethi, line 84: are we sure this is always a # or ^?

packages/typescript-docs-example-generator/src/analyzer/template-analyzer.ts
```diff
@@ -0,0 +1,319 @@
...
+} from "../types/index.js";
+import type { MustacheToken } from "../types/mustache-tokens.js";
+import {
+  isBlockToken,
+  isInvertedSectionToken,
+  isSectionToken,
+  isVariableToken,
+} from "../types/mustache-tokens.js";
+
+export class TemplateAnalyzer {
```
KuberSethi, line 33: I don't love how tightly coupled this is to Mustache in case we decide to change the lib we use down the road but for now this is fine, just calling out for SA

packages/typescript-docs-example-generator/src/analyzer/template-analyzer.ts
```diff
@@ -0,0 +1,319 @@
...
+   * Parses a template string into Mustache tokens
+   */
+  private parseTemplate(
+    template: string,
+  ): Result<MustacheToken[], ProcessingError> {
+    try {
+      // Parse template using Mustache
+      const tokens = Mustache.parse(template) as MustacheToken[];
+
+      // Also try to render the template to catch semantic errors
```
KuberSethi, line 92: Is this standard practice with Mustache usage? Seems a bit odd but if they don't expose a better way I suppose its fine

packages/typescript-docs-example-generator/src/analyzer/template-analyzer.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  /**
+   * Extracts all variables from Mustache tokens with type inference
+   */
+  private extractVariables(tokens: MustacheToken[]): TemplateVariable[] {
+    const variables = new Map<string, TemplateVariable>();
+
+    this.traverseTokens(tokens, (token) => {
+      if (isVariableToken(token)) {
+        const name = token[1]; // variable name is at index 1
```
KuberSethi, line 117: nit: might be worth structuring / typing these, if the ordering ever changes this would break and be annoying to debug potentially

packages/typescript-docs-example-generator/src/analyzer/template-analyzer.ts
```diff
@@ -0,0 +1,319 @@
...
+            required: true,
+            defaultValue: undefined,
+          });
+        }
+      } else if (isBlockToken(token)) {
+        const name = token[1]; // section name is at index 1
+        if (!variables.has(name)) {
+          variables.set(name, {
+            name,
+            type: "boolean", // Block variables are typically boolean
```
KuberSethi, line 131: nit: typically? when would they not be booleans?

packages/typescript-docs-example-generator/src/utils/batch-processor.ts
```diff
@@ -0,0 +1,250 @@
...
+  continueOnError?: boolean;
+  /** Custom error message prefix */
+  errorPrefix?: string;
+}
+
+/**
+ * Utility class for consolidating common batch processing and error handling patterns.
+ * Reduces code duplication across multiple modules that follow similar error collection patterns.
+ */
+export class BatchProcessor {
```
KuberSethi, line 55: Thoughts on some tests here? I think for this PR leaving many of these simple util / i/o functions untested is fine but I'm typically a little more cautious when it comes to concurrency code as concurrency bugs can manifest strangely so its often best to just invest in tests early.

packages/typescript-docs-example-generator/src/analyzer/template-analyzer.ts
ewitkon: @KuberSethi This is VERY coupled with Mustache because this is what Forge is using. I started using `handlebars` but the block logic if it is very different (include scope of variable inside block and so) => As long as Forge is using Mustache we should be using the same.

packages/typescript-docs-example-generator/src/validation/template-validator.ts
ewitkon: Yes, this is the syntax of handlebars

packages/typescript-docs-example-generator/src/utils/generateFileHeader.ts
ewitkon: Are you sure this is the copyright message?

packages/typescript-sdk-docs-examples/src/examples/typescript/2.0.0/applyAction_#hasAttachmentUpload.ts
```diff
@@ -0,0 +1,47 @@
...
+import { createAttachmentUpload } from "@osdk/client";
+import { documentEquipment } from "../../../generatedNoCheck/index.js";
+
+async function callActionWithAttachmentUpload() {
+    async function uploadMyFile() {
+        const file = await fetch("file.json");
+        const blob = await file.blob();
+        return createAttachmentUpload(blob, "myFile");
+    }
+    const attachment: AttachmentUpload = await uploadMyFile();
```
nihalbhatnagar, line 34: One nit: This doesn't actually upload attachments (doesn't send a network request). Maybe we should just have `attachment` below `createAttachmentUpload(blob, "myFile");

packages/typescript-sdk-docs/src/documentation.yml
ewitkon: Not sure what do you mean by generate docs?

packages/typescript-sdk-docs/src/documentation.yml
ewitkon: At least in compile this does not break
```
import { Employee } from "../../../generatedNoCheck/index.js";
// Edit this import if your client location differs
import { client } from "./client.js";

const numEmployee = await client(Employee)
    .where({ department: { $isNull : false }})
    .aggregate({
        $select: { $count: "unordered" },
        $groupBy: { managedEmployeeIds: "exact" },
    });
```
managedEmployeeIds is an integer type

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -2061,14 +2106,16 @@ versions:
                   .withProperties({
                     "newPropertyName": (baseObjectSet) =>
                         {{^isUnary}}
-                        baseObjectSet.pivotTo("{{linkName}}")
-                            .selectProperty("{{property}}").{{operation}}(
-                                baseObjectSet.pivotTo("{{linkName}}").aggregate("$count"))
+                        baseObjectSet.pivotTo("{{linkApiName}}")
+                            .aggregate("{{property}}").{{operation}}(
```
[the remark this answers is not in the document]
ewitkon: Why? 
The issue with select property is that you can only show the template for links which are from N-> 1 and not for all link types.

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -2077,17 +2124,18 @@ versions:
 
               const {{objectType}}WithExpression = await client({{objectType}})
                   .withProperties({
-                    "newPropertyName": (baseObjectSet) =>
                         {{^isExtractPart}} 
-                        baseObjectSet.pivotTo("{{linkName}}")
-                            .selectProperty("{{property}}")
-                            .{{operation}}(baseObjectSet.pivotTo("{{linkName}}").selectProperty("{{property}}"))
+                        "{{operation}}_between_{{property}}_and_{{otherProperty}}": (baseObjectSet) =>
```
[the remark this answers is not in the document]
ewitkon: This is just me trying to give the property a name which explains the operation

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -2077,17 +2124,18 @@ versions:
...
+                        baseObjectSet.pivotTo("{{linkApiName}}")
+                            .aggregate("{{property}}")
+                            .{{operation}}(baseObjectSet.pivotTo("{{linkApiName}}").aggregate("{{otherProperty}}"))
```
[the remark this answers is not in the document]
ewitkon: Maybe I don't understand the operation here.

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -1759,14 +1839,11 @@ versions:
             const sum{{objectType}} = await client({{objectType}})
                 .withProperties({
                   "newPropertyName": (baseObjectSet) =>
-                    baseObjectSet.pivotTo("fooLink").pivotTo("barLink").selectProperty("foo")
+                    baseObjectSet.pivotTo("{{linkApiName}}").aggregate("$count")
```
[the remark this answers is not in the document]
nihalbhatnagar: Can we leave this change out for now? I see your relevant change to pass in linkApiName from that component in forge, but I want to really minimize the complexity fo these PRs and track these as FLUPs

packages/typescript-docs-example-generator/src/baseContext.ts
```diff
@@ -0,0 +1,764 @@
...
+  // For nearest neighbors (2.4.0+)
+  vectorProperty: "skillVector",
+  vectorDimensionSize: 128,
+};
+
+/**
+ * Complete template configuration registry
+ * Easy lookup: TEMPLATE_REGISTRY[templateName] gives you all parameters
+ */
+const TEMPLATE_REGISTRY: TemplateRegistry = {
```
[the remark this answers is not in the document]
ewitkon: I will look into this as a FLUP

packages/typescript-sdk-docs/src/documentation.yml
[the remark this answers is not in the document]
nihalbhatnagar: We have the selector in forge that `should` narrow it down to just `one` links. Maybe we should just another section thats `numeric operations` and make this one `numeric operations on linked properties`?

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -951,118 +951,100 @@ versions:
...
+                if (firstPage.nextPageToken === undefined) {
+                    // eslint-disable-next-line no-console
+                    console.log(firstPage.data);
+                }
                 const secondPage: PageResult<Osdk.Instance<{{objectType}}>>
-                 = await client({{objectType}}).fetchPage({ $pageSize: 30, $nextPageToken: firstPage.nextPageToken });
-                const objects = [...firstPage.data, ...secondPage.data];
-                const object = objects[0];
+                = await client({{objectType}}).fetchPage({ $pageSize: 30, $nextPageToken: firstPage.nextPageToken });
+                // eslint-disable-next-line no-console
```
nihalbhatnagar, line 980: This should not be in our docs I think

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -2077,17 +2169,18 @@ versions:
...
               const {{objectType}}WithExpression = await client({{objectType}})
                   .withProperties({
-                    "newPropertyName": (baseObjectSet) =>
                         {{^isExtractPart}} 
-                        baseObjectSet.pivotTo("{{linkName}}")
+                        "derivedPropertyDatetime_{{operation}}": (baseObjectSet) =>
+                        baseObjectSet.pivotTo("{{linkApiName}}")
                             .selectProperty("{{property}}")
-                            .{{operation}}(baseObjectSet.pivotTo("{{linkName}}").selectProperty("{{property}}"))
+                            .{{operation}}(baseObjectSet.selectProperty("{{property}}"))
```
nihalbhatnagar, line 2176: This should also be pivot to

packages/typescript-sdk-docs/src/documentation.yml
```diff
@@ -2059,16 +2145,22 @@ versions:
...
+                        baseObjectSet.pivotTo("{{linkApiName}}")
+                            .aggregate("{{property}}").{{operation}}(
+                                baseObjectSet.pivotTo("{{linkApiName}}").aggregate("$count"))
+                    }).fetchPage();
+                    {{/isUnary}}
+                    {{#isUnary}}
+                    "abs_{{property}}": (baseObjectSet) =>
+                        baseObjectSet.pivotTo("{{linkApiName}}")
+                        .selectProperty("{{property}}").abs(),
+                    "negate_{{property}}": (baseObjectSet) =>
```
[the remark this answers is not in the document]
ewitkon: Good catch!

PR #2033, opened 2025-10-03
Add object set support to widgets config

amishshah:
Allows specifying an `objectSet` parameter in config:

```ts
import { Employee } from "@widget/sdk";

defineConfig({
  parameters: {
    myObjectSet: {
      type: "objectSet",
      displayName: "Object set",
      objectType: Employee,
    }
  },
});
```

and usage:

```ts
const { emitEvent, parameters } = useWidgetContext();

parameters.values.myObjectSet.objectSetRid === "ri.object-set.provided-by-host-application";
```

Didn't want to take an explicit dependency on `@osdk/api` (since I don't want there to be breaks between different versions of the API for unrelated properties) so I just use the `ObjectDefinition` type in tests to make sure the type we use is consistent with the type in our widget API. 

I've left it at just `objectSetRid` for now, need to reason more about how/where/if we should expose object type. Wrapped `objectSetRid` in an object to allow for future extension e.g. adding `objectSet` as a helper.

This is enough to unblock actual usage though, the syntactic sugar improvements will come in a follow-up PR!

### DIFF
--- a/packages/widget.vite-plugin/src/build-plugin/__tests__/buildWidgetSetManifest.test.ts
+++ b/packages/widget.vite-plugin/src/build-plugin/__tests__/buildWidgetSetManifest.test.ts
@@ -20,7 +20,7 @@ import { buildWidgetSetManifest } from "../buildWidgetSetManifest.js";
 import type { WidgetBuildOutputs } from "../getWidgetBuildOutputs.js";
 
 describe("buildWidgetSetManifest", () => {
-  const WIDGET_SET_RID = "ri.widgetregistry.widget-set..test-widget-set";
+  const WIDGET_SET_RID = "ri.widgetregistry..widget-set.test-widget-set";
   const WIDGET_SET_VERSION = "1.0.0";
 
   test("builds a valid manifest with multiple widgets", () => {
--- a/packages/widget.vite-plugin/src/build-plugin/buildWidgetSetManifest.ts
+++ b/packages/widget.vite-plugin/src/build-plugin/buildWidgetSetManifest.ts
@@ -80,13 +80,9 @@ function convertParameter(
   parameter: ParameterDefinition,
 ): ManifestParameterDefinition {
   if (parameter.type === "objectSet") {
-    // We expect to always have a rid here because the config has already been validated
-    if (
-      typeof parameter.objectType.internalDoNotUseMetadata?.rid !== "string"
-    ) {
-      throw new Error(
-        `Missing object type rid for object set parameter '${parameter.displayName}'`,
-      );
+    // Config has already been validated so rid must be present
+    if (parameter.objectType.internalDoNotUseMetadata == null) {
+      throw new Error("Expected internal metadata to be present");
     }
     return {
       type: "objectSet",


.changeset/ten-guests-study.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/widget.api": minor
```
tzyl, line 2: Are we intending to backport all of these to release or wait to next release? Usual stuff about will need to change to patch on backport I think

packages/widget.api/src/config.ts
```diff
@@ -26,9 +26,15 @@ interface ArrayParameterDefinition<S extends ParameterValue.PrimitiveType> {
...
+interface ObjectSetParameterDefinition<T extends ObjectType> {
+  type: "objectSet";
+  displayName: string;
+  objectType: T;
+}
 export type ParameterDefinition =
   | PrimitiveParameterDefinition<ParameterValue.PrimitiveType>
-  | ArrayParameterDefinition<ParameterValue.PrimitiveType>;
+  | ArrayParameterDefinition<ParameterValue.PrimitiveType>
+  | ObjectSetParameterDefinition<ObjectType>;
```
tzyl, line 37: Have you thought about what an event with updateParameterIds that includes an object set parameter will do?

tzyl (commented):
Nice looks great just had a few questions on what the plans are

packages/widget.api/src/config.ts
amishshah: I've played around with the existing [bidirectional iframe widgets](https://github.com/palantir/workshop-iframe-custom-widget) and they can actually pass a new object set rid up to Workshop which Workshop is able to use.

I think we could allow custom widgets to also do the same, i.e. emit an update event that sets the object set variable in Workshop to a new object set rid value. What are your thoughts?

packages/widget.vite-plugin/src/common/validateWidgetConfig.ts
```diff
@@ -75,5 +75,15 @@ function validateWidgetParameters(parameters: ParameterConfig): void {
         `Parameter id "${parameterId}" does not match allowed pattern (must be camelCase)`,
       );
     }
+    if (parameterConfig.type === "objectSet") {
```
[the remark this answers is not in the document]
amishshah: Yep exactly!

packages/widget.vite-plugin/src/common/validateWidgetConfig.ts
```diff
@@ -75,5 +75,15 @@ function validateWidgetParameters(parameters: ParameterConfig): void {
...
+      if (
+        typeof parameterConfig.objectType.experimentalDoNotUseMetadata?.rid
```
tzyl, line 80: Just adding to the PR as well that at whatever point we announce this to users then we may be locking ourselves into experimentalDoNotUseMetadata to not break library versions that users depend on

tzyl (dismissed):

packages/widget.api/src/config.ts
```diff
@@ -26,9 +26,21 @@ interface ArrayParameterDefinition<S extends ParameterValue.PrimitiveType> {
...
+
+export type ManifestParameterDefinition =
+  | Exclude<ParameterDefinition, ObjectSetParameterDefinition<ObjectType>>
+  | (Omit<ObjectSetParameterDefinition<ObjectType>, "objectType">) & {
+    objectTypeRids: [string];
```
tzyl, line 42: Do you prefer to keep the manifest format array as well or start it off as string? Right now we have:
- Config format single object type
- Manifest format array
- Stored format array

For the manifest we have some flexibility to change it like allowing `string | string[]` in the future but starting it off as array is fine as well if it's clear that it will be enforced to be non-empty with one rid to start with per your types ([string])

packages/widget.api/src/config.ts
```diff
@@ -26,9 +26,21 @@ interface ArrayParameterDefinition<S extends ParameterValue.PrimitiveType> {
...
+  displayName: string;
+  objectType: T;
+}
 export type ParameterDefinition =
   | PrimitiveParameterDefinition<ParameterValue.PrimitiveType>
-  | ArrayParameterDefinition<ParameterValue.PrimitiveType>;
+  | ArrayParameterDefinition<ParameterValue.PrimitiveType>
+  | ObjectSetParameterDefinition<ObjectType>;
+
+export type ManifestParameterDefinition =
```
tzyl, line 39: It wouldn't directly be mapped from ParameterDefinition but do you think it's clearer to follow if we just redefined the types:
```
interface ManifestObjectSetParameterDefinition<T extends ObjectType> {
  type: ObjectSetParameterDefinition<T>["type"];
  displayName: string;
  objectTypeRids: [string];
}
export type ManifestParameterDefinition =
  | PrimitiveParameterDefinition<ParameterValue.PrimitiveType>
  | ArrayParameterDefinition<ParameterValue.PrimitiveType>
  | ManifestObjectSetParameterDefinition<ObjectType>;
```

packages/widget.vite-plugin/src/common/validateWidgetConfig.ts
```diff
@@ -75,5 +75,15 @@ function validateWidgetParameters(parameters: ParameterConfig): void {
         `Parameter id "${parameterId}" does not match allowed pattern (must be camelCase)`,
       );
     }
+    if (parameterConfig.type === "objectSet") {
+      if (
+        typeof parameterConfig.objectType.internalDoNotUseMetadata?.rid
+          !== "string"
+      ) {
+        throw new Error(
```
tzyl, line 83: Should we include some more detail on parameterConfig / parameterConfig.objectType when this happens so we know exactly what the shape looked like?

packages/widget.vite-plugin/src/build-plugin/buildWidgetSetManifest.ts
```diff
@@ -58,11 +60,36 @@ function buildWidgetManifest(
...
+
+function convertParameter(
+  parameter: ParameterDefinition,
+): ManifestParameterDefinition {
+  if (parameter.type === "objectSet") {
+    return {
+      type: "objectSet",
+      displayName: parameter.displayName,
+      // Config has already been validated so rid must be present
+      objectTypeRids: [parameter.objectType.internalDoNotUseMetadata!.rid],
```
tzyl, line 87: Can we throw instead rather than silently allowing undefined through at runtime? The validation is defined elsewhere so would be good to quickly check that assumption here

packages/widget.vite-plugin/src/build-plugin/__tests__/buildWidgetSetManifest.test.ts
```diff
@@ -0,0 +1,236 @@
...
+ * limitations under the License.
+ */
+
+import type { ParameterDefinition, WidgetSetInputSpec } from "@osdk/widget.api";
+import { describe, expect, test } from "vitest";
+import { buildWidgetSetManifest } from "../buildWidgetSetManifest.js";
+import type { WidgetBuildOutputs } from "../getWidgetBuildOutputs.js";
+
+describe("buildWidgetSetManifest", () => {
+  const WIDGET_SET_RID = "ri.foundry.main.widget-set.test-widget-set";
```
tzyl, line 23: if you want this to be more representative of the real world:
```suggestion
  const WIDGET_SET_RID = "ri.widgetregistry..widget-set.test";
```

tzyl (commented):
Thanks for the tests! Some last small comments

packages/widget.api/src/config.ts
amishshah: I'd like to keep it `string[]` in case we expand in the future. I think the backend change could look ugly if we have to start supporting `string | string[]` in the types? +1 on validating that it's a `[string]` for now though!

packages/widget.vite-plugin/src/build-plugin/__tests__/buildWidgetSetManifest.test.ts
```diff
@@ -20,7 +20,7 @@ import { buildWidgetSetManifest } from "../buildWidgetSetManifest.js";
 import type { WidgetBuildOutputs } from "../getWidgetBuildOutputs.js";
 
 describe("buildWidgetSetManifest", () => {
-  const WIDGET_SET_RID = "ri.foundry.main.widget-set.test-widget-set";
+  const WIDGET_SET_RID = "ri.widgetregistry.widget-set..test-widget-set";
```
tzyl, line 23: This is the correct format sorry `ri.widgetregistry..widget-set.test-widget-set`

tzyl (approved):

PR #2034, opened 2025-10-03
Intermediary links in OAC

ethanai22:
e2e: https://swirl.palantirfoundry.com/workspace/ontology/link-type/wzvk7f92.fizz-to-foo/overview


packages/maker/src/api/links/LinkType.ts
```diff
@@ -104,6 +113,37 @@ export interface ManyToManyObjectLinkReferenceUserDefinition {
...
+}
+
+export interface IntermediaryObjectLinkReference {
+  object: ObjectTypeDefinition;
+  metadata: LinkTypeMetadata;
+}
+
+export interface IntermediaryLinkTypeUserDefinition {
+  apiName: string;
+  manyObjectA: IntermediaryObjectLinkReferenceUserDefinition;
```
ethanai22, line 135: open to ideas on the syntax for this. `many` vs `manyObjectA` could be confusing, but I think it's a good idea to keep the fields for many-to-many and intermediary links different

PR #2040, opened 2025-10-06
Fix vector property types for EditBatch API

marshallavail:
When then `EditBatch` API was introduced, vector property types were not yet supported in the TS OSDK.
Support was added recently, but there is a currently a bug that causes them to be typed as `number` and not `number[]`.
```
const batch = createEditBatch<Edits.Object<VectorObjectType>>(client);
batch.create(VectorObjectType, { vectorProperty: [1, 2, 3] }) // complains that "vectorProperty" should just be "number"
```

This seems to stem from the usage of the type utility `Converted` in `OsdkObjectCreatePropertyType` [here](https://github.com/palantir/osdk-ts/blob/3c96a2dcff8a87d08383db77724dde90cafdac9f/packages/api/src/Definitions.ts#L59C13-L68), which flattens out any array types.

From what I can tell both `Converted` and `Raw` are an anachronism. They were introduced [here](https://github.com/palantir/osdk-ts/commit/0860ae94b590ab345a7aae9d353caa875624a55b#diff-e93f74e301b57d81b61bab153df75f41b4c8734f2aada16d1020118151cd99db) and were meant to index into `ValidPropertyTypes`, which doesn't exist anymore.

### DIFF
--- a/etc/api.report.api.md
+++ b/etc/api.report.api.md
@@ -1081,7 +1081,6 @@ export type OsdkObject<N extends string> = {
 // Warning: (tsdoc-param-tag-with-invalid-type) The @param block should not include a JSDoc-style '{type}'
 // Warning: (tsdoc-param-tag-with-invalid-type) The @param block should not include a JSDoc-style '{type}'
 // Warning: (ae-forgotten-export) The symbol "MaybeArray" needs to be exported by the entry point index.d.ts
-// Warning: (ae-forgotten-export) The symbol "Converted" needs to be exported by the entry point index.d.ts
 // Warning: (ae-forgotten-export) The symbol "GetCreatePropertyValueFromWire" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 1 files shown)


packages/api/src/Definitions.ts
```diff
@@ -38,19 +35,19 @@ export type OsdkObjectPropertyType<
   T extends ObjectMetadata.Property,
   STRICTLY_ENFORCE_NULLABLE extends boolean = true,
 > = STRICTLY_ENFORCE_NULLABLE extends false ?
-    | MaybeArray<T, Converted<GetClientPropertyValueFromWire<T["type"]>>>
+    | MaybeArray<T, GetClientPropertyValueFromWire<T["type"]>>
```
nihalbhatnagar: This is fine because any types indexed by GetClientPropertyValueFromWire would have always resolved to T. The only place this is used otherwise is for aggregation results and these would be incorrect for vectors.

PR #2046, opened 2025-10-07
Heartbeat Websocket Connections

nihalbhatnagar:
This PR introduces WebSocket heartbeating. The browser tries to close websockets after a period of inactivity. We automatically retry websocket connections when dropped, leading to multiple onOutOfDate messages where the client needs to reload all data but is likely unnecessary.

This doesn't need a flag and isn't a break because fundamentally resending the same subscriptions will not change any subscription IDs or drop any subscriptions.

Tested via a local webapp

### DIFF
--- a/packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
+++ b/packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
@@ -43,6 +43,7 @@ const EXPONENTIAL_BACKOFF_MAX_DELAY_MS = 60000;
 const EXPONENTIAL_BACKOFF_MULTIPLIER = 2;
 const EXPONENTIAL_BACKOFF_JITTER_FACTOR = 0.3;
 const WEBSOCKET_IDLE_DISCONNECT_DELAY_MS = 15000;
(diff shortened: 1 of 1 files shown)


packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -43,6 +43,7 @@ const EXPONENTIAL_BACKOFF_MAX_DELAY_MS = 60000;
 const EXPONENTIAL_BACKOFF_MULTIPLIER = 2;
 const EXPONENTIAL_BACKOFF_JITTER_FACTOR = 0.3;
 const WEBSOCKET_IDLE_DISCONNECT_DELAY_MS = 15000;
+const WEBSOCKET_HEARTBEAT_INTERVAL_MS = 45 * 1000;
```
nihalbhatnagar, line 46: This number was chosen based on some information I found online about how often the browser culls websockets when under load and claude. However, there isn't an actual default that the browser will use, so its hard for me to get a reading of a good default here. I will experiment in an internal repository before GAing this workflow

bryantpark04 (approved):

Issue #2081, opened 2025-10-14
Docs for actions that take an object input are missing the apiName


joekane3:
The generated action docs mention the primary key, but neglect to mention that a necessary field for an object input is the object apiName, and not just the primaryKey

Using the docs as is:
<img width="643" height="171" alt="Image" src="https://github.com/user-attachments/assets/28b7c31f-73d5-4129-833f-afcb3e0407ca" />

Here are the docs generated, missing the mention of the apiName
<img width="1191" height="664" alt="Image" src="https://github.com/user-attachments/assets/b40782d7-34e3-4e0f-859d-a2e2230984da" />

Here is the manual action needed, including looking up the API name to fill in this red box
<img width="727" height="323" alt="Image" src="https://github.com/user-attachments/assets/709f2b97-bb96-4911-8f7a-d15ac2d50e97" />

PR #2082, opened 2025-10-14
Adding API name for object input

joekane3:
fixes https://github.com/palantir/osdk-ts/issues/2081

I think this template change would fix this.


joekane3:
switched to: https://github.com/palantir/osdk-ts/pull/2188/files

PR #2085, opened 2025-10-15
Apply boolean formatting rule

vhellem:
Apply formatting rule for boolean properties and rules

### DIFF
--- a/packages/client/src/object/formatting/applyPropertyFormatter.test.ts
+++ b/packages/client/src/object/formatting/applyPropertyFormatter.test.ts
@@ -122,7 +122,7 @@ describe("getFormattedValue", () => {
     expect(formatted).toBeUndefined();
   });
 
-  it("returns undefined for null values", () => {
+  it("returns undefined for undefined values", () => {
     const obj = getObject();
     const formatted = obj.$__EXPERIMENTAL__NOT_SUPPORTED_YET__getFormattedValue(
       "nullableBooleanProperty",


packages/client/src/object/formatting/applyPropertyFormatter.ts
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectMetadata, PropertyValueFormattingRule } from "@osdk/api";
+import type { SimpleOsdkProperties } from "../SimpleOsdkProperties.js";
+import { formatBoolean } from "./formatBoolean.js";
+
+export interface FormatPropertyOptions {
```
vhellem: Github wasn't smart enough to understand a lot of this just moved from `applyPropertyFormatters`

packages/client/src/object/formatting/applyPropertyFormatter.test.ts
```diff
@@ -0,0 +1,133 @@
...
+  it("returns undefined for property without formatting rules", () => {
+    const obj = getObject();
+    const formatted = obj.$__EXPERIMENTAL__NOT_SUPPORTED_YET__getFormattedValue(
+      "stringPropertyNoFormatting",
+    );
+
+    expect(formatted).toBeUndefined();
+  });
+
+  it("returns undefined for null values", () => {
```
jakejrichards: just a small point, we call it `null` here but it's `undefined` above -- do we actually care to test `null` specifically or is all absence treated the same?

PR #2086, opened 2025-10-15
Apply numeric formatting rule

vhellem:
Apply numeric formatting rule to property formatters.

### DIFF
--- a/packages/client/src/object/formatting/formatNumber.ts
+++ b/packages/client/src/object/formatting/formatNumber.ts
@@ -57,7 +57,7 @@ export function formatNumber(
       );
 
     case "fixedValues":
-      return numberType.values[value] ?? value.toString();
+      return numberType.values[value];
 
     case "currency":
       return formatCurrency(value, numberType, objectData, locale);


packages/client/src/object/formatting/propertyFormattingUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+    const value = objectData[ref.propertyApiName];
+    return value != null ? String(value) : undefined;
+  }
+  return undefined;
+}
+/**
+ * Gets the browser's current locale
+ */
+export function getBrowserLocale(): string {
+  if (typeof navigator !== "undefined" && navigator.language != null) {
```
nihalbhatnagar: Is this always unavailable in a node runtime? Should we throw instead if not available

packages/client/src/object/formatting/formatNumber.ts
```diff
@@ -0,0 +1,398 @@
...
+  switch (numberType.type) {
+    case "standard":
+      return formatStandardNumber(
+        value,
+        numberType.baseFormatOptions,
+        locale,
+      );
+
+    case "fixedValues":
+      return numberType.values[value] ?? value.toString();
```
nihalbhatnagar: Should we throw here/return undefined?

packages/client/src/object/formatting/propertyFormattingUtils.ts
nihalbhatnagar: I think we its fine to default to english, is there a good way to document this at the top level

PR #2087, opened 2025-10-15
Add date and time formatting

vhellem:
Add date and time formatting

### DIFF
--- a/packages/client/src/object/formatting/formatDateTime.ts
+++ b/packages/client/src/object/formatting/formatDateTime.ts
@@ -57,6 +57,8 @@ export function formatDateTime(
   }
 }
 
+const INVALID_DATE_STRING = "Invalid date or timezone";
+
 function formatLocalized(
   date: Date,
   format: DatetimeLocalizedFormat,
@@ -72,11 +74,15 @@ function formatLocalized(
   }
 
   const options = getLocalizedFormatOptions(format.format);
-
-  return new Intl.DateTimeFormat(
-    locale,
-    timezone ? { ...options, timeZone: timezone } : options,
-  ).format(date);
+  try {
+    return new Intl.DateTimeFormat(
+      locale,
+      timezone ? { ...options, timeZone: timezone } : options,
+    ).format(date);
+  } catch (_e) {
+    // If a property reference is an invalid timezone we specifically say that it's invalid instead of returning undefined;
+    return INVALID_DATE_STRING;
+  }
 }
 
 function getLocalizedFormatOptions(


packages/client/src/object/formatting/applyPropertyFormatter.test.ts
```diff
@@ -493,4 +622,162 @@ describe("getFormattedValue", () => {
...
+    it("formats date without timezone", () => {
+      const obj = getObject();
+      const formatted = obj
+        .$__EXPERIMENTAL__NOT_SUPPORTED_YET__getFormattedValue(
+          "createdDate",
+          EN_US,
+        );
+
+      // Date format should show: Wed, Jan 15, 2025
+      expect(formatted).toMatch(/Wed.*Jan.*15.*2025/);
```
jakejrichards: Any reason to use regex for these rather than an exact string match?

packages/client/src/object/formatting/formatDateTime.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (format.format === "DATE_FORMAT_RELATIVE_TO_NOW") {
+    // TODO (relative time formatting)
+    return undefined;
+  }
+
+  const options = getLocalizedFormatOptions(format.format);
+
+  return new Intl.DateTimeFormat(
+    locale,
+    timezone ? { ...options, timeZone: timezone } : options,
```
nihalbhatnagar: What does the error look like if the timezone isn't a valid value? Does the backend provide validation here?

packages/client/src/object/formatting/formatDateTime.ts
```diff
@@ -0,0 +1,167 @@
...
+  date: Date,
+  format: DatetimeLocalizedFormat,
+  locale: string,
+  timezone: string | undefined,
+): string | undefined {
+  if (format.format === "DATE_FORMAT_ISO_INSTANT") {
+    return date.toISOString();
+  }
+  if (format.format === "DATE_FORMAT_RELATIVE_TO_NOW") {
+    // TODO (relative time formatting)
```
nihalbhatnagar: These won't do what is needed but we have other datetime helpers in `datetimeConverters.ts`. Would be great if there was some overlap there to connect these, but I don't actually think there is

vhellem: I'll take a look, thanks for the pointer!

packages/client/src/object/formatting/formatDateTime.ts
vhellem: That's a great point, it will throw a potential error here. I'll take a look at how we could handle this more gracefully.

jakejrichards (approved):

packages/client/src/object/formatting/formatDateTime.ts
vhellem: Yeah, nothing that seemed obvious we could reuse

PR #2091, opened 2025-10-16
Add relative time formatting

vhellem:
Adds formatting relative time

### DIFF
--- a/packages/client/src/object/formatting/formatDateTime.ts
+++ b/packages/client/src/object/formatting/formatDateTime.ts
@@ -69,8 +69,7 @@ function formatLocalized(
     return date.toISOString();
   }
   if (format.format === "DATE_FORMAT_RELATIVE_TO_NOW") {
-    // TODO (relative time formatting)
-    return undefined;
+    return formatRelativeToNow(date, locale, timezone);
   }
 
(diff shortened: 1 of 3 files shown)


jakejrichards (dismissed):

packages/client/src/object/formatting/formatDateTime.ts
```diff
@@ -171,3 +179,43 @@ function resolveTimezone(
...
+function formatRelativeToNow(
+  date: Date,
+  locale: string,
+  timezone: string | undefined,
+): string {
+  const now = Date.now();
+  const diff = date.valueOf() - now;
+  const absDiff = Math.abs(diff);
+
+  // More than 1 day - fall back to absolute date formatting
```
nihalbhatnagar: Just curious, why do we do this? Seems like RelativeFormat has support for "1 day ago", etc

vhellem: That's is what we internally do for this specific formatting in the ontology, so just keeping them in sync. (I think it's also about 15 days ago is just harder to reason about for a user than a specific date at a certain point, although I'd probably myself put that a >1 week.

nihalbhatnagar (approved):

PR #2098, opened 2025-10-17
Upgrade to latest spec and fix template

ewitkon:
This PR upgrade to the latest docs spec removing unused property and fix a bug in the attachment template


packages/typescript-docs-example-generator/src/types/context.ts
```diff
@@ -95,7 +95,6 @@ export interface PropertyV2 {
 export interface ActionParameterSampleValue {
   key: string;
   value: string;
-  last: boolean;
```
ewitkon: We don't use the last property

PR #2105, opened 2025-10-21
Formatting options for OAC interface actions

ethanai22:
Default format + table configuration for interface actions. Also removed the conjure types from the action user definitions in favor of our own wrappers with easier types and better names (only one team uses these and it's a trivial break)

### DIFF
--- a/.changeset/sour-buckets-happen.md
+++ b/.changeset/sour-buckets-happen.md
@@ -1,0 +1,5 @@
+---
+"@osdk/maker": patch
+---
+
+Add default format and table config options to OAC interface actions


packages/maker/src/api/test/actions.test.ts
```diff
@@ -5286,1412 +5316,1419 @@ describe("Action Types", () => {
               text: "Custom button",
             },
           },
+          tableConfiguration: {
+            columnWidthByParameterRid: {},
+            enableFileImport: false,
+            fitHorizontally: true,
+            frozenColumnCount: 1,
+            rowHeightInLines: 10,
+          },
```
ethanai22: the relevant snapshot changes end here. The rest is the same, but linting messed with the indentation

PR #2110, opened 2025-10-22
context in git

ewitkon:
Track generated context to make sure context is released when documentation example are changing


packages/osdk-docs-context/.gitignore
```diff
@@ -1,2 +1 @@
 # Generated files
-src/generated/
```
KuberSethi: nit: newline

PR #2113, opened 2025-10-23
implement shape types

KuberSethi:
Adds the core types that will power shapes. 

Shapes are intended to be the tool for when users want frontend-only types that are derived from ontology types but with nullability checks, certain fields applied, and OSDK operations performed on top of them. By providing a first-class way to handle this, users will no longer need to maintain many permutations of the same types. 

Future PRs will use these types to implement shape behavior and functionality. 

Tried to keep changes mostly constrained to the shapes package - would love more review attention to the seams where this interacts with the rest of the codebase.

### DIFF
--- a/packages/api/src/objectSet/ShapeObjectSet.ts
+++ b/packages/api/src/objectSet/ShapeObjectSet.ts
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-
-
 import type { FetchPageResult } from "../object/FetchPageResult.js";
 import type { Result } from "../object/Result.js";
 import type {


packages/api/src/objectSet/ShapeObjectSet.ts
```diff
@@ -0,0 +1,277 @@
...
+ * const SlimPlayer = defineShape<typeof Player>()
+ *   .requiredOrThrow("name", "age")
+ *   .asType(Player);
+ *
+ * const objectSet: ShapeObjectSet<...> = client(SlimPlayer);
+ * const player = await objectSet.fetchOne("pk");
+ * // player.name is string (not string | undefined)
+ * ```
+ */
+export interface ShapeObjectSet<
```
KuberSethi: This approach results in some duplication but I couldn't think of a better way to do this that didn't result in infinite type explosion or edits to `ObjectSet` typing

packages/api/src/shapes/defineShapeAsType.ts
```diff
@@ -0,0 +1,135 @@
...
+  L extends PropertyKeys<BASE>,
+  MAKE_REQUIRED extends L,
+>(
+  shape: ShapeDefinition<BASE, L, MAKE_REQUIRED>,
+): ReadonlyArray<MAKE_REQUIRED> {
+  const gathered: MAKE_REQUIRED[] = [];
+
+  for (
+    const list of [
+      shape.__requiredOrThrowProps,
```
KuberSethi: I think it makes sense to support both of these but happy to just have one for now if we want to make things simpler

ericanderson: I think doing both makes sense here. I think both will be needed/wanted

KuberSethi:
Implementing in separate PR

PR #2123, opened 2025-10-28
add osdk toolkit monitor v0

KuberSethi:
Adds a v0 for OSDK Devtools. 
- Allows users to track cache hits / cache performance
- Allows users to track action / optimistic update performance
- Allows users to track compute usage (only for OSS right now unfortunately, this is a backend constraint)
- Lazily loaded and easily enableable (but I still might revisit the installation story, ideally this is out-of-the-box for toolkit users)

image

image

image

image


packages/react-toolkit-monitor/package.json
```diff
@@ -0,0 +1,101 @@
+{
+  "name": "@osdk/react-toolkit-monitor",
+  "version": "0.0.1",
```
ericanderson: Once this merges, this will publish. Is this what you want or do you want a beta package?

ericanderson (approved):
Giving a lot of room due to this being new/experimental and not affecting much.

PR #2130, opened 2025-10-30
Add cli logic to generate code snippets

chengJazhua:
Should be non-breaking - generating code snippets based off of the ontology for documentation purposes

### DIFF
--- a/packages/maker/src/cli/main.ts
+++ b/packages/maker/src/cli/main.ts
@@ -37,9 +37,9 @@ export default async function main(
     valueTypesOutput: string;
     outputDir?: string;
     dependencies?: string;
-    generateCodeSnippets?: boolean;
-    codeSnippetPackageName?: string;
-    codeSnippetDir?: string;
+    generateCodeSnippets: boolean;
+    codeSnippetPackageName: string;
+    codeSnippetDir: string;
     randomnessKey?: string;
   } = await yargs(hideBin(args))
     .version(process.env.PACKAGE_VERSION ?? "")
@@ -93,6 +93,7 @@ export default async function main(
       generateCodeSnippets: {
         describe: "Enable code snippet files creation",
         type: "boolean",
+        default: false,
       },
       codeSnippetPackageName: {
         describe:
@@ -129,7 +130,7 @@ export default async function main(
   if (
     !commandLineOpts.generateCodeSnippets
     && (commandLineOpts.codeSnippetPackageName !== ""
-      || commandLineOpts.codeSnippetDir !== "./")
+      || commandLineOpts.codeSnippetDir !== path.resolve("./"))
   ) {
     consola.info(
       "Package name and/or directory supplied for code snippets, but code snippet generation is false.",
@@ -184,9 +185,9 @@ async function loadOntology(
   apiNamespace: string,
   outputDir: string | undefined,
   dependencyFile: string | undefined,
-  generateCodeSnippets: boolean | undefined,
-  snippetPackageName: string | undefined,
-  codeSnippetDir: string | undefined,
+  generateCodeSnippets: boolean,
+  snippetPackageName: string,
+  codeSnippetDir: string,
   randomnessKey?: string,
 ) {
   const q = await defineOntology(


packages/maker/src/api/defineOntology.ts
```diff
@@ -594,3 +614,108 @@ addDependency("${namespaceNoDot}", new URL(import.meta.url).pathname);
...
+function generateActionSnippet(actionType: ActionType, packageName: string) {
+  const actionContext = {
+    "actionApiName": actionType.apiName,
+    "packageName": packageName,
+  };
+  const allSnippets = getSnippets(objectSnippets, actionContext);
+  return allSnippets;
+}
+
+function getSnippets(snippetType: any, context: {}) {
```
davidpaquin: might be worth defining some types and unions so `snippetType` isn't an `any`, but I'll leave that up to you

packages/maker/src/cli/main.ts
```diff
@@ -87,6 +90,23 @@ export default async function main(
...
+        describe:
+          "The package name that will be displayed in the code snippets",
+        type: "string",
+      },
+      snippetFileOutputDir: { // what should the default be?
+        alias: "f",
+        describe: "Directory for generated code snippet files",
+        type: "string",
+        coerce: path.resolve,
+      },
```
davidpaquin: - Let's remove the aliases here, since I don't think they're going to be used extensively outside of other code calling the cli
- For the default output dir, I think conventionally it's just whatever the working directory is (`./`)
- `codeSnippetFiles` -> `generateCodeSnippets` and `snippetFileOutputDir` -> `codeSnippetDir`
- Should we throw if the packagename and outputdir are supplied but the switch is false? log a warning? no-op? up to you on what to do here

package.json
```diff
@@ -83,6 +85,7 @@
   },
   "packageManager": "pnpm@9.5.0+sha512.140036830124618d624a2187b50d04289d5a087f326c9edfc0ccd733d76c4f52c3a313d4fc148794a2a9d81553016004e6742e8cf850670268a7387fc220c903",
   "dependencies": {
+    "@osdk/maker": "^0.11.3",
```
davidpaquin: I don't think this is needed

packages/maker/src/api/defineOntology.ts
```diff
@@ -43,6 +44,15 @@ import type { OntologyDefinition } from "./common/OntologyDefinition.js";
...
 import type { OntologyEntityType } from "./common/OntologyEntityTypeMapping.js";
 
+// Added
+import { TYPESCRIPT_OSDK_SNIPPETS } from "@osdk/typescript-sdk-docs";
+import Mustache from "mustache";
+import type { InterfaceType } from "./interface/InterfaceType.js";
+import type { OneToManyLinkTypeDefinition } from "./links/LinkType.js";
+import type { ObjectType } from "./object/ObjectType.js";
+import { interfaceSnippets, objectSnippets } from "./snippetTypes.js";
+// import { curlYml, javaYml, pythonYml, unityYml } from "@foundry/documentation-snippets" // this package does not exist
```
davidpaquin: nit: remove the comments here

packages/maker/src/api/defineOntology.ts
```diff
@@ -594,3 +614,108 @@ addDependency("${namespaceNoDot}", new URL(import.meta.url).pathname);
 export function addNamespaceIfNone(apiName: string): string {
   return apiName.includes(".") ? apiName : namespace + apiName;
 }
+
+function convertCardinality(
```
davidpaquin: This function already exists in `convertCardinality.ts`, I assume it's just here from a semantic merge conflict, so you can remove it

packages/maker/src/cli/main.ts
```diff
@@ -152,13 +184,19 @@ async function loadOntology(
   apiNamespace: string,
   outputDir: string | undefined,
   dependencyFile: string | undefined,
+  generateCodeSnippets: boolean | undefined,
+  snippetPackageName: string | undefined,
+  codeSnippetDir: string | undefined,
```
davidpaquin: Because these have defaults we can drop the `| undefined`

davidpaquin (commented):
Looks good, one nit on typing then I'll approve

packages/maker/src/cli/main.ts
chengJazhua: Fixed!

PR #2132, opened 2025-10-31
Fix 2d array generation

ssanjay1:


packages/client/src/queries/applyQuery.ts
```diff
@@ -125,10 +125,8 @@ async function remapQueryResponse<
     }
   }
 
-  if (
-    responseDataType.multiplicity != null
-    && responseDataType.multiplicity
-  ) {
+  // TODO(bryantp) fix logic to not use multiplicity and unwrap inner type
+  if (responseDataType.multiplicity != null && responseDataType.multiplicity) {
```
nihalbhatnagar: Whats the blocker here on doing this rn?

bryantpark04: Or we should keep the multiplicity logic to maintain backcompat with SDKs generated with an older version? But I need to figure out how to make the lint step warn instead of error on deprecated, that was what was causing the failed checks earlier

PR #2133, opened 2025-10-31
add subscription support for useObjectSet

KuberSethi:
Adds subscription support for `useObjectSet` following a user request. Also refactors subscription logic out of ListQuery into BaseListQuery for reuse across different implementations.


KuberSethi:
> Looks fine however I am unsure if the subscription can work on links?

Yep you're right, removed them. Didn't quite understanding the backend limitation but dug in a bit and yeah sadly not supported.

PR #2141, opened 2025-11-06
Add Slate config and emit Slate events

zaini:


packages/widget.client/src/client.ts
```diff
@@ -134,6 +156,15 @@ export function createFoundryWidgetClient<
         },
       });
     },
+    emitSlateEvent: (eventId, payload) => {
```
zaini: Was having a lot of issues with types when trying to just reuse the existing `emitEvent` action.

packages/widget.api/src/messages/widgetMessages.ts
```diff
@@ -62,11 +82,20 @@ export namespace WidgetMessage {
   export interface EmitEvent<C extends WidgetConfig<C["parameters"]>>
     extends WidgetBaseMessage<"widget.emit-event", Payload.EmitEvent<C>>
   {}
+
+  /**
+   * Slate event payload that the widget sends to the host Foundry UI
+   */
+  export interface EmitSlateEvent<C extends WidgetConfig<C["parameters"]>>
+    extends
+      WidgetBaseMessage<"widget.emit-slate-event", Payload.EmitSlateEvent<C>>
```
tzyl: Wondering if we considered reusing `widget.emit-event` and moving the type union / guards onto the payload itself which may also make it easier to reuse a single `emitEvent` method and have the types block the code from actually sending a wrongly typed workshop vs slate event based on the config

packages/widget.api/src/manifest.ts
```diff
@@ -49,7 +50,9 @@ export interface WidgetSetManifestContentV1 {
   inputSpec?: WidgetSetInputSpecV1;
 }
 
-export interface WidgetManifestConfigV1 {
+export type WidgetManifestConfigType = "workshopWidgetV1" | "slateWidgetV1";
```
tzyl: I'm sure you're tracking the next step to also implement the manifest parsing on publish?

PR #2149, opened 2025-11-07
Use new ephemeral media upload

ssanjay1:
We now have a public API that allows us to upload media directly to an ephemeral storage place without needing to associated to a mediaset, object + property, or even an action!

We only expose the direct upload function through the functions package because within the context of an app, users will probably want to associate this with an object and you can only do that by running an action.

### DIFF
--- a/etc/functions.report.api.md
+++ b/etc/functions.report.api.md
@@ -217,7 +217,7 @@ export type TimestampISOString<T extends string = string> = T & {
 export { TwoDimensionalAggregation }
 
 // @public (undocumented)
-export function uploadMedia(_client: Client, mediaUpload: MediaUpload): Promise<MediaReference>;
+export function uploadMedia(client: Client, mediaUpload: MediaUpload): Promise<MediaReference>;
 
 // @public (undocumented)
 export interface UrlLinkTarget {


packages/faux/src/errors.ts
```diff
@@ -40,6 +40,8 @@ export function ObjectTypeDoesNotExistError(
   return {
     errorCode: "NOT_FOUND",
     errorName: "ObjectTypeNotFound",
+    errorDescription:
+      "The requested object type is not found, or the client token does not have access to it.",
```
KuberSethi: some of these descriptions include "client token", others mention it's due to "the user", just double checking this discrepancy is intentional

packages/faux/src/errors.ts
```diff
@@ -127,13 +139,16 @@ export function InvalidRequest(errorName: string): BaseAPIError {
 export const ApplyActionFailedError: ApplyActionFailed = {
   errorCode: "INVALID_ARGUMENT",
   errorName: "ApplyActionFailed",
+  errorDescription: "",
```
KuberSethi: Is this intentional?

nihalbhatnagar (approved):

bryantpark04 (approved):

Issue #1923, opened 2025-09-08
Throw exception if filters are not properly structured


tompersona:
Currently, if individual filters are included at the top level of a where clause alongside an and clause, those filters are ignored but the user is not notified.

For example, this –

```
{type: { $eq: "MyType" }, cancelled: { $eq: false }, $and: [/* date filters */] }
```

– succeeds but silently doesn't apply the `type` or `cancelled` filter.

Whereas this succeeds _and_ works as expected:

```
{ $and: [ { type: { $eq: "MyType" } }, { cancelled: { $eq: false } }, /* date filters */ ] }
```

Issue #2075, opened 2025-10-13
Filters should type error if more than one key in an and object


nihalbhatnagar:
The below should not work (ignore the RDP, not relevant could be any property) because we nest two ands (we will wrap the initial and, see there are multiple keys in the same object, and wrap it around another and). We should typeguard against this
```
  const result2 = await client(Country_1).withProperties({
    stateCount: (base) => base.pivotTo("stateTerritory").aggregate("$count"),
  }).where({$and: [{"airportCountryIsoCode": {$eq: "hey"}, "stateCount": {$eq: 2}}}]).fetchPage();
```

PR #2152, opened 2025-11-07
Fix where clauses where multiple entries in a top level filter

nihalbhatnagar:
Fixes https://github.com/palantir/osdk-ts/issues/1923. No type changes needed, this is already supported by our types. Potentially need a type change to address #2075

### DIFF
--- a/packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
+++ b/packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
@@ -99,6 +99,10 @@ export function modernToLegacyWhereClauseInner<
   objectOrInterface: T,
   rdpNames?: Set<string>,
 ): SearchJsonQueryV2 {
+  const parts = Object.entries(whereClause);
+
+  invariant(parts.length === 1, "Invalid where clause provided.");
+
   if (isAndClause(whereClause)) {
     return {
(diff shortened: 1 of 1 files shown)


packages/client/src/internal/conversions/modernToLegacyWhereClause.test.ts
```diff
@@ -991,32 +1069,36 @@ describe(modernToLegacyWhereClause, () => {
...
+              },
+              "type": "gte",
+              "value": 5,
             },
-            value: "John",
-          },
-        ],
-      });
+            {
+              "field": undefined,
```
nihalbhatnagar: This is fine, but this is why I added the invariant to check field and propertyIdentifer aren't defined together

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -63,14 +63,40 @@ function isNotClause<
   return "$not" in whereClause && whereClause.$not !== undefined;
 }
 
-/** @internal */
```
KuberSethi: was removing this intentional? seems its still internal unless i missed smth

KuberSethi (approved):

PR #2166, opened 2025-11-12
Expand available struct array filters

nihalbhatnagar:
All array properties can have filters applied based on its sub-type. This also applies to struct arrays, where you can apply the struct sub properties filter to any property on the array. We already added support for basic `$eq` filters, this PR expands the types and the parser.

### DIFF
--- a/packages/e2e.sandbox.catchall/src/runStructsTest.ts
+++ b/packages/e2e.sandbox.catchall/src/runStructsTest.ts
@@ -14,42 +14,46 @@
  * limitations under the License.
  */
 
-import { OsdkTestObject } from "@osdk/e2e.generated.catchall";
-import { client } from "./client.js";
+import {
+  BgaoNflPlayer,
+  McAirportStruct,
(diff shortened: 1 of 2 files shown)


packages/e2e.sandbox.catchall/src/runStructsTest.ts
```diff
@@ -14,53 +14,50 @@
...
-} from "@osdk/e2e.generated.catchall";
-import { client, dsClient } from "./client.js";
+import { OsdkTestObject } from "@osdk/e2e.generated.catchall";
+import { client } from "./client.js";
 
 export async function runStructsTest(): Promise<void> {
-  const player = await dsClient(BgaoNflPlayer).fetchOne(
-    "50A409AB-C909-453A-A61A-31B51324C8E3",
-  );
+  // const player = await dsClient(BgaoNflPlayer).fetchOne(
```
ssanjay1: Can we uncomment again

PR #2168, opened 2025-11-12
feature: expose foundry hostname util

overtonch:
Adds a utility to the functions package which allows clients to access the API gateway base URL from within a function.  This is intended to enable manually constructing a client within a function to use the Platform SDK.


.changeset/fuzzy-melons-go.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/functions": minor
```
ssanjay1: can we make this a patch please

ssanjay1 (approved):

PR #2170, opened 2025-11-12
Get latest instead of first snippet and update code snippet names

chengJazhua:
Originally only got the first existence of a snippet - change now gets the latest snippet

### DIFF
--- a/.changeset/cool-ghosts-prove.md
+++ b/.changeset/cool-ghosts-prove.md
@@ -1,5 +1,0 @@
----
-"@osdk/maker": patch
----
-
-Get latest snippet


packages/maker/src/api/code-snippets/createCodeSnippets.ts
```diff
@@ -114,12 +116,21 @@ function getSnippets(
...
-      Object.values(TYPESCRIPT_OSDK_SNIPPETS.versions).find(v =>
-        v.snippets[templateName]
-      )?.snippets[templateName][0].template ?? "";
+    const versions = Object.values(TYPESCRIPT_OSDK_SNIPPETS.versions);
+    const latestTemplate = versions
+      .slice()
+      .reverse()
+      .find(v => v.snippets[templateName])
+      ?.snippets[templateName]
+      .at(-1)?.template ?? "";
```
davidpaquin: you `.reverse()` but also index `-1` instead of `0`, is that going to have the opposite effect of what you want?

PR #2176, opened 2025-11-13
Pipe through transactionId to data loading endpoints when present

nihalbhatnagar:
Endpoints that load from an object set now take a transaction ID, and if the OSDK client is created with one, we want to pass it through to those requests

### DIFF
--- a/.changeset/fruity-yaks-exist.md
+++ b/.changeset/fruity-yaks-exist.md
@@ -10,4 +10,4 @@
 "@osdk/faux": patch
 ---
 
-Pipe through transactionId to OSS endpoints when present
+Pipe through transactionId to endpoints that load data via object sets


packages/client/src/createClient.ts
```diff
@@ -129,7 +129,7 @@ export function createClientInternal(
     {
       ...options,
       logger: options?.logger ?? new MinimalLogger(),
-      transactionRid: transactionRid,
+      transactionId: transactionRid,
```
nihalbhatnagar: This is still RID because I'll need to bump the functions runtime once I break this. I'll do it eventually just want to get e2e stuff working

packages/functions/src/transactions/createWriteableClient.ts
KuberSethi: Do we want to make the same primaryKey change here? Don't have full context but I'd presume we're reading the primary key from the property map here as well?

nihalbhatnagar: Nope, modify and delete edits still require an explicit primary key. We don't want to suggest for modify edits that you can change the PK of the object from the property map, and for delete edits, there is no property map

PR #2192, opened 2025-11-17
Update Media Upload docs

bryantpark04:


packages/typescript-sdk-docs/src/docs.ts
```diff
@@ -389,7 +389,7 @@ function renderType(
       return "{}";
     case "mediaReference":
       return context === "actionParameter"
-        ? "mediaReferenceOrUpload"
+        ? "mediaUpload"
```
bryantpark04: oddly this code never gets called when the local tests run

PR #2199, opened 2025-11-20
Validate widget sets before building their manifest

gpirindev:
Add a validation step when constructing the build manifest for a custom widget. Widget ids will be checked to confirm there are no duplicates.

### DIFF
--- a/.changeset/cold-hairs-fall.md
+++ b/.changeset/cold-hairs-fall.md
@@ -1,0 +1,5 @@
+---
+"@osdk/widget.vite-plugin": minor
+---
+
+Validate widget sets before building their manifest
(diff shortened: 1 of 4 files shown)


tzyl (dismissed):
If you haven't already can you also make sure to test this out in an actual widget project?

- Build the vite plugin e.g. `pnpm build --filter=@osdk/widget.vite-plugin`
- Manually change temporarily one of the example widget sets to include the duplicate widget id https://github.com/palantir/osdk-ts/tree/main/examples/example-widget-react-sdk-2.x/src and build e.g. `pnpm build --filter=@osdk/examples.widget-react-sdk-2.x`
- Or use an external repo and use the local copy of the vite plugin in package.json e.g. `"@osdk/widget.vite-plugin": "file:/...",`

PR #2216, opened 2025-11-28
[widgets] expand ErrorBoundary to catch all errors

amishshah:
Expand the error boundary to now catch all uncaught errors and give users tips on what to improve


packages/widget.client-react/src/ErrorBoundary.tsx
```diff
@@ -43,17 +43,51 @@ export class ErrorBoundary extends React.Component<Props, State> {
...
-          <h3>Widget failed to start</h3>
-          <pre>
-            {this.state.error instanceof Error
-              ? this.state.error.stack
-              : "See browser console for more details."}
+        <section style={{ padding: "16px" }}>
+          <h3 style={{ margin: "0 0 12px 0", color: "#c00" }}>
+            {this.state.caughtBeforeReady
+              ? "Widget failed to start"
+              : "An uncaught error occurred"}
```
tzyl: Hmm, do you think there's value in differentiating to the developer "Widget failed to start"? With the error boundary now doesn't the widget still start in the sense that we never block the ready event being sent?

tzyl (approved):

PR #2222, opened 2025-12-01
Widget resize message

tzyl:
This PR introduces a new `widget.resize` message type for notifying the host application when the content size of the widget changes so it can sync the iframe height for auto sizing. The widget react library automatically handles setting up a resize observer on the widget document body to send this message.

**NOTE:**

This is targeted at requiring the widget developer to ensure their content affects the height of the document body element their widget is rendered in which I think is a fair expectation to have.

There are other cases a widget could be styled such as some of our templates had `height: 100%` to try to fit to the space given to the widget. This is harder to detect comprehensively as the true height of the content might be the scroll height of the body, or it could be the (scroll) height of an inner container for which a resize event on the top level may not occur and you may have to best effort detect this with a combination like a mutation observer ([stackoverflow](https://stackoverflow.com/questions/44428370/detect-scrollheight-change-with-mutationobserver)).

After thinking about it I think keeping the contract clear and simple is best with the widget requiring its content to affect the body height in order to be automatically resized.

### DIFF
--- a/.changeset/chilly-wasps-slide.md
+++ b/.changeset/chilly-wasps-slide.md
@@ -1,0 +1,9 @@
+---
+"@osdk/create-widget.template.minimal-react.v2": patch
+"@osdk/create-widget.template.react.v2": patch
+"@osdk/widget.client-react": patch
+"@osdk/widget.client": patch
+"@osdk/widget.api": patch
+---
+
+Introduce widget resize message for automatic heights


packages/create-widget.template.minimal-react.v2/templates/src/main.css
```diff
@@ -1,4 +1,3 @@
 html, body {
-    height: 100%;
```
tzyl: I removed this so that both templates start with taking up as much space as their content which will work better with the default layout of auto in workshop

PR #2225, opened 2025-12-01
refactor OSDK React docs

KuberSethi:
Refactors OSDK React docs into separate files per workflow/use case and also moves them to the docs root folder. Also adds an index and config file to the root docs folder such that we publish to Github pages (this should already be configured for our repo in https://github.com/palantir/osdk-ts/settings/pages)


ssanjay1 (approved):
Rubber stamping!

PR #2234, opened 2025-12-03
[create-app] Add CLI support for bootstrapping without an OSDK

samialfhaily:
## Changes
- Add CLI support for bootstrapping a project without an OSDK
- Add examples and tests for projects without an OSDK

Supported templates:
- react.v2
- vue.v2
- expo.v2

## Screenshots
Screenshot 2025-12-04 at 11 54 43 AM
Screenshot 2025-12-04 at 11 55 32 AM
Screenshot 2025-12-04 at 1 16 19 PM

### DIFF
--- a/examples/example-react-sdk-2.x-no-osdk/README.md
+++ b/examples/example-react-sdk-2.x-no-osdk/README.md
@@ -5,7 +5,7 @@ This project was generated with [`@osdk/create-app`](https://www.npmjs.com/packa
 To quickly create your own version of this template run the following command and answer the prompts based on your Developer Console application:
 
 ```
-npm create @osdk/app@latest -- --template react --sdkVersion 2.x
+npm create @osdk/app@latest -- --template react --sdkVersion 2.x --skipOsdk
 ```
 
 Alternatively check out the Developer Console docs for a full guide on creating and deploying frontend applications with the Ontology SDK.
--- a/packages/example-generator/src/run.ts
+++ b/packages/example-generator/src/run.ts
@@ -42,23 +42,51 @@ interface RunArgs {
 export async function run({ outputDirectory, check }: RunArgs): Promise<void> {
   const resolvedOutput = path.resolve(outputDirectory);
   const tmpDir = createTmpDir();
-  await generateCreateAppExamples(tmpDir);
+
+  await generateCreateAppExamplesWithOsdk(tmpDir);
+  await generateCreateAppExamplesWithoutOsdk(tmpDir);
   await generateCreateWidgetExamples(tmpDir);
+
   await fixMonorepolint(tmpDir);
-  if (check) {
-    checkExamples(resolvedOutput, tmpDir, [
+
+  const templatesWithSdkVersionWithOsdk = templatesWithSdkVersions(
+    [
       ...TEMPLATES,
       ...WIDGET_TEMPLATES,
-    ], true);
-    checkExamples(resolvedOutput, tmpDir, [...TEMPLATES], false);
+    ],
+    true,
+  );
+  const templatesWithSdkVersionWithoutOsdk = templatesWithSdkVersions(
+    TEMPLATES,
+    false,
+  );
+
+  if (check) {
+    checkExamples(
+      resolvedOutput,
+      tmpDir,
+      templatesWithSdkVersionWithOsdk,
+      true,
+    );
+    checkExamples(
+      resolvedOutput,
+      tmpDir,
+      templatesWithSdkVersionWithoutOsdk,
+      false,
+    );
   } else {
     copyExamples(
       resolvedOutput,
       tmpDir,
-      [...TEMPLATES, ...WIDGET_TEMPLATES],
+      templatesWithSdkVersionWithOsdk,
       true,
     );
-    copyExamples(resolvedOutput, tmpDir, [...TEMPLATES], false);
+    copyExamples(
+      resolvedOutput,
+      tmpDir,
+      templatesWithSdkVersionWithoutOsdk,
+      false,
+    );
   }
 }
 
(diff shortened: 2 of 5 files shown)


packages/create-app.template.react.beta/templates/src/client.ts.hbs
```diff
@@ -1,5 +1,10 @@
+{{#if osdkPackage}}
 import { createClient, type Client } from "@osdk/client";
 import { createPublicOauthClient, type PublicOauthClient } from "@osdk/oauth";
+{{else}}
+import { createPlatformClient, type PlatformClient } from "@osdk/client";
+import { createPublicOauthClient, type PublicOauthClient } from "@osdk/oauth";
```
zeyadkhaled: This line can be out of the if/else clause given its identical in both

packages/create-app.template.react.beta/templates/src/components/NextSteps.tsx.hbs
```diff
@@ -2,6 +2,14 @@ import React from "react";
 import css from "./NextSteps.module.css";
 
 const STEPS: INextStep[] = [
+  {{^if osdkPackage}}
+  {
+    icon: "cube",
+    title: "Configure your Ontology and Platform SDKs",
+    subtitle:
+      "Use the Ontology SDK to interact with your Ontology and the Platform SDK to access Foundry APIs. Configure your SDKs from the Ontology SDK and Platform SDK tabs in Developer Console.",
```
zeyadkhaled: I am wondering if referencing Ontology SDK here is confusing given that we have updated the imports and comments here? https://github.com/palantir/osdk-ts/pull/2234/files?new_files_changed=true#diff-28224c1f27733a82e65d0244f9ac097714db2ca98f87ef634e4588a386a4cf24R58

zeyadkhaled (commented):
[1/n] Templates review. I think template changes look mostly good with few comments.

packages/create-app/src/cli.test.ts
```diff
@@ -62,20 +67,74 @@ describe.each(TEMPLATES)("template $id", (template) => {
...
+        skipOsdk: true,
+      });
+    });
   },
+);
+
+test(`CLI rejects no OSDK with 1.x`, async () => {
+  await expect(runTest({
+    project: `expected-${TEMPLATES.filter(template => !template.hidden)[0].id}`,
+    template: TEMPLATES.filter(template => !template.hidden)[0],
```
zeyadkhaled: This is a bit hard to follow. Can you extract the specific template to a meaningful variable and reuse it for project and template?

packages/example-generator/src/run.ts
```diff
@@ -57,20 +68,33 @@ function createTmpDir(): tmp.DirResult {
...
+  isUsingOsdk: boolean,
 ) {
-  return templates.flatMap((template) =>
+  let templatesWithSdkVersions = templates.flatMap((template) =>
     Object.keys(template.files).map((sdkVersion) =>
       [template, sdkVersion as SdkVersion] as const
     )
   );
+
+  if (!isUsingOsdk) {
```
zeyadkhaled: Can you add a comment here explaining the filtering we are doing here and why we are omitting 1.x

packages/example-generator/src/run.ts
```diff
@@ -259,18 +339,41 @@ function getContents(aPath: string | null) {
...
+    const [template, sdkVersion] of templatesWithSdkVersions(
+      TEMPLATES,
+      false,
+    )
+  ) {
+    const exampleId = sdkVersionedTemplateExampleId(
+      template,
+      sdkVersion,
+      false,
+    );
```
zeyadkhaled: I don't understand why are we calling a block with isUsingOsdk and the other with false? So whats the use of having the `isUsingOsdk` flag at all why not pass true and false?

packages/create-app.template.react.beta/templates/src/main.tsx.hbs
```diff
@@ -1,12 +1,22 @@
+{{#if osdkPackage}}
 import { OsdkProvider } from "@osdk/react";
+{{/if}}
 import ReactDOM from "react-dom/client";
 import { RouterProvider } from "react-router-dom";
 import "./index.css";
+{{#if osdkPackage}}
 import client from "./client";
```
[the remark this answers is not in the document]
samialfhaily: It's intentionally placed here because the first `{{#if osdkPackage}}` is a package import and the second one is a relative import.

packages/create-app.template.react.beta/templates/src/components/NextSteps.tsx.hbs
samialfhaily: This is meant to tell people that they can configure their own Ontology SDK in Developer Console where we would also show them how to add it to their project. I think it's a useful—at worst, harmless—step.

examples/example-react-sdk-2.x-no-osdk/README.md
```diff
@@ -0,0 +1,11 @@
+# example-react
+
+This project was generated with [`@osdk/create-app`](https://www.npmjs.com/package/@osdk/create-app) from the `react` template. It is built against a locally generated SDK and a non-existent Foundry stack, so it is intended for reference purposes only.
+
+To quickly create your own version of this template run the following command and answer the prompts based on your Developer Console application:
+
+```
+npm create @osdk/app@latest -- --template react --sdkVersion 2.x
```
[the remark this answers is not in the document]
samialfhaily: That's not necessary. There's a prompt where the user can select "I don't want to use an OSDK". This also matches the same behavior of `applicationUrl` where there's a prompt where the user can say "I'll fill it out later" and we don't pass `--skipApplicationUrl` here

packages/example-generator/src/run.ts
```diff
@@ -98,9 +122,48 @@ async function generateExamples(tmpDir: tmp.DirResult): Promise<void> {
...
+      template,
+      sdkVersion,
+      foundryUrl: "https://fake.palantirfoundry.com",
+      applicationUrl: "https://example.com",
+      application: "ri.third-party-applications.main.application.fake",
+      ontology: undefined,
+      clientId: "123",
+      osdkPackage: undefined,
+      osdkRegistryUrl:
+        "https://fake.palantirfoundry.com/artifacts/api/repositories/ri.artifacts.main.repository.fake/contents/release/npm",
```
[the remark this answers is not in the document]
samialfhaily: Not necessarily, but it can still be passed in. I've removed it from here though.

examples/example-react-sdk-2.x-no-osdk/README.md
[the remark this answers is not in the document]
zeyadkhaled: In this case though, this example's sole purpose is to illustrate the `--skipOntology` part of the CLI. At least as a user coming to check this repository and looking at the examples, I would assume this to ReadMe to showcase this difference.

zeyadkhaled (commented):
Let's update the README of create-app https://github.com/palantir/osdk-ts/blob/aa18e33042781874530abc8bf4e6bbe36ea76726/packages/create-app/README.md to include the new args/update existing ones that became optional.

packages/example-generator/src/run.ts
```diff
@@ -42,12 +42,23 @@ interface RunArgs {
...
+    checkExamples(resolvedOutput, tmpDir, [...TEMPLATES], false);
   } else {
-    await copyExamples(resolvedOutput, tmpDir);
+    copyExamples(
+      resolvedOutput,
+      tmpDir,
+      [...TEMPLATES, ...WIDGET_TEMPLATES],
+      true,
+    );
+    copyExamples(resolvedOutput, tmpDir, [...TEMPLATES], false);
```
zeyadkhaled: I think the code doesn't read as clean as it did before and the concept of no-osdk templates is spread implicitly all over the code. We could do pre-filtering on the templates before this code block and split them into (templates using osdk and templates not using osdk). Then pass those respectively in each of the function calls. 

That way the checkExamples and copyExamples function doesn't have this additional logic it shouldn't need to know about and we won't need to keep passing true,false down the chain.

packages/example-generator/src/run.ts
```diff
@@ -166,18 +229,29 @@ async function fixMonorepolint(tmpDir: tmp.DirResult): Promise<void> {
...
+  const mrlPathsWithoutOsdk = templatesWithSdkVersions(TEMPLATES, false)
+    .map((
+      [template, sdkVersion],
+    ) =>
+      path.join(
+        tmpDir.name,
+        sdkVersionedTemplateExampleId(template, sdkVersion, false),
+        "package.json",
+      )
+    );
```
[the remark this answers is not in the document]
zeyadkhaled: May be I am miss understanding an implicit logic, but if true is passed we simply return all examples. This [block](https://github.com/palantir/osdk-ts/pull/2234/files#diff-8350165c025c2245f683642f527e263a0fe9c79ca99a205ff55831f230bf1f99R81-R87) just restricts the result down.

packages/example-generator/src/run.ts
```diff
@@ -57,20 +68,34 @@ function createTmpDir(): tmp.DirResult {
...
+    templatesWithSdkVersions = templatesWithSdkVersions.filter((
+      [template, sdkVersion],
+    ) => !template.hidden && sdkVersion === "2.x");
+  }
+
+  return templatesWithSdkVersions;
 }
 
-async function generateExamples(tmpDir: tmp.DirResult): Promise<void> {
+async function generateCreateAppExamples(
```
zeyadkhaled: I like how you split this into two function which aided with readability. But `generateCreateAppExamples` is now duplicating some work. Let's have a shared "internal" generateCreateAppExamples and introduce a generateCreateAppExamplesWithOsdk and generateCreateAppExamplesWithoutOsdk and that calls the common one with the extra flag passed. That way it keeps this function smaller and easier for reader to understand the different paths.

zeyadkhaled (commented):
Everything looks good overall. I still think we could polish `packages/example-generator/src/run.ts` more and make it easier to read, so I left some suggestions.

PR #2241, opened 2025-12-04
Create foundryWidgetTokenProvider helper

tzyl:
Currently it is expected for the user code to provide a dummy token provider when constructing OSDK clients:

```
import { createClient } from "@osdk/client";
import { $ontologyRid } from "@custom-widget/sdk";

const client = createClient(
    window.location.origin,
    $ontologyRid,
    () => Promise.resolve("widgets-auth"),
);
```

This PR provides a helper as some syntax sugar to avoid user code needing to be aware of this:

```
import { createClient } from "@osdk/client";
import { createFoundryWidgetTokenProvider } from "@osdk/widget.client";
import { $ontologyRid } from "@custom-widget/sdk";

const client = createClient(
    window.location.origin,
    $ontologyRid,
    createFoundryWidgetTokenProvider(),
);
```

Before this can be used in the templates it'll need to be released

### DIFF
--- a/.changeset/wild-streets-sneeze.md
+++ b/.changeset/wild-streets-sneeze.md
@@ -2,4 +2,4 @@
 "@osdk/widget.client": patch
 ---
 
-Create foundryWidgetTokenProvider helper
+Create createFoundryWidgetTokenProvider helper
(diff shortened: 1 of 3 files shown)


amishshah (dismissed):
~Should we also update the OSDK template?~

Nevermind, just saw your description!

PR #2256, opened 2025-12-08
[Components] Setup @osdk/components package

ziyunp:
- Setup @osdk/components package
- Added APIs for ObjectTable and FilterList

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1886,37 +1886,6 @@ importers:
         specifier: ~5.5.4
         version: 5.5.4
 
-  packages/components:
-    dependencies:
-      react-dom:
-        specifier: ^17 || ^18 || ^19
-        version: 18.3.1(react@18.3.1)
-    devDependencies:
-      '@osdk/api':
-        specifier: workspace:*
-        version: link:../api
-      '@osdk/client':
-        specifier: workspace:*
-        version: link:../client
-      '@osdk/monorepo.api-extractor':
-        specifier: workspace:~
-        version: link:../monorepo.api-extractor
-      '@osdk/monorepo.tsconfig':
-        specifier: workspace:~
-        version: link:../monorepo.tsconfig
-      '@osdk/react':
-        specifier: workspace:*
-        version: link:../react
-      '@types/react':
-        specifier: ^18.3.24
-        version: 18.3.24
-      react:
-        specifier: ^18.3.1
-        version: 18.3.1
-      typescript:
-        specifier: ~5.5.4
-        version: 5.5.4
-
   packages/create-app:
     dependencies:
       consola:
@@ -4121,6 +4090,37 @@ importers:
       tiny-invariant:
         specifier: ^1.3.3
         version: 1.3.3
+      typescript:
+        specifier: ~5.5.4
+        version: 5.5.4
+
+  packages/react-components:
+    dependencies:
+      react-dom:
+        specifier: ^17 || ^18 || ^19
+        version: 18.3.1(react@18.3.1)
+    devDependencies:
+      '@osdk/api':
+        specifier: workspace:*
+        version: link:../api
+      '@osdk/client':
+        specifier: workspace:*
+        version: link:../client
+      '@osdk/monorepo.api-extractor':
+        specifier: workspace:~
+        version: link:../monorepo.api-extractor
+      '@osdk/monorepo.tsconfig':
+        specifier: workspace:~
+        version: link:../monorepo.tsconfig
+      '@osdk/react':
+        specifier: workspace:*
+        version: link:../react
+      '@types/react':
+        specifier: ^18.3.24
+        version: 18.3.24
+      react:
+        specifier: ^18.3.1
+        version: 18.3.1
       typescript:
         specifier: ~5.5.4
         version: 5.5.4


packages/components/src/filter-list/FilterListItemApi.ts
```diff
@@ -0,0 +1,193 @@
...
+> {
+  /**
+   * The property key to filter on
+   */
+  key: K;
+
+  /**
+   * Display label for the filter
+   */
+  label: string;
```
KuberSethi: Should this be required?

packages/components/src/filter-list/FilterListItemApi.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  onFilterStateChanged: (state: FilterStateByComponentType[C]) => void;
+
+  onFilterRemoved?: (key: PropertyKeys<Q>) => void;
+}
+
+/**
+ * A filter definition specifies configuration for a single filter
+ *
+ * The component type C must be compatible with the property type derived from the key.
```
KuberSethi: What happens if this isn't the case? Is that worth documenting?

packages/components/src/filter-list/FilterListItemApi.ts
```diff
@@ -0,0 +1,193 @@
...
+export interface DateRangeFilterState extends BaseFilterState {
+  type: "DATE_RANGE";
+  /**
+   * Minimum date value (inclusive)
+   */
+  minValue?: Date;
+  /**
+   * Maximum date value (inclusive)
+   */
+  maxValue?: Date;
```
KuberSethi: nit: Is there a reason we're keeping both min and max inclusive? Maybe I'm biased but I always thought min should be inclusive while max exclusive when it comes to dates.

packages/components/src/object-table/ObjectTableApi.ts
```diff
@@ -0,0 +1,209 @@
...
+  orderable?: boolean;
+  /**
+   * The current order by clause to sort the objects in the table.
+   *
+   * If not supplied, sorting is managed internally, else `onOrderByChanged` is required.
+   */
+  orderBy?: Array<{
+    property: PropertyKeys<Q>;
+    direction: "asc" | "desc";
+  }>;
```
KuberSethi: nit: insert newline below

packages/components/src/filter-list/FilterListItemApi.ts
```diff
@@ -0,0 +1,193 @@
...
+  | "LISTOGRAM"
+  | "DATE_RANGE"
+  | "NUMBER_RANGE"
+  | "TEXT_TAGS"
+  | "CONTAINS_TEXT";
+
+/**
+ * Gets valid component types for a given property type
+ */
+export type ValidComponentsForPropertyType<P extends WirePropertyTypes> =
```
KuberSethi: Just to confirm, we're good to return never for the other `WirePropertyTypes` right?

packages/components/src/object-table/ObjectTableApi.ts
```diff
@@ -0,0 +1,209 @@
...
+  /**
+   * The set of objects to show in the table
+   */
+  objectSet: ObjectSet;
+  /**
+   * Ordered list of column definitions to show in the table
+   *
+   * If not provided, all of the properties of the object type will be shown in default order.
+   */
+  columnDefinitions?: Array<ColumnDefinition<Q, RDPs, FunctionColumns>>;
```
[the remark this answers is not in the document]
KuberSethi: What is default order? I think if we're being opinionated about the ordering we should state what ordering we're providing is

packages/components/vitest.config.mts
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { configDefaults, defineConfig } from "vitest/config";
+
+export default defineConfig({
```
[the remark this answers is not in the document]
KuberSethi: I think if we're not actively using it we should delete it but it's no big deal if we're planning on adding tests soon

packages/components/src/object-table/ObjectTableApi.ts
```diff
@@ -0,0 +1,224 @@
...
+   * Required when sorting is controlled.
+   *
+   * @param newOrderBy The new order by clause
+   */
+  onOrderByChanged?: (
+    newOrderBy: Array<{
+      property: PropertyKeys<Q>;
+      direction: "asc" | "desc";
+    }>,
+  ) => void;
```
KuberSethi: nit: insert newline below

packages/components/src/object-table/ObjectTableApi.ts
```diff
@@ -0,0 +1,224 @@
...
+  /**
+   * Called when a column is resized.
+   *
+   * @param columnId The ID of the resized column
+   * @param newWidth The new width of the column
+   */
+  onColumnResize?: (
+    columnId: PropertyKeys<Q> | keyof RDPs | keyof FunctionColumns,
+    newWidth: number,
+  ) => void;
```
KuberSethi: nit: insert newline below

KuberSethi (approved):

PR #2289, opened 2025-12-16
[Components] Object table implementation

ziyunp:
Feature: A basic ObjectTable that 
- renders the first 50 rows (the page size) of the objectSet and fetches more rows when scrolled to the end
- renders rows with rowVirtualizer where row height is fixed (?)
- displays columns based on columnDefinitions (obj properties and derived properties are supported)
- displays all object property in the columns if columnDefinitions not provided
- calls onRowClick when clicking on a row

### DIFF
--- a/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
+++ b/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
@@ -31,7 +31,7 @@ const columnDefinitions: ColumnDefinition<
     locator: { type: "property", id: "firstFullTimeStartDate" },
     width: 300,
     renderHeader: () => "Start Date",
-    renderCell: (object: Osdk.Instance<Employee>, _locator) => {
+    renderCell: (object: Osdk.Instance<Employee>) => {
       return (
         <div>
           {object["firstFullTimeStartDate"]
@@ -50,7 +50,7 @@ const columnDefinitions: ColumnDefinition<
         baseObjectSet.pivotTo("lead").selectProperty("fullName"),
     },
     renderHeader: () => "Derived Manager Name",
-    renderCell: (object: Osdk.Instance<Employee>, _locator) => {
+    renderCell: (object: Osdk.Instance<Employee>) => {
       if ("managerName" in object) {
         return object["managerName"] as string;
       }


packages/components/src/object-table/README.md
```diff
@@ -0,0 +1,64 @@
+# ObjectTable Component
+
+A headless table component for displaying OSDK object sets using React TanStack Table.
```
KuberSethi: If we're interested in adding documentation alongside these components it might be better to do so in the docs directory in the root level!

KuberSethi (commented):
Generally looks good, need to rebase I think though!

packages/e2e.sandbox.peopleapp/src/generatedNoCheck2/OntologyMetadata.ts
ziyunp: @KuberSethi Should we add this file to .gitignore?

packages/react-components/src/object-table/TableBody.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface TableBodyProps<TData extends RowData> {
+  rows: Array<Row<TData>>;
+  tableContainerRef: React.RefObject<HTMLDivElement>;
+}
+
+export function TableBody<TData extends RowData>({
+  rows,
+  tableContainerRef,
+}: TableBodyProps<TData>): React.ReactElement {
+  // TODO: Allow user to pass in a custom row height?
```
ziyunp: Q: Do we want to add a rowHeight prop?

packages/react-components/src/object-table/TableHeader.tsx
```diff
@@ -0,0 +1,67 @@
...
+    >
+      {headerGroups.map((headerGroup) => (
+        <tr
+          key={headerGroup.id}
+          style={{
+            display: "flex",
+          }}
+        >
+          {headerGroup.headers.map((header) => (
+            // Q: Should we restrict inline styling?
```
ziyunp: Styling will be handled in a separate PR. Doing inline styling in this PR for now, and will move these to the css file later

packages/react-components/src/object-table/hooks/useDefaultTableStates.ts
```diff
@@ -0,0 +1,89 @@
...
+  >();
+
+  useEffect(() => {
+    if (columnDefinitions) {
+      const colVisibility: VisibilityState = columnDefinitions.reduce(
+        (acc, colDef) => {
+          if (colDef.isVisible !== undefined) {
+            const { locator } = colDef;
+            // TODO: Extract a helper function to get colKey from locator
+            const colKey = locator.type === "property"
```
ziyunp: The TODO is not required if we standardise to use "id"

packages/react-components/src/public/experimental.ts
```diff
@@ -17,4 +17,8 @@
 export type { FilterListProps } from "../filter-list/FilterListApi.js";
 export type { FilterListItemProps } from "../filter-list/FilterListItemApi.js";
 
-export type { ObjectTableProps } from "../object-table/ObjectTableApi.js";
+export { ObjectTable } from "../object-table/ObjectTable.js";
```
ziyunp: For future, I'm considering to export the plain Table component in case anyone wants to create their own table instance from useReactTable hook. They would benefit from the themable table and all the default table behaviors

packages/react-components/src/object-table/hooks/useColumnDefs.ts
```diff
@@ -0,0 +1,123 @@
...
+          header: renderHeader ?? propertyMetadata?.displayName,
+          size: width,
+          minSize: minWidth,
+          maxSize: maxWidth,
+          enableResizing: resizable,
+          enableSorting: orderable,
+          enableColumnFilter: filterable,
+          cell: (cellContext) => {
+            const object: Osdk.Instance<Q> = cellContext.row.original;
+            // TODO: Should we rename object to "row" and pass the cellValue instead of locator as locator is not providing much value?
```
KuberSethi: This seems fine to do and I believe is also in-line with typical table implementations in other libraries

packages/react-components/src/object-table/TableBody.tsx
KuberSethi: Yeah I think this is likely going to be asked for so we should add one. I'm ok leaving this as a FLUP item though!

packages/react-components/src/object-table/hooks/useColumnDefs.ts
```diff
@@ -0,0 +1,123 @@
...
+    return Object.entries(metadata?.properties).map(([key, property]) => {
+      const colDef: AccessorColumnDef<
+        Osdk.Instance<Q>
+      > = {
+        accessorKey: key,
+        header: property.displayName ?? key,
+      };
+      return colDef;
+    });
+  }, [metadata?.properties]);
```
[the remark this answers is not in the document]
KuberSethi: Also this `useMemo` is quite long, thoughts on breaking its contents up into helper functions?

packages/react-components/src/object-table/TableBody.tsx
ziyunp: Agree to leave this as FLUP. I was also thinking if this value can be read from the css variable. 

Claude suggests this:
```
const ROW_HEIGHT = parseInt(
  getComputedStyle(document.documentElement)
    .getPropertyValue('--object-table-row-height')
    .trim()
) || 40; // fallback to 40 if CSS variable is not defined
```

packages/react-components/src/object-table/TableHeader.tsx
```diff
@@ -0,0 +1,67 @@
...
+    >
+      {headerGroups.map((headerGroup) => (
+        <tr
+          key={headerGroup.id}
+          style={{
+            display: "flex",
+          }}
+        >
+          {headerGroup.headers.map((header) => (
+            // TODO: Move inline styling to CSS file
```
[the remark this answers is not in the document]
ziyunp: Agree, I'll leave this as inline until we decided on a styling method

packages/react-components/src/object-table/ObjectTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+      objectType,
+      columnDefinitions,
+    );
+
+  const { columnVisibility } = useDefaultTableStates({ columnDefinitions });
+
+  const table = useReactTable<
+    Osdk.Instance<Q>
+  >({
+    data: (data ?? []) as Array<Osdk.Instance<Q>>,
```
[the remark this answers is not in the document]
ziyunp: The cast is needed because useObjectSet is returning data is of type Osdk.Instance<
      Q,
      **"$allBaseProperties",**
      PropertyKeys<Q>,
      RDPs
    >.
and that means all our table types will need to add "$allBaseProperties" to the Osdk.Instance template type. I'm not sure if we want to expose that in the table API?

packages/react-components/src/object-table/hooks/useColumnDefs.ts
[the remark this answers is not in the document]
ziyunp: Good spot, I have to set up my editor to flag the missing deps.

I'll add helper funcs too

packages/react-components/src/object-table/hooks/useDefaultTableStates.ts
```diff
@@ -0,0 +1,89 @@
...
+              ? locator.propertyKey
+              : locator.id as string;
```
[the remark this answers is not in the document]
ziyunp: The issue is that locator.id has a union type of PropertyKeys<Q> | keyof FunctionColumns | keyof RDPs, which TypeScript sees as string | number | symbol since it's a keyof of generic types.

packages/react-components/src/object-table/TableBody.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <tbody
+      style={{
+        display: "grid",
+        position: "relative",
+        height: `${rowVirtualizer.getTotalSize()}px`,
+      }}
+    >
+      {rowVirtualizer.getVirtualItems().map((virtualRow) => {
+        const row = rows[virtualRow.index] as Row<TData>;
```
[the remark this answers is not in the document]
ziyunp: Good catch, it's not needed at all. Removed typecast

packages/react-components/src/object-table/hooks/__tests__/useObjectTableData.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+    });
+  });
+
+  it("calls useObjectSet without withProperties when no columnDefinitions provided", () => {
+    const { result } = renderHook(
+      () => useObjectTableData(mockObjectSet, undefined),
+      { wrapper },
+    );
+
+    expect((result.current as any)._testOptions.withProperties).toBeUndefined();
```
KuberSethi: hmm i'm okay manually casting to make scaffolding fixtures easier but i think we should hard cast assertions more cautiously. Is this needed? Here and in the other places we do this

packages/react-components/src/object-table/TableRow.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function TableRow<TData extends RowData>({
+  row,
+  virtualRow,
+  onRowClick,
+}: TableRowProps<TData>): React.ReactElement {
+  const handleClick = useCallback(() => {
+    onRowClick?.(row.original);
+  }, []);
```
KuberSethi:
```suggestion
  }, [onRowClick, row.original]);
```

packages/react-components/src/object-table/Table.tsx
```diff
@@ -0,0 +1,88 @@
...
+          }
+        }
+      }
+    },
+    [fetchNextPage, isLoading],
+  );
+
+  const handleScroll = useCallback(
+    async (e: React.UIEvent<HTMLDivElement>) => {
+      await fetchMoreOnEndReached(e.currentTarget);
```
KuberSethi: Have we tested this btw? We might also want to debounce scroll events in case there's a ton generated

ziyunp: Yes I have tested that with a smaller table height and page size of 2 and it works fine

packages/react-components/src/object-table/hooks/__tests__/useObjectTableData.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+} from "@osdk/api";
+import type { Client } from "@osdk/client";
+import { OsdkProvider } from "@osdk/react";
+import { renderHook } from "@testing-library/react";
+import * as React from "react";
+import { describe, expect, it, vi } from "vitest";
+import type { ColumnDefinition } from "../../ObjectTableApi.js";
+import { useObjectTableData } from "../useObjectTableData.js";
+
+// Mock the useObjectSet hook from @osdk/react/experimental
```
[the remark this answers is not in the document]
ziyunp: yeah no, will remove

packages/react-components/src/object-table/TableRow.tsx
ziyunp: Added the exhaustive-deps rule to the package, can't risk this kind of mistake

packages/react-components/src/object-table/TableBody.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+export function TableBody<TData extends RowData>({
+  rows,
+  tableContainerRef,
+}: TableBodyProps<TData>): React.ReactElement {
+  // TODO: Allow user to pass in a custom row height?
+  const ROW_HEIGHT = 40;
+
+  // Important: Keep the row virtualizer in the lowest component possible to avoid unnecessary re-renders.
```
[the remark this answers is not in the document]
ziyunp: I left that comment intentionally so that it's not being moved to other components

packages/react-components/src/object-table/ObjectTable.tsx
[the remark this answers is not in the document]
ziyunp: I updated the Osdk.Instance type to include $allBaseProperties

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
```diff
@@ -0,0 +1,616 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectTypeDefinition, Osdk, PropertyKeys } from "@osdk/api";
+import type { Client } from "@osdk/client";
+import { OsdkProvider } from "@osdk/react";
+import type { AccessorKeyColumnDef } from "@tanstack/react-table";
+import { renderHook, waitFor } from "@testing-library/react";
+import pDefer from "p-defer";
```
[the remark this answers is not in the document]
KuberSethi: Yup that's fine then! Thanks for checking

PR #2294, opened 2025-12-16
Improved object set writes

amishshah:
You can now call e.g.
```ts
emitEvent("updateObjectSet", {
  parameterUpdateIds: {
    myObjectSetA: newObjectSetDefinition,
  }
}
```

And we will create the temporary object set for you and pass that along the wire as `objectSetRid` so you don't have to do it manually.

### DIFF
--- a/packages/widget.client-react/package.json
+++ b/packages/widget.client-react/package.json
@@ -34,6 +34,7 @@
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
+    "test": "vitest run",
     "transpileBrowser": "monorepo.tool.transpile -f esm -m normal -t browser",
     "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
     "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
--- a/packages/widget.client/package.json
+++ b/packages/widget.client/package.json
@@ -34,6 +34,7 @@
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "fix-lint": "eslint . --fix && dprint fmt --config $(find-up dprint.json)",
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
+    "test": "vitest run",
     "transpileBrowser": "monorepo.tool.transpile -f esm -m normal -t browser",
     "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
     "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",


packages/widget.client-react/src/utils/transformEmitEventPayload.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (osdkClient == null) {
+          throw new Error(
+            `Cannot emit event "${eventId.toString()}" with ObjectSet parameter "${paramId}" without an osdk client`,
+          );
+        }
+        const objectSetRid = await createAndFetchTempObjectSetRid(
+          osdkClient,
+          paramValue as ObjectSet,
+        );
+        return [paramId, { objectSetRid }];
```
tzyl: Can we at least type the `{ objectSetRid }` part to check that it matches the original widget.api type? I think for the other ones as there a direct pass through it doesn't matter as much but it feels like here could easily make a mistake if it didn't actually match the correct shape

packages/widget.client-react/src/context.ts
```diff
@@ -15,22 +15,76 @@
...
+  payload: AugmentedEmitEventIdMap<C>[EventId<C>];
+}
+
+/**
+ * Emits a widget event with parameter updates to the host Foundry UI.
+ *
+ * For object set parameters, pass `ObjectSet<T>` directly.
+ * If the same event is emitted multiple times in quick succession, only the last call will be sent.
+ *
+ * Errors (e.g., network failures when materializing ObjectSet parameters)
```
tzyl: let's avoid confusion with other meaning of materialize object set
```suggestion
 * Errors (e.g., network failures when serializing ObjectSet parameters)
```

packages/widget.client-react/src/client.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+    // First call
+    act(() => {
+      emit("eventA", { parameterUpdates: { myParam: "first" } });
+    });
+
+    // Complete first call before making second
+    await act(async () => {
+      deferred.resolveNext();
+      await vi.runAllTimersAsync();
+    });
```
tzyl: From the other comment is this actually representative of the real world? It seems hopeful but not guaranteed that the promise will complete before another event could be emitted. I think we'd definitely want a predictable result if "first" and "second" are set in order to not get different results each time

packages/widget.client-react/src/utils/transformEmitEventPayload.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    internalDoNotUseMetadata: { rid },
+  });
+
+  const mockObjectSet = {} as ObjectSet;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("should transform ObjectSet parameters and pass through others", async () => {
```
tzyl: Should there be another case for full pass through without any object set parameters?

tzyl (approved):
Thanks

PR #2308, opened 2025-12-18
React toolkit - Add platform api hooks

brandonwisnicki:
Adds simple platform api hooks for the following endpoints
- get (gets user by ID)
- getCurrent (gets current user)
- list (gets a list of users)

### DIFF
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -18,6 +18,7 @@ React framework for building front-end applications with OSDK.
 - [Actions](/react/actions) - useOsdkAction, validation, optimistic updates
 - [Advanced Queries](/react/advanced-queries) - useObjectSet, derived properties, aggregations
 - [Cache Management](/react/cache-management) - Cache behavior and invalidation
+- [Platform APIs](/react/platform-apis) - useCurrentFoundryUser, useFoundryUser, useFoundryUsersList
 
 ### Other Guides
 
--- a/docs/react/getting-started.md
+++ b/docs/react/getting-started.md
@@ -28,6 +28,7 @@ Using `latest` doesn't always install actual latest versions for beta packages. 
 ```
 
 Check for newer versions on npm:
+
 - [@osdk/react versions](https://www.npmjs.com/package/@osdk/react?activeTab=versions)
 - [@osdk/client versions](https://www.npmjs.com/package/@osdk/client?activeTab=versions)
 - [@osdk/api versions](https://www.npmjs.com/package/@osdk/api?activeTab=versions)
@@ -56,11 +57,11 @@ If you haven't generated an SDK yet:
 
 `@osdk/react` has two entry points:
 
-| | `@osdk/react` | `@osdk/react/experimental` |
-|---|---|---|
-| **Use when** | You only need client access | You want reactive data hooks (most apps) |
-| **Features** | Client provider, metadata | Queries, actions, caching, optimistic updates |
-| **API stability** | Stable | APIs may change between releases |
+|                   | `@osdk/react`               | `@osdk/react/experimental`                    |
+| ----------------- | --------------------------- | --------------------------------------------- |
+| **Use when**      | You only need client access | You want reactive data hooks (most apps)      |
+| **Features**      | Client provider, metadata   | Queries, actions, caching, optimistic updates |
+| **API stability** | Stable                      | APIs may change between releases              |
 
 **We recommend starting with experimental.** The "experimental" label indicates the API surface may evolve, not that the features are unstable or buggy. Most applications benefit from the reactive hooks, automatic caching, and optimistic update support.
 
@@ -80,6 +81,7 @@ ReactDOM.createRoot(document.getElementById("root")!).render(
 ```
 
 Stable exports (use with `OsdkProvider`):
+
 - `OsdkProvider` - Provider component for basic client access
 - `useOsdkClient` - Access the OSDK client directly
 - `useOsdkMetadata` - Fetch type metadata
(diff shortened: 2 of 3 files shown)


packages/react/src/new/platform-apis/admin/useCurrentFoundryUser.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function useCurrentFoundryUser(
+  { enabled = true }: UseCurrentFoundryUserOptions = {},
+): UseCurrentFoundryUserResult {
+  const { client } = React.useContext(OsdkContext2);
+
+  const handleQuery = React.useCallback(
+    () => Admin.Users.getCurrent(client),
+    [],
```
KuberSethi: any reason we're not passing in `client` as a dep here (surprised linter didn't complain / maybe my brain is off and we can indeed skip passing it in?)

packages/react/src/new/platform-apis/admin/useFoundryUsersList.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export function useFoundryUsersList(
+  { enabled = true, include = "ACTIVE", pageSize = 1000, pageToken }:
+    UseCurrentFoundryUserOptions = {},
+): UseFoundryUsersListResult {
+  const { client } = React.useContext(OsdkContext2);
+
+  const handleQuery = React.useCallback(() => {
+    return Admin.Users.list(client, { include, pageSize, pageToken });
+  }, [client]);
```
KuberSethi: likewise here I believe we should pass in our deps exhaustively

packages/react/src/utils/usePlatformQuery.ts
```diff
@@ -0,0 +1,119 @@
...
+      });
+  }, [query]);
+
+  const { subscribe, getSnapShot } = React.useMemo(
+    () => {
+      if (!enabled) {
+        return makeExternalStore<QueryPayload<T>>(
+          () => ({ unsubscribe: () => {} }),
+          process.env.NODE_ENV !== "production"
+            ? `Current Foundry User Query [DISABLED]`
```
KuberSethi: should this be more generic considering this is in `usePlatformQuery`?

packages/react/src/new/platform-apis/admin/useCurrentFoundryUser.ts
```diff
@@ -0,0 +1,70 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Admin } from "@osdk/foundry";
+
+import React from "react";
+import { usePlatformQuery } from "../../../utils/usePlatformQuery.js";
```
KuberSethi: is this the right import (`usePlatformQuery.js`)?

packages/react/src/new/platform-apis/admin/useCurrentFoundryUser.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface UseCurrentFoundryUserResult {
+  currentUser: Admin.User | undefined;
+  isLoading: boolean;
+
+  error: Error | undefined;
+
+  refetch: () => void;
+}
+
+export function useCurrentFoundryUser(
```
KuberSethi: I think users (esp those newer or w/o Foundry context) will be a little confused at the distinction between `useCurrentFoundryUser` and `useFoundryUser`. We should update our documentation (can be in a FLUP) but a short comment would help here

packages/react/src/utils/usePlatformQuery.ts
```diff
@@ -0,0 +1,119 @@
...
+        (observer: Observer<QueryPayload<T> | undefined>) => {
+          observerRef.current = observer;
+          handleQuery();
+          return {
+            unsubscribe: () => {
+              observerRef.current = undefined;
+            },
+          };
+        },
+        "foundry-current-user",
```
KuberSethi: likewise here regarding genericizing

packages/react/src/new/platform-apis/admin/useCurrentFoundryUser.ts
brandonwisnicki: yeah i think so - it's what i added to this PR

packages/react/src/new/platform-apis/admin/useFoundryUser.ts
```diff
@@ -0,0 +1,78 @@
...
+export interface UseFoundryUserResult {
+  user: Admin.User | undefined;
+  isLoading: boolean;
+
+  error: Error | undefined;
+
+  refetch: () => void;
+}
+
+export function useFoundryUser(
```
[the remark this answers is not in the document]
brandonwisnicki: main thought is around the slightly different options that are passed in. the contract is different, so felt like it would be more clear to the consumer to have different hooks to use

packages/react/src/new/platform-apis/admin/useFoundryUsersList.ts
```diff
@@ -0,0 +1,94 @@
...
+   * @default 1000
+   */
+  pageSize?: number;
+
+  /**
+   * The page token indicates where to start paging. This should be omitted from the first page's request.
+   * To fetch the next page, clients should take the value from the nextPageToken field of the previous
+   * response and use it to populate the pageToken field of the next request.
+   */
+  pageToken?: string;
```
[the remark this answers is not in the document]
brandonwisnicki: good call, added!

packages/react/src/new/platform-apis/admin/useCurrentFoundryUser.ts
brandonwisnicki: yeah - it's the correct file reference if that's what your asking?

KuberSethi: Whoops forgot about how imports work, yeah ignore

PR #2321, opened 2026-01-05
[oac] support custom delete object parameter

ethanai22:
Also adds handling for allowing multiple values of target parameters

### DIFF
--- a/packages/maker/src/api/defineDeleteObjectAction.ts
+++ b/packages/maker/src/api/defineDeleteObjectAction.ts
@@ -20,29 +20,25 @@ import {
   convertValidationRule,
   createParameters,
   defineAction,
+  DELETE_OBJECT_PARAMETER,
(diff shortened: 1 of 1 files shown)


packages/maker/src/api/defineDeleteObjectAction.ts
```diff
@@ -16,31 +16,28 @@
...
+  kebab,
+} from "./defineAction.js";
 
 export function defineDeleteObjectAction(
   def: ActionTypeUserDefinition,
 ): ActionType {
+  const parameters = createParameters(
+    def,
+    {},
+    new Set(["objectToDeleteParameter"]),
```
davidpaquin: nit: use the constant

ethanai22: done

PR #2322, opened 2026-01-05
Add testing for node 24 

nihalbhatnagar:
Updates OSDK node usages in CI pipelines


.github/workflows/release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 24
```
ssanjay1: Why are some of these 24 and some 20?

PR #2327, opened 2026-01-06
Add rids to objects returned from OSW

nihalbhatnagar:
OSW supports returning RIDs on object updates for non-reference updates. This means the rid property must be optional on the returned object as we don't differentiate between the reference updates and full updates.

### DIFF
--- a/packages/api/src/objectSet/ObjectSetListener.ts
+++ b/packages/api/src/objectSet/ObjectSetListener.ts
@@ -19,6 +19,7 @@ import type {
   PropertyKeys,
 } from "../ontology/ObjectOrInterface.js";
 import type { CompileTimeMetadata } from "../ontology/ObjectTypeDefinition.js";
+import type { WirePropertyTypes } from "../ontology/WirePropertyTypes.js";
 import type { Osdk } from "../OsdkObjectFrom.js";
 
 export namespace ObjectSetSubscription {
@@ -76,7 +77,9 @@ export namespace ObjectSetSubscription {
      * contains a new value for a geotime series reference property, in which case the RID will be undefined. RIDs will not be included
      * on the objects themselves.
      */
-    includeRid?: AllFalse<HasGeotimeSeriesReference<O, P>> extends true ? R
+    includeRid?: AllFalse<
+      PropertyTypesOnDefMatchesType<O, P, "geotimeSeriesReference">
+    > extends true ? R
       : false;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/api/src/OsdkObjectFrom.ts
```diff
@@ -259,6 +262,8 @@ export namespace Osdk {
     & (IsNever<OPTIONS> extends true ? {}
       : IsAny<OPTIONS> extends true ? {}
       : "$rid" extends OPTIONS ? { readonly $rid: string }
+      : "$ridOrUndefined" extends OPTIONS
```
nihalbhatnagar: As explained in the description, not all subscription updates will come back with RIDs defined on the property(specifically reference updates, which is a very small minority of the returned requests). This means that we need to allow an undefined rid on the OSDK object. However, if we just wanted to override the returned objects in the onChange method via `& {$rid": string | undefined}`, then returned objects wouldn't have a clean type which would make assignment difficult

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -270,8 +271,9 @@ export function createObjectSet<Q extends ObjectOrInterfaceDefinition>(
       ).subscribe(
         objectType,
         objectSet,
-        listener,
+        listener as ObjectSetSubscription.Listener<Q, any>,
```
KuberSethi: Any way we can be more specific than `any` here?

nihalbhatnagar: Its doesn't matter too much since the values can be initialized with anything at the end of the day

packages/api/src/objectSet/ObjectSetListener.ts
```diff
@@ -52,21 +53,37 @@ export namespace ObjectSetSubscription {
...
+   * @param includeRid Whether to include the $rid property in the subscription. Defaults to false. RIDs will be returned on all updates unless the update
+   * contains a new value for a geotime series reference property, in which case the RID will be undefined.
    */
   export interface Options<
     O extends ObjectOrInterfaceDefinition,
     P extends PropertyKeys<O> = PropertyKeys<O>,
+    R extends boolean = false,
   > {
+    /**
+     * The properties to request a subscription for. Requesting specific properties limits the possible properties
```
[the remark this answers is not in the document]
nihalbhatnagar: Everything is nullable so theres cases where you still may not get any values back (the backend case is when a GTSR is returned)

ssanjay1 (changes requested):
Think we should think about my earlier comment before merging ^

PR #2329, opened 2026-01-07
[components] filter list v0

KuberSethi:
Initial implementation of Filter List implementation. 

image

1. What it supports: Filtering OSDK object lists with multiple input types, histogram bucketing, etc to try and approach parity with the Workshop Filter List
2. Architecture: Three-layer design—a headless API layer (FilterListApi, FilterListItemApi) defining filter configuration and state, a base layer with unstyled composable React components and hooks (useFilterListState), and a Blueprint layer that wraps base components with Blueprint.js styling
3. Styling: Standalone SCSS system with shared variables, color aliases in
shared/styles/ plus filter-specific styles in base/styles/ and Blueprint overrides in
blueprint/styles/—enabling theming at each layer independently.

image

### DIFF
--- a/packages/react-components/src/filter-list/__tests__/LinkedPropertyInput.test.tsx
+++ b/packages/react-components/src/filter-list/__tests__/LinkedPropertyInput.test.tsx
@@ -55,15 +55,32 @@ function createMockObjectSet() {
 }
 
 function createDefinition(
-  linkedFilterComponent: "CHECKBOX_LIST" | "MULTI_SELECT" | "SINGLE_SELECT" | "CONTAINS_TEXT" | "TOGGLE" | "NUMBER_RANGE" | "DATE_RANGE",
-): LinkedPropertyFilterDefinition<ObjectTypeDefinition, string, ObjectTypeDefinition, PropertyKeys<ObjectTypeDefinition>> {
+  linkedFilterComponent:
+    | "CHECKBOX_LIST"
+    | "MULTI_SELECT"
+    | "SINGLE_SELECT"
+    | "CONTAINS_TEXT"
+    | "TOGGLE"
+    | "NUMBER_RANGE"
+    | "DATE_RANGE",
+): LinkedPropertyFilterDefinition<
+  ObjectTypeDefinition,
+  string,
+  ObjectTypeDefinition,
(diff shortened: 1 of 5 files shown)


packages/react-components/package.json
```diff
@@ -82,5 +99,10 @@
   "main": "./build/cjs/index.cjs",
   "module": "./build/esm/index.js",
   "types": "./build/cjs/index.d.cts",
-  "type": "module"
+  "type": "module",
+  "dependencies": {
+    "@blueprintjs/core": "^5.19.1",
```
ziyunp: Do we want to use blueprint or another lib?

packages/react-components/src/filter-list/types/ClassNameOverrides.ts
```diff
@@ -0,0 +1,171 @@
...
+  checkboxIndicator?: string;
+  label?: string;
+  count?: string;
+}
+
+/**
+ * Utility function to merge class names, filtering out undefined values.
+ * Returns undefined if no class names are provided (avoids empty className="" in DOM).
+ */
+export function mergeClassNames(
```
ziyunp: Why do we use this instead of [classnames](https://www.npmjs.com/package/classnames)?

ziyunp:
Regarding the 3-layer design, I agree on the headless layer and the base layer. I don't think we should have a Blueprint layer. In the base layer, we will use Base UI components and offer a default Blueprint styling from a separate styling package.

packages/react-components/src/filter-list/FilterListApi.ts
```diff
@@ -92,11 +146,139 @@ export interface FilterListProps<Q extends ObjectTypeDefinition> {
...
+   * Show add filter button at bottom
+   */
+  showAddFilterButton?: boolean;
+
+  /**
+   * Available filter templates for the "Add filter" menu.
+   * When provided along with showAddFilterButton, clicking the button
+   * shows a popover with these templates organized by category.
+   */
+  filterTemplates?: FilterTemplate[];
```
ziyunp: Agree that we need this, my intention was to use filterDefinitions for this. Can we add a isVisible prop to filterDefinitions so user can define the full list of available filters and isVisible decides whether it's added to the list? Similar to ObjecTableApi

PR #2340, opened 2026-01-13
add useOsdkFunction

KuberSethi:
This PR adds a new `useOsdkFunction` hook to wrap the existing client's `executeFunction` with the advantages that the client observable and OSDK React layers provide.

Key Implementation Details

- Auto-invalidation: Functions declare dependencies via dependsOn (object types) or dependsOnObjects (specific objects); when actions modify those objects, the function automatically refetches
- Manual control: refresh() forces re-execution; results are cached and shared across all subscribers with the same query+params
- Architecture: Follows existing Store/Query patterns used by useOsdkObject and useOsdkObjects

### DIFF
--- a/packages/client/src/observable/internal/function/FunctionParamsCanonicalizer.test.ts
+++ b/packages/client/src/observable/internal/function/FunctionParamsCanonicalizer.test.ts
@@ -92,4 +92,34 @@ describe("FunctionParamsCanonicalizer", () => {
     // Type distinction
     expect(c.canonicalize({ val: 1 })).not.toBe(c.canonicalize({ val: "1" }));
   });
+
+  it("handles Set with non-primitive values", () => {
+    const c = new FunctionParamsCanonicalizer();
+    const obj1 = { id: 1, name: "first" };
+    const obj2 = { id: 2, name: "second" };
+    expect(c.canonicalize({ s: new Set([obj1, obj2]) }))
+      .toBe(c.canonicalize({ s: new Set([obj2, obj1]) }));
+  });
+
+  it("handles Map with non-primitive keys", () => {
+    const c = new FunctionParamsCanonicalizer();
+    const key1 = { id: 1 };
+    const key2 = { id: 2 };
+    expect(c.canonicalize({ m: new Map([[key1, "a"], [key2, "b"]]) }))
+      .toBe(c.canonicalize({ m: new Map([[key2, "b"], [key1, "a"]]) }));
+  });
+
+  it("handles Map with non-primitive values", () => {
+    const c = new FunctionParamsCanonicalizer();
+    const val1 = { data: "first" };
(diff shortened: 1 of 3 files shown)


packages/client/src/observable/internal/Changes.ts
```diff
@@ -79,6 +83,10 @@ export class Changes {
     this.modified.add(key);
   };
 
+  registerFunction = (key: FunctionCacheKey): void => {
```
ziyunp: Why add to "modified" instead of "added"? When do we add to "added" set?

packages/client/src/observable/internal/function/FunctionParamsCanonicalizer.ts
```diff
@@ -0,0 +1,215 @@
...
+// Path markers use "$:" prefix. User data with this prefix is unlikely but could
+// theoretically cause collisions if it matches the exact marker sequence.
+function isPrimitiveValue(value: unknown): value is PrimitiveValue {
+  if (value == null) return true;
+  const t = typeof value;
+  return t === "string" || t === "number" || t === "boolean" || t === "bigint";
+}
+
+export class FunctionParamsCanonicalizer {
+  #inputCache = new WeakMap<
```
ziyunp: Just for my understanding, when do we use '#'?

packages/client/src/observable/internal/function/FunctionParamsCanonicalizer.ts
```diff
@@ -0,0 +1,215 @@
...
+      return null;
+    }
+
+    if (value === undefined) {
+      path.push(undefined);
+      return undefined;
+    }
+
+    if (
+      typeof value === "boolean"
```
ziyunp: Why not use isPrimitive function above?

packages/client/src/observable/internal/function/FunctionParamsCanonicalizer.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    expect(c.canonicalize(input)).toBe(c.canonicalize(input));
+  });
+
+  it("returns undefined for null/undefined params", () => {
+    const c = new FunctionParamsCanonicalizer();
+    expect(c.canonicalize(null)).toBeUndefined();
+    expect(c.canonicalize(undefined)).toBeUndefined();
+  });
+
+  it("throws on circular references", () => {
```
ziyunp: When will this happen?

KuberSethi: Users can accidentally pass self-referential objects as function parameters:

```
const obj = { data: 1 };
obj.self = obj; // circular
useOsdkFunction(myQuery, { param: obj });
```

This would break our canonicalization logic

PR #2344, opened 2026-01-13
[Components] Async loading of derived properties columns

ziyunp:
- When loading data with derived properties, the base properties columns are returned first and derived properties shows loading state.
- When derived properties return, show the value in the column.
- The recording shows scrolling with a page size = 2, the loaded rows preserve the loaded values while the new rows show loading state.

The useAsyncColumnData hook is the starting point for how we would enrich data from slower loading columns, including function-backed columns. 
I would like to introduce generator functions later to asynchronously load individual columns, especially when we have multiple function-backed columns. 
With the way I fetch derived properties now, all derived properties will be loaded the same time. 

**Q: Would love some feedback on whether fetching derived properties using the useObjectSet hook is ok or should I call the client fetchPage function directly for each derived property so each derived property column can be loaded separately.**

https://github.com/user-attachments/assets/97d82253-25f2-4198-966c-6008c3242aae

### DIFF
--- a/.changeset/major-trams-win.md
+++ b/.changeset/major-trams-win.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": minor
+---
+
+Async loading of derived properties in ObjectTable
--- a/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
+++ b/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
@@ -59,7 +59,7 @@ export function EmployeesTable() {
   return (
     <div
       style={{
-        height: "120px",
+        height: "500px",
         overflow: "auto",
       }}
     >


packages/react-components/src/object-table/hooks/useAsyncColumnData.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  // Fetch derived properties for current baseRows
+  const {
+    data: derivedData,
+    isLoading: isDerivedPropertiesLoading,
+    error: derivedPropertiesError,
+  } = useObjectSet<Q, never, RDPs>(objectSet, {
+    withProperties,
+    where: filter,
+    enabled: filter != null,
```
KuberSethi: should we also check if `derivedPropertyKeys` is empty to be extra defensive?

KuberSethi:
> Q: Would love some feedback on whether fetching derived properties using the useObjectSet hook is ok or should I call the client fetchPage function directly for each derived property so each derived property column can be loaded separately.

I personally believe we should either keep calling `useObjectSet` (or preferably `useOsdkObjects` actually) as we then get these values cached. Alternatively if we want more granular loading we can look into a solution at the osdk-react layer. 

This is a good concern though!

PR #2346, opened 2026-01-13
update useOsdkAggregation to support composed object sets, async

KuberSethi:
Before this PR:
- `useOsdkAggregation` only worked on the base object type, didn't support aggregating on pivoted, filtered, or composed ObjectSets
- The underlying observeAggregation client method was synchronous-only and therefore couldn't handle things like links

What this PR does:
- Adds an objectSet parameter to useOsdkAggregation, enabling aggregations on pivoted/composed sets (e.g., $(Employee).pivotTo("primaryOffice")).
- Adds an async overload to observeAggregation in the client layer to support ObjectSet-based aggregation with proper cache invalidation.
  - This PR deprecates the old sync way of doing things. A future PR will remove it. osdk-react is still in beta so I feel comfortable with this approach of breaking changes

### DIFF
--- a/packages/client/src/observable/ObservableClient.ts
+++ b/packages/client/src/observable/ObservableClient.ts
@@ -170,6 +170,20 @@ export interface ObserveObjectSetArgs<
   totalCount?: string;
 }
 
+interface ObserveAggregationBaseOptions<
+  T extends ObjectOrInterfaceDefinition,
+  A extends AggregateOpts<T>,
+  RDPs extends Record<string, SimplePropertyDef> = {},
+> extends CommonObserveOptions, ObserveOptions {
+  type: T;
+  where?: WhereClause<T, RDPs>;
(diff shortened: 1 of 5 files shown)


packages/client/src/observable/internal/aggregation/AggregationCacheKey.ts
```diff
@@ -27,9 +27,11 @@ import type { AggregationQuery } from "./AggregationQuery.js";
 
 export const TYPE_IDX = 0;
 export const API_NAME_IDX = 1;
-export const WHERE_IDX = 2;
```
bryantpark04: out of curiosity, why is `CacheKey.otherKeys` an `any[]` that we use these constants to index into instead of an object with properties for type, API name, etc.?

bryantpark04 (approved):

bryantpark04 (approved):

PR #2347, opened 2026-01-13
[Transactions] Allow flushing edits

nihalbhatnagar:
Adds ability to flush edits for transactions. Additionally, adds it to an experimental export for beta

### DIFF
--- a/packages/client/src/createClient.test.ts
+++ b/packages/client/src/createClient.test.ts
@@ -147,6 +147,7 @@ describe(createClient, () => {
       const transactionId = "test-transaction-id";
       const clientWithTransaction = createClientWithTransaction(
         transactionId,
+        async () => {},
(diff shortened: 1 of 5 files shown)


packages/functions/src/transactions/flushEdits.ts
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+import {
+  type WriteableClient,
+  writeableClientContext,
+} from "./WriteableClient.js";
+
+export async function flushEdits(client: WriteableClient<any>): Promise<void> {
+  await client[writeableClientContext].editRequestManager.flushPendingEdits();
```
KuberSethi: What if this fails?

KuberSethi (approved):

PR #2351, opened 2026-01-13
Support IDPs in OAC

ethanai22:
bluf:
keys in defineInterface properties map must be a SPT or in line IDP (removed raw PropertyTypeType usage). For example:

```
const spt = defineSharedPropertyType({
  apiName: "spt",
  type: "string",
});

const iface = defineInterface({
  apiName: "bar",
  displayName: "Bar",
  properties: {
    "foo": { type: "boolean" },
    "spt": spt,
  },
});
```
produces an IDP for foo and a spt for spt. An interface property output shape is created for both. A spt shape is still created for the spt. We maintain `propertiesV2` on IR interface types, but they now only include spts.

IDPs are referenced by non namespaced api names on actions (this only has to be unique within the owning interface and extended interfaces), while spts must be referenced by full api name

### DIFF
--- a/packages/maker/src/conversion/toMarketplace/convertInterfacePropertyType.ts
+++ b/packages/maker/src/conversion/toMarketplace/convertInterfacePropertyType.ts
@@ -15,7 +15,10 @@
  */
 
 import type { OntologyIrMarketplaceInterfacePropertyType } from "@osdk/client.unstable";
-import type { InterfacePropertyType } from "../../api/interface/InterfacePropertyType.js";
+import {
+  type InterfacePropertyType,
+  isInterfaceSharedPropertyType,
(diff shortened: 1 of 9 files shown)


packages/maker/src/conversion/toMarketplace/convertInterfacePropertyType.ts
```diff
@@ -0,0 +1,74 @@
...
+      },
+    }];
+  } else {
+    return [apiName, {
+      type: "interfaceDefinedPropertyType",
+      interfaceDefinedPropertyType: {
+        apiName: apiName,
+        displayMetadata: {
+          displayName: prop.displayName ?? apiName,
+          visibility: prop.visibility ?? "NORMAL",
```
aep000: We should probably share default constants with convertSpt

PR #2352, opened 2026-01-14
Secure Property Value Loading

nihalbhatnagar:
Secure Property Values allow users to fetch [markings](https://www.palantir.com/docs/foundry/security/markings) at the property level.

The OSDK implements this feature by exposing an additional `$propertySecurities` method off of objects.

```
const objects = await objectSet.fetchPage({$loadPropertySecurityMetadata: true});
const securities: PropertySecurity[] = osdkObject.$propertySecurities.myProperty
```

### DIFF
--- a/packages/monorepo.cspell/dict.osdk.txt
+++ b/packages/monorepo.cspell/dict.osdk.txt
@@ -22,3 +22,4 @@ unistyles
 widgetregistry
 widgetset
 dsmtClient
+sidc


packages/e2e.sandbox.catchall/src/runWithPropertiesTest.ts
```diff
@@ -35,7 +35,7 @@ export async function runWithPropertiesTest(): Promise<void> {
       base.pivotTo("stateTerritory").aggregate("airportStateName:collectSet", {
         "limit": 10,
       }),
-  }).fetchPage();
+  }).fetchPage({ $loadPropertySecurityMetadata: true });
```
ssanjay1: Do we actually have a test anywhere where we look at the property securties that come back?

ssanjay1 (dismissed):

PR #2363, opened 2026-01-15
add interface support to useOsdkObject

KuberSethi:
Adds interface support to useOsdkObject, which previously lacked it vs other hooks


bryantpark04 (approved):
+1 to unblock

PR #2364, opened 2026-01-16
Add optional struct params for actions

ssanjay1:
Now, if you have an action that takes a struct param, we support optional fields in the struct. This means you will not be required to pass in that field.

`client(myAction).applyAction({structParam: {"required" : 1, "notRequired": undefined}}`

In that example, `notRequired` could also just be excluded from the entry.


packages/api/src/__quickinfo_snapshot__/__snapshots__/actions.snap
```diff
@@ -33,7 +33,9 @@ type actionParam_primitiveType_attachment =
 type actionParam_primitiveType_string = string;
 
 /** A struct-typed action parameter — fields are mapped through DataValueClientToWire. */
-type actionParam_structType = { name: string; age: number; joined: string };
+type actionParam_structType =
+  & { name: string; age: number; joined: string }
+  & {};
```
bryantpark04, line 38: might be nicer if this didn't change

bryantpark04: should these snapshots be located in `src`?

packages/client/src/actions/actions.test.ts
```diff
@@ -206,14 +206,14 @@ describe.each([
...
-        address: { city: string; state: string; zipcode: number };
+        address: { city: string; state: string; zipcode?: number | undefined };
       }
     >()
       .toMatchTypeOf<
         InferredParamType
       >();
 
-    const result = await client(createStructPerson).applyAction({
+    const result = void await client(createStructPerson).applyAction({
```
bryantpark04, line 216: why did we void this?

packages/client/src/derivedProperties/createWithPropertiesObjectSet.test.ts
```diff
@@ -32,7 +32,7 @@ describe(createWithPropertiesObjectSet, () => {
         base.pivotTo("lead").selectProperty("employeeId"),
     } satisfies DerivedProperty.Clause<Employee>;
 
-    const result = clause["derivedPropertyName"](deriveObjectSet);
```
bryantpark04, line 35: seems like dprint changes running on the pre-commit hook. they caused a diff in unrelated packages which I had to reset.

bryantpark04 (approved):

nihalbhatnagar (approved):

joshbouganim:
👋 Pardon the drive-by,  I'm working with Max Magid to get this onto our stack and was reviewing the change ahead of that. One thing I wanted to confirm: it looks like this types optional struct fields as `T | undefined`, which means an explicit null is no longer assignable. Is that intentional?

The reason I ask is that null and undefined aren't interchangeable on the wire per the client's own serializer. `toDataValue` in `packages/client/src/util/toDataValue.ts`:

```ts
if (value == null) {
  // typeof null is 'object' so do this first
  // Sending null over the wire clears the data, whereas undefined is dropped at request time.
  return value;
}
```

So for an optional struct field: omitting it (or undefined) drops the key from the payload, whereas null is sent as `{"field": null}` and clears the value. I confirmed server-side that the (recently fixed) validator accepts an explicit null for an optional struct field. It validates VALID, same as omitting, so the runtime supports the "clear" semantic; it's only the generated type that blocks it.

The optional half of `StructType` in `packages/api/src/actions/Actions.ts` resolves nullable fields to `DataValueClientToWire[U] | undefined`. For contrast, top-level optional params in the same generated file are typed `| null`, e.g. in createStructPerson.ts:

`readonly address?: ActionParam.StructType<{ city: 'string'; state: 'string' }> | null;`

So within one generated action, the top-level optional param allows null but a nullable struct field wouldn't, which seems like an unintended asymmetry. Would changing the optional branch to `| null` (mirroring top-level optionals) be reasonable?

bryantpark04:
Thanks for flagging! That sounds reasonable to me, will leave to @ssanjay1 to include it in a FLUP to address the other nits on this PR.

PR #2369, opened 2026-01-16
Add tokenStorage option for configurable token persistence

srhodes-palantir:
On shared devices/workstations, cached refresh tokens in localStorage can persist causing problems with rapid user swapping scenarios. This change allows applications to opt into sessionStorage for cleaner token clearing when the browser tab/window closes.

- Adds tokenStorage option to createPublicOauthClient() with three modes:
    - 'localStorage' (default) - tokens persist across browser sessions
    - 'sessionStorage' - tokens cleared when browser/tab closes
    - 'none' - tokens never stored, user re-authenticates on every page load

Enabling does not affect previously configured applications with localStorage instead of sessionStorage.

### DIFF
--- a/packages/oauth/src/index.ts
+++ b/packages/oauth/src/index.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-
 export type { TokenStorageType } from "./common.js";
 export type { ConfidentialOauthClient } from "./ConfidentialOauthClient.js";
 export { createConfidentialOauthClient } from "./createConfidentialOauthClient.js";


packages/oauth/src/publicOauth.test.ts
```diff
@@ -532,4 +533,113 @@ describe(createPublicOauthClient, () => {
...
+        BASE_CLIENT_ARGS.foundryUrl,
+        BASE_CLIENT_ARGS.redirectUrl,
+        { tokenStorage: "sessionStorage" },
+      );
+
+      const lastArg = vi.mocked(commonJs.common).mock.calls[0][7];
+      expect(lastArg).toBe(mockSessionStorage);
+    });
+
+    it("should pass undefined to common when tokenStorage is 'none'", () => {
```
nihalbhatnagar: Can we add an explicit test for undefined

PR #2393, opened 2026-01-22
add streaming support to useOsdkObject

KuberSethi:
Adds the ability to stream updates to useOsdkObject, so changes to the object are reflected in real-time without needing to refetch.


bryantpark04 (dismissed):
+1 to unblock

PR #2416, opened 2026-01-28
[2/4] [components] add base ui components for filter-list

KuberSethi:
Adding some reusable base components that the filter inputs will use.

- combobox component with search, multi-select, and tag display
- select component for single value dropdowns
- switch component for toggle inputs


KuberSethi:
Superseded by #2422 after restack

PR #2417, opened 2026-01-28
[3/4] [components] add filter-list hooks

KuberSethi:
filter list component hooks

- useFilterListState tracks filter values and converts them to where clauses
- usePropertyAggregation loads value counts for checkbox lists and histograms


KuberSethi:
Superseded by #2423 after restack

PR #2418, opened 2026-01-28
[4/4] [components] add filter-list ui components

KuberSethi:
filter list rendering and input components

- FilterList, FilterListContent, FilterListItem for the main layout
- input components for each filter type (checkbox list, date range, number range, etc)
- sandbox demo in the people app


KuberSethi:
Superseded by #2424 after restack

PR #2419, opened 2026-01-28
support interface link constraints

KuberSethi:
Adds support for interface link constraints in the observable client.

- Enables observing links from interface instances (not just object types)
- Properly handles cache key construction and invalidation for interface-sourced links
- Updates query construction to use interface object sets with intersect for filtering


bryantpark04 (dismissed):
+1 to unblock

PR #2461, opened 2026-02-03
fix pagesize caching bug

KuberSethi:
This PR introduces a new `View` abstraction layer to fix a pageSize caching bug where multiple subscribers with different pageSize values would share cached data incorrectly. The solution:

1. Creates `ListQueryView` - a per-subscriber view that slices shared cache data to each subscriber's `viewLimit`
2. Tracks per-subscriber pageSize in BaseListQuery to compute the effective fetch pageSize (max across all subscribers)
3. Modifies AbstractHelper._subscribe() to wrap list queries in views when pageSize or autoFetchMore is specified

This is arguably controversial, but I believe we should decouple rendering/view details from caching concerns. This PR / abstraction codifies that.


ziyunp (dismissed):

ssanjay1 (approved):
Stamping

PR #2462, opened 2026-02-03
Relax interface generation validation

ssanjay1:
We now specifically check for cases where interface properties have no namespace and conflict with the root of a property that does have an api namespace which matches that of the interface.


KuberSethi (commented):
Looks good just two small thoughts

PR #2471, opened 2026-02-05
support listening to cache invalidations

KuberSethi:
support listening to cache invalidations


KuberSethi:
No pressing use-case to justify this at the moment

PR #2472, opened 2026-02-05
Add support for query functions

schmelczer:
## Changes

- Add new `@osdk/generator-converters.preview` package with `PreviewOntologyIrConverter` that returns `ActionTypeFullMetadata` (with `fullLogicRules`) instead of `ActionTypeV2`
- Add `ActionLogicRuleConverter` to convert IR logic rules (create/modify/delete object, create/modify interface) to `ActionLogicRule` format
- Add TypeScript and Python function discovery to `@osdk/generator-converters.ontologyir` via `getOsdkQueryTypes` — discovers functions and converts them to `QueryTypeV2` for OSDK generation
- Add `generate-sdk` CLI to the preview package for converting OntologyIR JSON + discovered functions into a generated SDK
- Add function discovery and intermediate OSDK generation to `@osdk/maker` CLI (`--functionsRootDir`, `--functionsOutput`, `--generateFunctionsOsdk`)
- Resolve object type property RIDs to apiNames for primaryKey/titleProperty in the base converter

## Notes

- Function discovery dependencies (`@foundry/functions-typescript-osdk-discovery`, etc.) are optional — loaded lazily at runtime via dynamic imports
- Python function discovery invokes `python3 -m functions.bin parse` and parses JSON from stdout
- The preview converter uses deterministic UUID v5-like RIDs derived from SHA-256 hashes of apiNames

### DIFF
--- a/packages/react-components-storybook/.storybook/util-polyfill.ts
+++ b/packages/react-components-storybook/.storybook/util-polyfill.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-
 export const inspect = (obj: any, options?: any): string => {
   // Simple fallback implementation for browser
   try {
--- a/packages/react-components-storybook/eslint.config.mjs
+++ b/packages/react-components-storybook/eslint.config.mjs
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-
 import storybook from "eslint-plugin-storybook";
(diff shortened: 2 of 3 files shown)


packages/maker/src/cli/main.ts
ethanai22: are there plans to add python discovery capability here as well?

PR #2474, opened 2026-02-05
[custom widgets] support interfaces in object set parameters

amishshah:
This is a **breaking change**, but it allows users to start specifying OSDK interfaces as the base/allowed type for their object sets, rather than just object types.

I actually started a PR that implemented this in a backwards-compatible way, see https://github.com/palantir/osdk-ts/pull/2444, but I decided against it.

I made this breaking because I didn't think there was a good upgrade story for our users in the future. Decorating `objectType` with `@deprecated` only served to prevent new typings of that field. Existing usage in VS Code (at least for me) didn't show me any warnings. I also added logging to the build script to show you a warning if you're using the old field. I then considered that probably very few people are actually going to see that warning if their build succeeds.

At that point, I decided to make this PR instead. I imagined the other route would have users for a few months not realising they're using deprecated fields, and then all of a sudden they disappear. I'd rather be noisy and make them fix it now.

The fix is very easy on the user's end, and I've done my best to make the types guide them:

image

image

### DIFF
--- a/.changeset/ready-roses-relate.md
+++ b/.changeset/ready-roses-relate.md
@@ -1,7 +1,7 @@
 ---
-"@osdk/widget.client-react": patch
-"@osdk/widget.vite-plugin": patch
-"@osdk/widget.api": patch
+"@osdk/widget.client-react": minor
+"@osdk/widget.vite-plugin": minor
+"@osdk/widget.api": minor
 ---
 
 Support interfaces in object set parameters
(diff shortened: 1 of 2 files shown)


gpirindev (dismissed):
Looks good! I think it's fine to introduce this breaking change and customers/AI agents should be able to fix it relatively easy as they upgrade their repos. It's probably nice to introduce it in a minor version bump, rather than a patch and have it noted in a changelog.

PR #2494, opened 2026-02-10
add react-sdk-docs snippets package

KuberSethi:
Adds react-sdk-docs snippets package for osdk-react dev console docs, this contains code snippets for osdk-react usage (see current docs if you're curious https://palantir.github.io/osdk-ts/react/getting-started). The goal is to ensure parity between existing TS snippets and React snippets so we can enable an easy toggle on the FE in Dev Console to let users pick which snippets they'd like to view.

Will update our templates in a separate PR.

### DIFF
--- a/packages/react-sdk-docs/codegen.mjs
+++ b/packages/react-sdk-docs/codegen.mjs
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


packages/react-sdk-docs/codegen.mjs
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
```
zeyadkhaled: Update the header date here and the other newly introduced files
```suggestion
 * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
```

bryantpark04 (approved):

PR #2510, opened 2026-02-11
Support reducers and struct main values in oac

ethanai22:


packages/maker/src/api/properties/PropertyTypeType.ts
```diff
@@ -57,6 +57,11 @@ export type PropertyTypeTypeStruct = {
       | StructPropertyType
       | Exclude<PropertyTypeTypesWithoutStruct, PropertyTypeTypeMarking>;
   };
+  mainValue?: {
+    fields: Array<string>;
```
davidpaquin: A main value can be made of multiple fields?

PR #2513, opened 2026-02-11
filter derived property keys in websocket subscribe to prevent client crash

KuberSethi:
subscribe method crashes when derived property keys appear in the properties array because they don't exist in objOrInterfaceDef.properties

• filter out property keys not present in the object/interface definition before the geotime type check in ObjectSetListenerWebsocket
• derived properties are resolved server-side from the plan, so they don't need to be in the client's propertySet


bryantpark04 (dismissed):
+1 to unblock

PR #2430, opened 2026-01-29
add e2e sandbox filter list demos and remove unused types

KuberSethi:
sandbox demos for BaseFilterList in both peopleapp and officenetwork e2e apps

• add officenetwork "Filters" lens with department, city, title, employee number, and start date filters
• add peopleapp EmployeesWithFilterList page with department checkbox filter
• remove unused AddFilterMenuProps and filter template types from API surface

PR #2429, opened 2026-01-29
[5/6] [components] add filter-list complex and aggregation inputs

KuberSethi:
Adds complex filter inputs that use aggregation queries and linked objects.

- LinkedPropertyInput for filtering on linked object properties
- Aggregation inputs: CheckboxListInput, ListogramInput, MultiSelectInput, SingleSelectInput
- Range inputs: RangeInput base, NumberRangeInput, DateRangeInput
- Additional inputs: TextTagsInput, TimelineInput, NullValueWrapper
- createHistogramBuckets utility for histogram visualization

PR #2428, opened 2026-01-29
[4/6] [components] add filter-list base containers and simple inputs

KuberSethi:
Adds base filter list container components and simple input types.

- FilterList, FilterListHeader, FilterListContent, FilterListItem containers
- FilterInput router component with error boundary
- Simple inputs: ContainsTextInput, SingleDateInput, MultiDateInput, ToggleInput
- Shared SCSS variables for filter-list styling

PR #2529, opened 2026-02-12
restructure filter list into base and osdk-aware layers

KuberSethi:
base input components no longer depend on osdk types or hooks, enabling standalone use with custom data sources

• split FilterList into BaseFilterList (data-agnostic) and FilterList (osdk-aware) via a renderInput callback pattern, move data fetching into the osdk layer and pass pre-fetched values/counts to base inputs as plain props
• tokenize ~250 hardcoded css values into filter-list.css custom properties for theming, use transform scaleX for listogram bar animation
• add per-filter where clauses so each filter's aggregation excludes its own state
• add drag-and-drop test suite, sandbox demo page, clean up dead type files

### DIFF
--- a/.changeset/filter-list-base-osdk-layers.md
+++ b/.changeset/filter-list-base-osdk-layers.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": minor
+---
+
+Restructure FilterList into base and osdk layers
--- a/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesWithFilterList.tsx
+++ b/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesWithFilterList.tsx
@@ -21,7 +21,7 @@ import {
 } from "@osdk/react-components/experimental";
 import "@osdk/react-components/styles.css";
 import { useOsdkObjects } from "@osdk/react/experimental";
-import { useCallback, useState } from "react";
+import { useState } from "react";
 
 import { List } from "../../components/List.js";
 import { ListItem } from "../../components/ListItem.js";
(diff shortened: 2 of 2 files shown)


KuberSethi:

- #2557
    - #2558
        - #2428
            - #2429
                - #2430
                    - #2529 ◀

packages/react-components/src/filter-list/base/BaseFilterList.tsx
```diff
@@ -81,20 +60,20 @@ export function FilterList<Q extends ObjectTypeDefinition>(
         <FilterListHeader
           title={title}
           titleIcon={titleIcon}
+          collapsed={collapsed}
```
nfdiop: question (non-blocking): is this the only place where we handle `collapsed`?

packages/react-components/src/filter-list/base/inputs/ListogramInput.module.css
```diff
@@ -22,67 +22,71 @@
...
-  background: var(--osdk-filter-listogram-bar-color);
-  transition: width var(--osdk-emphasis-transition-duration) var(--osdk-emphasis-ease-default);
+  width: 100%;
+  background: var(--bar-fill-color, var(--osdk-filter-listogram-bar-color));
+  transform-origin: left;
+  transition: transform var(--osdk-filter-listogram-bar-fill-transition);
 }
 
 .row[aria-pressed="true"] .barFill {
   background: var(--osdk-filter-listogram-selected-color);
```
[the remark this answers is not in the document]
KuberSethi: It's set inline per row in `ListogramInput.tsx`! Arguably not the best to be doing it inline but since we need runtime-level CSS configuration/transformation seemed to be the best way that we have given we don't have a more expressive CSS library/configuration

packages/react-components/src/filter-list/base/BaseFilterList.tsx
nfdiop: I was also expecting to check it before rendering the add button

nfdiop (approved):

PR #2530, opened 2026-02-12
[Components] OSDK Components Storybook

ziyunp:
Screenshot 2026-02-12 at 16 49 32
Screenshot 2026-02-12 at 18 29 50


packages/react-components-storybook/public/mockServiceWorker.js
```diff
@@ -0,0 +1,343 @@
...
+
+    case "PASSTHROUGH": {
+      return passthrough();
+    }
+  }
+
+  return passthrough();
+}
+
+/**
```
nfdiop: maybe we could add a function description in cases like this

packages/react-components-storybook/README.md
```diff
@@ -0,0 +1,35 @@
...
+The Storybook will be available at http://localhost:6006
+
+## Features
+
+- Interactive component demonstrations
+- Multiple variations and states for each component
+- Mock data using FauxFoundry for realistic examples
+- MSW (Mock Service Worker) integration for API mocking
+
+## Components.npmrc
```
nfdiop: extra nit: additional `.npmrc` added here

packages/react-components-storybook/.storybook/manager.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { addons } from "@storybook/manager-api";
+
+// Redirect to the first story if we're at the root
+addons.register(
+  "redirect-to-first-story",
+  (api: { on: (arg0: string, arg1: () => void) => void }) => {
+    // Check if we're at the root path (no story selected)
+    api.on("STORY_RENDERED", () => {
+      const url = new URL(window.location.href);
```
nfdiop: Maybe pull the logic into a function since it's used in two places?
Also, instead of "// Redirect to the first story if we're at the root," let's use "Redirect to the object table story if we're at the root"—it's more accurate.

nfdiop (approved):

KuberSethi (approved):

PR #2536, opened 2026-02-14
unify RDP canonicalization between useOsdkObjects and useObjectSet

iw-an:
## Changes

- Unify `withProperties` canonicalization so both `useOsdkObjects `and `useObjectSet` produce the same canonical RDP object, causing fetched objects to share cache slots regardless of which hook loaded them
- Fix cross-variant propagation to preserve non-null RDP field values when merging between variants
- Defer subscription cleanup to a microtask to prevent `propagateWrite` from skipping keys during React unmount-remount cycles

## Problem

`useOsdkObjects` (via `ListsHelper`) and `useObjectSet` (via `ObjectSetHelper`) used fundamentally different strategies to canonicalize `withProperties`:

- ListsHelper used `RdpCanonicalizer` which executes builder functions, extracts wire-format definitions, and structurally deduplicates
- ObjectSetHelper only used `Object.keys(options.withProperties).sort()` - sorted property names with no structural identity

Additionally, `BaseListQuery.rdpConfig` reads `this.cacheKey.otherKeys[RDP_IDX]` where `RDP_IDX = 4`. For `ObjectSetCacheKey`, `otherKeys` only has 2 elements (`[baseObjectSetWire, operations]`), so `otherKeys[4]` was always undefined. This meant `storeOsdkInstances` was called with null rdpConfig from `ObjectSetQuery`, creating separate cache slots for the same physical objects.

### DIFF
--- a/packages/client/src/observable/internal/Store.ts
+++ b/packages/client/src/observable/internal/Store.ts
@@ -432,8 +432,6 @@ export class Store {
       } else if (cacheKey.type === "aggregation") {
         return cacheKey.otherKeys[AGGREGATION_RDP_IDX];
       } else if (cacheKey.type === "objectSet") {
-        // ObjectSetQuery stores its canonical RDP on the query instance,
-        // not in the cache key's otherKeys array.
         const query = this.queries.peek(cacheKey);
         if (query) {
           return query.rdpConfig;
--- a/packages/client/src/observable/internal/actions/ActionApplication.ts
+++ b/packages/client/src/observable/internal/actions/ActionApplication.ts
@@ -122,10 +122,7 @@ export class ActionApplication {
         }
       }
 
-      // Use the registry to find all RDP variant cache keys for each deleted
-      // object. We can't use cacheKeys.get() here because that would create
-      // and register an orphan key with no subject, which blows up the
-      // invariant in Store.#cleanupCacheKey when GC runs.
+      // Use the registry to find all RDP variant cache keys for each deleted object.
       this.store.batch({}, (batch) => {
         for (const { objectType, primaryKey } of deletedObjects ?? []) {
           for (
--- a/packages/client/src/observable/internal/object/ObjectsHelper.test.ts
+++ b/packages/client/src/observable/internal/object/ObjectsHelper.test.ts
@@ -28,6 +28,7 @@ import {
   updateObject,
   waitForCall,
 } from "../testUtils.js";
+
 function createFakeRdpConfig(...fields: string[]): Canonical<Rdp> {
   const rdp: Rdp = {};
   for (const field of fields) {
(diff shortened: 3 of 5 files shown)


packages/client/src/observable/internal/Store.ts
```diff
@@ -417,6 +431,13 @@ export class Store {
         return cacheKey.otherKeys[LIST_RDP_IDX];
       } else if (cacheKey.type === "aggregation") {
         return cacheKey.otherKeys[AGGREGATION_RDP_IDX];
+      } else if (cacheKey.type === "objectSet") {
```
KuberSethi: hmm might be misremembering types but I think this could potentially just be

```
const query = this.queries.peek(cacheKey);
if (query) {
  return query.rdpConfig;
}
```

`rdpConfig` is already on `BaseListQuery` so I think the guard is a little redundant

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -83,6 +87,10 @@ export class ObjectSetQuery extends BaseListQuery<
     }
   }
 
+  public override get rdpConfig(): Canonical<Rdp> | null {
```
KuberSethi: I tested this PR out pretty thoroughly (via clicking around and re-orging the e2e sandbox officenetwork app) and found that this causes

```
Uncaught Error: Invariant failed
    at invariant (tiny-invariant.js:12:11)
    at #cleanupCacheKey (Store.ts:205:7)
    at #cleanupCacheKey (CacheKeys.ts:148:10)
    at #refCounts [as cleanup] (CacheKeys.ts:56:17)
    at RefCounts.gc (RefCounts.ts:94:14)
    at CacheKeys.ts:76:23
``` 

I'm still digging to confirm since this might've uncovered other issues but I think this is likely the root cause: now that ObjectSetQuery stores objects with RDP-tagged cache keys (instead of plain ones), any code path that calls cacheKeys.get("object", type, pk, undefined) creates an orphan cache key and since we never clean it up the invariant blows up.

two call sites hit by this:
- ActionApplication.ts:128 — looks up plain keys for deleted objects during action results
- BaseListQuery.onOswRemoved (line 779) — looks up plain keys for websocket REMOVED events

both call cacheKeys.get() but never create a query/subject for the resulting key.

I think to fix we could add a CacheKeys.peek() method that looks up the key without calling register() then  use it in these places so they don't create orphans.

packages/client/src/observable/internal/actions/ActionApplication.ts
```diff
@@ -123,18 +122,23 @@ export class ActionApplication {
         }
       }
 
+      // Use the registry to find all RDP variant cache keys for each deleted
```
KuberSethi: nit: this could just be `Use the registry to find all RDP variant cache keys for each deleted object` I think, the rest won't make as much sense without the context of this PR

packages/client/src/observable/internal/object/ObjectsHelper.test.ts
```diff
@@ -0,0 +1,236 @@
...
+import { createClient } from "../../../createClient.js";
+import type { Canonical } from "../Canonical.js";
+import type { Rdp } from "../RdpCanonicalizer.js";
+import { Store } from "../Store.js";
+import {
+  mockSingleSubCallback,
+  updateObject,
+  waitForCall,
+} from "../testUtils.js";
+function createFakeRdpConfig(...fields: string[]): Canonical<Rdp> {
```
KuberSethi: nit: insert newline above

packages/client/src/observable/internal/Store.ts
```diff
@@ -417,6 +431,13 @@ export class Store {
...
+        // ObjectSetQuery stores its canonical RDP on the query instance,
```
KuberSethi: nit: this comment isn't necessary

KuberSethi (approved):

PR #2542, opened 2026-02-16
package and bundle size optimizations

KuberSethi:
CJS build was bundling the entire TypeScript compiler into internal-node.cjs because typescript is a devDependency and tsup inlines devDeps by default. npm tarball went from 5.3MB to 732KB compressed.

• externalize devDependencies in tsup CJS builds to prevent bundling test/dev tooling into production output
• add sideEffects: false to @osdk/client and @osdk/react to enable tree-shaking in webpack/rollup/vite
• exclude test files and test utilities from ESM, browser, and types build outputs
• remove unused fetch-retry production dependency from @osdk/client

### DIFF
--- a/packages/react/src/new/useOsdkAggregation.ts
+++ b/packages/react/src/new/useOsdkAggregation.ts
@@ -231,7 +231,7 @@ export function useOsdkAggregation<
       }
       return makeExternalStore<ObserveAggregationArgs<Q, A>>(
         (observer) =>
-           
+          // eslint-disable-next-line @typescript-eslint/no-deprecated
           observableClient.observeAggregation(
             {
               type: type,


packages/client/src/createGeotimeSeriesProperty.ts
```diff
@@ -19,7 +19,7 @@ import type {
   TimeSeriesPoint,
   TimeSeriesQuery,
 } from "@osdk/api";
-import * as OntologiesV2 from "@osdk/foundry.ontologies";
+import * as TimeSeriesValueBankProperties from "@osdk/foundry.ontologies/TimeSeriesValueBankProperty";
```
ssanjay1: Yea I guess we did the base API originally because its a more stable surface area, but we control the platform apis as well, so switching away from the barrel imports to the subpaths should be fine so long as we dont break things

PR #2566, opened 2026-02-19
filter list feature completions

KuberSethi:
collapsible panel, select all, include/exclude toggle, per-item colors, histogram display modes, and session persistence

• add collapsible panel with smooth css grid transition and collapse/expand props
• add select all checkbox, include/exclude toggle, per-item color mapping, and listogram display modes
• add session persistence support via initialFilterStates and serialization utilities


KuberSethi:
superseded by https://github.com/palantir/osdk-ts/pull/2594

PR #2581, opened 2026-02-23
[Components] Update Editable Table UX and API to pass rowData

ziyunp:
In this PR:
- Added enableEditModeByDefault prop and updated the table interaction according to [UX](https://www.figma.com/design/6K3jfj7ZMVyF3rN5DsR1e0/-CC--FDP-Design-System?node-id=2539-11921&t=vKXiHSYFlC9SEfIH-1)
- Updated the edit API to pass row data in the callback
- Disable row click in edit mode

Next PR will handle edit validation and add callbacks to revert changes.

When enableEditModeByDefault = true, click on row/cell to enable table editing:

https://github.com/user-attachments/assets/dd27fbd1-97b3-4d2b-9f62-cd3cec5607a9

When enableEditModeByDefault = false, click Edit Table to enable table editing:

https://github.com/user-attachments/assets/6dbd2f02-8ac9-47d1-b8e2-50c76e7ab450

### DIFF
--- a/packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
@@ -1121,6 +1121,7 @@ export const WithSubmitEditsButton: Story = {
     ],
     onSubmitEdits: (edits: CellEditInfo<Osdk.Instance<Employee>>[]) => {
       alert("Submitting edits");
+      return true;
     },
   },
   parameters: {
@@ -1142,6 +1143,8 @@ export const WithSubmitEditsButton: Story = {
     columnDefinitions={columnDefinitions} 
     onSubmitEdits={(edits) => {
         alert("Submitting edits");
+        // Return true to indicate edits were successfully submitted and can be cleared from the table's edit state
+        return true;
     }}
   />`,
       },
--- a/packages/react-components/src/object-table/EditableCell.tsx
+++ b/packages/react-components/src/object-table/EditableCell.tsx
@@ -100,7 +100,7 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
       return;
     }
     const parsedValue = parseValueByType(inputValue, dataType);
-    onCellEdit?.(cellId, {
+    onCellEdit(cellId, {
       rowId,
       columnId,
       newValue: parsedValue as CellValue,
--- a/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
+++ b/packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
@@ -33,7 +33,6 @@ const columnDefinitions: Array<
     columnName: "Employee Number",
     editable: false,
   },
-  // With isVisible prop
   {
     locator: { type: "property", id: "jobTitle" },
     editable: true,


packages/react-components-storybook/.storybook/styles.css
```diff
@@ -0,0 +1,6 @@
+@layer storybook, tokens, components, themes;
+
+@import "../src/styles/storybook.css" layer(storybook);
+@import "@osdk/react-components/styles.css" layer(components);
+@import "@osdk/react-components-styles" layer(tokens);
+@import "./themes.css" layer(themes);
```
KuberSethi: nit: newline at bottom

packages/react-components/src/object-table/hooks/useEditableTable.ts
```diff
@@ -49,12 +50,14 @@ export interface UseEditableTableProps<
 }
 
 export interface UseEditableTableResult {
-  cellEdits: Record<string, CellValueState>;
+  isInEditMode: boolean;
+  handleEnableEditMode?: (enabled: boolean) => void;
+  cellEdits: Record<string, CellEditEvent<any, unknown>>;
```
KuberSethi: It looks like `CellEditEvent` has a `TData` generic, is it possible to use that to type this better?

packages/react-components/src/object-table/Table.tsx
```diff
@@ -42,16 +42,21 @@ declare module "@tanstack/react-table" {
   interface TableMeta<TData extends RowData = unknown> {
     onCellEdit?: (
       cellId: string,
-      state: CellValueState,
+      event: CellEditEvent<TData, unknown>,
     ) => void;
-    cellEdits?: Record<string, CellValueState>;
+    cellEdits?: Record<string, CellEditEvent<TData, unknown>>;
+    isInEditMode?: boolean;
+    focusedRowId?: string | null;
```
KuberSethi: Hmm it looks like we're adding this property but not assigning it wherever `TableMeta` is used? Or maybe I'm misunderstanding its purpose?

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -69,13 +73,16 @@ function parseValueByType(
   return parsedNumber;
 }
 
-export function EditableCell({
+export function EditableCell<TData = unknown>({
```
KuberSethi: Can we memoize this component? I think right now when `focusedRowId` changes that'll cause `TableRow` to rerender and then every `EditableCell` to also rerender, even ones that haven't been changed.

KuberSethi (commented):
Looks good, just a few minor things!

packages/react-components/src/object-table/hooks/useEditableTable.ts
```diff
@@ -69,31 +72,37 @@ export function useEditableTable<
...
+  const [cellEdits, setCellEdits] = useState<
+    Record<string, CellEditEvent<any, unknown>>
+  >(
     {},
   );
 
-  const handleCellEdit = useCallback(
-    (cellId: string, state: CellValueState) => {
-      const cellIdentifier = getCellIdentifier(cellId);
+  const handleEnableEditMode = useCallback((enabled: boolean) => {
```
nfdiop: nit: can we drop this function and use `setIsInEditMode` directly?

packages/react-components/src/object-table/hooks/useEditableTable.ts
```diff
@@ -103,13 +112,18 @@ export function useEditableTable<
...
 
   const handleSubmitEdits = useCallback(async () => {
-    await onSubmitEdits?.(cellEdits);
+    const edits = Object.values(cellEdits);
+    await onSubmitEdits?.(edits);
   }, [cellEdits, onSubmitEdits]);
 
   return {
+    isInEditMode,
+    handleEnableEditMode: !enableEditModeByDefault
```
nfdiop: should we use `isInEditMode` here instead?

packages/react-components-storybook/.storybook/styles.css
ziyunp: :( Is this something we can format with dprint / add to lint?

packages/react-components/src/object-table/Table.tsx
```diff
@@ -131,21 +128,63 @@ export function BaseTable<
...
+  const {
+    cellEdits,
+    enableEditModeByDefault,
+    isInEditMode,
+    onSubmitEdits,
+    onEnableEditMode,
+    clearEdits,
+  } = editableConfig ?? {};
+
+  const hasEditableColumns = table
```
KuberSethi: Sorry, think I missed on the first read but could we memoize this computation?

packages/react-components/src/object-table/hooks/useEditableTable.ts
```diff
@@ -69,31 +62,48 @@ export function useEditableTable<
...
+    (
+      cellId: string,
+      info: CellEditInfo<
+        Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>,
+        unknown
+      >,
+    ) => {
       // If value is changed back to original, remove it from edits
-      if (state.newValue === state.oldValue) {
+      if (info.newValue === info.oldValue) {
```
nfdiop: What happens when I do the following changes: "original" -> "new value" -> "new value" -> "original"?
I also assumed we wanted to store `rowData` to use it here?

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -113,6 +114,14 @@ const meta: Meta<EmployeeTableProps> = {
         defaultValue: { summary: "true" },
       },
     },
+    enableEditModeByDefault: {
+      description: "Whether the table is in edit mode by default.",
+      control: "boolean",
+      defaultValue: true,
```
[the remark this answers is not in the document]
nfdiop: Got it! Thanks for clarifying. In that case, happy to have default: true

packages/react-components/src/object-table/hooks/useEditableTable.ts
```diff
@@ -103,13 +113,19 @@ export function useEditableTable<
...
-    await onSubmitEdits?.(cellEdits);
+    const edits = Object.values(cellEdits);
+    return onSubmitEdits ? onSubmitEdits(edits) : false;
   }, [cellEdits, onSubmitEdits]);
 
+  const editModeConfig: EditMode = editMode === "always"
+    ? { type: "always", isActive: true }
+    : { type: "manual", isActive, setActive };
+
   return {
```
nfdiop: FLUP: could we memoize the return value?

PR #2620, opened 2026-02-27
Add support for stubbing queries and add exports

nihalbhatnagar:
Allows users to stub query outputs specified to certain parameters


packages/functions-testing.experimental/src/mock/__tests__/createMockClient.test.ts
nihalbhatnagar: Will probs flup and rename to whenObjectSet

PR #2647, opened 2026-03-04
add associationMap to link observation responses

KuberSethi:
link observations now include an associationMap so callers can determine which source object links to which targets

• add associationMap (source pk -> linked objects) to observeLinks callback args at all layers
• expose associationMap in useLinks react hook
• fix useObjectSet isLoading to respect enabled prop


lukeMuratore12 (approved):
I think this makes sense and would fit our use case very well! 

Leaving some thoughts here:

- From my understanding even though `useLinks` and `useOsdkObjects` with `pivotTo` do not share a cache, we'll still likely want to use `useLinks` for the benefit of this association map here
- Note for myself confirming that loading by rid with multiple objects (which was an issue for us in the past) was fixed here https://github.com/palantir/osdk-ts/pull/2481/

bryantpark04 (approved):
stamping to unblock

PR #2648, opened 2026-03-04
Preserve derived property values when subscription updates write to RDP cache keys

iw-an:
### Changes

- Fixes a bug where derived property (RDP) values were silently dropped from the object cache when subscription updates arrived with only base properties, causing sort failures and missing data on paginated lists using withProperties

### Problem

When observeList() is used with withProperties, objects fetched via fetchPage are stored in the cache with their derived property values. However, when a websocket subscription update arrives for the same object, the update payload only contains base properties. propagateWrite() wrote this base-only object directly to the RDP cache key, overwriting the previously cached derived property values.

This caused page 1 objects to lose their derived properties between page fetches. When page 2 arrived and triggered a client-side re-sort, the comparator saw undefined for the sort property on page 1 items, breaking sort order.

### DIFF
--- a/packages/client/src/observable/internal/object/ObjectsHelper.ts
+++ b/packages/client/src/observable/internal/object/ObjectsHelper.ts
@@ -139,12 +139,14 @@ export class ObjectsHelper extends AbstractHelper<
           }
         }
 
-        valueToWrite = mergeObjectFields(
-          valueToWrite,
-          actualRdpFields,
-          expectedRdpFields,
-          existing.value,
-        );
+        if (actualRdpFields.size !== expectedRdpFields.size) {
+          valueToWrite = mergeObjectFields(
+            valueToWrite,
+            actualRdpFields,
+            expectedRdpFields,
+            existing.value,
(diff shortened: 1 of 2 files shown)


packages/client/src/observable/internal/object/ObjectsHelper.ts
```diff
@@ -116,7 +117,37 @@ export class ObjectsHelper extends AbstractHelper<
...
+    // property values. Merge with the existing cached value so that RDP fields
+    // not present in the incoming object are preserved.
+    if (
+      valueToWrite !== tombstone
+      && existing?.value
+      && this.isObjectHolder(existing.value)
+    ) {
+      const expectedRdpFields = this.store.objectCacheKeyRegistry
+        .getRdpFieldSet(sourceCacheKey);
+      if (expectedRdpFields.size > 0) {
```
KuberSethi: should we also check `actualRdpFields.size === expectedRdpFields.size` and short circuit if they're the same?

PR #2657, opened 2026-03-05
Array Reducers and Struct Main Values in the OSDK

nihalbhatnagar:
## Summary

Adds `$applyModifiers` to `fetchPage` to support three property modifiers:

**`applyMainValue`** — extract the main value from a struct property (scalar if single field, subset struct if multiple):
```ts
await client(Employee).fetchPage({
  $applyModifiers: {
    addressStruct: "applyMainValue",
  },
});
// addressStruct: { city: string; zipCode: string } | undefined
```

**`applyReducers`** — reduce an array property to its single element type:
```ts
await client(Employee).fetchPage({
  $applyModifiers: {
    salaryHistory: "applyReducers",
  },
});
// salaryHistory: number | undefined (was number[] | undefined)
```

**`applyReducersAndExtractMainValue`** — combine both on arrays of structs (scalar if the struct's main value has a single field):
```ts
await client(Employee).fetchPage({
  $applyModifiers: {
    bonusHistory: "applyReducersAndExtractMainValue",
  },
});
// bonusHistory: number | undefined (single main-value field "amount")
```

## Test plan

- [ ] `pnpm check` passes
- [ ] `runReducerAndMainValueTest` e2e sends correct `selectV2` wire format
- [ ] Invalid modifier keys in `$applyModifiers` produce TypeScript errors


packages/e2e.sandbox.catchall/src/runReducerAndMainValueTest.ts
```diff
@@ -0,0 +1,29 @@
+import { ReducerTest } from "@osdk/e2e.generated.catchall";
+import { assert } from "console";
+import { client } from "./client.js";
+
+export async function runReducerAndMainValueTest(): Promise<void> {
+  const reducerTestObject = await client(ReducerTest).fetchPage(
+    {
+      $applyModifiers: {
```
nihalbhatnagar, line 8: One flag is that $select and $applyModifiers kind of override each other, so one of them is not going to redline if I select some and then apply modifiers to a property that wasn't selected

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -134,12 +141,17 @@ export interface AsyncIterArgs<
...
   PROPERTY_SECURITIES extends boolean = false,
+  MODIFIERS extends ApplyModifiersArg<Q, PropertyKeys<Q>> = {},
 > extends
   SelectArg<Q, K, R, S, RDP_KEYS, PROPERTY_SECURITIES>,
   OrderByArg<Q, PropertyKeys<Q> | RDP_KEYS, ORDER_BY_OPTIONS>
 {
   $__UNSTABLE_useOldInterfaceApis?: boolean;
   $includeAllBaseObjectProperties?: PropertyKeys<Q> extends K ? T : never;
+  $applyModifiers?:
+    & ApplyModifiersArg<Q, PropertyKeys<Q>>
```
nihalbhatnagar: Type intellisense and checking dont work without these, helps TS narrow

ssanjay1: huh interesting

packages/api/src/ontology/PropertyModifiers.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export type PropsWithOnlyMainValue<Q extends ObjectOrInterfaceDefinition> =
+  Exclude<PropsWithMainValue<Q>, PropsWithReducers<Q>>;
+
+export type PropsWithOnlyReducers<Q extends ObjectOrInterfaceDefinition> =
+  Exclude<PropsWithReducers<Q>, PropsWithMainValue<Q>>;
+
+export type ApplyModifiersArg<
+  Q extends ObjectOrInterfaceDefinition,
+  K extends PropertyKeys<Q> = PropertyKeys<Q>,
```
ssanjay1: Why did you capture K here?

packages/client/src/object/fetchPage.ts
```diff
@@ -46,6 +51,94 @@ import { extractObjectOrInterfaceType } from "../util/extractObjectOrInterfaceTy
 import { extractRdpDefinition } from "../util/extractRdpDefinition.js";
 import { resolveBaseObjectSetType } from "../util/objectSetUtils.js";
 
+/**
+ * Converts a PropertyModifierValue to the corresponding wire format loadLevel type.
+ */
+function modifierToLoadLevelType(
+  modifier: PropertyModifierValue,
+): "extractMainValue" | "applyReducers" | "applyReducersAndExtractMainValue" {
```
ssanjay1: can we use `LoadLevelType` as return type?

packages/client/src/object/fetchPage.ts
```diff
@@ -46,6 +51,94 @@ import { extractObjectOrInterfaceType } from "../util/extractObjectOrInterfaceTy
...
+          propertyIdentifier: { type: "property", apiName: prop },
+          loadLevel: { type: modifierToLoadLevelType(modifiersMap[prop]) },
+        });
+      } else {
+        entries.push({ type: "property", apiName: prop });
+      }
+    }
+
+    for (const [prop, modifier] of Object.entries(modifiersMap)) {
+      if (!select.includes(prop)) {
```
ssanjay1: So if we don't explicitly select a property but modify it we still return? Assuming you did this because couldn't get the types working to redline if select didn't include the modifier props? Should we just throw at runtime here? Feels weird that we're silently asking for a property someone didn't select

packages/api/src/ontology/PropertyModifiers.ts
```diff
@@ -0,0 +1,145 @@
...
+  PropertyKeys,
+} from "./ObjectOrInterface.js";
+import type {
+  CompileTimeMetadata,
+  ObjectMetadata,
+} from "./ObjectTypeDefinition.js";
+import type { WirePropertyTypes } from "./WirePropertyTypes.js";
+
+/** Modifier values users can specify */
+export type PropertyModifierValue =
```
ssanjay1, line 29: Should we expand the fetch page arg here to be an object instead of just a string? Trying to see if we need to future proof if more options are added later, e.g. how should the reducers work. Though I guess with those things configured at runtime our ability to do any compile time stuff also goes out the window

nihalbhatnagar: We can always override it if those options come, same with how we did eq. better to keep a simple api imo

packages/client/src/observable/OptimisticBuilder.ts
```diff
@@ -28,7 +23,7 @@ export interface OptimisticBuilder {
   createObject: <T extends ObjectTypeDefinition>(
     type: T,
     primaryKey: PrimaryKeyType<T>,
-    properties: Pick<Osdk.Instance<T>, PropertyKeys<T>>,
+    properties: Partial<Osdk.Instance<T>>,
```
ssanjay1: was this intentional?

packages/api/src/OsdkObjectFrom.ts
```diff
@@ -177,11 +184,11 @@ export type IsAny<T> = unknown extends T
 
 export type GetPropsKeys<
   Q extends ObjectOrInterfaceDefinition,
-  P extends PropertyKeys<Q>,
+  P,
```
ssanjay1: Assuming this broke things if you kept the narrowing?

nihalbhatnagar: Updated, remnant from previous APIs

packages/client/src/observable/OptimisticBuilder.ts
nihalbhatnagar: Yes the compiler can't resolve PropertyKeys<Q> is subset of the instance that easily anymore, but updated to the equivalent of what this was before

ssanjay1 (approved):
Approving to unblock, would be nice to explore types later if we can try redlining if we don't select properties we do modifiers on

PR #2748, opened 2026-03-16
Add media transformation api

lucy-k-wang:
## Summary
- Add experimental `transformAndWait` method on the OSDK client for media transformations
- Define fully-typed `MediaTransformation` types using the `$`-prefixed pattern, matching `GeoFilter`
- Add runtime-safe conversion from user-facing types to wire format via `makeMediaTransformation`

## Usage
```typescript
import { __EXPERIMENTAL__NOT_SUPPORTED_YET__transformAndWait } from "@osdk/api/unstable";

const result = await client(
  __EXPERIMENTAL__NOT_SUPPORTED_YET__transformAndWait,
).transformAndWait({
  mediaReference,
  transformation: {
    $image: {
      $encoding: "png",
      $operations: [{ $resize: { $width: 50, $height: 50 } }],
    },
  },
});
```


packages/client/src/util/transformAndWaitInternal.ts
```diff
@@ -0,0 +1,90 @@
...
+  const jobId = job.jobId;
+
+  if (status === "FAILED") {
+    throw new MediaTransformationFailedError(jobId);
+  }
+
+  const deadline = Date.now() + pollTimeoutMs;
+
+  while (status !== "SUCCESSFUL") {
+    if (Date.now() >= deadline) {
```
wxyangf: while date is fine for now, we should replace with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal#browser_compatibility temporal once it's in baseline, date has odd semantics that can result in unpredictable behavior.

wxyangf (commented):
I don't like this while loop construction, i think it makes the flow very difficult to reason about, especially since it's now a clean sleep but a no-op promise await?

ssanjay1 (commented):
Also, failing type checks looks like

ssanjay1 (approved):

PR #2797, opened 2026-03-20
devtools registration, vite plugin, and e2e wiring

KuberSethi:
glue that connects devtools to the app via registration and vite plugin

• registerDevTools entry point that wraps client and injects devtools UI
• vite plugin (osdkDevTools) that auto-injects fiber hook script
• e2e sandbox wired up with devtools dependency and plugin


KuberSethi:

- #2795
    - #2796
        - #2797 ◀

ziyunp (dismissed):

packages/react-devtools/package.json
```diff
@@ -26,7 +32,7 @@
     "lint": "eslint . && dprint check  --config $(find-up dprint.json)",
     "preview": "vite preview",
     "test": "vitest run",
-    "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
+    "transpileEsm": "rm -rf build/esm && vite build",
```
nfdiop, line 35: curious why only build/esm now?

packages/react-devtools/src/vite/hookInstaller.ts
```diff
@@ -0,0 +1,110 @@
...
+    window.__REACT_DEVTOOLS_GLOBAL_HOOK__ = {
+      // Renderer registry
+      renderers: new Map(),
+
+      // Indicate we support React Fiber architecture
+      supportsFiber: true,
+
+      // Called by React when a renderer initializes
+      inject(renderer) {
+        const id = this.renderers.size + 1;
```
nfdiop: using `this` could fail here depending on how this is invoked. Can we define renderers close to fiberRootsByRenderer then call here?

packages/react-devtools/package.json
KuberSethi: it's an esm-only package so there's no need for the others, the rm change here is more of a DX thing to avoid nuking your built types which will happen if you're rebuilding many things at once

packages/react-devtools/src/vite/hookInstaller.ts
KuberSethi: good catch, moved the renderers out

nfdiop (approved):

PR #2816, opened 2026-03-23
Add recursive query type support

nihalbhatnagar:
Summary

  Adds support for recursive/self-referential types in query definitions via the new typeReferences map on QueryTypeV2.

Generates types in a non-exported namespaced by their type ID. Expands type for intellisense


packages/generator/src/shared/getInterfaceTypesFromQueryDataType.ts
```diff
@@ -60,12 +60,10 @@ export function getInterfaceTypesFromQueryDataType(
     case "threeDimensionalAggregation":
     case "timestamp":
     case "twoDimensionalAggregation":
-    case "entrySet":
```
nihalbhatnagar, line 63: Duplicates

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -127,6 +130,10 @@ async function generateV2QueryFile(
     query.fullApiName,
   );
 
+  const typeRefs = query.raw.typeReferences ?? {};
+  const customTypesNs = generateCustomTypesNamespace(ontology, typeRefs);
+  const typeRefNames = buildTypeRefNames(typeRefs, true);
```
KuberSethi, line 135: could we just run this once and then pass the result into `generateCustomTypesNamespace` + reuse?

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -390,21 +395,67 @@ export function getQueryParamType(
...
   return paramType;
 }
+
+function buildTypeRefNames(
+  typeRefs: Record<string, QueryDataType>,
+  expand: boolean,
+): Map<string, string> {
+  const names = new Map<string, string>();
+  for (const id of Object.keys(typeRefs)) {
+    const sanitized = `$${id.replace(/-/g, "_")}`;
```
KuberSethi, line 433: Here and below in `generateCustomTypesNamespace`, unless this is widely known to be our format can we add a comment explaining why we're using this ID sanitization scheme

packages/generator/src/shared/getObjectTypesFromQueryDataType.ts
```diff
@@ -64,12 +64,10 @@ export function getObjectTypesFromQueryDataType(
     case "threeDimensionalAggregation":
     case "timestamp":
     case "twoDimensionalAggregation":
-    case "entrySet":
+    case "typeReference":
```
KuberSethi, line 67: this makes sense but why the other removals?

packages/generator/src/shared/getInterfaceTypesFromQueryDataType.ts
```diff
@@ -60,12 +60,10 @@ export function getInterfaceTypesFromQueryDataType(
...
+    case "typeReference":
```
KuberSethi, line 63: this makes sense but why the other removals?

packages/generator-converters/src/wireQueryDataTypeToQueryDataTypeDefinition.ts
KuberSethi: nit: remove

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -390,21 +395,67 @@ export function getQueryParamType(
...
+      paramType = `Partial<Record<${recurse(input.keyType, true)}, ${
+        recurse(input.valueType)
+      }>>`;
+      break;
+
+    case "typeReference":
+      if (!typeRefNames?.has(input.typeId)) {
+        throw new Error(`Unknown typeReference: ${input.typeId}`);
+      }
+      paramType = typeRefNames.get(input.typeId)!;
```
KuberSethi: do we need the `!` assertion?

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -390,21 +395,67 @@ export function getQueryParamType(
...
+function generateCustomTypesNamespace(
+  ontology: EnhancedOntologyDefinition,
+  typeRefs: Record<string, QueryDataType>,
+): string {
+  const names = buildTypeRefNames(typeRefs, false);
+  if (names.size === 0) return "";
+
+  const interfaces: string[] = [];
+  for (const [id, dt] of Object.entries(typeRefs)) {
+    if (dt.type !== "struct") continue;
```
KuberSethi: could you help me understand why we're skipping structs here but not in `buildTypeRefNames`? Guessing I'm missing something but just want to understand

packages/generator/src/v2.0/generatePerQueryDataFiles.test.ts
```diff
@@ -529,4 +529,587 @@ describe("generatePerQueryDataFiles", () => {
         "
       `);
   });
+
+  it("generates interfaces for recursive type references", async () => {
```
KuberSethi, line 533: Is it worth adding more tests? These are good but I think given the size/complexity of this change it could benefit from additional testing. Will leave it up to you though!

packages/generator-converters/src/wireQueryDataTypeToQueryDataTypeDefinition.ts
nihalbhatnagar: flop thanks

packages/generator/src/shared/getInterfaceTypesFromQueryDataType.ts
nihalbhatnagar: Media reference also a duplicate

packages/generator/src/v2.0/generatePerQueryDataFiles.test.ts
nihalbhatnagar: Is there a specific edge case you want a test for?  This feels like good coverage

packages/generator/src/shared/getObjectTypesFromQueryDataType.ts
nihalbhatnagar: same as above media reference above

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
nihalbhatnagar: No we need both the unexpanded names (because we only want to expand the top layer for intellisense and put the raw reference inside the actual struct

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
nihalbhatnagar: Sure I'll add the comment, but IDs start with numbers so we need the prefix

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
nihalbhatnagar: Oversight in buildTypeRefNames, but whats coming back over the wire is always struct. So im just gonna throw if we see a non struct in either place

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
nihalbhatnagar: just made a regular {} since map doesnt have type assertion and we needed it

packages/generator/src/v2.0/generatePerQueryDataFiles.test.ts
KuberSethi: Think it's fine, although potential improvements could be:
 - non-struct typeRefs
 - different positions of typeRef inside the struct

KuberSethi (approved):

PR #2837, opened 2026-03-24
wire shapes into existing hooks, add test utilities and shape tests

KuberSethi:
integrate shapes into useOsdkObject and useOsdkObjects with inline config overloads

• shape overloads on useOsdkObject and useOsdkObjects with inline and pre-built configs
• createMockObservableClient test utility with correct observeLinks payload shape
• useOsdkObject.shape.test.tsx covering single object shape observation

### DIFF
--- a/packages/react/src/testing/createMockObservableClient.ts
+++ b/packages/react/src/testing/createMockObservableClient.ts
@@ -258,7 +258,15 @@ export function createMockObservableClient(
 
     observeLinks: ((_objects, _linkName, _options, subFn): Unsubscribable => {
       setTimeout(() => {
-        (subFn as Observer<unknown>).next(makeListPayload("loaded"));
+        (subFn as Observer<unknown>).next({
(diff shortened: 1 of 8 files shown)


KuberSethi:

- #2836
    - #2837 ◀
        - #2890
            - #3091

packages/react/src/testing/createMockObservableClient.ts
```diff
@@ -0,0 +1,293 @@
...
+interface Subscription {
+  type: "object" | "list";
+  key: string;
+  observer: Observer<unknown>;
+}
+
+const NOOP_ASYNC = async () => {};
+
+function makeListPayload(
+  status: string,
```
nfdiop: suggestion (non-blocking): the `status` parameter in `makeListPayload` and `makeObjectPayload` is typed as `string` — a union type like `"loading" | "loaded" | "error" | "init"` would catch typos at compile time and make the mock's behavior self-documenting

PR #2929, opened 2026-04-03
add resolveToObjectType option for interface queries

KuberSethi:
## Summary

when loading objects via interface, consumers get interface-narrowed views by default (only interface properties, $apiName set to interface name). this forces apps to manually resolve interface link constraints to concrete link names via metadata calls and extra state management.

• add `resolveToObjectType` option to `useOsdkObjects` and `observeList` — skips `$as()` wrapping in `InterfaceListQuery`, returning full concrete OT instances with `$title` and all properties
• include `resolveToObjectType` in `ListCacheKey` to prevent cache collisions

## Test plan

- 9 unit tests in `useOsdkObjects.interfacePivotTo.test.tsx`
- e2e verified via office network app: Worker/Person interfaces with resolveToObjectType show 33 keys (full OT) vs 10 keys (interface view)


packages/client/src/observable/internal/list/InterfaceListQuery.ts
```diff
@@ -120,11 +118,13 @@ export class InterfaceListQuery extends ListQuery {
     return reloadDataAsFullObjects(this.store.client, data);
   }
 
-  protected createPayload(
-    params: CollectionConnectableParams,
-  ): ListPayload {
-    const resolvedList = params.resolvedData?.map((obj: ObjectHolder) =>
-      obj.$as(this.apiName)
+  private wrapObject(object: ObjectHolder): ObjectHolder | InterfaceHolder {
```
KuberSethi: here and below in this file are actual functional changes from this PR, calling out explicitly to help review bc of all the lint noise

lukeMuratore12 (dismissed):
LGTM! Thank you so much!

Out of curiosity, will this propogate to `useLinks` as well, or will that require separate changes? 

We use `useOsdkObjects` more heavily in our codebase so this is excellent but just curious.

packages/react/src/new/useOsdkObjects.ts
```diff
@@ -162,6 +163,31 @@ export interface UseOsdkObjectsOptions<
...
+   *
+   * @default false
+   *
+   * @example
+   * useOsdkObjects(MyInterface, {
+   *   pivotTo: "linkedItems",
+   *   resolveToObjectType: true,
+   * })
+   */
+  resolveToObjectType?: boolean | ObjectTypeDefinition;
```
nfdiop: I don’t think `resolveToObjectType: ObjectTypeDefinition` is sound. Querying an interface can return multiple implementing object types, and this option does not filter results at runtime. So `useOsdkObjects(Person, { resolveToObjectType: Employee })` can type `data` as `Employee[]` even if the server returns Contractors too.

Could we make `resolveToObjectType` boolean-only and return a broad concrete object type? Consumers that need a specific object type should be responsible for asserting the type 

PS:  Could we also change types so that object queries cannot pass the option?

Something like:

```
type ResolveToObjectTypeOption<Q extends ObjectOrInterfaceDefinition> =
    Q extends { type: "interface" }
      ? { resolveToObjectType?: boolean }
      : { resolveToObjectType?: never };

 export type UseOsdkObjectsOptions<
    T extends ObjectOrInterfaceDefinition,
    RDPs extends Record<string, SimplePropertyDef> = {},
  > = {
    // existing options except resolveToObjectType
    rids?: readonly string[];
    where?: WhereClause<T, RDPs>;
    // ...
    $includeAllBaseObjectProperties?: boolean;
  } & ResolveToObjectTypeOption<T>;
```

KuberSethi: Great feedback, I appreciate the thought that went into this!! Good ideas, incorporated them 😄

nfdiop (approved):

PR #2934, opened 2026-04-05
add resolveToObjectType support to useLinks

KuberSethi:
extends resolveToObjectType to the useLinks hook for interface link targets

• extract reloadDataAsFullObjects to shared utility used by both InterfaceListQuery and SpecificLinkQuery
• thread resolveToObjectType through ObserveLinks.Options, cache key, LinksHelper, and SpecificLinkQuery
• add option to useLinks hook with pass-through to observeLinks

### DIFF
--- a/.changeset/resolve-to-object-type-links.md
+++ b/.changeset/resolve-to-object-type-links.md
@@ -1,0 +1,6 @@
+---
+"@osdk/client": minor
+"@osdk/react": minor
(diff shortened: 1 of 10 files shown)


KuberSethi:
addressed in the last PR this is on top of, did the same thing here

nfdiop (approved):

PR #2947, opened 2026-04-06
allow pivotTo and useLinks to accept ILT names on object type sources

KuberSethi:
consumers loading by concrete object type can now pass interface link type (ILT) names to pivotTo and useLinks without needing additional requests to discover concrete link names. This is helpful for users who are loading by object type but in the context of heavy interface usage. 

### pivotTo with ILT names

```typescript
// BEFORE: additional requests to discover the concrete link name if working in interfaces
const concreteLinkName = useLinkNameResolver(sourceType, targetInterface);
const results = useOsdkObjects(ObjectTypeA, {
  pivotTo: concreteLinkName,
  enabled: !!concreteLinkName, // can't render until resolved
});

// AFTER: just use the ILT name directly
const results = useOsdkObjects(ObjectTypeA, {
  pivotTo: "com.example.InterfaceA.linkToB",
});
```

### useLinks with ILT names

```typescript
// BEFORE
const linkName = useLinkNameResolver(sourceType, targetInterface);
const { links } = useLinks(myObject, linkName, {
  enabled: !!linkName,
});

// AFTER
const { links } = useLinks(myObject, "com.example.InterfaceA.linkToB");
```

### concrete link names still work exactly as before

```typescript
// still fully typed, still autocompletes
const { links } = useLinks(employee, "lead");
const offices = useOsdkObjects(Employee, { pivotTo: "primaryOffice" });
```

## how it works

• on first fetch, `ListQuery.#resolveIltPivot` loads the OT definition and checks if the link name is in the concrete links map
• if not, it validates the link exists as an ILT via `findIltLinkDef`, then rebuilds the wire object set as `interfaceLinkSearchAround` (which the server already knows how to resolve)
• `SpecificLinkQuery` does the same check inline in `fetchPageData` for `useLinks`
• websocket subscriptions are deferred until after ILT resolution to avoid subscribing to a stale object set
• concrete link names skip all of this and behave identically to before


lukeMuratore12 (dismissed):
LGTM! This will work great for us, thank you!

PR #2963, opened 2026-04-08
Add form field validation to BaseForm/ActionForm

nfdiop:
## Summary
- Wire react-hook-form validation using `onTouched` mode (validate on blur, revalidate on change after first error)
- Add `extractValidationRules` utility that derives RHF rules from field constraint props (`min`/`max`, `minLength`/`maxLength`, `maxSize`, `required`)
- Show red border + error message on invalid fields via `data-invalid` attribute and CSS tokens
- Submit button stays enabled with error summary tooltip when form has validation errors
- Export `ValidationError` discriminated union so users can customize messages via `onValidationError`
- Clamp NumberInputField stepper to min/max bounds

## Test plan
- [x] Unit tests for `extractValidationRules` (19 tests covering all field types, custom validate, onValidationError overrides)
- [x] Integration tests in `BaseForm.test.tsx` (required error on blur, minLength error, error clears on fix, prevents submit when invalid)
- [x] Typecheck passes (`pnpm turbo typecheck --filter=@osdk/react-components`)
- [x] All existing tests still pass
- [ ] Manual testing with form page for visual error states

## Notes
- Tracking more validation to do as a FLUP. For example for integer fields, we need to validate the shape

Screenshot 2026-04-10 at 10 11 41
Screenshot 2026-04-10 at 10 11 50
Screenshot 2026-04-10 at 10 11 54
Screenshot 2026-04-10 at 10 13 41
Screenshot 2026-04-10 at 10 13 50


packages/react-components/src/action-form/ActionFormApi.ts
```diff
@@ -135,7 +135,7 @@ export type BaseFormProps =
 interface BaseFormCommonProps {
   formTitle?: string;
   fieldDefinitions: ReadonlyArray<RendererFieldDefinition>;
-  onSubmit: (formState: Record<string, unknown>) => void;
+  onSubmit: (formState: Record<string, unknown>) => Promise<void> | void;
```
nfdiop, line 138: so that we can await

packages/react-components/src/action-form/BaseForm.module.css
```diff
@@ -30,9 +30,40 @@
...
+  cursor: help;
+
+  /*
+ * FIX: The SVG icon appears truncated/invisible because the path coordinates
+ * (approx x:80, y:100) are far outside the defined viewBox="0 0 16 16".
+ * Since the SVG markup/viewBox cannot be modified directly, we force
+ * overflow: visible to allow the path to be rendered outside the
+ * clipping boundary of the SVG container.
+ */
+  & svg {
```
nfdiop: Curious if there is a better way to fix

packages/react-components/src/action-form/BaseForm.tsx
```diff
@@ -42,28 +48,78 @@ export const BaseForm: React.FC<BaseFormProps> = memo(function BaseFormFn({
...
+  const {
+    isPending: isSubmitting,
+    error: submissionError,
+    execute: executeSubmit,
+    clearError,
+  } = useAsyncAction(onSubmit);
+  const submissionErrorMessage = submissionError != null
+    ? submissionError instanceof Error
+      ? submissionError.message
+      // TODO: provide better error message
```
nfdiop, line 75: Tracking

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -81,20 +81,20 @@ export interface FormFieldDefinition<
...
   /**
-   * Additional function to validate the field
+   * Additional function to validate the field.
    *
-   * @param value the current field value
-   * @returns a boolean promise indicating whether the value is valid
+   * Return `undefined` if valid, or an error message string if invalid.
    */
-  validate?: (value: FieldValueType<Q, K>) => Promise<boolean>;
+  validate?: (value: FieldValueType<Q, K>) => Promise<string | undefined>;
```
nfdiop, line 97: I prefer when users return a string we can use to provide an explanation. Fine to change since the API is not released yet

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -112,16 +112,18 @@ export interface FormFieldDefinition<
...
-  | "max"
-  | "minLength"
-  | "maxLength"
-  | "pattern"
-  | "validate";
+/**
+ * A discriminated union describing which validation rule failed and the
+ * constraint data the user needs to build a meaningful error message.
+ */
+export type ValidationError =
```
nfdiop, line 119: Changed so the error message can use the validation config

packages/react-components/src/shared/hooks/useIsMounted.ts
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+import { useEffect, useRef } from "react";
+
+/**
+ * Returns a ref whose `.current` is `true` while the component is mounted
+ * and `false` after unmount. Use to guard async state updates.
+ */
+export function useIsMounted(): Readonly<React.RefObject<boolean>> {
```
nfdiop, line 23: Simple hook so no tests added

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -112,16 +112,18 @@ export interface FormFieldDefinition<
...
+  | { type: "required" }
+  | { type: "min"; min: number | Date }
+  | { type: "max"; max: number | Date }
+  | { type: "minLength"; minLength: number }
+  | { type: "maxLength"; maxLength: number }
+  | { type: "maxSize"; maxSize: number }
```
ziyunp, line 125: Nit: Can actually rename minLength and maxLength to minSize and maxSize so we don't need to have maxLength & maxSize

packages/react-components/src/action-form/BaseForm.tsx
```diff
@@ -14,16 +14,22 @@
...
+import { Tooltip } from "../base-components/tooltip/Tooltip.js";
+import { useAsyncAction } from "../shared/hooks/useAsyncAction.js";
 import type { BaseFormProps } from "./ActionFormApi.js";
 import styles from "./BaseForm.module.css";
 import { FieldBridge } from "./fields/FieldBridge.js";
 import type { RendererFieldDefinition } from "./FormFieldApi.js";
 import { FormHeader } from "./FormHeader.js";
 
+// Using a lower trigger delay for tooltips in forms so the user can see the errors quickly
+const TOOLTIP_TRIGGER_DELAY_MS = 200;
```
ziyunp, line 31: Nit: Move to shared constants file

packages/react-components/src/action-form/BaseForm.tsx
```diff
@@ -42,28 +48,83 @@ export const BaseForm: React.FC<BaseFormProps> = memo(function BaseFormFn({
...
+  const {
+    control,
+    trigger,
+    getValues,
+    formState: { errors },
+  } = useForm<Record<string, unknown>>({
+    // Validate on blur first, then revalidate on change after the first
+    // error. This gives the user a chance to finish typing before seeing
+    // errors, while staying responsive once an error is surfaced.
+    mode: "onTouched",
```
ziyunp, line 60: This seems like a good feature, do you think we can wrap the editable table with RHF? We're now handling this ourselves

packages/react-components/src/action-form/BaseForm.module.css
ziyunp: We've used icons in multiple places and haven't seen the need for this... I'd also suggest to see if Blueprint applies this css in their global file

packages/react-components/src/action-form/utils/extractValidationRules.ts
```diff
@@ -0,0 +1,180 @@
...
+      if (max != null) {
+        const msg = getMessage(fieldDef, { type: "max", max });
+        validateFns.max = (value) =>
+          value instanceof Date && value.getTime() > max.getTime()
+            ? msg
+            : true;
+      }
+      break;
+    }
+    case "FILE_PICKER": {
```
ziyunp, line 97: For future, add validation for accepted file type

packages/react-components/src/action-form/utils/extractValidationRules.ts
```diff
@@ -0,0 +1,180 @@
...
+): RhfRules {
+  const rules: RhfRules = {};
+
+  if (fieldDef.isRequired) {
+    rules.required = getMessage(fieldDef, { type: "required" });
+  }
+
+  const validateFns: Record<
+    string,
+    (value: unknown) => string | true | Promise<string | true>
```
ziyunp, line 44: Q: Why return true and string? Can we match the userValidate function where we return null if no error?

packages/react-components/src/action-form/fields/TextAreaField.tsx
```diff
@@ -48,6 +49,7 @@ export function TextAreaField({
       minLength={minLength}
       maxLength={maxLength}
       render={renderTextarea}
+      aria-invalid={error != null || undefined}
```
ziyunp, line 52: Q: Why need to set to undefined instead of let it be false when error == null?

.changeset/salty-pandas-jump.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/react-components": patch
```
ziyunp: Nit: I'd set to minor for changes visible to users

ziyunp (dismissed):

packages/react-components/src/action-form/BaseForm.tsx
nfdiop: Will it be used in other places? If not, I prefer to keep here or use inside the Tooltip component

packages/react-components/src/action-form/BaseForm.tsx
nfdiop: I'm not sure but if that is possible, we should do it since it simplifies

packages/react-components/src/action-form/utils/extractValidationRules.ts
nfdiop: Apparently that is what RHF does so I thought it could be valuable. Making the change

packages/react-components/src/action-form/BaseForm.module.css
nfdiop: Looks like it had been introduced in the last version of BP (see usage in pdf viewer below). Pushed a fix for all icons.

Image

packages/react-components/src/action-form/BaseForm.tsx
ziyunp: There are other Tooltip usage, e.g. in EditableCell and I'd want to apply the same default delay

ziyunp (approved):

packages/react-components/src/action-form/FormFieldApi.ts
nfdiop: So I kept it because it looks clearer to me that size is for the files., length is for the number of characters

packages/react-components/src/action-form/BaseForm.tsx
nfdiop: Sounds good. Let me FLUP

packages/react-components/src/action-form/fields/TextAreaField.tsx
nfdiop: Without || undefined, you'd get aria-invalid="false" on every valid field, which is unnecessary DOM noise

PR #2965, opened 2026-04-08
Use v2 dev mode manifest endpoint in widget vite plugin

amishshah:
## Summary

- Switch the widget dev mode setup flow to use the v2 `setWidgetSetManifest` and `enable` endpoints, which accept a rich manifest instead of just script/stylesheet overrides
- Dev mode now sends the full widget configuration (name, description, parameters, events, permissions, OSDK input spec) to Foundry — matching what production builds publish
- Extract shared `buildWidgetManifestConfig` function used by both build and dev manifest construction, eliminating duplicated widget-config-to-manifest conversion logic

## Motivation

Previously, the dev mode `/finish` handler only sent script entrypoint URLs per widget to Foundry, discarding all other widget metadata. This meant dev mode preview couldn't display widget parameters, events, or descriptions. The new v2 endpoint accepts a manifest in the same format as `.palantir/widgets.config.json`, enabling a richer dev mode preview experience.

## Changes

- **`buildWidgetSetManifest.ts`** — Extract `buildWidgetManifestConfig` as a shared export; refactor `buildWidgetManifest` to use it
- **`buildDevModeManifest.ts`** (new) — Builds a dev settings manifest using `buildWidgetManifestConfig`, `extractWidgetConfig`, and `getWidgetSetInputSpec`
- **`network.ts`** — Replace `setWidgetSetSettings` with `setWidgetSetManifest` targeting v2 endpoints; update `enableDevMode` URL
- **`publishDevModeSettings.ts`** — Accept manifest instead of override map; add `InvalidManifest` error hint
- **`FoundryWidgetDevPlugin.ts`** — Accept `pluginOptions`, use `buildDevModeManifest` instead of `getWidgetIdOverrideMap`
- **`index.ts`** — Pass plugin options to dev plugin
- Delete `getWidgetIdOverrideMap.ts` (replaced by `buildDevModeManifest.ts`)


packages/widget.vite-plugin/src/build-plugin/__tests__/buildWidgetSetManifest.test.ts
```diff
@@ -14,248 +14,217 @@
...
-      true,
-    );
+    expect(manifest.manifestVersion).toBe("1.0.0");
+    expect(manifest.widgetSet.rid).toBe(WIDGET_SET_RID);
+    expect(manifest.widgetSet.version).toBe(WIDGET_SET_VERSION);
+    expect(manifest.widgetSet.inputSpec).toBe(widgetSetInputSpec);
+    expect(Object.keys(manifest.widgetSet.widgets)).toEqual([
+      "widgetOne",
+      "widgetTwo",
+    ]);
```
tzyl, line 199: Is the intention that this replaces "builds a valid manifest with multiple widgets" and only checks the minimal structure and everything else like parameters and events are now covered in the buildWidgetManifestConfig tests?

packages/widget.vite-plugin/src/dev-plugin/buildDevModeManifest.ts
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
```
tzyl: Couple of these new files are out of date we can fix up

packages/widget.vite-plugin/src/dev-plugin/FoundryWidgetDevPlugin.ts
```diff
@@ -176,19 +180,15 @@ export function FoundryWidgetDevPlugin(): Plugin {
...
             getBaseHref(server),
+            pluginOptions,
           );
-          await publishDevModeSettings(
-            server,
-            widgetIdToOverrides,
-            getBaseHref(server),
-            res,
-          );
+          await publishDevModeSettings(server, manifest, res);
```
tzyl, line 191: I've noticed an unrelated bug which is more visible when running the plugin locally:

Image

There's a poll interval for finishing setting up dev mode from the plugin app and if this takes >250ms to respond then the interval will trigger again leading to setting dev mode settings multiple times:
https://github.com/palantir/osdk-ts/blob/main/packages/widget.vite-plugin/src/client/app.tsx#L36

It looks like it's been there for a long time and is relatively harmless but worth flagging and we can fix separately!

packages/widget.vite-plugin/src/dev-plugin/network.ts
```diff
@@ -74,7 +49,10 @@ export function enableDevMode(
   foundryUrl: string,
   viteMode: string | undefined,
 ): Promise<Response> {
-  const url = new URL("api/v2/widgets/devModeSettings/enable", foundryUrl);
+  const url = new URL(
+    "api/v2/widgets/devModeSettingsV2/enable",
+    foundryUrl,
+  );
   url.searchParams.set("preview", "true");
   return fetch(
```
tzyl, line 57: Is it easy to unit test the URL we fetch doesn't have the double slash?

tzyl (dismissed):

tzyl (approved):

PR #2970, opened 2026-04-08
[Storybook] Add experimental tag and accessible plugins

ziyunp:
- added experimental tag
- added accessible plugin
- reorganised storybook
- added ColumnConfigDialog and MultiColumnSortDialog stories to object table building blocks
- export MultiColumnSortDialog in @osdk/react-components

Screenshot 2026-04-09 at 13 12 02

Screenshot 2026-04-08 at 15 14 36


Leolide (dismissed):
LGTM

nfdiop (commented):
Could we add an `Experimental` section instead of tags? In the future, we'll have many experimental components and it would be weird to see so many tags.

ziyunp:
> Could we add an `Experimental` section instead of tags? In the future, we'll have many experimental components and it would be weird to see so many tags.

Updated Components/ section to Experimental/ and keeping the tags

nfdiop (approved):

PR #2973, opened 2026-04-09
add cbac classification constraint, validation, and tooltip features

KuberSethi:
cbac picker polish pass FLUPS

• add MaxClassificationConstraint type and MaxClassificationField component with validationCallouts slot on BaseCbacPicker
• replace generic validation messages with priority-ordered error messages using userSatisfiesMarkings and disallowedMarkingIds from restrictions API to match existing error messages in-platform
• add disallowed marking tooltips with aria-disabled pattern for hover accessibility on disabled buttons
• misc minor CSS tweaks and hover/disable behavior updates

image


packages/cbac-components/src/cbac-picker/base/MarkingButton.tsx
```diff
@@ -39,24 +42,79 @@ const selectionStateClassMap: Record<
...
-  return (
+  const handleToggle = React.useCallback(() => {
+    onToggle(id);
+  }, [onToggle, id]);
+
+  const hasDescription = description !== undefined
+    && description.length > 0;
+  const isButtonDisabled = disabled ?? isDisallowed(selectionState);
+  const showTooltip = hasDescription || isDisallowed(selectionState);
+  const hint = getSelectionHint(selectionState);
```
nfdiop, line 77: question (non-blocking): `showTooltip` gates on `isDisallowed(selectionState)`, but `getSelectionHint` returns text for `SELECTED` ("Click to deselect") and `IMPLIED` ("Implied by another marking") too. Since the tooltip wrapper won't render for those states, those hints are unreachable.
 Is that intentional?

packages/cbac-components/src/cbac-picker/base/MarkingButton.module.css
```diff
@@ -74,3 +73,32 @@
 .markingButton:disabled {
   cursor: not-allowed;
 }
+
+.tooltip {
+  z-index: var(--osdk-cbac-marking-tooltip-z-index);
```
nfdiop, line 78: suggestion (non-blocking): This `.tooltip` block (z-index, max-width, padding, background, border, border-radius, box-shadow, font-size, line-height) is copy-pasted identically in `OverflowItem.module.css` and again as `.infoTooltip` in `CategoryMarkingGroup.module.css`. Would it make sense to extract a shared `CbacTooltip.module.css` or a shared tooltip component that owns these styles? Fine as a FLUP.

packages/cbac-components/src/cbac-picker/useCbacPickerState.ts
```diff
@@ -64,27 +67,38 @@ export function useCbacPickerState(
...
     error: bannerError,
+    refetch: refetchBanner,
   } = useCbacBanner({ markingIds: stableSelectedIds });
+
+  const bannerRef = React.useRef(latestBanner);
+  if (latestBanner != null) {
+    bannerRef.current = latestBanner;
+  }
+  const banner = latestBanner ?? bannerRef.current;
   const {
```
nfdiop, line 90: nice: smart pattern — keeping the last non-null banner in a ref so the UI doesn't flash empty while the new banner loads. Clean way to handle the transient `undefined` during refetch.

packages/cbac-components/src/cbac-picker/utils/validationMessages.test.ts
```diff
@@ -0,0 +1,119 @@
...
+const validBase: ValidationContext = {
+  isValid: true,
+  requiredMarkingGroups: [],
+  selectedIds: ["m1"],
+  disallowedMarkingIds: [],
+  userSatisfiesMarkings: true,
+};
+
+describe("getSubmitDisabledReason", () => {
+  it("returns undefined when isValid is true", () => {
```
nfdiop, line 30: nice tests!

nfdiop (dismissed):

packages/cbac-components/src/cbac-picker/base/MarkingButton.module.css
KuberSethi: I was considering it but we don't typically pull out CSS for things like this right? I haven't seen so at least in internal codebases or in this repo.

nfdiop (approved):

packages/cbac-components/src/cbac-picker/base/MarkingButton.tsx
KuberSethi: yeah this is intentional, we only show hints when the markings have descriptions which is not always guaranteed. This is effectively the same behavior as what exists internally. If this is problematic though happy to change later if users complain.

PR #2974, opened 2026-04-09
[Components] Optimise function columns queries

ziyunp:
1. Added maxConcurrentRequests config to allow users to limit the number of parallel network requests
2. Added pageSize config to customise ObjectTable page size
3. Optimised the function columns queries to batch query by page size
4. Removed the grouping of queries by apiName as the same request should be cached if within the dedupe interval

Tested with 10k rows and this works well. Previously 3 out of 5 requests failed, and all are passing now.

https://github.com/user-attachments/assets/e8b98542-a0b5-4c9c-9caf-0eef3123dde1


packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
```diff
@@ -45,7 +45,6 @@ const columnDefinitions: Array<
       getFunctionParams: (objectSet: any) => ({ employees: objectSet }),
       getKey: (obj: any) => `${obj.$objectType}:${obj.$primaryKey}`,
       getValue: (data: { daysSinceStart: any }) => data?.daysSinceStart,
-      dedupeIntervalMs: 10_000, // Short dedupe for sandbox testing
```
nfdiop, line 48: is this intentional?

packages/react-components/src/object-table/hooks/useFunctionColumnsData.ts
```diff
@@ -44,27 +50,20 @@ export interface FunctionColumnData {
...
+  getKey: (
+    object: Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>,
+  ) => string;
 };
 
-// Function column data is readOnly and can be cached aggressively,
-// so we set a longer dedupe interval to maximize cache hits
-export const DEFAULT_DEDUPE_INTERVAL_MS = 300_000; // 5 minutes
+// Re-export for backward compatibility
+export const DEFAULT_DEDUPE_INTERVAL_MS: number =
```
nfdiop: Hum, do we need to since I don't expect users to start relying on this constant yet?

packages/react-components/src/object-table/hooks/useObjectTableData.ts
```diff
@@ -72,9 +74,12 @@ export function useObjectTableData<
   objectSet?: ObjectSet<Q>,
   objectSetOptions?: ObjectSetOptions<Q>,
   dedupeIntervalMs?: number,
+  maxConcurrentRequests?: number,
+  pageSize?: number,
```
nfdiop: nit: can you specify pageSize = DEFAULT_PAGE_SIZE here so users won't accidentally use pageSize vs resolvedPageSize. Same for the dedupeInterval

packages/react-components/src/object-table/hooks/useObjectTableData.ts
```diff
@@ -157,11 +162,18 @@ export function useObjectTableData<
   // Get the result from the appropriate hook
   const baseResult = shouldUseObjectSet ? objectSetResult : osdkObjectsResult;
 
+  const primaryKeyApiName = objectOrInterfaceType.type === "object"
+    ? (objectOrInterfaceType as ObjectTypeDefinition).primaryKeyApiName
```
nfdiop: question (blocking): why do we need the type assertion when we have objectOrInterfaceType.type === "object"?

packages/react-components/src/object-table/utils/constants.ts
nfdiop, line 1: Nice centralizing magic numbers

packages/react/src/new/__tests__/useOsdkFunctions.test.tsx
```diff
@@ -551,8 +571,239 @@ describe("useOsdkFunctions", () => {
...
+      {
+        dedupeInterval: 5000,
+        dependsOn: [TestObjectType],
+        dependsOnObjects: [testObject],
+      },
       expect.objectContaining({ next: expect.any(Function) }),
     );
   });
+
+  describe("maxConcurrent", () => {
```
nfdiop, line 583: nice tests

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -209,6 +211,24 @@ export interface ObjectTableProps<
...
+  /**
+   * Maximum number of concurrent requests for function-backed columns.
+   * Limits the number of function queries that can run in parallel.
+   *
+   * If undefined, all function-backed column requests will run concurrently
+   * without a limit.
+   *
+   * @default undefined
+   */
+  maxConcurrentRequests?: number;
```
nfdiop: Should we define a good default here?

packages/react/src/new/useOsdkFunctions.ts
```diff
@@ -48,6 +48,16 @@ export interface UseOsdkFunctionsProps {
    * @default true
    */
   enabled?: boolean;
+
+  /**
+   * Maximum number of queries to execute concurrently.
+   * When set, queries are subscribed in batches — the next query starts
+   * only after a running one completes (status "loaded" or "error").
+   * Useful for large payloads that may exceed server time limits.
```
nfdiop, line 55: Do we know the server limit?

packages/react/src/new/useOsdkFunctions.ts
```diff
@@ -56,6 +66,129 @@ export type UseOsdkFunctionsResult = Array<
...
 
+/**
+ * Creates a composite external store that manages N function subscriptions
+ * through a single `useSyncExternalStore` interface.
+ *
+ * Unlike `makeExternalStore` (which wraps a single Observer), this manages
+ * N observers funneled into one snapshot array. Supports optional concurrency
+ * limiting via `maxConcurrent` to stagger subscriptions.
+ */
+function createCompositeExternalStore(
```
nfdiop: Can we move this to a different file and add tests?

packages/react-components/src/object-table/hooks/useFunctionColumnsData.ts
```diff
@@ -79,164 +78,251 @@ export function useFunctionColumnsData<
...
+  }
+  if (objectKey in merged.functionsMap) {
+    const rawData = merged.functionsMap[objectKey];
+    return { isLoading: false, data: getValue ? getValue(rawData) : rawData };
+  }
+  // Key not in results — still loading, or query returned no data for this object
+  return { isLoading: merged.isLoading, data: prevData };
+}
+
+function chunk<T>(array: T[], size: number): T[][] {
```
nfdiop: Could we use lodash since it is already here?

nfdiop (commented):
Thanks for handling this! I like how we now show the first pages first :) 
Left few comments/question. 
Could you extract some of the logic inside useFunctionsColumnData into pure functions for better readability? Some suggestion here https://github.com/palantir/osdk-ts/pull/2977/changes#diff-0552bc1ffa338037f7413b5becc3750048b8aec75837b10b1ef684ad88e29944

packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
ziyunp: Yes I wanted to test the default dedupe interval

packages/react-components/src/object-table/ObjectTableApi.ts
ziyunp: I think undefined is a good default because it means to parallelise all requests

packages/react/src/new/useOsdkFunctions.ts
ziyunp: I don't, but when I was testing a 10k objects object set with 5 parallel requests, 3 of them were timing out

packages/react-components/src/object-table/hooks/useFunctionColumnsData.ts
ziyunp: agree!

packages/react-components/src/object-table/hooks/useFunctionColumnsData.ts
ziyunp: good catch

packages/react/src/new/useOsdkFunctions.ts
nfdiop: Can we check with them so we provide good defaults? if the user doesn't pass `maxConcurrent`, this should work

ziyunp: After I paginate the object set for queries, this works with all 5 queries undefined.

nfdiop (dismissed):

nfdiop (approved):

PR #2975, opened 2026-04-09
Replace DatetimePickerField button with text input + DateRangeInput

nfdiop:
## Summary

- Replace button trigger with text input — users can type dates directly or pick from calendar
- Add `DateRangeInputField` with two-input layout, two-month calendar, and range selection
- Add `showTime` prop for both components with time picker below calendar
- Locale-friendly idle display format (e.g. "Jan 15, 2024"), editable ISO format on focus
- Shared `useDateEditState` hook for two-phase display pattern (idle vs editing)
- Full a11y: `role="combobox"`, `aria-expanded`, `aria-controls`, focus sentinels for Tab cycling
- CSS tokens for all calendar styling (`--osdk-datetime-*`), including day hover/selected/range states

## Test plan

- [x] Typecheck passes (`pnpm turbo typecheck --filter=@osdk/react-components`)
- [x] All 525 tests pass (`pnpm turbo test --filter=@osdk/react-components`)
- [x] Lint clean on changed files
- [x] dprint formatted

Screenshot 2026-04-13 at 15 43 46

Screenshot 2026-04-13 at 16 29 47

Screenshot 2026-04-13 at 16 31 39


packages/react-components-styles/src/tokens/datetime-picker.css
```diff
@@ -1,28 +1,19 @@
 /* Datetime Picker Styling */
 :root {
-  /* Input */
-  --osdk-datetime-input-font-family: var(--osdk-typography-family-default);
```
nfdiop, line 4: moved below

packages/react-components/src/action-form/fields/useDateEditState.ts
```diff
@@ -0,0 +1,144 @@
...
+  max,
+}: UseDateEditStateConfig): UseDateEditState {
+  const [isEditing, setIsEditing] = useState(false);
+  const [inputValue, setInputValue] = useState("");
+
+  // Sync inputValue when external value changes (e.g. parent resets)
+  const prevValueTimeRef = useRef<number | null>(value?.getTime() ?? null);
+  const currentValueTime = value?.getTime() ?? null;
+  if (prevValueTimeRef.current !== currentValueTime) {
+    prevValueTimeRef.current = currentValueTime;
```
ziyunp: question: The external value sync compares `value?.getTime()` but doesn't track `editFormatFn` identity. If a parent re-renders with the same `value` but a different `editFormatFn` (e.g. toggling `showTime`), `inputValue` won't update until the next value change. Is that the intended behavior?

packages/react-components/src/shared/dateUtils.ts
```diff
@@ -96,11 +96,22 @@ export function parseDatetimeFromDisplay(
   value: string | undefined | null,
 ): Date | undefined {
   if (!value) return undefined;
-  const normalized = value.includes("T") ? value : value.replace(" ", "T");
+  const normalized = value.includes("T") ? value : value.replace(/\s/, "T");
```
ziyunp, line 84: Question: I'm always concerned when we manipulate a date string like this. This means to replace any space with T. What is the intention here?

packages/react-components-styles/src/tokens/datetime-picker.css
```diff
@@ -1,28 +1,19 @@
...
   /* Popover */
   --osdk-datetime-popover-bg: var(--osdk-background-primary);
-  --osdk-datetime-popover-border: var(--osdk-surface-border);
+  --osdk-datetime-popover-border: var(--osdk-surface-border-width) solid
+    var(--osdk-surface-border-color-strong);
   --osdk-datetime-popover-border-radius: var(--osdk-surface-border-radius);
   --osdk-datetime-popover-padding: calc(var(--osdk-surface-spacing) * 2);
   --osdk-datetime-popover-z-index: var(--osdk-surface-z-index-3);
+  /* TODO: replace with a shared --osdk-surface-shadow-* token for dark mode */
+  --osdk-datetime-popover-shadow: 0 4px 20px rgba(0, 0, 0, 0.2),
```
ziyunp: Suggestion: There are blueprint tokens for --bp-surface-shadow-*. You can just map them to --osdk-surface-shadow-* in base.css

packages/react-components/src/action-form/fields/DateRangeInputField.tsx
```diff
@@ -0,0 +1,496 @@
...
+    value: endDate,
+    displayFormatFn,
+    editFormatFn,
+    parseFn,
+    min,
+    max,
+  });
+
+  // --- Cross-input error: overlapping range ---
+  const overlappingError: "overlapping" | null = (() => {
```
ziyunp: Question: Why not return boolean?

nfdiop: Good catch. Because previously we were returning another type of error. Cleaning up..

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -172,8 +173,63 @@ export interface DatetimePickerFieldProps extends BaseFormFieldProps<Date> {
...
+   * When typing, the input shows the parsable format (YYYY-MM-DD or YYYY-MM-DD HH:mm).
+   * Provide a matching `parseDate` if using a custom format.
+   */
+  formatDate?: (date: Date) => string;
+
+  /**
+   * Parses a user-typed string back into a Date.
+   * Must be the inverse of `formatDate` — if `formatDate(d)` produces string `s`,
+   * then `parseDate(s)` must return an equivalent Date.
+   * When omitted, defaults to parsing "YYYY-MM-DD" (date-only) or "YYYY-MM-DD HH:mm" (with time).
```
ziyunp, line 189: 👍 very clear instruction here

packages/react-components-styles/src/tokens/datetime-picker.css
nfdiop: thanks!

packages/react-components/src/shared/dateUtils.ts
nfdiop: added comment

packages/react-components/src/action-form/FormFieldApi.ts
ziyunp, line 149: Suggestion: Add a description about how we handle timezone, like DayPicker stated their default here https://daypicker.dev/localization/setting-time-zone

packages/react-components/src/action-form/fields/DateRangeInputField.tsx
```diff
@@ -0,0 +1,506 @@
...
+    <Popover.Root open={isOpen} onOpenChange={handleOpenChange}>
+      <Popover.Trigger
+        nativeButton={false}
+        render={<div className={styles.osdkDateRangeContainer} />}
+      >
+        <div
+          className={classnames(
+            commonStyles.osdkDatePickerInputWrapper,
+            styles.osdkDateRangeInputWrapper,
+            (startInvalid || hasOverlapError)
```
ziyunp, line 428: It's possible to get into invalid state (end date < start date) without error

Image

packages/react-components/src/action-form/fields/useDateEditState.ts
```diff
@@ -0,0 +1,159 @@
...
+   * Formats the date when the input is idle (not focused). Can be locale-friendly
+   * like "Jan 15, 2024" — it doesn't need to be parsable since the user isn't typing.
+   * When no custom formatDate prop is provided, this is the same as editFormatFn.
+   */
+  displayFormatFn: (date: Date) => string;
+  /**
+   * Formats the date when the user starts typing. Must produce a string that
+   * parseFn can round-trip back to a Date (e.g. "2024-01-15" or "2024-01-15 14:30").
+   */
+  editFormatFn: (date: Date) => string;
```
ziyunp, line 33: Q: what's the expected behaviour when typing? Trying to test this

packages/react-components/src/action-form/fields/DatetimePickerField.tsx
```diff
@@ -17,46 +17,159 @@
...
+  const handleKeyDown = useCallback(
+    (e: React.KeyboardEvent<HTMLInputElement>) => {
+      if (e.key === "Enter") {
+        e.preventDefault();
+        if (inputValue === "") {
+          onChange?.(null);
+        } else if (validatedDate != null) {
+          onChange?.(validatedDate);
+        }
+        stopEditing();
```
ziyunp: Issue (not blocking): When I type in a string that isn't formatted to a valid date, it reverts to the previous date, and if I type "backspace" after the revert, the input doesn't respond to it.

https://github.com/user-attachments/assets/4c4db38c-1d49-437d-8010-46722e056b39

ziyunp (dismissed):

packages/react-components/src/action-form/fields/useDateEditState.ts
nfdiop: when typing, it should look like 2024-01-15

packages/react-components/src/action-form/FormFieldApi.ts
nfdiop: I can FLUP and add support for timezone picker like in Blueprint

ziyunp: A description should be enough for now

ziyunp (dismissed):

ziyunp (approved):

PR #2976, opened 2026-04-09
[Components] Merge @osdk/react-components-styles into @osdk/react-components

ziyunp:
- Merge @osdk/react-components-styles into @osdk/react-components
- Keep local copy of blueprint tokens
- Updated packages/react-components/scripts/build-css.mjs to include tokens.css


.monorepolint.config.mjs
```diff
@@ -358,7 +358,7 @@ const archetypeRules = archetypes(
     {
       ...LIBRARY_RULES,
       react: true,
-      cssExport: true,
+      cssExport: ["styles.css"],
```
ziyunp, line 373: Updated this to enable export of multiple css files later

packages/e2e.sandbox.peopleapp/package.json
```diff
@@ -38,12 +38,11 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
-    "@osdk/api": "workspace:*",
-    "@osdk/client": "workspace:*",
```
nfdiop: Help me understand why we made these changes?

packages/react-components/README.md
```diff
@@ -84,48 +71,35 @@ Because `osdk.components` is declared after `osdk.tokens`, component styles take
...
 @import "./user-brand.css" layer(user.brand);
 ```
 
 #### Portal isolation (required)
 
+Add `isolation: isolate` to your app's root element. This is required for Base UI portals. See https://base-ui.com/react/overview/quick-start#portals
+
 ```css
-.root {
+#root {
```
nfdiop, line 98: Is this intentional?

nfdiop (dismissed):
Looks good to me! Thanks for doing this!

packages/e2e.sandbox.peopleapp/package.json
ziyunp: It's probably missed when merging conflicts. Let me update

nfdiop (commented):
Should we keep the components style package and update the [README](https://www.npmjs.com/package/@osdk/react-components-styles) to redirect users?

packages/react-components-styles/README.md
```diff
@@ -1,5 +1,23 @@
 # @osdk/react-components-styles
 
+> **Deprecated:** This package has been merged into `@osdk/react-components` and will be removed in a future release.
```
nfdiop, line 3: nice! I think we should remove the entire readme contend and just say deprecated + link to the styles doc

olliejenningsPT: So one thing is we need to publish this package for the readme to be updated on NPM. In the open-source world most users would expect the latest published version to work. If we remove the entire contents then the latest published version would not work.

nfdiop (approved):

PR #2993, opened 2026-04-10
Add checks for bundles

ssanjay1:
Checks entry points for our bundle clis in an isolated environment (so no dev deps present) to make sure everything required for normal operation is correctly bundled in

### DIFF
--- a/packages/monorepo.tool.check-bundle/bin/check-bundle.mjs
+++ b/packages/monorepo.tool.check-bundle/bin/check-bundle.mjs
@@ -1,0 +1,169 @@
+#!/usr/bin/env node
+// @ts-check
+
+import { execFileSync } from "node:child_process";
+import {
+  cpSync,
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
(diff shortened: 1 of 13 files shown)


packages/monorepo.tool.check-bundle/bin/check-bundle.mjs
```diff
@@ -0,0 +1,171 @@
...
+  readFileSync,
+  realpathSync,
+  rmSync,
+  statSync,
+  symlinkSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import * as path from "node:path";
+
+const pkgJson = JSON.parse(readFileSync("package.json", "utf-8"));
```
bryantpark04, line 19: If this script checks for `bin` entries, can we just add it to every package in the monorepo? Could probably have it run as part of the other tool.monorepo things if we don't want to add it to every package.json

bryantpark04 (dismissed):

bryantpark04 (approved):

PR #2999, opened 2026-04-13
pin @osdk/react version to @osdk/client fixed group

KuberSethi:
@osdk/react was versioned independently at 0.10.0 while @osdk/client and @osdk/api share a version line at 2.8.0, making it hard for consumers to know which versions are compatible

• add @osdk/react to the changesets fixed group so it always shares the same version as @osdk/client and @osdk/api
• tighten react dev deps from workspace:* to workspace:~ to match the client→api pattern
• include a react-components changeset so its peer dep range recomputes correctly after the version jump


ziyunp (dismissed):

ziyunp (approved):

bryantpark04 (dismissed):
Are there any other pins we need to update from exact to ^ now that we are GA?

bryantpark04 (approved):

KuberSethi:
> Are there any other pins we need to update from exact to ^ now that we are GA?

We're going to do the same for react-components but this isn't important right now, mostly important down the road for stability reasons

PR #3000, opened 2026-04-13
fix invalidateObjectType for queries with result type transformations

KuberSethi:
invalidateObjectType wasn't matching queries where the result type differs from apiName

• wire up the existing revalidateObjectType method as the single decision point, resolving the actual fetched result type via getObjectTypesThatInvalidate so link-traversal queries match against the target type
• fix pre-existing bug where InterfaceListQuery.invalidateObjectType never checked if a concrete type implements the queried interface
• update Store change propagation and maybeUpdateAndRevalidate to use the new objectTypes getter

### DIFF
--- a/packages/client/src/observable/internal/list/InterfaceListQuery.ts
+++ b/packages/client/src/observable/internal/list/InterfaceListQuery.ts
@@ -129,23 +129,9 @@ export class InterfaceListQuery extends ListQuery {
   protected createPayload(
     params: CollectionConnectableParams,
   ): ListPayload {
-    const resolvedList = params.resolvedData?.map((obj: ObjectHolder) => {
-      if (process.env.NODE_ENV !== "production") {
-        const missing = !obj || !(ObjectDefRef in obj);
-        if (missing) {
-          // eslint-disable-next-line no-console
-          console.error(
-            `[@osdk/client] InterfaceListQuery.createPayload: object missing ObjectDefRef`,
-            {
-              apiName: this.apiName,
-              objectType: (obj as ObjectHolder | undefined)?.$objectType,
-              primaryKey: (obj as ObjectHolder | undefined)?.$primaryKey,
-            },
-          );
-        }
-      }
-      return obj.$as(this.apiName);
-    });
+    const resolvedList = params.resolvedData?.map((obj: ObjectHolder) =>
+      obj.$as(this.apiName)
+    );
 
     return {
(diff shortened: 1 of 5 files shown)


packages/client/src/observable/internal/list/InterfaceListQuery.ts
```diff
@@ -103,14 +103,20 @@ export class InterfaceListQuery extends ListQuery {
...
+    // implements this query's interface. e.g. invalidating "Employee"
+    // should revalidate a query for "Assignable" if Employee implements it.
+    try {
+      const objectMetadata = await this.store.client.fetchMetadata({
+        type: "object",
+        apiName: objectType,
+      });
+      return this.apiName in objectMetadata.interfaceMap;
+    } catch {
+      return true;
```
ziyunp, line 119: Q: Just checking u want to default to revalidating if fetchMetadata fails?

packages/client/src/observable/internal/list/ListQuery.ts
```diff
@@ -173,6 +184,17 @@ export abstract class ListQuery extends BaseListQuery<
     return this.#pivotInfo;
   }
 
+  get objectTypes(): ReadonlySet<string> {
+    return this.#objectTypesCache ?? new Set([this.apiName]);
+  }
+
+  #updateFetchedObjectType(fetchedApiName: string): void {
+    this.#fetchedObjectType = fetchedApiName;
+    this.#objectTypesCache = fetchedApiName !== this.apiName
```
ziyunp, line 193: Question: Why not do this.#objectTypesCache.add? Can there only be 2 objectTypes max?

packages/client/src/observable/internal/list/ListQuery.ts
```diff
@@ -357,6 +402,31 @@ export abstract class ListQuery extends BaseListQuery<
...
+      && this.#fetchedObjectType !== this.apiName
+    ) {
+      const fetchedType = this.#fetchedObjectType;
+      if (
+        (changes.addedObjects.get(fetchedType)?.length ?? 0) > 0
+        || (changes.modifiedObjects.get(fetchedType)?.length ?? 0) > 0
+      ) {
+        return this.revalidate(true);
+      }
+      for (const key of changes.deleted) {
```
ziyunp, line 420: Q: Why revalidate if it's been deleted?

ziyunp (approved):

PR #3020, opened 2026-04-14
Use workspace:^ for peer deps to prevent major version bumps

nihalbhatnagar:
## Summary
- Changes all `workspace:~` peer dependencies to `workspace:^` across the repo
- Adds an MRL rule requiring `workspace:^` for peer deps (and auto-fixing `workspace:~`)

**Root cause:** Tilde ranges (`~2.8.0`) exclude minor updates, so when a peer dep gets a minor bump (e.g. `2.8.0` → `2.9.0`), changesets sees it as "out of range" and applies a major bump to the dependent. Caret ranges (`^2.8.0`) include minor updates, keeping the dep in range and avoiding the unwanted major propagation.

## Test plan
- [ ] Run `pnpm run check-mrl` — passes with no errors
- [ ] Run release script (`scripts/createReleasePr.sh`) — packages like `@osdk/functions-testing.experimental` get minor bumps instead of major

### DIFF
--- a/packages/cli.cmd.typescript/src/generate/handleGenerate.mts
+++ b/packages/cli.cmd.typescript/src/generate/handleGenerate.mts
@@ -221,6 +221,8 @@ async function generateClientSdk(
           dependencyVersions.osdkApiVersion = "workspace:~";
           dependencyVersions.osdkClientVersion = "workspace:~";
           dependencyVersions.osdkLegacyClientVersion = "workspace:~";
+          dependencyVersions.osdkApiPeerVersion = "workspace:^";
+          dependencyVersions.osdkClientPeerVersion = "workspace:^";
         }
 
         const expectedDeps = getExpectedDependencies(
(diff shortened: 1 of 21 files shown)


packages/tool.release/src/mutateReleasePlan.ts
nihalbhatnagar, line 101: This only looked at the actual changestes, not what was versioned. Worth keeping both IMO

packages/cli.cmd.typescript/src/generate/handleGenerate.mts
```diff
@@ -235,10 +237,19 @@ async function generateClientSdk(
           }
         }
 
-        updateVersionsIfTheyExist(packageJson, {
-          "@osdk/client": dependencyVersions.osdkClientVersion,
-          "@osdk/api": dependencyVersions.osdkApiVersion,
-        });
+        updateVersionsIfTheyExist(
```
nihalbhatnagar, line 240: Needed to doc this because 5 e2e packages were regenerating in CI have osdk-unstable-typescript generate --internal as their codegen, and that CLI hardcoded workspace:~ for  api/client — applying that one value to dep, devDep, AND peerDep sections via updateVersionsIfTheyExist.

bryantpark04 (dismissed):
update the PR title/desc before merging

turbo.json
```diff
@@ -44,6 +44,7 @@
 
     "check-spelling": {
       "outputLogs": "new-only",
+      "dependsOn": ["codegen"],
```
nihalbhatnagar, line 47: cspell's inputs include src/generatedNoCheck/*.ts, and codegen starts with rm -rf src/generatedNoCheck/* before regenerating. turbo was scheduling check-spelling in parallel with codegen, so cspell would open a file that codegen was mid-delete/regenerate on and fail with "Linter: File not found"

bryantpark04 (approved):

PR #3023, opened 2026-04-15
Support reloading widget on vite HMR full reload

tzyl:
Currently when vite HMR triggers a full page reload (for example changing a top level entrypoint file which does not have a HMR boundary) this reloads the widget into a blank screen without content requiring manually refreshing in the parent window.

This PR uses the [vite HMR API](https://vite.dev/guide/api-hmr) if detected to propagate a reload message up so the parent can reload automatically when a full reload is required. I considered in the vite plugin we have the [handleHotUpdate](https://vite.dev/guide/api-plugin#handlehotupdate) hook but no direct information on whether the update is a full reload and also the plugin runs in the vite dev server outside of the browser and so can't easily notify the parent window in the browser. Using the HMR client API instead runs in the browser environment but does require making client code aware of this.

We also could use a transform hook in the vite plugin to inject this code to the client but this has the downside of not reusing the @osdk/widget.client typed wrapper and is separate from the lifecycle in @osdk/widget.client-react


packages/widget.client-react/src/client.tsx
```diff
@@ -254,9 +254,17 @@ export const FoundryWidget = <C extends WidgetConfig<C["parameters"]>>({
     });
     resizeObserver.observe(document.body, { box: "border-box" });
 
+    const handleFullReload = () => client.reload();
+    if (import.meta.hot?.on) {
```
tzyl, line 258: This code gets tree shaken so isn't actually present in production: https://vite.dev/guide/api-hmr#required-conditional-guard

I put the additional guard on `?.on` / `?.off` because this interferes with the vitest environment which has its own HMR API

packages/widget.client-react/src/vite-hmr.d.ts
```diff
@@ -14,12 +14,16 @@
...
+}
 
-const sidebars: SidebarsConfig = {
-  docs: [
-    "CbacPicker",
-  ],
-};
-
-export default sidebars;
+interface ImportMeta {
```
tzyl, line 27: I think we had another use case before on making the widget client library aware of vite and rather than requiring a vite dev dependency just for the types this optionally provides support for vite environments to propagate the message

amishshah (dismissed):

amishshah (approved):

PR #3026, opened 2026-04-15
Split experimental barrel into individual subpath exports

olliejenningsPT:
## Summary

- Splits the single `@osdk/react-components/experimental` barrel export (~60 symbols from 5 component groups) into individual subpath exports so consumers only pay for what they import
- Adds 5 new entry points: `experimental/filter-list`, `experimental/object-table`, `experimental/pdf-viewer`, `experimental/markdown-renderer`, `experimental/action-form`
- The existing `@osdk/react-components/experimental` import continues to work unchanged (re-exports from the sub-modules)
- Adds explicit `package.json` export entries for each new sub-path (required by monorepolint, in addition to the existing `./*` wildcard)
- Updates `.size-limit.cjs` to resolve wildcard `package.json` exports by scanning the build directory, so new subpath exports automatically get individual bundle-size tracking in CI

## Why

Today, importing _any_ experimental component (e.g. just `MarkdownRenderer`) pulls the entire experimental surface into the consumer's bundle — including ObjectTable, PdfViewer, ActionForm, FilterList, and all their types. This is wasteful for apps that only use one or two component groups.

With individual subpath exports, bundlers can tree-shake at the entry-point level:
```ts
// Before: imports everything
import { MarkdownRenderer } from "@osdk/react-components/experimental";

// After: imports only the markdown renderer group
import { MarkdownRenderer } from "@osdk/react-components/experimental/markdown-renderer";
```

## What changed

| File | Change |
|------|--------|
| `src/public/experimental/filter-list.ts` | New — exports BaseFilterList, FilterList, serialization utils, types |
| `src/public/experimental/object-table.ts` | New — exports ObjectTable, BaseTable, ColumnConfigDialog, MultiColumnSortDialog, types |
| `src/public/experimental/pdf-viewer.ts` | New — exports all PDF components, hooks, context, types |
| `src/public/experimental/markdown-renderer.ts` | New — exports MarkdownRenderer, types |
| `src/public/experimental/action-form.ts` | New — exports ActionForm, BaseForm, types |
| `src/public/experimental.ts` | Re-exports from the 5 sub-modules (backwards compatible) |
| `package.json` | Explicit export entries for each new sub-path (required by monorepolint) |
| `.size-limit.cjs` | Resolves wildcard exports via filesystem scan for automatic per-entry bundle tracking |

## Bundle size visibility

`.size-limit.cjs` previously skipped wildcard exports entirely. Now it resolves them by scanning the filesystem, which means every new file added to `src/public/experimental/` (or any other wildcard-matched path) automatically gets its own CI bundle-size check — no manual config needed.

## Test plan

- [x] `pnpm turbo typecheck --filter=@osdk/react-components` — passes
- [x] `pnpm turbo transpile --filter=@osdk/react-components` — all 5 new `.js` files generated
- [x] `pnpm exec mrl check` — passes (explicit exports in package.json)
- [x] `node -e "require('./.size-limit.cjs')"` — resolves 8 react-components entries (3 existing + 5 new, no duplicates)
- [x] `pnpm turbo test --filter=@osdk/react-components` — 617 tests pass (2 pre-existing failures unrelated to this change)


ziyunp (dismissed):

.monorepolint.config.mjs
```diff
@@ -353,13 +353,31 @@ const archetypeRules = archetypes(
...
+    "reactComponentsLibrary",
+    [
       "@osdk/react-components",
     ],
     {
       ...LIBRARY_RULES,
       react: true,
       cssExport: true,
       extraPublishFiles: ["AGENTS.md", "docs"],
+      attwExcludeEntrypoints: [
```
nfdiop, line 375: Could you elaborate on what this is doing?

.size-limit.cjs
```diff
@@ -89,7 +74,63 @@ module.exports = TRACKED_PACKAGES.flatMap((pkg) => {
...
         return config;
       },
-    });
+    };
+  }
+
+  const entries = [];
+
+  // Collect explicit (non-wildcard) export paths to avoid duplicates
+  const explicitExports = new Set(
```
nfdiop, line 83: Curious why we need this?

nfdiop (commented):
Thanks for doing this! I wonder how we can communicate to existing users "hi import from these new paths"

.monorepolint.config.mjs
olliejenningsPT: sure, node10 doesn't support the nested exports, so we this to enable us to ignore them. Am not entirely sure why we check for node10 support or not in builds.

.size-limit.cjs
olliejenningsPT: So if you have multiple wildcard exports in the package.json,its only unique, granted that package.json exports is unique itself, but this just makes it a quick lookup

nfdiop (commented):
`export *` in `experimental.ts` means any new symbol added to a sub-module auto-appears at the top level. Desired long-term, or should it eventually be a curated list?

.monorepolint.config.mjs
nfdiop: If we could get rid of this it would be great. Otherwise we have to update this file whenever we add a new component

nfdiop (dismissed):
Left a comment about attwExcludeEntrypoints. Should we update any docs?

.monorepolint.config.mjs
olliejenningsPT: @nfdiop l would love to get rid of the `attwExcludeEntrypoints` but whilst we have this strict mono-repolint we cannot, unless we loosen it / remove the requirements around exports etc.

I am all for this, but not trying to introduce too many big changes at once

nfdiop (approved):

PR #3032, opened 2026-04-16
chore: remove transpileAllDeps

ziyunp:
Cleaned up transpileAllDeps script, which was added to get every dependent packages transpiled before starting the peopleapp and storybook. I learned the proper way to configure it in turbo.json and with the right config, we don't need the script.


packages/react-components-storybook/turbo.json
```diff
@@ -7,6 +7,7 @@
     "dev": {
       "dependsOn": [
```
nfdiop, line 8: If we consume a transitive @osdk/* package that isn't a direct dependency, will it get transpiled before build?

turbo.json
```diff
@@ -166,6 +166,7 @@
       "dependsOn": [
         "//#babel-config",
         "@osdk/monorepo.tsconfig#typecheck",
+        "^transpileEsm",
```
nfdiop, line 169: This is affecting the whole repo and not just the two packages. Is that expected?

ziyunp:
## Local test results after merging main

### 1. `pnpm turbo check` (full monorepo)
- **121/122 tasks passed**, 1 failed
- Only failure: `//#check-mrl` — `vitest.config.mts` content mismatch in `@osdk/react-components`. **Pre-existing on main.**

### 2. Typecheck on affected packages
| Package | Result |
|---|---|
| `@osdk/react-components` | :white_check_mark: Passed |
| `@osdk/react-components-storybook` | :white_check_mark: Passed |
| `@osdk/e2e.sandbox.peopleapp` | :white_check_mark: Passed |

### 3. Turbo dependency resolution
- Verified `@osdk/react-components-storybook` build/dev correctly resolves the new `@osdk/cbac-components#transpileBrowser` dependency
- Verified `@osdk/e2e.sandbox.peopleapp` build/dev correctly resolves `^transpileBrowser`

### 4. `@osdk/react-components` tests
- **651/651 test cases passed**
- 1 test file failed: `TiffRenderer.test.tsx` — **pre-existing on main** (introduced by #3004)

### 5. `@osdk/oauth` tests (local)
- **39/39 tests passed** across 5 files — no issues locally

### 6. CI failure investigation
- `Build and Test (20)` failed on `@osdk/oauth#test` — `createPublicOauthClient.test.ts` with "Failed to resolve entry for package @osdk/oauth"
- Same job on **main** fails earlier on `check-mrl` and never reaches the oauth test, so can't confirm if it's also broken there
- The test passes locally because transpile outputs are cached; in CI it's a clean build so ordering matters more
- Rerun triggered for the failed job

PR #2984, opened 2026-04-09
Add source export condition for dev-time HMR

nfdiop:
## Summary
- Add `"source"` export condition to `@osdk/react`, `@osdk/react-components`, `@osdk/api`, `@osdk/client`, `@osdk/functions`, `@osdk/functions-testing.experimental`, and `@osdk/cbac-components`
- Configure peopleapp and storybook Vite to use the `"source"` condition with `define` entries for `process.env.PACKAGE_VERSION` and `process.env.MODE` (used by `@osdk/client` via Babel transforms at build time)
- Changes to these packages are now picked up automatically via HMR — no manual rebuild or dev server restart needed

## Notes
I tried `turbo watch` first but I kept getting `failed to connect to daemon` and none of the approaches I tried [here](https://github.com/vercel/turborepo/issues/9394) worked. Upgrading turbo could be a solution here.

## Test plan
- [x] `pnpm turbo typecheck` passes for all modified packages
- [x] `check-mrl` passes with updated exports convention
- [x] Run `pnpm --filter @osdk/e2e.sandbox.peopleapp dev` and verify HMR on source changes
- [x] Run storybook `pnpm dev` and verify HMR on source changes

PR #3036, opened 2026-04-16
Add turbo watch to peopleapp + storybook dev script

nfdiop:
## Summary
- Adds `turbo watch transpileBrowser` to `dev` scripts in peopleapp and storybook so upstream package changes are automatically rebuilt during local development
- Disables the Vite HMR error overlay in both apps — turbo watch rebuilds briefly produce incomplete builds that Vite picks up as errors, causing the overlay to flash on every cycle
- Uses `trap 'kill 0' EXIT` to cleanly kill the turbo watch process when the dev server exits

## Important
Updates aren't instant, unlike what we have [here](https://github.com/palantir/osdk-ts/pull/2984). But at least we're using the build outputs like our users would. Couldn't find a quick way to make this smoother, punting for now.

## Test plan
- [x] Run `pnpm dev` in peopleapp and verify turbo watch starts alongside vite
- [x] Run `pnpm dev` in react-components-storybook and verify turbo watch starts alongside storybook
- [x] Edit an upstream package and verify the change is picked up without manual rebuild
- [x] Ctrl+C the dev server and verify the turbo watch process is also killed

### DIFF
--- a/.changeset/fuzzy-moose-grin.md
+++ b/.changeset/fuzzy-moose-grin.md
@@ -1,6 +1,2 @@
 ---
-"@osdk/e2e.sandbox.peopleapp": patch
-"@osdk/react-components-storybook": patch
 ---
-
-Add turbo watch to dev scripts for automatic upstream rebuilds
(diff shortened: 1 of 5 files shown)


ziyunp (approved):

.changeset/fuzzy-moose-grin.md
ziyunp, line 2: nit: remove

PR #3037, opened 2026-04-16
[Object Table] Add dropdown field to editable table

ziyunp:
- Add dropdown field component to editable ObjectTable

https://github.com/user-attachments/assets/bbd1e1cc-3e70-4f07-a9eb-e612809fd866


packages/react-components/src/object-table/components/DropdownCellField.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {...editFieldConfig.fieldComponentProps}
+        portalRef={portalRef}
+        value={inputValue}
+        onChange={onChange}
+      />
+    </div>
+  );
+}
+
+export const DropdownCellField: React.MemoExoticComponent<
```
nfdiop, line 87: nit: I prefer to drop  DropdownCellFieldInner and use React.memo directly. It is faster to jump through the code this way

packages/react-components/src/object-table/utils/types.ts
```diff
@@ -59,3 +59,80 @@ export interface EditableConfig<
   validationErrors: Map<string, string>;
   clearCellValidationError: (cellId: string) => void;
 }
+
+/**
```
nfdiop, line 63: nice comment! But I don't think we should duplicate since we're already using DropdownField

packages/react-components/src/object-table/DefaultCellRenderer.tsx
```diff
@@ -49,15 +49,20 @@ export function renderDefaultCell<TData extends RowData>(
 
   const cellEdits = meta.cellEdits;
   const editedValue = cellEdits?.[cellId];
-  const currentValue = editedValue?.newValue ?? cellValue;
+  // If newValue is explicitly set to null, treat it as null. Otherwise, fall back to the original cell value.
```
nfdiop, line 52: wait we are checking for undefined here, not null

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -214,31 +229,50 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
     [currentValue],
   );
 
+  // Dropdown: commit on selection
+  const handleDropdownChange = useCallback(
+    (newValue: unknown) => {
+      commitEdit(newValue as CellValue);
```
nfdiop, line 235: could we get rid of the type assertion or at least add a comment?

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -259,3 +293,4 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
 export const EditableCell = React.memo(
   EditableCellInner,
 ) as typeof EditableCellInner;
+"";
```
nfdiop, line 296: nit

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -214,31 +229,50 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
...
-            />
-            {hasValidationError && <Error className={styles.errorIcon} />}
-          </div>
+        <Tooltip.Trigger
+          className={styles.osdkEditableCellTrigger}
+        >
+          {isDropdown
+            ? (
+              <DropdownCellField
+                editFieldConfig={editFieldConfig as EditFieldConfig & {
```
nfdiop, line 255: I wonder if could we get rid of the type assertion?

packages/react-components/src/object-table/utils/PortalTracker.tsx
nfdiop, line 1: nit: maybe move to a context folder

packages/react-components/src/object-table/utils/PortalTracker.tsx
```diff
@@ -0,0 +1,95 @@
...
+      {children}
+    </PortalTrackerContext.Provider>
+  );
+}
+
+export function usePortalTracker(): PortalTrackerContextValue | undefined {
+  return useContext(PortalTrackerContext);
+}
+
+/**
```
nfdiop, line 69: Very interesting!

packages/react-components/src/object-table/components/DropdownCellField.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <div
+      className={classNames(
+        styles.osdkEditableCellDropdown,
+        {
+          [styles.error]: hasValidationError,
+          [styles.osdkEditedInput]: isEdited,
+        },
+      )}
+    >
+      <DropdownField
```
nfdiop, line 77: maybe we should  aria-invalid too here?

nfdiop (approved):

PR #3038, opened 2026-04-16
Show "No value" for empty filter values and fix $isNull where clauses

nfdiop:
## Summary
- Display "No value" with muted italic styling for empty/null values in ListogramInput instead of showing a blank row
- Fix where-clause generation to produce `$isNull: true` instead of `$eq: ""` when empty string values are selected (affects both EXACT_MATCH and SELECT filter states)
- Fix FauxFoundry's `filterObjects` to treat empty strings as null for `$isNull` checks, matching real Foundry backend behavior

## Test plan
- [x] 6 new tests in `filterStateToWhereClause.test.ts` covering EXACT_MATCH and SELECT with empty strings, mixed values, and isExcluding
- [x] All tests pass
- [x] Visual check in Storybook: null rows show "No value" in muted italic
- [x] Verify selecting "No value" in a listogram correctly filters the ObjectTable


packages/faux/src/FauxFoundry/filterObjects.ts
```diff
@@ -149,7 +149,10 @@ export function filterObjects(
       }
       invariant(field);
       return objects.filter((obj) => {
-        return value ? obj[field] == null : obj[field] != null;
+        // Empty strings are also considered null to match the behavior of the backend
```
nfdiop, line 152: To replicate the backend behaviour

packages/react-components-storybook/src/mocks/employeeData.ts
nfdiop, line 1: making changes so I can test the empty strings

packages/react-components-storybook/src/mocks/employeeData.ts
```diff
@@ -89,7 +89,7 @@ export const employeeData: BaseServerObject[] = [
     "locationCity": "New York",
     "workerType": "Employee",
     "employeeNumber": 657495106,
-    "department": "Operations",
+    "department": "   ",
```
nfdiop, line 92: Intentional: I wanted to test when we have a string with spaces

ziyunp (approved):

olliejenningsPT:
@nfdiop Do you want to allow users to pass in a "no values" string to customise the default behaviour

nfdiop:
> @nfdiop Do you want to allow users to pass in a "no values" string to customise the default behaviour

 I hesitated but since that was a feature punted to later. Happy to follow up if someone asks :)

PR #3042, opened 2026-04-17
[Components] Editable ObjectTable with date picker

ziyunp:
- Added date picker field to ObjectTable
- Had to modify TimePicker to not call onChange until user blurs on the input. It holds a local value. The bug was if user is typing 12, it calls onChange on each input of 1 and 2, resulting to user seeing 01:00 and 02:00

To check:
1. default formatter in onChange
2. invalid state

https://github.com/user-attachments/assets/d0724a49-cebc-4571-bb38-c5031edfd712


packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -1056,22 +1087,57 @@ return (
...
-        }}
-      >
-        Try changing edit mode to "always" to enable inline editing without
-        needing to toggle edit mode on.
+  render: (args) => {
+    const [lastEdit, setLastEdit] = useState<
+      CellEditInfo<Osdk.Instance<typeof Employee>> | undefined
+    >();
+
+    const handleCellValueChanged = useCallback(
```
nfdiop, line 1095: nit: can be deleted

packages/react-components/src/object-table/components/DatePickerCellField.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+function DatePickerCellFieldInner({
+  fieldComponentProps,
+  inputValue,
+  hasValidationError,
+  isEdited,
+  onChange,
+  dataType,
+}: DatePickerCellFieldProps): React.ReactElement {
+  const portalRef = useRegisterPortal();
```
nfdiop, line 55: question (non-blocking): `isRowFocused` is accepted as a prop but I don't see it used in the component body. `DropdownCellField` uses it to toggle between a `ReadonlyDisplayCell` and the actual dropdown when the row isn't focused. Should `DatePickerCellField` do something similar, or does the `readOnly` prop on `DatetimePickerField` handle that case differently?

If it's intentionally unused right now and planned for a follow-up, maybe worth a quick comment so future readers don't wonder.

packages/react-components/src/action-form/fields/DatetimePickerField.module.css
```diff
@@ -5,6 +5,22 @@
...
+  background-color: transparent;
+  cursor: default;
+
+  &:hover {
+    background-color: transparent;
+  }
+
+  &:focus-within {
+    outline: none;
+    border-color: transparent;
```
nfdiop: question (non-blocking): `outline: none` in `:focus-within` for the read-only variant -- is there a replacement focus indicator? Since the popover won't open and the input has `readOnly`, users probably can't meaningfully interact with it, but if someone Tabs into it they'd lose the focus ring. Just wanted to double-check this is intentional.

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -241,38 +246,64 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
     ? "number"
     : "text";
 
-  const isDropdown = editFieldConfig?.fieldComponent === "DROPDOWN";
+  const renderFieldInput = () => {
+    switch (editFieldConfig?.fieldComponent) {
+      case "DROPDOWN":
```
nfdiop, line 251: nice: the switch from ternary to a `switch` statement here is much cleaner -- makes it trivial to add new cell types in the future without growing a nested ternary chain.

packages/react-components/src/action-form/fields/TimePicker.tsx
```diff
@@ -29,11 +29,40 @@ export function TimePicker({
...
+
+  const handleFocus = useCallback(() => {
+    isFocusedRef.current = true;
+  }, []);
+
+  const handleBlur = useCallback(() => {
+    isFocusedRef.current = false;
+    onChange(localValueRef.current);
+  }, [onChange]);
+
```
nfdiop, line 54: suggestion (non-blocking): the ref-based approach works, but I wonder if reading `localValue` directly in `handleBlur` (and adding it to the dependency array) would be simpler:

```ts
const handleBlur = useCallback(() => {
  isFocusedRef.current = false;
  onChange(localValue);
}, [onChange, localValue]);
```

The tradeoff is `handleBlur` recreates on each keystroke, but since `Input` likely isn't memoized on its `onBlur` prop, it shouldn't cause extra renders. Up to you -- the ref approach is also fine, just a bit more indirection to follow.

packages/react-components/src/action-form/fields/DatetimePickerField.module.css
```diff
@@ -5,6 +5,22 @@
   outline: none;
 }
 
+/* Read-only mode: hide borders and make it look like plain text */
+.osdkDatetimeReadOnly {
```
nfdiop: should we move to DatePickerCommon.module so we can also use it for the DateRange?

packages/react-components/src/action-form/fields/DatetimePickerField.tsx
```diff
@@ -87,9 +89,10 @@ export const DatetimePickerField: React.NamedExoticComponent<
   // --- Input event handlers ---
 
   const handleFocus = useCallback(() => {
+    if (readOnly) return;
```
nfdiop: Can we let the hook `useDateEditState` handle this?

packages/react-components/src/action-form/fields/TimePicker.tsx
```diff
@@ -29,11 +29,40 @@ export function TimePicker({
...
+    setLocalValue(newValue);
+  }, []);
+
   return (
     <Input
       type="time"
-      value={value}
-      onValueChange={onChange}
+      value={localValue}
+      onValueChange={handleChange}
```
nfdiop, line 63: can we use setLocalValue directly here?

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
ziyunp: I think it's good to see it. To know when the value is actually committed

packages/react-components/src/action-form/fields/TimePicker.tsx
```diff
@@ -29,11 +29,40 @@ export function TimePicker({
   onChange,
   label = "Time",
 }: TimePickerProps): React.ReactElement {
+  const [localValue, setLocalValue] = useState(value);
+  const isFocusedRef = useRef(false);
+  const localValueRef = useRef(localValue);
+  localValueRef.current = localValue;
+
+  // Sync from prop when not focused (parent changed the value externally)
```
nfdiop, line 37: why only when we are not focused?

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -189,6 +189,19 @@ export interface DatetimePickerFieldProps extends BaseFormFieldProps<Date> {
...
+   * Used to track portaled content for click-outside detection.
+   */
+  portalRef?: React.Ref<HTMLDivElement>;
+
+  /**
+   * When true, the field displays its value as plain text without
+   * opening the popover on focus. The input border is hidden so it
+   * visually blends with surrounding read-only content.
+   */
+  readOnly?: boolean;
```
nfdiop, line 204: I think this should belong to BaseFormFieldProps. We can move it there then support the rest as FLUP

packages/react-components/src/object-table/components/DatePickerCellField.tsx
```diff
@@ -0,0 +1,96 @@
...
+    ?? dataType === "timestamp";
+
+  const handleChange = useCallback((newValue: Date | null) => {
+    if (newValue == null) {
+      onChange(null);
+      return;
+    }
+    onChange(
+      showTime
+        ? newValue.toISOString()
```
nfdiop, line 68: nit: maybe use formatDatetimeForInput for consistency

packages/react-components/src/object-table/components/ReadonlyDisplayCell.tsx
```diff
@@ -0,0 +1,58 @@
...
+    >
+      <div
+        className={classNames(
+          styles.osdkEditableInput,
+          styles.readonlyDisplayCell,
+        )}
+      >
+        {inputValue}
+      </div>
+      {hasValidationError && <Error className={styles.errorIcon} />}
```
nfdiop, line 51: question (not-blocking): should we show the error?

packages/react-components/src/object-table/utils/types.ts
```diff
@@ -108,11 +108,59 @@ export interface DropdownEditConfig<V = unknown> {
...
 }
 
+/**
+ * User-facing configuration for a date picker editor in a table cell.
+ *
+ * This is intentionally a standalone interface rather than re-exporting
+ * `DatetimePickerFieldProps` from ActionForm, so the table API doesn't break
+ * when ActionForm's prop shape changes.
+ */
+export interface DatePickerEditConfig {
```
nfdiop, line 118: same comment as in other PR. Do we need this?

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -241,38 +246,64 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
...
-                inputType={inputType}
-                inputValue={inputValue}
-                hasValidationError={hasValidationError}
-                isEdited={isEdited}
-                onValueChange={handleInputChange}
-                onBlur={handleBlur}
-                onKeyDown={handleKeyDown}
-              />
-            )}
+          {renderFieldInput()}
```
nfdiop, line 306: nice!

packages/react-components/src/action-form/fields/TimePicker.tsx
```diff
@@ -29,11 +29,40 @@ export function TimePicker({
...
+    }
+  }
+
+  const handleFocus = useCallback(() => {
+    isFocusedRef.current = true;
+  }, []);
+
+  const handleBlur = useCallback(() => {
+    isFocusedRef.current = false;
+    onChange(localValueRef.current);
```
nfdiop, line 52: Does the UI still feels smooth with changing on blur?

packages/react-components/src/object-table/components/DatePickerCellField.tsx
ziyunp: I removed the usage as overriding the DatetimePickerField styling was enough to make it look like the ReadonlyDisplayCell in Dropdown field. It's better UX without switching btw 2 components like in Dropdown. I'll remove

packages/react-components/src/action-form/fields/DatetimePickerField.module.css
ziyunp: Don't think this class is used anymore, will remove

packages/react-components/src/object-table/components/DatePickerCellField.tsx
ziyunp: I used that initially but it doesn't format it to be the expected OSDK timestamp format

packages/react-components/src/object-table/components/ReadonlyDisplayCell.tsx
ziyunp: no, it's to show in the cell when a field doesn't match the validation. I'll add screenshot of the error state

packages/react-components/src/action-form/fields/TimePicker.tsx
ziyunp: I didn't notice any UX issue, u can see it in the video above or feel free to test it out!

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
nfdiop: I mean you can use `setLastEdit` directly

nfdiop (approved):

packages/react-components/src/action-form/fields/TimePicker.tsx
ziyunp: so that we don't overwrite what the user is actively typing

PR #3052, opened 2026-04-20
[FilterList] Add STATIC_VALUES filter definition type

olliejenningsPT:
## Summary

- Adds a new `STATIC_VALUES` filter definition type that allows consumers to provide a fixed list of string values instead of fetching them via OSDK aggregation queries
- Supports `LISTOGRAM`, `SINGLE_SELECT`, `MULTI_SELECT`, and `TEXT_TAGS` filter components
- Auto-generates WHERE clauses from the `key` property, with an optional `toWhereClause` override for custom mapping (e.g. mapping "Yes"/"No" to `{ active: true/false }`)
- Includes storybook story demonstrating all four component types and custom WHERE clause generation

## Motivation

Currently all filter values must come from OSDK aggregation. This makes it impossible to use filters with fixed/known values like "Yes"/"No", "Active"/"Inactive", or custom categories without a backing property aggregation. `STATIC_VALUES` fills this gap.

## Usage

```typescript
{
  type: "STATIC_VALUES",
  key: "priority",
  label: "Priority",
  filterComponent: "LISTOGRAM",
  values: ["High", "Medium", "Low"],
  filterState: { type: "EXACT_MATCH", values: [] },
}
```

## Screenshots

image

image

## Test plan

- [x] Unit tests for `getFilterKey` with STATIC_VALUES definitions (key and id override)
- [x] Unit tests for `buildWhereClause` covering EXACT_MATCH, SELECT, isExcluding, toWhereClause override, and empty state
- [x] All 46 existing + new tests pass
- [x] Verify storybook `WithStaticValues` story renders correctly
- [x] Verify WHERE clause JSON output updates when selecting static values

### DIFF
--- a/packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
+++ b/packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
@@ -19,6 +19,7 @@ import { FilterList, ObjectTable } from "@osdk/react-components/experimental";
 import type {
   FilterDefinitionUnion,
   FilterListProps,
+  FilterState,
 } from "@osdk/react-components/experimental";
 import type { Meta, StoryObj } from "@storybook/react-vite";
 import { useCallback, useMemo, useState } from "react";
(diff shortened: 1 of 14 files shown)


packages/react-components/src/filter-list/inputs/StaticValuesFilterInput.tsx
```diff
@@ -0,0 +1,259 @@
...
+  coerceToString,
+  coerceToStringArray,
+} from "../utils/coerceFilterValue.js";
+
+interface StaticValuesFilterInputProps<Q extends ObjectTypeDefinition> {
+  definition: StaticValuesFilterDefinition<Q>;
+  filterState: FilterState | undefined;
+  onFilterStateChanged: (state: FilterState) => void;
+  searchQuery?: string;
+  excludeRowOpen?: boolean;
```
ziyunp, line 43: Suggestion: Add a JSDoc for each prop, especially excludeRowOpen. I wouldn't know what it's for without checking the code

ziyunp: Do users get exclude row interaction OOTB?

packages/react-components/src/filter-list/types/StaticValuesTypes.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Filter definition for static (non-OSDK) value lists.
+ * Renders built-in filter components with user-provided values instead of
+ * fetching values via OSDK aggregation queries.
+ */
+export interface StaticValuesFilterDefinition<
+  Q extends ObjectTypeDefinition,
+  C extends StaticValuesComponentType = StaticValuesComponentType,
+> {
+  type: "STATIC_VALUES";
```
ziyunp, line 42: Can we (and should we) extend an existing type definition from non-static filter component?

packages/react-components/src/filter-list/inputs/StaticValuesFilterInput.tsx
olliejenningsPT: Yes, users get the exclude row interaction out of the box. `FilterInputExcludeRow` wraps each input and provides the include/exclude toggle, same as the existing property filter inputs. The `excludeRowOpen` prop is passed through from `BaseFilterList` → `FilterListItem` → `renderInput` → here.

packages/react-components/src/filter-list/types/StaticValuesTypes.ts
olliejenningsPT: Considered this — the shared fields between `PropertyFilterDefinition` and `StaticValuesFilterDefinition` are: `id`, `key`, `label`, `filterComponent`, `filterState`, `colorMap`, `listogramConfig`, `renderValue`, `isVisible`.

However, `PropertyFilterDefinition` has strongly-typed generics that tie `key` to `PropertyKeys<Q>` and `filterComponent` to `ValidComponentsForPropertyType<PropertyTypeFromKey<Q, K>>`, while `StaticValuesFilterDefinition` uses `string` for `key` (since there's no OSDK property backing it) and `StaticValuesComponentType` for `filterComponent` (a subset excluding DATE_RANGE, NUMBER_RANGE, etc. which don't make sense without aggregation data).

Extracting a shared base type would either require complex conditional generics that hurt readability, or a loose base that loses the type safety `PropertyFilterDefinition` currently provides. Given that, I think keeping them as separate interfaces with documented parallel structure is the better tradeoff here.

packages/react-components/src/filter-list/inputs/StaticValuesFilterInput.tsx
```diff
@@ -0,0 +1,259 @@
...
+        >
+          <TextTagsInput
+            suggestions={aggregationValues}
+            isLoading={false}
+            error={null}
+            tags={exactMatch.selectedValues}
+            onChange={exactMatch.handleChange}
+          />
+        </FilterInputExcludeRow>
+      );
```
ziyunp, line 258: Nit: Is this not flagged for not having a default clause? (Pretty sure you've covered all cases by TS just curious)

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1077,6 +1078,164 @@ const handleFilterRemoved = (filterKey) => {
...
+  const filterDefinitions = useMemo(
+    (): FilterDefinitionUnion<Employee>[] => [
+      {
+        type: "STATIC_VALUES",
+        key: "department",
+        label: "Department (static)",
+        filterComponent: "LISTOGRAM",
+        values: ["Marketing", "Operations", "Finance", "Product"],
+        filterState: { type: "EXACT_MATCH", values: [] },
+        listogramConfig: { displayMode: "minimal" },
```
ziyunp, line 1095: As discussed in person, we'll raise an issue to fix the API of this so it's a more generic filterComponentConfig

packages/react-components/src/filter-list/inputs/StaticValuesFilterInput.tsx
olliejenningsPT: Nope, but have added exhaustive checks now to it

packages/react-components/src/filter-list/inputs/StaticValuesFilterInput.tsx
ziyunp: Got it, it's not exposed publicly

ziyunp (approved):

PR #3053, opened 2026-04-20
Improve URL handling

kurzdev:
Prior to this PR, there were a bunch of cases in our code where we constructed a URL...
- ...with POSIX joins
- ...with string concatentations

This is not a pattern that should be used. String concatenations in particular are error prone, particularly in cases where, for example we have something like `${baseUrl}/api/...`. If `baseUrl` contains a trailing `/`, this breaks since we end up with a path like `https://baseUrl.foo//api`, which is invalid and breaks calls. This happened in practice for users of the `language-models` package, if they were to use the `baseUrl` from the OSDK client:
```ts
async function createAnthropicClient(
  client: Client
): Promise<Anthropic> {
  const clientContext: SharedClientContext = osdkClient[symbolClientContext]

  return new Anthropic({
    authToken: await getFoundryToken(clientContext),
    baseURL: getAnthropicBaseUrl(clientContext),
    fetch: createFetch(clientContext),
  });
}
```

We should use the JS-built-in [`URL`](https://datatracker.ietf.org/doc/html/rfc3986) builder to construct URLs. Since this is strictly [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) compliant, we need to ensure that each `baseUrl` carries a trailing `/`. This is already done today in the widely-used [`createMinimalClient`](https://github.com/palantir/osdk-ts/blob/da434c9adc015453696d5a2c37dc7f65bef25b41/packages/client/src/createMinimalClient.ts#L70-L73) (which is also used within [`createClient`](https://github.com/palantir/osdk-ts/blob/da434c9adc015453696d5a2c37dc7f65bef25b41/packages/client/src/createClient.ts#L136), hence causing the issue mentioned above).

This PR introduces this pattern throughout the codebase. It
- Ensures trailing `/` always exist on base URLs
- Uses URL builders where possible
- Migrates manual query parameter addition to use built-in functionality of `URL`

### DIFF
--- a/packages/cli.common/src/util/ensureTrailingSlash.ts
+++ b/packages/cli.common/src/util/ensureTrailingSlash.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,11 +14,6 @@
  * limitations under the License.
  */
 
-export type { CliCommonArgs } from "./CliCommonArgs.js";
-export type { CommonAuthArgs } from "./commands/auth/CommonAuthArgs.js";
-export type { LoginArgs } from "./commands/auth/login/LoginArgs.js";
-export { ExitProcessError } from "./ExitProcessError.js";
-export { getYargsBase } from "./getYargsBase.js";
-export { ensureTrailingSlash } from "./util/ensureTrailingSlash.js";
-export { isValidSemver } from "./util/isValidSemver.js";
-export { YargsCheckError } from "./YargsCheckError.js";
+export function ensureTrailingSlash(input: string): string {
+  return input.endsWith("/") ? input : input + "/";
+}
--- a/packages/client/src/createClient.test.ts
+++ b/packages/client/src/createClient.test.ts
@@ -122,7 +122,7 @@ describe(createClient, () => {
         undefined,
         fetchFunction,
       );
-      expect(spy.mock.calls[0][0]).toBe("https://mock.com/");
+      expect(spy.mock.results[0].value.baseUrl).toBe("https://mock.com/");
 
       createClient(
         "https://mock1.com/",
@@ -131,7 +131,7 @@ describe(createClient, () => {
         undefined,
         fetchFunction,
       );
-      expect(spy.mock.calls[1][0]).toBe("https://mock1.com/");
+      expect(spy.mock.results[1].value.baseUrl).toBe("https://mock1.com/");
 
       createClient(
         "https://mock2.com/stuff/first/foo",
@@ -140,7 +140,9 @@ describe(createClient, () => {
         undefined,
         fetchFunction,
       );
-      expect(spy.mock.calls[2][0]).toBe("https://mock2.com/stuff/first/foo/");
+      expect(spy.mock.results[2].value.baseUrl).toBe(
+        "https://mock2.com/stuff/first/foo/",
+      );
 
       createClient(
         "https://mock3.com/stuff/first/foo/",
@@ -149,7 +151,9 @@ describe(createClient, () => {
         undefined,
         fetchFunction,
       );
-      expect(spy.mock.calls[3][0]).toBe("https://mock3.com/stuff/first/foo/");
+      expect(spy.mock.results[3].value.baseUrl).toBe(
+        "https://mock3.com/stuff/first/foo/",
+      );
 
       const conjureContextSpy = vi.spyOn(
         MakeConjureContext,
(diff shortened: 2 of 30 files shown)


packages/client/src/createPlatformClient.ts
```diff
@@ -38,8 +38,12 @@ export function createPlatformClient(
   options: undefined = undefined,
   fetchFn: typeof globalThis.fetch = fetch,
 ): PlatformClient {
+  const processedBaseUrl = new URL(baseUrl);
```
nihalbhatnagar: Should we do this in createSharedClientContext itself?

packages/widget.vite-plugin/src/dev-plugin/network.ts
```diff
@@ -69,8 +72,11 @@ export function enableDevMode(
   foundryUrl: string,
   viteMode: string | undefined,
 ): Promise<Response> {
+  const base = foundryUrl.endsWith("/") ? foundryUrl : foundryUrl + "/";
```
nihalbhatnagar: Can we modify where foundryUrl is set initially?

packages/widget.vite-plugin/src/dev-plugin/network.ts
```diff
@@ -45,8 +45,11 @@ export function setWidgetSetSettings(
       ] as const),
     ),
   );
+  const base = foundryUrl.endsWith("/") ? foundryUrl : foundryUrl + "/";
+  const url = new URL("api/v2/widgets/devModeSettings/setWidgetSetById", base);
+  url.searchParams.set("preview", "true");
   return fetch(
-    `${foundryUrl}/api/v2/widgets/devModeSettings/setWidgetSetById?preview=true`,
+    url.toString(),
```
gpirindev: nit: You can directly pass the url parameter here. No need to cast it to string.

packages/widget.vite-plugin/src/dev-plugin/network.ts
```diff
@@ -69,8 +72,11 @@ export function enableDevMode(
   foundryUrl: string,
   viteMode: string | undefined,
 ): Promise<Response> {
+  const base = foundryUrl.endsWith("/") ? foundryUrl : foundryUrl + "/";
+  const url = new URL("api/v2/widgets/devModeSettings/enable", base);
+  url.searchParams.set("preview", "true");
   return fetch(
-    `${foundryUrl}/api/v2/widgets/devModeSettings/enable?preview=true`,
+    url.toString(),
```
gpirindev: same here, you can use the URL object directly

gpirindev (commented):
Overall looks good from the custom widgets side. Leaving a few stylistic comments.

Would it be possible to define a shared util, e.g. `ensureTrailingSlash` that appends a slash at the end of a string if it is missing, instead of repeating
`foundryUrl.endsWith("/") ? foundryUrl : foundryUrl + "/"`

packages/cli.common/src/util/ensureTrailingSlash.ts
kurzdev, line 1: I placed this `cli.common` since it felt more natural here (+ we also have to use it for `loginFlow.ts`) than keeping it as a util function in `cli`. But happy to move it to `cli` and add the ternary into `loginFlow.ts` if you're not a fan of exporting this here.

kurzdev:
> Would it be possible to define a shared util, e.g. `ensureTrailingSlash` that appends a slash at the end of a string if it is missing, instead of repeating `foundryUrl.endsWith("/") ? foundryUrl : foundryUrl + "/"`

@gpirindev I didn't explicitly add a helper function within `share.client.impl` and `widget.vite-plugin` since there is only a single use, and we cannot re-use the helper from `cli`. However, if you want me to add separate helpers to these packages, let me know.

packages/client/src/createClient.test.ts
```diff
@@ -131,7 +131,7 @@ describe(createClient, () => {
         undefined,
         fetchFunction,
       );
-      expect(spy.mock.calls[1][0]).toBe("https://mock1.com/");
+      expect(spy.mock.results[1].value.baseUrl).toBe("https://mock1.com/");
```
nihalbhatnagar, line 134: Why did we need this change?

gpirindev:
Looks good from the widgets side 👍

packages/client/src/createClient.test.ts
kurzdev: Since we we spy on `createSharedClientContext` here and moved the normalization of the URL from `createMinimalClient` into `createSharedClientContext`.
So before, `createSharedClientContext` already received the normalized URL as input from `createMinimalClient` (which we checked via `spy.mock.calls[1][0]`). But now, that input URL is not normalized anymore, so we need to check for the result returned by `createSharedClientContext` to be normalized (which we check via `spy.mock.results[1].value.baseUrl`).

nihalbhatnagar (approved):

PR #3058, opened 2026-04-20
Add OBJECT_SELECT field for selecting object instances in action forms

nfdiop:
## Summary
- Add a new `OBJECT_SELECT` form field type to `@osdk/react-components` that lets users search and select individual object instances in action forms
- Introduces `AsyncDropdownField` — a reusable async-loading dropdown built on the existing `Combobox`, with search-as-you-type and loading states
- `ObjectSelectField` wires the async dropdown to OSDK object search via `useOsdkObjects`
- Includes default field definition mapping so `OBJECT_SELECT` is automatically used for object reference action parameters

## Test plan
- [x] Unit tests for `ObjectSelectField` (rendering, search, selection, clearing)
- [x] Unit tests for `AsyncDropdownField` (loading states, option rendering, callbacks)
- [x] Manual verification in e2e sandbox people app (Manager field on form page)

Screenshot 2026-04-30 at 10 19 31


packages/react-components/src/action-form/fields/AsyncDropdownField.tsx
```diff
@@ -0,0 +1,122 @@
...
+      );
+    },
+    [isLoading, fetchError, hasMore, sentinelRef, searchQuery],
+  );
+
+  return (
+    <DropdownField
+      {...dropdownProps}
+      isSearchable={true}
+      filter={dropdownProps.onQueryChange != null ? null : undefined}
```
nfdiop: this is how we disable the default filtering on the ComboBox

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -290,6 +291,45 @@ export interface DropdownFieldProps<V, Multiple extends boolean = false>
...
+
+/**
+ * Props passed to the `itemListRenderer` callback on `DropdownFieldProps`.
+ * Contains pre-built pieces that the renderer can compose with custom UI.
+ */
+export interface ItemListRendererProps {
+  /** Pre-rendered item list (e.g. `<Combobox.List>` with all items). */
+  itemList: React.ReactNode;
+  /** Wraps children in the combobox empty state container (styled, hidden when items exist). */
+  renderEmpty: (children: React.ReactNode) => React.ReactNode;
```
nfdiop: Passing a function because I want to customise how we render empty values

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -433,6 +473,34 @@ export interface ObjectSetFieldProps<T extends ObjectTypeDefinition>
...
+/**
+ * Object select field props for selecting object instances.
+ * Used for action parameters that accept a single object or multiple objects.
+ */
+export interface ObjectSelectFieldProps extends BaseFormFieldProps<unknown> {
+  /**
+   * The API name of the object type to search within.
+   * Resolved from the action parameter metadata (e.g. "Employee").
+   */
+  objectTypeApiName: string;
```
nfdiop: How could we make this work for functions?

packages/react-components/src/base-components/combobox/Combobox.module.css
```diff
@@ -354,8 +354,8 @@
 }
 
 .osdkComboboxEmpty {
-  padding: calc(var(--osdk-combobox-spacing, var(--osdk-surface-spacing)) * 3)
-    calc(var(--osdk-combobox-spacing, var(--osdk-surface-spacing)) * 2);
+  padding: var(--osdk-combobox-spacing, var(--osdk-surface-spacing))
```
nfdiop, line 537: The padding was too big

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
nfdiop, line 1: Copied

packages/react-components/src/shared/hooks/useEventCallback.ts
nfdiop, line 1: Copied

packages/react-components/src/shared/hooks/useDebouncedValue.ts
nfdiop, line 1: copied

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -78,10 +93,16 @@ export function DropdownField<V, Multiple extends boolean = false>({
...
+        query={query}
+        onQueryChange={onQueryChange}
+        disableClientSideFiltering={disableClientSideFiltering}
+        popupStatus={popupStatus}
+        trailingItem={trailingItem}
       />
     );
   }
 
+  // TODO: Support trailingItem
```
nfdiop, line 104: Not supporting now since that would require more testing

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
```diff
@@ -211,6 +224,16 @@ function coerceToFileValue(value: unknown): File | File[] | null {
   return null;
 }
 
+/** Narrows the untyped form value to an OsdkObject by checking for $primaryKey. */
+function narrowToOsdkObject(
+  value: unknown,
+): Osdk.Instance<ObjectOrInterfaceDefinition> | null {
+  if (value != null && typeof value === "object" && "$primaryKey" in value) {
```
nfdiop, line 231: Not entirely sure about this check

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -311,9 +330,14 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
                   <Combobox.SearchInput placeholder="Search…" />
                 </div>
               )}
-              <Combobox.Empty>No results</Combobox.Empty>
+              {popupStatus}
+              {/* Hide "No results" when popupStatus provides its own message (e.g. "Searching…") */}
+              {popupStatus == null && (
```
nfdiop, line 336: I don't like this coupling but cna be refactored later

packages/react-components/src/action-form/utils/getDefaultFieldDefinitions.ts
```diff
@@ -62,12 +60,18 @@ function buildFieldDefinition(
...
+            objectTypeApiName: paramType.object,
+          },
+        };
       case "interface":
-        // TODO: provide correct items
         return {
           ...base,
-          fieldComponent: "DROPDOWN",
-          fieldComponentProps: { items: EMPTY_ITEMS },
+          fieldComponent: "TEXT_INPUT",
```
nfdiop, line 79: Will figure out later

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
```diff
@@ -0,0 +1,102 @@
...
+      targetRef.current = node;
+      if (observer.current == null) {
+        observer.current = new IntersectionObserver(
+          handleObserverUpdate,
+          OPTIONS,
+        );
+      }
+      observer.current.observe(node);
+    } else if (targetRef.current != null) {
+      observer.current?.unobserve(targetRef.current);
```
ziyunp, line 97: Is the else if statement for when setRef is called with null? Should we always unobserve before observing a new target? 

Besides, do we need to set targetRef.current to null after unobserving?

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
```diff
@@ -0,0 +1,102 @@
...
+   * The current count of the number of loaded items in the infinite scroll. This is used to
+   * detect changes when the height of the container is large and can hold multiple initial
+   * loaded pages without making the target element scroll out of the viewport.
+   */
+  loadedCount: number;
+}
+
+/**
+ * This hooks handles the logic of triggering a callback when the user reaches the target element (e.g. spinner).
+ * It returns a callback that must be set as the ref of the target element at the bottom of the
```
ziyunp, line 42: Q: Is it a common pattern to observe intersection with a target element? Asking because what I did was to trigger on scroll to end

packages/react-components/src/shared/hooks/useEventCallback.ts
```diff
@@ -0,0 +1,57 @@
...
+ * The callback is only updated after render, so any usage of the callback during render will operate on the
+ * previous render's callback. Callbacks created by this hook should **only** be used for event handling.
+ * If the callback is called on the first render of a component, it will log an error. In the future,
+ * this will throw.
+ *
+ * ```
+ *
+ * @param callback
+ */
+/* eslint-disable-next-line @typescript-eslint/no-unsafe-function-type */
```
ziyunp, line 31: Why disable this lint rule?

packages/react-components/src/shared/hooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Returns a debounced version of `value` that only updates after `debounceMs`
+ * of inactivity. Each change to `value` resets the timer.
+ */
+export function useDebouncedValue<T>(value: T, debounceMs: number): T {
+  const [debouncedValue, setDebouncedValue] = useState<T>(value);
+
+  useEffect(function debounceValue() {
+    const timer = setTimeout(() => {
```
ziyunp, line 27: Can we use lodash debounce?

packages/react-components/src/base-components/combobox/Combobox.module.css
ziyunp: I prefer to define the default in tokens/* so we can get a list of all css variables and their default values in one place

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -290,6 +292,34 @@ export interface DropdownFieldProps<V, Multiple extends boolean = false>
...
    */
   portalRef?: React.Ref<HTMLDivElement>;
+
+  /**
+   * Controlled search input value. Must be provided together with `onQueryChange`.
+   */
+  query?: string;
+
+  /**
+   * Callback when the search input value changes. Must be provided together with `query`.
```
ziyunp: Issue: Don't have to be provided with query, this can act as an event listener for user to add extra handlers on query change

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -433,6 +463,34 @@ export interface ObjectSetFieldProps<T extends ObjectTypeDefinition>
...
+    | "onQueryChange"
+    | "disableClientSideFiltering"
+    | "renderItemList"
+  >
+{
+  /**
+   * The API name of the object type to search within.
+   * Resolved from the action parameter metadata (e.g. "Employee").
+   */
+  objectTypeApiName: string;
```
ziyunp: issue (blocking): Can we follow all other API to pass the objectType? We can get the api name from it

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -290,6 +292,34 @@ export interface DropdownFieldProps<V, Multiple extends boolean = false>
...
+  /**
+   * Callback when the search input value changes. Must be provided together with `query`.
+   */
+  onQueryChange?: (query: string) => void;
+
+  /**
+   * When true, disables the combobox's built-in client-side filtering.
+   * Use when items are already filtered server-side (e.g. via `onQueryChange`).
+   */
+  disableClientSideFiltering?: boolean;
```
ziyunp, line 314: Suggestion: Can u add some docs on the default behaviour pls?

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
```diff
@@ -0,0 +1,143 @@
...
+    (newValue: OsdkObject | null) => {
+      onChange?.(newValue);
+      setQuery("");
+    },
+    [onChange],
+  );
+
+  // At runtime, useOsdkObjects only reads apiName and type from the definition.
+  // The full ObjectTypeDefinition is for compile-time type safety only.
+  const objectTypeDef = useMemo(
```
ziyunp: Issue (blocking): We don't want to construct objectType like this. Let user pass in the objectType

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
```diff
@@ -0,0 +1,143 @@
...
+  // name at runtime from metadata, not from compile-time type info.
+  const where = useMemo(() => {
+    const trimmed = debouncedQuery.trim();
+    if (trimmed === "" || titleProperty == null) {
+      return undefined;
+    }
+    return {
+      [titleProperty]: { $containsAllTermsInOrder: trimmed },
+    } as Parameters<typeof useOsdkObjects>[1] extends
+      | { where?: infer W }
```
ziyunp: Suggestion: Use WhereClause type from @osdk/api (I use it in ObjectTable too)

packages/react-components/src/action-form/fields/AsyncDropdownField.tsx
```diff
@@ -0,0 +1,107 @@
...
+  return (
+    <DropdownField
+      {...dropdownProps}
+      isSearchable={true}
+      popupStatus={popupStatus}
+      trailingItem={
+        // Adding a trailing item that will trigger a reload when it becomes visible
+        hasMore
+          ? (
+            <div key="sentinel" ref={infiniteScrollRef} role="presentation">
```
ziyunp: Suggestion: Can we move this component into its own component function?

packages/react-components/src/shared/hooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+import { useEffect, useState } from "react";
+
+/**
+ * Returns a debounced version of `value` that only updates after `debounceMs`
+ * of inactivity. Each change to `value` resets the timer.
+ */
+export function useDebouncedValue<T>(value: T, debounceMs: number): T {
```
KuberSethi, line 23: `osdk/react` has a `useDebouncedCallback` which might be helpful here!

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
KuberSethi: hmm, definitely agree this is weird and am wondering if there's a cleaner way to do this. at the very least is it worth putting this into a util?

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
```diff
@@ -0,0 +1,129 @@
...
+  // matches the same text displayed to the user via obj.$title.
+  // The where clause is loosely typed because we resolve the property
+  // name at runtime from metadata, not from compile-time type info.
+  const where: WhereClause<ObjectTypeDefinition> | undefined = useMemo(() => {
+    const trimmed = debouncedQuery.trim();
+    if (trimmed === "" || titleProperty == null) {
+      return undefined;
+    }
+    return {
+      [titleProperty]: { $containsAllTermsInOrder: trimmed },
```
KuberSethi, line 75: just confirming this is the right search filter you want to use - have more enumerated in https://palantir.github.io/osdk-ts/react/querying-data#text-search-filters

packages/react-components/src/shared/hooks/useEventCallback.ts
KuberSethi: +1

packages/react-components/src/shared/hooks/useEventCallback.ts
```diff
@@ -0,0 +1,57 @@
...
+export function useEventCallback<T extends Function>(callback: T): T {
+  const latestCallbackValueWrapper = React.useRef<T>(
+    ((...args: unknown[]) => {
+      if (typeof globalThis.console !== "undefined") {
+        globalThis.console.error(
+          "A callback created with `useEventCallback` was called during the initial render of a component. This will throw an error in the future.",
+        );
+      }
+      return (callback as unknown as (...a: unknown[]) => unknown)(...args);
```
KuberSethi, line 40: hmm this is really suspicious typing, is there any way to clean this up?

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
KuberSethi: Also I'm not sure checking pk is enough here we should be more cautious and also check for apiName

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -311,9 +331,14 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
+                <Combobox.Empty>No results</Combobox.Empty>
+              )}
               <Combobox.List>
-                {isMultiple ? renderItemWithIndicator : renderItem}
+                {items.map(renderItem)}
```
KuberSethi, line 340: Will this break list virtualization as `items.map` will eagerly evaluate?

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export function useInfiniteScroll(
+  { callback, loadedCount }: UseInfiniteScrollOptions,
+): LegacyRef<HTMLDivElement> {
+  const observer = useRef<IntersectionObserver>();
+  const targetRef = useRef<Element | undefined>();
+
+  const eventCallback = useEventCallback(callback);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const throttledCallback = useCallback(throttle(eventCallback), []);
```
KuberSethi, line 53: Do we want to pass a value to `throttle`?

packages/react-components/src/action-form/fields/AsyncDropdownField.tsx
```diff
@@ -0,0 +1,118 @@
...
+      disableClientSideFiltering={dropdownProps.onQueryChange != null}
+    />
+  );
+});
+
+/** Skeleton placeholder that triggers the next page fetch when it scrolls into view. */
+function InfiniteScrollSentinel({
+  infiniteScrollRef,
+}: {
+  infiniteScrollRef: React.LegacyRef<HTMLDivElement>;
```
KuberSethi, line 111: nit: I could be wrong but surely there's a modern / non-legacy equivalent vs `LegacyRef`?

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
```diff
@@ -0,0 +1,129 @@
...
+    } as WhereClause<ObjectTypeDefinition>;
+  }, [debouncedQuery, titleProperty]);
+
+  const {
+    data,
+    isLoading,
+    error: fetchError,
+    fetchMore,
+    hasMore,
+  } = useOsdkObjects(objectType, {
```
KuberSethi, line 85: Could we also perhaps use `enabled` to ensure we actually call this only when we need to vs on mount? Or somehow lazily evaluate?

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
nfdiop: Makes sense! Let me do so. I wasn't entirely sure here. Let me check for that

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
nfdiop: So as discussed in person, copied from another place and this looks to be working well. So I didn't want to make some changes

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
nfdiop: Yes! Very common. I'm adding an element to the end and it is useful for showing a skeleton but also for handling reloading

packages/react-components/src/shared/hooks/useEventCallback.ts
nfdiop: Good question. Copied over and I didn't want to make changes to reduce testing area. But let me look

packages/react-components/src/shared/hooks/useDebouncedValue.ts
nfdiop: Let me track a refactor of these copied hooks

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
nfdiop: Yes so I originally used containsAnyTerm but I found it weird that "Fatou Diop" would also find "Fatou D" and decided to be stricter here. Happy to change!

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
nfdiop: How would you set the `enabled` here? I liked that you already saw some data when opening the field and because it's only 50, I was happy to do the request.

packages/react-components/src/shared/hooks/useInfiniteScroll.ts
nfdiop: So I copied over the hook and since I knew it worked I didn't want to make changes to it. 
Happy to do so

packages/react-components/src/action-form/fields/DropdownField.tsx
nfdiop: Yes it is not ideal but the alternative is more complex code. I tried to use tanstack virtualizer but then it was making the code complex for no real value. Here because we are using paging, I don't expect the data to be that big so just removed it. Happy to add some back

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
KuberSethi: Yup that makes sense, just wanted to confirm since it's not intuitive imo which filter to use when so if this works well that's great!

KuberSethi (approved):
Approving, rest like refactoring hooks can be done in a FLUP

PR #3060, opened 2026-04-20
[FilterList] Call onFilterClauseChanged when initialFilterStates is set

ziyunp:
Previously, onFilterClauseChanged was called manually in setFilterState, clearFilterState, and reset — but was missing from the initialFilterStates code path, causing consumers to be out of sync on mount.

This replaces the scattered manual calls with a single useEffect that watches the derived whereClause. This:

- Fixes the initialFilterStates bug (callback now fires on mount)
- Also fires onFilterClauseChanged on mount for all cases, so consumers always receive the initial where clause without computing it themselves
- Eliminates the class of bug where new mutation paths forget to notify
- Fixes a stale closure bug in clearFilterState (now uses functional state updater)
- Simplifies setFilterState, clearFilterState, and reset

The useEffect approach is the correct React pattern here: onFilterClauseChanged is a side effect derived from state, so it belongs in an effect rather than being manually interleaved with state updates.

### DIFF
--- a/packages/react-components/src/filter-list/hooks/__tests__/useFilterListState.test.tsx
+++ b/packages/react-components/src/filter-list/hooks/__tests__/useFilterListState.test.tsx
@@ -135,7 +135,7 @@ describe("useFilterListState", () => {
     expect(result.current.whereClause).toEqual({ name: "John" });
   });
 
-  it("does not call onFilterClauseChanged on mount", () => {
+  it("calls onFilterClauseChanged on mount with initial where clause", () => {
     const onFilterClauseChanged = vi.fn();
     const nameDef = createPropertyFilterDef(
       "name",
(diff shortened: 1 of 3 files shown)


packages/react-components/src/filter-list/hooks/__tests__/useFilterListState.test.tsx
```diff
@@ -135,7 +135,7 @@ describe("useFilterListState", () => {
     expect(result.current.whereClause).toEqual({ name: "John" });
   });
 
-  it("does not call onFilterClauseChanged on mount", () => {
```
ziyunp, line 138: Intentionally updated this test because I don't think we should skip an onFilterClauseChanged call. Users will need to construct the filter clause themselves if this is not called, just like the bug we are trying to solve here.

An example use case is user will need the filter clause to filter down the object table.

olliejenningsPT (approved):

PR #3065, opened 2026-04-21
[Metrics] Add extra user agents to fetch header 

ziyunp:
Add extra user agents when requests originate from react-components and react hooks.

If fetch originates from @osdk/react, it will have `osdk-react/${process.env.PACKAGE_VERSION}` in the header.

If fetch originates from @osdk/react-components, it will have `osdk-react-components/${process.env.PACKAGE_VERSION}/ComponentName`

Screenshot 2026-04-21 at 12 53 04

### DIFF
--- a/packages/react-components/README.md
+++ b/packages/react-components/README.md
@@ -177,6 +177,7 @@ When building new components:
 3. Keep the Base component API simple using primitive types
 4. For complex components, consider a building blocks tier with sub-components and hooks
 5. Document all layers for users who want to customize
+6. **Register a user agent for metrics** — every OSDK component must call `useRegisterUserAgent` at the top of its render body (see [Metrics](#metrics) below)
 
 ## Folder Structure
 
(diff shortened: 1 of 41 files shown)


packages/react-components/README.md
```diff
@@ -177,6 +177,7 @@ When building new components:
 3. Keep the Base component API simple using primitive types
 4. For complex components, consider a building blocks tier with sub-components and hooks
 5. Document all layers for users who want to customize
+6. **Register a user agent for metrics** — every OSDK component must call `useRegisterUserAgent` at the top of its render body (see [Metrics](#metrics) below)
```
nfdiop: I wonder if higher order components could be clearer here? https://www.patterns.dev/react/hoc-pattern/ 

We would have `withMetrics(Component, "name")` then the component code stays clean?

nfdiop: thanks for updating the docs

packages/react/src/new/OsdkProvider2.tsx
```diff
@@ -45,8 +47,22 @@ export function OsdkProvider2({
...
+    return () => {
+      userAgentsRef.current.delete(agent);
+    };
+  }, []);
+
   const baseObservableClient = useMemo(
-    () => observableClient ?? createObservableClient(client),
+    () =>
+      observableClient
+        ?? createObservableClient(
```
nfdiop, line 62: Couple of questions:
- Is it fine to create an observable agent multiple times? If yes, any clean up to do? 
- I'm not sure we want to use a ref here. Updating the array won't trigger a re-render (if I'm not mistaken)

packages/react/src/new/useRegisterUserAgent.ts
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { useEffect } from "react";
+import { UserAgentContext } from "./UserAgentContext.js";
+
+export function useRegisterUserAgent(agent: string): void {
+  const addUserAgent = React.useContext(UserAgentContext);
+  useEffect(() => {
```
nfdiop: Let's introduce a useOnMount hook to communicate why we do 
// eslint-disable-next-line react-hooks/exhaustive-deps

packages/react-components/README.md
ziyunp: Yeah I did explore this idea. However, there's a significant tradeoff: this adds an extra React component layer to every OSDK component, which complicates devtools and can interfere with typedReactMemo and generics (like ObjectTable's generic parameter).

If you think it's fine with that, I can do that

packages/react/src/new/OsdkProvider2.tsx
ziyunp: `Is it fine to create an observable agent multiple times? If yes, any clean up to do?`
This is what has been in the osdk/react setup. We should check with @KuberSethi when he's back. 

`I'm not sure we want to use a ref here` 
If we don't use ref, the baseObservableClient will be created everytime the userAgents array change no?

nfdiop: Oh nice: I get it now. I thought we were passing userAgentsRef.current directly and didn't realise we had a function

nfdiop (approved):

PR #3066, opened 2026-04-21
Seed Data compiler for Superrepo

humblefoolboi:
- Add new @osdk/seed-helpers package providing a type-safe SeedBuilder API for authoring seed data files for the local ontology server
  - Add --compileSeed CLI flag to @osdk/maker-experimental that compiles seed .mts files into JSON
  - Seed files use the same generated SDK types developers already use in their functions — compile-time validation on property names, value types, link names, and link target types via TypeScript generics
  - Schema-aware string format validation catches malformed timestamps, dates, longs, and decimals before the Rust CLI writes to SQLite


packages/maker-experimental/src/cli/compileSeedData.test.ts
```diff
@@ -0,0 +1,351 @@
...
+    });
+    expect(() => validateStringFormats(output, schema)).not.toThrow();
+  });
+
+  it("should skip unknown object types silently", () => {
+    const output = makeOutput({ Ghost: [{ id: "1" }] });
+    expect(() => validateStringFormats(output, schema)).not.toThrow();
+  });
+
+  it("should skip unknown property names silently", () => {
```
aep000: Do we want this behavior? I feel we should be pretty strict about this.

packages/maker-experimental/src/cli/compileSeedData.test.ts
```diff
@@ -0,0 +1,351 @@
...
+  const schema = buildSchemaMap(mockOntology());
+
+  it("should pass for valid seed data", () => {
+    const output = makeOutput({
+      Employee: [{ employeeId: "emp-001", firstName: "Alice", age: 30 }],
+    });
+    expect(() => validateStringFormats(output, schema)).not.toThrow();
+  });
+
+  it("should skip unknown object types silently", () => {
```
aep000: Definitely should not do that.

packages/maker-experimental/src/cli/compileSeedData.test.ts
```diff
@@ -0,0 +1,351 @@
...
+  });
+
+  it("should skip null property values", () => {
+    const output = makeOutput({
+      Employee: [{ employeeId: "emp-001", firstName: null }],
+    });
+    expect(() => validateStringFormats(output, schema)).not.toThrow();
+  });
+
+  it("should skip non-string values", () => {
```
aep000: Why would this skip?

packages/maker-experimental/src/cli/main.ts
```diff
@@ -87,6 +89,19 @@ export default async function main(
         describe: "Value used to assure uniqueness of entities",
         type: "string",
       },
+      seedDir: {
```
aep000: I think building this into maker-experimental is a mistake. We won't be able to support imported ontologies. We should have a separate TS cli that can take in these files. Then a library that uses OSDK types and OAC types (ideally we migrate exclusively to OSDK) to typecheck the seed data. We probably don't need the full ontology IR.

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -0,0 +1,301 @@
...
+    dstOrLinkName: SeedRef | ObjectTypeDefinition | string,
+    typeOrLinkName?: ObjectTypeDefinition | string,
+    dstPk?: string | number,
+    _dstLinkName?: string,
+  ): void {
+    const resolved = isSeedRef(srcOrType)
+      ? resolveFromRefs(
+        srcOrType,
+        linkNameOrPk as string,
+        dstOrLinkName as SeedRef,
```
aep000, line 195: Some of these casts seem unsafe.

packages/seed-helpers/src/SeedBuilder.test.ts
```diff
@@ -0,0 +1,386 @@
...
+
+const Product = {
+  type: "object" as const,
+  apiName: "Product",
+  primaryKeyApiName: "productId",
+  primaryKeyType: "integer" as const,
+};
+
+// Shorthand for calling link() without compile-time link-name constraints.
+const linkByRef = (
```
aep000: These wrappers make tests confusing remove and use the methods directly.

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -0,0 +1,301 @@
...
+    this.#refs.add(`${apiName}:${pk}`);
+
+    return Object.freeze({
+      __objectTypeApiName: apiName,
+      __primaryKey: pkValue as string | number,
+    }) as SeedRef<Q>;
+  }
+
+  /**
+   * Register a link between two seed objects.
```
aep000, line 147: Can we move some of this documentation to a readme that explains how to use this?

aep000 (commented):
I think there are some architectural changes we should make. Principals are largely ok.

packages/seed-helpers/src/SeedBuilder.ts
humblefoolboi: We deliberately keep `link()` as a single method with two calling shapes — by-reference or by-type+PK — because that's what the public API and README document. They produce the same output and are explained as one method.
The casts in the body come from a TypeScript limitation: when you have an overloaded method, the implementation signature has to cover both overloads, so inside the body the parameters look like a wide union (e.g. `SeedRef | ObjectTypeDefinition`). The `isSeedRef(...)` check tells us which overload the caller used, but TypeScript doesn't carry that information into the rest of the function body. So we have to use `as` casts to tell the compiler "we already know the shape from the runtime check."
In other words, the casts aren't unsafe — they're guarded by `isSeedRef` plus the public overload signatures. The only way to remove them entirely would be to split `link` into two separate methods (e.g. `linkByRef` and `linkByPk`), which we don't want because it splits the API.

aep000 (dismissed):
LGTM

ethanai22 (approved):

PR #3067, opened 2026-04-21
Attach SharedClientContext to mock client so platform fetches don't crash

nihalbhatnagar:
## Summary
- Platform API helpers (e.g. `Users.getCurrent(client)`) read `ctx.baseUrl` off the client before making a request. `createMockClient()` never populated that context, so platform calls crashed inside `foundryPlatformFetch` on `ctx.baseUrl.endsWith(...)` before the fetch mock could see them.
- Attach a minimal `SharedClientContext` (baseUrl=`https://mock.invalid/`, stub token provider, `globalThis.fetch`) via the `__osdkClientContext` symbol so callers can intercept with MSW / `vi.spyOn(globalThis, "fetch")` / etc. as they would for any other fetch.
- No options exposed — mocks never actually hit the network, so there's nothing for the caller to configure.

## Test plan
- [x] `pnpm check --filter=@osdk/functions-testing.experimental` passes (lint, typecheck, test, check-attw, check-api, check-spelling).
- [x] New unit test verifies the context shape (baseUrl, fetch, tokenProvider) is attached and `.endsWith` doesn't throw.
- [x] New example test `platformApiWithMsw` drives `Users.getCurrent(mockClient)` end-to-end through MSW.

### DIFF
--- a/packages/functions-testing.experimental/src/example-function-tests/platformApiWithMsw/platformApiWithMsw.test.ts
+++ b/packages/functions-testing.experimental/src/example-function-tests/platformApiWithMsw/platformApiWithMsw.test.ts
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


packages/functions-testing.experimental/src/example-function-tests/platformApiWithMsw/platformApiWithMsw.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+beforeAll(() => server.listen({ onUnhandledRequest: "error" }));
+afterEach(() => server.resetHandlers());
+afterAll(() => server.close());
+
+describe("platformApiWithMsw", () => {
+  it("resolves the current user when MSW handles the platform call", async () => {
+    server.use(
+      http.get(
+        "https://mock.invalid/api/v2/admin/users/getCurrent",
```
nihalbhatnagar, line 36: We will have to document that this is the URL our mock client will hit. We can change this name if we want

bryantpark04 (approved):

packages/functions-testing.experimental/src/example-function-tests/platformApiWithMsw/platformApiWithMsw.test.ts
ssanjay1: Hmm, is there a reason we can't make this configurable?

PR #3070, opened 2026-04-21
Support asymmetric matchers in mock stub argument matching

nihalbhatnagar:
`deepEqual` in `@osdk/functions-testing.experimental` now recognizes Vitest/Jest asymmetric matchers so they can be used as stub parameters.

- `expect.any(...)`, `expect.anything()`, `expect.stringContaining(...)`, `expect.objectContaining(...)` etc. now match against stub args
- Previously matchers were compared by raw object shape and silently never matched
- Works in either position (matcher on the stub side or on the actual-arg side)

### DIFF
--- a/packages/functions-testing.experimental/src/mock/__tests__/createMockClient.test.ts
+++ b/packages/functions-testing.experimental/src/mock/__tests__/createMockClient.test.ts
@@ -254,6 +254,56 @@ describe("createMockClient", () => {
     });
   });
 
+  describe("asymmetric matchers", () => {
+    it("matches whenQuery params with expect.any(...)", async () => {
+      const mockClient = createMockClient();
+
+      mockClient
+        .whenQuery(addOne, { n: expect.any(Number) as unknown as number })
+        .thenReturn(42);
+
+      expect(await mockClient(addOne).executeFunction({ n: 1 })).toBe(42);
+      expect(await mockClient(addOne).executeFunction({ n: 99 })).toBe(42);
+    });
+
+    it("matches where clauses with expect.objectContaining(...)", async () => {
+      const mockClient = createMockClient();
+      const emp = createMockOsdkObject(Employee, { employeeId: 7 });
+
+      mockClient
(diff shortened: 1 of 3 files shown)


packages/functions-testing.experimental/src/mock/__tests__/createMockClient.test.ts
```diff
@@ -238,19 +238,53 @@ describe("createMockClient", () => {
...
-      expect(typeof ctx.baseUrl).toBe("string");
-      expect(ctx.baseUrl.length).toBeGreaterThan(0);
-      // foundryPlatformFetch reads `ctx.baseUrl.endsWith("/")` — verify it doesn't throw.
-      expect(() => ctx.baseUrl.endsWith("/")).not.toThrow();
-      expect(typeof ctx.fetch).toBe("function");
-      expect(typeof ctx.tokenProvider).toBe("function");
-      await expect(ctx.tokenProvider()).resolves.toBeTypeOf("string");
+
+      mockClient
+        .whenQuery(addOne, { n: expect.any(Number) as unknown as number })
```
nihalbhatnagar, line 262: I think I'll force users to have this cast and won't support this first class. My main motivation behind adding this is seeing its use in actual repositories and if this actually delivers value. I think hiding it behind requiring casts means we can continue to keep this hidden and theoretically unsupported

packages/functions-testing.experimental/src/mock/createMockObjectSetWithResolver.ts
```diff
@@ -154,3 +162,12 @@ export function deepEqual(a: unknown, b: unknown): boolean {
   }
   return false;
 }
+
+function isAsymmetricMatcher(
+  x: unknown,
+): x is { asymmetricMatch: (other: unknown) => boolean } {
```
nihalbhatnagar, line 168: Dont know if its worth to make this strongly typed and dont know how

ssanjay1 (approved):

PR #3073, opened 2026-04-21
Add thenThrow to whenQuery stub builder

nihalbhatnagar:
Adds `thenThrow(error)` to `whenQuery(...)` in `@osdk/functions-testing.experimental` so tests can explicitly configure the error a query's `executeFunction` rejects with.

- Previously the only way to hit the rejected-promise branch was to not register a stub at all (which produces the generic "No stub for query '<name>'" error)
- `thenThrow` rejects with the exact Error instance the test passed in, making error-branch coverage easy
- Only matching params trigger the thrown error; unrelated params still surface the default "No stub" message

### DIFF
--- a/.changeset/mock-query-then-throw.md
+++ b/.changeset/mock-query-then-throw.md
@@ -1,0 +1,5 @@
+---
+"@osdk/functions-testing.experimental": patch
+---
+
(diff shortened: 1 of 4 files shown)


nihalbhatnagar:
I'm not adding this to object set methods because maybe you should be using `fetchOneWithErrors` instead of fetchOne and try catching? maybe not

bryantpark04 (approved):

PR #3076, opened 2026-04-22
[Components] Fix stableObjectSet

ziyunp:
Created a useStableObjectSet hook to handle stabilising object set properly. 

Problem
Function-backed columns show data for only 1 row after: applying a filter to show only 1 row → switching tabs → returning → removing the filter. The remaining 49 rows have empty function columns.

Root Cause Analysis
We traced the data flow from ObjectTable → useObjectTableData → useObjectSet → useFunctionColumnsData → useOsdkFunctions → createCompositeExternalStore and identified three compounding bugs:

Bug 1: useObjectSet returns stale objectSet from previous payload
In useObjectSet.js, the returned objectSet comes from lastLoaded?.objectSet (the last completed payload). When query params change (filter removed), lastLoaded falls back to previousCompletedPayloadRef.current while the new subscription is loading. This payload has the old where clause baked into its objectSet structure. The returned objectSet should always reflect the current query intent, not the cached payload's objectSet.

Location: @osdk/react — useObjectSet.js: objectSet: lastLoaded?.objectSet

Bug 2: useFunctionColumnsData memoizes objectSet with JSON.stringify instead of getWireObjectSet
stableObjectSet is memoized via JSON.stringify(objectSet), but direct JSON serialization of an OSDK ObjectSet object does not include where clauses — only getWireObjectSet() produces the full wire representation. This means even if a corrected objectSet arrives (with different where), the memo key doesn't change, so stableObjectSet remains stale.

Bug 3 (consequence): Function queries sent to backend with stale filter
Because of bugs 1 and 2, buildPagedObjectSets receives the stale objectSet and constructs function query params that include the old filter. The backend function only returns results for the 1 row matching both the stale filter AND the $in clause, leaving all other rows empty.

**Fix: Added useStableObjectSet hook that uses getWireObjectSet for proper memoization, so objectSet changes involving where clauses are correctly detected.**

Upstream Fixes Needed
@osdk/react (useObjectSet): Look for a correct value to return in objectSet when the hook is loading. It's now only clearing the previousCompletedPayloadRef when objectType changed. Should we also check for changes in the options?

### DIFF
--- a/packages/react/src/new/core/useStableObjectSet.ts
+++ b/packages/react/src/new/core/useStableObjectSet.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type {
+  ObjectOrInterfaceDefinition,
+  ObjectSet,
+  SimplePropertyDef,
+} from "@osdk/api";
+import { getWireObjectSet } from "@osdk/client/unstable-do-not-use";
+import { useMemo } from "react";
+
+/**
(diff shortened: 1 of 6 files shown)


packages/react/src/new/core/useStableObjectSet.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import type {
+  ObjectOrInterfaceDefinition,
+  ObjectSet,
+  SimplePropertyDef,
+} from "@osdk/api";
+import { getWireObjectSet } from "@osdk/client/unstable-do-not-use";
+import { useMemo } from "react";
+
+/**
```
nfdiop, line 25: nice comment

nfdiop (approved):

PR #3081, opened 2026-04-23
Improve DropdownField: searchable input, multi-select, clear button

nfdiop:
## Summary
- Add searchable text input inside the dropdown popup for filtering options
- Add multi-select mode with checkboxes
- Add clear button to reset selection
- Fix Select component positioning

## Test plan
- [x] Verify single-select dropdown still works as before
- [x] Verify multi-select mode with checkboxes
- [x] Verify search/filter within dropdown
- [x] Verify clear button resets selection
- [x] Run existing DropdownField tests

Screenshot 2026-04-23 at 13 54 16
Screenshot 2026-04-23 at 14 01 40
Screenshot 2026-04-23 at 14 01 58
Screenshot 2026-04-23 at 14 02 13

### DIFF
--- a/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
@@ -791,33 +791,33 @@ function ObjectSetFieldStory(): React.ReactElement {
   );
 }
 
-const searchableDropdownFieldDefinitions: ReadonlyArray<
+const singleSelectDropdownFieldDefinitions: ReadonlyArray<
   RendererFieldDefinition
 > = [
   {
     fieldKey: "department",
     fieldComponent: "DROPDOWN",
-    label: "Department",
+    label: "Department (Select)",
     fieldComponentProps: {
       items: DEPARTMENT_ITEMS,
-      isSearchable: true,
-      placeholder: "Search departments...",
+      placeholder: "Select department...",
     },
   },
   {
-    fieldKey: "name",
-    fieldComponent: "TEXT_INPUT",
-    label: "Name",
-    isRequired: true,
+    fieldKey: "team",
+    fieldComponent: "DROPDOWN",
+    label: "Team (Searchable)",
     fieldComponentProps: {
-      placeholder: "Enter your name",
+      items: DEPARTMENT_ITEMS,
+      isSearchable: true,
+      placeholder: "Search teams...",
     },
   },
 ];
 
-export const WithSearchableDropdown: Story = {
+export const WithDropdown: Story = {
   args: {
-    fieldDefinitions: searchableDropdownFieldDefinitions,
+    fieldDefinitions: singleSelectDropdownFieldDefinitions,
     onSubmit: handleSubmit,
   },
   parameters: {
@@ -827,17 +827,25 @@ export const WithSearchableDropdown: Story = {
   {
     fieldKey: "department",
     fieldComponent: "DROPDOWN",
-    label: "Department",
+    label: "Department (Select)",
+    fieldComponentProps: {
+      items: ["Engineering", "Marketing", "Sales", "Finance", "Operations", "Legal"],
+      placeholder: "Select department...",
+    },
+  },
+  {
+    fieldKey: "team",
+    fieldComponent: "DROPDOWN",
+    label: "Team (Searchable)",
     fieldComponentProps: {
       items: ["Engineering", "Marketing", "Sales", "Finance", "Operations", "Legal"],
       isSearchable: true,
-      placeholder: "Search departments...",
+      placeholder: "Search teams...",
     },
   },
 ];
 
-// isSearchable renders a Combobox with type-ahead filtering
-// instead of a plain Select dropdown.
+// Side-by-side comparison: plain Select vs searchable Combobox.
 <BaseForm
   fieldDefinitions={fieldDefinitions}
   onSubmit={(formState) => console.log("Submitted:", formState)}
@@ -851,23 +859,24 @@ const multiSelectDropdownFieldDefinitions: ReadonlyArray<
   RendererFieldDefinition
 > = [
   {
-    fieldKey: "title",
-    fieldComponent: "TEXT_INPUT",
-    label: "Title",
-    isRequired: true,
+    fieldKey: "categories",
+    fieldComponent: "DROPDOWN",
+    label: "Categories (Select)",
     fieldComponentProps: {
-      placeholder: "Enter a title",
+      items: TAG_ITEMS,
+      isMultiple: true,
+      placeholder: "Select categories...",
     },
   },
   {
     fieldKey: "tags",
     fieldComponent: "DROPDOWN",
-    label: "Tags",
+    label: "Tags (Searchable)",
     fieldComponentProps: {
       items: TAG_ITEMS,
       isMultiple: true,
       isSearchable: true,
-      placeholder: "Select tags...",
+      placeholder: "Search tags...",
     },
   },
 ];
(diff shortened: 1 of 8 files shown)


packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
```diff
@@ -791,33 +791,33 @@ function ObjectSetFieldStory(): React.ReactElement {
   );
 }
 
-const searchableDropdownFieldDefinitions: ReadonlyArray<
+const singleSelectDropdownFieldDefinitions: ReadonlyArray<
```
nfdiop, line 794: Made some changes so this page shows single select vs multiple ones

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -162,28 +216,95 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
       <Combobox.Root
         value={value}
         onValueChange={onChange}
+        open={open}
+        onOpenChange={setOpen}
         multiple={isMultiple}
         itemToStringLabel={itemToStringLabel}
         isItemEqualToValue={isItemEqual}
         items={items}
+        autoHighlight={isSearchable}
```
nfdiop, line 249: To highlight the first match

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -89,17 +101,50 @@ const SelectDropdown = typedReactMemo(function SelectDropdownFn<
...
   placeholder,
   portalRef,
-}: InnerDropdownProps<V, Multiple>): React.ReactElement {
+}: InnerSelectProps<V, Multiple>): React.ReactElement {
+  const [open, setOpen] = useState(false);
+
+  const hasValue = value != null;
+
+  const handleClear = useCallback(() => {
+    onChange?.(null as Parameters<NonNullable<typeof onChange>>[0]);
```
KuberSethi: this is quite suspicious looking, is there no way to make the typing or value used here clearer?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -128,29 +173,38 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
+    (itemToRemove: V) => {
+      if (!isMultiple || !Array.isArray(value)) {
+        return;
+      }
+      const next = value.filter((v) =>
+        isItemEqual != null
+          ? !isItemEqual(v, itemToRemove)
+          : v !== itemToRemove
+      );
+      onChange?.(next as Parameters<NonNullable<typeof onChange>>[0]);
```
KuberSethi: likewise here is there anyway to improve this typing?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -162,28 +216,95 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
+        <Combobox.Trigger
+          className={isMultiple
+            ? comboboxStyles.osdkComboboxTriggerMulti
+            : undefined}
+        >
+          <div className={comboboxStyles.osdkComboboxValueContainer}>
+            {isMultiple && hasValue
+              ? (
+                <div className={comboboxStyles.osdkComboboxTriggerChips}>
+                  {(value as V[]).map((item) => (
```
KuberSethi: is there no way to get rid of this `value as V[]` cast?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -192,6 +313,13 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
   );
 });
 
+// Prevent the clear/remove click from bubbling into the trigger
```
KuberSethi, line 326: nice! optionally is it perhaps worth pulling this out into a more common util? I expect we'll end up needing similar functions across our components?

packages/react-components/src/base-components/select/Select.tsx
```diff
@@ -105,6 +105,11 @@ function SelectPositioner({
     <BaseUISelect.Positioner
       className={classnames(styles.osdkSelectPositioner, className)}
       sideOffset={4}
+      /* Disable Base UI's default behavior of aligning the selected item in
```
KuberSethi, line 108: nice comment!

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -89,17 +101,50 @@ const SelectDropdown = typedReactMemo(function SelectDropdownFn<
...
       <Select.Root
         value={value}
         onValueChange={onChange}
+        open={open}
+        onOpenChange={setOpen}
         multiple={isMultiple}
         isItemEqualToValue={isItemEqual}
         itemToStringLabel={itemToStringLabel}
       >
-        <Select.Trigger id={id} placeholder={placeholder} />
```
KuberSethi, line 102: should we keep the `id={id}`?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -162,28 +216,95 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
         items={items}
+        autoHighlight={isSearchable}
       >
-        {isMultiple
-          ? (
-            <Combobox.Chips>
-              <Combobox.Value>{renderChips}</Combobox.Value>
-            </Combobox.Chips>
-          )
-          : <Combobox.Input id={id} placeholder={placeholder} />}
```
KuberSethi, line 181: do we also need to readd the `id={id}` for updated combobox input?

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -159,47 +218,110 @@ describe("DropdownField", () => {
...
+          isMultiple={true}
+        />,
+      );
+
+      const trigger = screen.getByRole("combobox");
+      fireEvent.click(trigger);
+
+      await vi.waitFor(() => {
+        const options = screen.getAllByRole("option");
+        expect(options.length).toBe(STRING_ITEMS.length);
```
KuberSethi: hm is this actually properly testing the presence of tick indicators? unless i'm misunderstanding i think the length would be the same regardless

packages/react-components/src/base-components/combobox/Combobox.module.css
```diff
@@ -16,6 +16,173 @@
...
+.osdkComboboxPlaceholder {
+  text-align: start;
+  color: var(
+    --osdk-combobox-placeholder-color,
+    var(--osdk-typography-color-muted)
+  );
+  pointer-events: none;
+}
+
+.osdkComboboxValue:not([data-placeholder]) ~ .osdkComboboxPlaceholder {
```
KuberSethi: is this ever used / I don't think it'll match unless I'm misreading bc `Comboxbox.Value` isn't wrapped with this

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -89,17 +101,50 @@ const SelectDropdown = typedReactMemo(function SelectDropdownFn<
...
+          <div className={selectStyles.osdkSelectValueContainer}>
+            <Select.Value />
+            {placeholder != null && (
+              <span className={selectStyles.osdkSelectPlaceholder}>
+                {placeholder}
+              </span>
+            )}
+          </div>
+          {hasValue && (
+            <Button
```
KuberSethi: I think this might be problematic, React doesn't like nested buttons and `Select.Trigger` is also a button, users will get warnings and I think this makes the component not a11y complaint / bad for screenreader users.

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -52,14 +62,16 @@ export function DropdownField<V, Multiple extends boolean = false>({
     [itemToKey, resolvedItemToStringLabel],
   );
 
-  if (isSearchable) {
+  // Multi-select always uses Combobox for the chip-based UI because it looks better
+  if (isSearchable || rest.isMultiple) {
```
KuberSethi: after adding this should also drop destructuring `isMultiple` below and or updating `InnerSelectProps` since it's not possible below? If I'm understanding correctly we basically have two divergent paths based off of `isMultiple` but in the latter path we're still typing/treating it as if `isMultiple` can be conditional.

KuberSethi (commented):
looks good, just some small things re: typing and cleanup. Only real blocking comment is around the nested button piece.

packages/react-components/src/action-form/fields/DropdownField.tsx
nfdiop: I was also hesitant but it would require re-checking Array.isArray: let me see what I can do

packages/react-components/src/action-form/fields/DropdownField.tsx
nfdiop: yes good flag

packages/react-components/src/action-form/fields/DropdownField.tsx
nfdiop: Good flag. Let me check!

packages/react-components/src/action-form/fields/DropdownField.tsx
nfdiop: Yes! Let me clarify

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
nfdiop: Good catch. Let me fix

KuberSethi (approved):

packages/react-components/src/action-form/fields/DropdownField.tsx
nfdiop: happy to do later :)

PR #3083, opened 2026-04-23
fix applyaction to auto invalidate lists and links after every edit

KuberSethi:
useList and useLinks used to stay stale after an action completed for link or non OT updates. also manual invalidation only touched the specific object keys reported by the server, never the list or link queries that should now gain or lose a member.

• tombstone deleted objects before invalidating so subscribers never see a spurious loading state for an object we already know is gone
• invalidate the union of object types from editedObjectTypes, added/modified/deleted objects, and both sides of added/deleted links so consumers no longer need manual invalidateObjectType calls
• narrow ActionResults.editedObjectTypes from Array<String> to Array<string>


.changeset/fix-action-auto-invalidation.md
```diff
@@ -0,0 +1,10 @@
+---
+"@osdk/client": patch
+"@osdk/api": patch
+---
+
+fix applyAction: process every ActionResults field so useList and useLinks refresh without a manual invalidateObjectType call after actions
+
+- tombstone deleted objects before invalidating so subscribers never observe a spurious "loading" state
```
nfdiop: what is tombstone?

packages/client/src/observable/internal/Store.invalidation.test.ts
```diff
@@ -46,6 +54,18 @@ const OFFICE_2_ID = "102";
 const TODO_1_ID = 201;
 const TODO_2_ID = 202;
 
+// Test-only actions not present in @osdk/client.test.ontology.
```
nfdiop: why not add them there?

packages/client/src/observable/internal/Store.invalidation.test.ts
```diff
@@ -647,4 +727,185 @@ describe("Store Invalidation Type Isolation", () => {
...
+        srcObject: office,
+        srcLinkName: "occupants",
+        targetType: Employee,
+        expected: [expect.objectContaining({ $primaryKey: EMPLOYEE_1_ID })],
+      });
+      linkSubFn.next.mockClear();
+      return linkSubFn;
+    }
+
+    it("modify invalidates list and link queries of the modified type", async () => {
```
nfdiop: nit: invalid?

packages/client/src/observable/internal/actions/ActionApplication.ts
```diff
@@ -104,46 +103,58 @@ export class ActionApplication {
...
+        ),
+      );
+      return;
+    }
+
     const {
-      deletedObjects,
-      modifiedObjects,
       addedObjects,
-      editedObjectTypes,
```
nfdiop, line 111: these are no longer considered. Is that intentional?

packages/client/src/observable/internal/actions/ActionApplication.ts
```diff
@@ -104,46 +103,58 @@ export class ActionApplication {
...
-        await this.store.invalidateObjectType(apiName as string, changes);
       }
+    });
+
+    const perObjectInvalidations: Array<Promise<unknown>> = [];
+    const touchedObjectTypes = new Set<string>();
+    for (const obj of [...(addedObjects ?? []), ...(modifiedObjects ?? [])]) {
+      touchedObjectTypes.add(obj.objectType);
+      perObjectInvalidations.push(
+        this.store.invalidateObject(obj.objectType, obj.primaryKey),
```
nfdiop, line 142: could we add a comment why we don't need this for deletedObjects anymore?

packages/client/src/observable/internal/Store.ts
```diff
@@ -585,6 +585,32 @@ export class Store {
     return Promise.allSettled(promises).then(() => void 0);
   }
 
+  /**
+   * Invalidate only specificLink queries for an object type. Object and list
+   * queries self-invalidate via the object-level propagation pipeline; link
```
nfdiop: Thanks for adding the comment. Can we expand more on it?

packages/client/src/observable/internal/actions/ActionApplication.ts
```diff
@@ -104,46 +103,58 @@ export class ActionApplication {
...
+      );
+    }
+    for (const obj of deletedObjects ?? []) {
```
nfdiop: is this the second time we loop through deletedObjects?

packages/client/src/observable/internal/Store.ts
```diff
@@ -567,7 +567,7 @@ export class Store {
 
     const promises: Array<Promise<void>> = [];
 
-    for (const cacheKey of this.layers.truth.keys()) {
+    for (const cacheKey of this.queries.keys()) {
```
nfdiop, line 575: can you expand on why this is fine?

packages/client/src/observable/internal/Store.invalidation.test.ts
nfdiop, line 1: nice tests. can we also test for deletedLinks and largeScaleEdits

packages/client/src/observable/internal/actions/ActionApplication.ts
KuberSethi: refactored to how it used to be / how it is now, in original implementation of this PR I was handling their cases elsewhere but yeah it wasn't the most clear

packages/client/src/observable/internal/actions/ActionApplication.ts
KuberSethi: good call, refactoring!

packages/client/src/observable/internal/Store.ts
KuberSethi: yup i'll add a comment!

PR #3084, opened 2026-04-23
fix react.beta template regressions from #2770

KuberSethi:
restore @osdk/react dep and OsdkProvider wrapping so the commented useOsdkClient example in Home.tsx works when uncommented

• add @osdk/react back to package.json.osdk.hbs, package.json.psdk.hbs, and both compiled examples (minimal + lohi)
• rename templates/src/main.tsx back to main.tsx.hbs with {{#if osdkPackage}} conditional OsdkProvider wrapping, preserving the ErrorBoundary/Suspense/Loading/path-alias improvements from #2770
• wrap RouterProvider in OsdkProvider in examples/example-react-sdk-2.x/src/main.tsx

### DIFF
--- a/.changeset/fix-react-beta-template-provider.md
+++ b/.changeset/fix-react-beta-template-provider.md
@@ -1,0 +1,5 @@
+---
+"@osdk/create-app.template.react.beta": patch
+---
+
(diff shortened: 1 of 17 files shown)


zeyadkhaled (approved):

samialfhaily (approved):
approving to unblock

PR #3090, opened 2026-04-24
Fix storybook and types

ziyunp:
- Fixed storybook displaying "object Object"
- Added more stories 
- FilterList stories - withObjectSet, HasLink filter, Custom filter
- ObjectTable stories - derived properties example with defaultOrderBy and filter, function column, interface type stories
- Fixed FilterList types that need casting
- Fixed ObjectTable orderBy type to accept RDP properties

Before:
Screenshot 2026-04-24 at 14 02 54

After:
Screenshot 2026-04-24 at 14 02 41

### DIFF
--- a/.changeset/nine-paws-sniff.md
+++ b/.changeset/nine-paws-sniff.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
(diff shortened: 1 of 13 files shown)


nfdiop (dismissed):
Thanks for fixing!

nfdiop (approved):

PR #3092, opened 2026-04-24
Improve ActionForm type safety with per-type field wrappers

nfdiop:
## Summary

- Make `RendererFieldDefinition<S>` generic over a form schema so `fieldKey` is constrained to `keyof S` and Dropdown/RadioButtons/Custom get typed items/options from `S[K]`
- Make `FormFieldDefinition` a discriminated union so choosing a `fieldComponent` narrows `fieldComponentProps` (e.g. `"DROPDOWN"` requires `items`)
- Make `BaseForm<S>` generic with typed `onSubmit(formState: S)` and `onFieldValueChange`
- Replace `coerceForRender` with per-type memoized wrapper components in `FormFieldRenderer` — each wrapper uses field extractors directly with `useMemo` and typed `useCallback` handlers, eliminating all `as` casts from the rendering pipeline
- Simplify `ObjectSetFieldProps` to use `ObjectSet` directly (instead of `BaseObjectSet<T>`) — removes the generic param, the runtime duck-typing check, and lets TypeScript guarantee the shape

## Test plan

- [ ] `pnpm turbo typecheck --filter=@osdk/react-components` passes
- [ ] `pnpm turbo typecheck --filter=@osdk/react-components-storybook` passes
- [ ] All action-form tests pass
- [ ] Verify that defining a `BaseForm` with a schema catches type mismatches
- [ ] Grep `FormFieldRenderer.tsx` for `as ` — zero matches


packages/e2e.sandbox.peopleapp/src/app/form/page.tsx
```diff
@@ -1,14 +1,33 @@
...
   BaseFormFieldProps,
+  DateRange,
   RendererFieldDefinition,
 } from "@osdk/react-components/experimental/action-form";
 import { useCallback, useMemo, useState } from "react";
 import { $ } from "../../foundryClient.js";
 import { Employee } from "../../generatedNoCheck2/index.js";
 import "./form-page.css";
 
+interface EmployeeFormSchema {
```
nfdiop: I am now requiring a state because I want stronger typing. Previously when users were using dropdown the props were not strongly typed (you could have a list of string, then have the renderItem take a number). 

The other alternative was that fieldComponent here would be a union like `{type: "dropdown", valueType: string} | {type: "textInput} | ...` This looked cleaner

packages/e2e.sandbox.peopleapp/src/app/form/page.tsx
```diff
@@ -190,23 +211,17 @@ const fieldDefinitions: ReadonlyArray<RendererFieldDefinition> = [
...
-  );
+  const handleSubmit = useCallback(async (formState: EmployeeFormSchema) => {
+    setSubmittedState(formState);
+  }, []);
 
-  const employeeObjectSet = useMemo(
-    () => $(Employee) as ObjectSet<ObjectTypeDefinition>,
-    [],
-  );
+  const employeeObjectSet = useMemo(() => $(Employee), []);
```
nfdiop, line 226: Requiring cast is cumbersome so I am using BaseObjectSet. I'm not 100% sure this is the right move. Would love to hear a different opinion

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -136,7 +115,7 @@ export interface FormFieldPropsByType {
   FILE_PICKER: FilePickerProps;
   NUMBER_INPUT: NumberInputFieldProps;
   OBJECT_SELECT: ObjectSelectFieldProps<ObjectTypeDefinition>;
-  OBJECT_SET: ObjectSetFieldProps<ObjectTypeDefinition>;
+  OBJECT_SET: ObjectSetFieldProps;
```
nfdiop, line 118: The generic is not adding any value so removed

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -18,100 +18,79 @@ import type {
...
+ *
+ * Implemented as a distributed mapped type so that choosing a `fieldComponent`
+ * narrows `fieldComponentProps` to the correct props — e.g. selecting
+ * `"DROPDOWN"` requires `items`, selecting `"RADIO_BUTTONS"` requires `options`.
+ *
+ * Dropdown, RadioButtons, and Custom fields receive the concrete value type
+ * from the action parameter (`FieldValueType<Q, K>`) instead of `unknown`.
  */
-export interface FormFieldDefinition<
+export type FormFieldDefinition<
```
nfdiop, line 41: without this change, it is possible to select a dropdown items and the typing would be wrong: we can have a list of strings and itemToStringLabel would take a number

PR #3094, opened 2026-04-24
fix infinite re-render loop with inline withProperties rdps

KuberSethi:
inline `withProperties` rdps caused `useOsdkObjects`, `useObjectSet`, and `useOsdkAggregation` to refetch on every render because canonicalization couldn't dedupe function-valued properties

• `canonicalizeOptions` was using `genericCanonicalizer` for `withProperties`, which compares with `fast-deep-equal` — that returns false for two different function references even when the source is identical, so every render produced a new canonical reference and the hook's `useMemo` deps churned
• swap to `rdpCanonicalizer` (already used by `ListsHelper`/`ObjectSetHelper`/`AggregationsHelper`) — it invokes each rdp function against a synthetic builder, extracts the resulting `DerivedPropertyDefinition`, and keys by `JSON.stringify` of those definitions, so semantically identical inline rdps return the same canonical reference
• consumers no longer need to wrap inline `withProperties` in `useMemo`


packages/react/test/useOsdkObjects.test.tsx
```diff
@@ -223,4 +223,43 @@ describe("useOsdkObjects enabled option", () => {
 
     expect(mockInvalidateObjectType).toHaveBeenCalledWith("MockObject");
   });
+
+  it("should not resubscribe when rerendered with a new inline withProperties of the same shape", () => {
```
nfdiop, line 227: thanks for adding tests

nfdiop (approved):

PR #3100, opened 2026-04-25
add enabled flag to useOsdkAggregation

KuberSethi:
useOsdkAggregation had no enabled flag, so callers had to use a spread-operator workaround when an upstream objectSet wasn't yet resolved

• add `enabled?: boolean` (default true) to UseOsdkAggregationBaseOptions, mirroring useOsdkObjects/useOsdkObject/useOsdkFunction
• short-circuit the store useMemo to a no-op when disabled so observeAggregation isn't called
• thread `enabled` through isPayloadLoading so isLoading reports false when disabled


KuberSethi:

- #3100 ◀

PR #3103, opened 2026-04-27
expose $includeAllBaseObjectProperties on react hooks

KuberSethi:
exposes $includeAllBaseObjectProperties on useOsdkObjects, useLinks, and useOsdkObject so callers can request all properties of the underlying concrete object type when querying interfaces

- option threads through observable client + cache keys + BulkObjectLoader, etc such that queries with different `$includeAllBaseObjectProperties` don't map to the same cache key.
- returned Osdk.Instance typing is conditional on the presence of `$includeAllBaseObjectProperties`, so obj.\$as(ConcreteType) works when the option is set
- ObserveObjectsCallbackArgs no longer hardcodes "\$allBaseProperties"


KuberSethi:
superseded by per-hook split:

- foundation: https://github.com/palantir/osdk-ts/pull/3134
- useOsdkObject: https://github.com/palantir/osdk-ts/pull/3135
- useOsdkObjects: https://github.com/palantir/osdk-ts/pull/3136
- useLinks: https://github.com/palantir/osdk-ts/pull/3137

leaving this open as a reference until the split lands. each new PR is sized for review independently; PRs 3135/3136/3137 all branch off 3134 so they auto-rebase to main once the foundation merges.

PR #3104, opened 2026-04-27
fix subscribe typing for interfaces

KuberSethi:
type-only fix: `client(genericInterfaceDefinition).subscribe(...)` failed to typecheck even though the runtime fully supports interface subscriptions

• move Subscribe from ObjectSetCleanedTypes into MinimalObjectSet so both the interface and object client paths get .subscribe at the type level

there should be no functional changes here, this is supported behavior that our types didn't adequately represent

added new intellisense tests for this as well as other interface-specific types

### DIFF
--- a/.changeset/interface-subscribe.md
+++ b/.changeset/interface-subscribe.md
@@ -1,0 +1,5 @@
+---
+"@osdk/api": minor
+---
+
+expose .subscribe() on interface-typed object sets returned by client()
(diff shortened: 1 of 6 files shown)


bryantpark04:
Does the implementation already remap from object properties to interface properties? OSW supports this, but it returns the object properties in the update. In Python I had to pipe the property mapping through.

nihalbhatnagar (approved):

KuberSethi:
> Does the implementation already remap from object properties to interface properties? OSW supports this, but it returns the object properties in the update. In Python I had to pipe the property mapping through.

Yeah I checked and we remap properly already!

PR #3105, opened 2026-04-27
Add support for streaming queries

tmisirpash:
This PR adds support for queries with streamed responses, which are currently supported in the platform SDK through the `streamingExecute` endpoint from the `@osdk/foundry.functions` package.

While the platform SDK requires you to read the stream from the `Response` manually and decode everything from UTF-8, the OSDK wraps that in a cleaner `AsyncIterable` interface that allows you to use `for await` syntax.

This currently marked as an experimental feature.

```typescript
const items: number[] = [];

const stream = client(__EXPERIMENTAL__NOT_SUPPORTED_YET__executeStreamingFunction)
              .executeStreamingFunction(addOne, { n: 2 });

for await (const item of stream) {
    items.push(item);
}
```


packages/client/src/util/streamutils.ts
```diff
@@ -121,3 +121,26 @@ export async function* iterateReadableStream(
...
+  const decoder = new TextDecoder("utf-8");
+  let buffer = "";
+  for await (const chunk of asyncIterable) {
+    buffer += decoder.decode(chunk, { stream: true });
+    let newlineIdx: number;
+    while ((newlineIdx = buffer.indexOf("\n")) !== -1) {
+      const line = buffer.slice(0, newlineIdx);
+      buffer = buffer.slice(newlineIdx + 1);
+      if (line.length > 0) {
+        yield JSON.parse(line);
```
ssanjay1, line 137: Is there any chance like we get a badly formatted line here from the backend data. Right now this would just break right, which seems like what we'd want? Or do we wanna skip with like try/catch and keep going on other potentially ok lines

tmisirpash: Hm, if the JSON we get is malformed then that indicates something has probably gone really wrong somewhere, since we should be enforcing the correct structure in the function runtime layer, function-executor and API Gateway. Breaking sounds right to me, and we would probably need to investigate further.

bryantpark04 (dismissed):

ssanjay1 (approved):

PR #3107, opened 2026-04-28
Export createMediaFromReference from @osdk/client/internal

nihalbhatnagar:
## Summary
- Adds `createMediaFromReference` to the `@osdk/client/internal` entry point so consumers can hydrate a `Media` object from a `MediaReference` outside of an object property context.

## Test plan
- [ ] CI passes

### DIFF
--- a/packages/react-components/vitest.config.mts
+++ b/packages/react-components/vitest.config.mts
@@ -18,7 +18,7 @@ import { configDefaults, defineConfig } from "vitest/config";
 
 export default defineConfig({
   test: {
-    pool: "forks",
+    pool: "threads",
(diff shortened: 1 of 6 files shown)


packages/react-components/vitest.config.mts
```diff
@@ -18,7 +18,7 @@ import { configDefaults, defineConfig } from "vitest/config";
 
 export default defineConfig({
   test: {
-    pool: "threads",
+    pool: "forks",
```
ssanjay1: Did you mean to do this?

ssanjay1 (dismissed):

KuberSethi (approved):

PR #3112, opened 2026-04-28
fix link query refresh for per-pk invalidation and interface targets

KuberSethi:
link queries weren't getting refreshed by per-pk invalidation, and interface-typed link targets were silently skipped during invalidation matching

• kick `specificLink` queries from `Store.invalidateObject` so per-pk invalidation refreshes link queries (previously only handled at type-level)
• fix `SpecificLinkQuery.invalidateObjectType` so interface-implementation matching on object-type targets isn't silently skipped
• fix `BaseListQuery.rdpConfig` index collision (`SpecificLinkCacheKey.otherKeys[4]` is the link name, not the rdp config); make `rdpConfig` abstract with concrete overrides per subclass

### DIFF
--- a/packages/client/src/observable/internal/Store.ts
+++ b/packages/client/src/observable/internal/Store.ts
@@ -309,7 +309,41 @@ export class Store {
       }
     }
 
+    // Per-PK invalidation doesn't propagate to specificLink queries (they're
+    // keyed on srcType+srcPk+linkName, not the linked object's pk).
+    promises.push(this.invalidateLinkQueriesForType(apiName));
+
     return Promise.allSettled(promises);
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


packages/client/src/observable/internal/Store.ts
```diff
@@ -309,9 +309,43 @@ export class Store {
       }
     }
 
+    // Per-PK invalidation doesn't propagate to specificLink queries (they're
+    // keyed on srcType+srcPk+linkName, not the linked object's pk).
+    promises.push(this.invalidateLinkQueriesForType(apiName));
```
ziyunp, line 314: Just to clarify my understanding, this line is the actual fix right? The changes in SpecificLinkQuery is just a refactor?

ziyunp (approved):

PR #3114, opened 2026-04-28
OAC edits history

ethanai22:


packages/maker/src/conversion/toMarketplace/convertObject.ts
```diff
@@ -292,3 +298,24 @@ function buildAggregation(
...
+          store: apiName,
+          storeAllPreviousProperties: config.storeAllPreviousProperties,
+        },
+      }
+      : {
+        type: "none",
+        none: {},
+      };
+  }
+  return undefined;
```
davidpaquin, line 320: is `undefined` separate from `none` here?

packages/maker/src/conversion/toMarketplace/convertObject.ts
```diff
@@ -292,3 +298,24 @@ function buildAggregation(
...
+function convertEditsHistory(
+  apiName: string,
+  config?: EditsHistoryConfig,
+): OntologyIrEditsHistory | undefined {
+  if (config) {
+    return config.enabled
+      ? {
+        type: "config",
+        config: {
+          store: apiName,
```
davidpaquin, line 311: interesting that they would have the api name / rid as a field here - is there any case where you'd want that to be different from the owning OT?

ethanai22: seems like they model the edits history config as a separate entity from the object itself. Skimmed through some code and they have dedicated queries for history

packages/maker/src/conversion/toMarketplace/convertObject.ts
ethanai22: this is mostly for safety to avoid changing IR from existing defineObject calls. In practice I think they resolve to the same thing

davidpaquin (approved):

PR #3120, opened 2026-04-29
ga @osdk/react: move hooks out of experimental, consolidate to one provider

KuberSethi:
promote modern hooks from /experimental to the main entry, rename admin to /platform-apis, consolidate to a single OsdkProvider, bump @osdk/react to 1.0.0

• `OsdkProvider`, `useOsdkObjects`, `useOsdkObject`, `useOsdkAction`, `useLinks`, `useObjectSet`, `useOsdkAggregation`, `useOsdkFunction`, `useOsdkFunctions`, `useStableObjectSet`, `useRegisterUserAgent`, `useDebouncedCallback` now live at `@osdk/react`. admin / cbac platform hooks (`useFoundryUser`, `useCurrentFoundryUser`, `useFoundryUsersList`, `useMarkings`, `useMarkingCategories`, `useUserViewMarkings`, `useCbacBanner`, `useCbacMarkingRestrictions`) move to `@osdk/react/platform-apis`
• consolidate to one provider: `OsdkProvider2` is renamed to `OsdkProvider`. `useOsdkClient` and `useOsdkMetadata` keep working since the new provider supplies the same `client` shape
• `@osdk/client/observable` is the new stable home for `createObservableClient`/`ObservableClient` and supporting types so the GA hooks no longer depend on `unstable-do-not-use`
• `@osdk/react/experimental`, `@osdk/react/experimental/admin`, `OsdkProvider2`, `useOsdkClient2`, and the observable symbols on `@osdk/client/unstable-do-not-use` are kept as `@deprecated` shims so 0.x consumers can upgrade with no code changes


packages/react/src/public/experimental.ts
nfdiop, line 1: Will this break user imports?

KuberSethi: yeah that's the point though - we're major versioning and this will contain breaking updates

nfdiop (dismissed):

packages/react/src/public/experimental.ts
nfdiop: Will defer to you but this will be a major pain point for users to update

.changeset/react-ga.md
```diff
@@ -0,0 +1,23 @@
+---
+"@osdk/react": major
+"@osdk/react-components": patch
+"@osdk/cbac-components": patch
+---
+
+GA: promote modern hooks from `@osdk/react/experimental` to the main entry, rename `@osdk/react/experimental/admin` → `@osdk/react/platform-apis`, consolidate to a single `OsdkProvider`
+
+#### `@osdk/react` (major)
```
ziyunp, line 16: Suggestion: Add clear migration steps

packages/client/src/public/unstable-do-not-use.ts
```diff
@@ -34,20 +51,9 @@ export type {
   ObserveObjectSetArgs,
   Unsubscribable,
 } from "../observable/ObservableClient.js";
+/** @deprecated Import from `@osdk/client/observable` instead. */
 export type { Observer } from "../observable/ObservableClient/common.js";
+/** @deprecated Import from `@osdk/client/observable` instead. */
```
nfdiop, line 56: interesting so if we add this on top, it will show up for the users as deprecated?

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
```diff
@@ -67,7 +68,10 @@ describe(useColumnDefs, () => {
   const createWrapper = (client: Client) => {
     return ({ children }: React.PropsWithChildren) => {
       return (
-        <OsdkProvider client={client}>
+        <OsdkProvider
+          client={client}
```
nfdiop, line 71: remark: I would expect a single client being passed either client={fakeObservableClient} or client={client}. This means it is possible that the two would be different. Is that expected?

packages/client/src/public/unstable-do-not-use.ts
KuberSethi: yeah, typescript and most IDEs render it with a strikethrough + a tooltip from the jsdoc marking it as deprecated. eslint-plugin-deprecation picks it up too if users are using that.

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
KuberSethi: this is for testing purposes, it lets tests stub out the observable layer when their client is a partial mock, arguably we should remove but this isn't anything new it exists on main now. I'd rather keep this interface intact as removing it would be a break that breaks compilation

packages/client/src/public/unstable-do-not-use.ts
nfdiop: nice! we should do the same for react components cc @olliejenningsPT

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
nfdiop: I think we should resolve this before GA. Once we ship with both `client` and `observableClient` as separate props on the provider, we're locked into supporting that surface. If they can diverge, that's a bug waiting to happen.

I know this exists on main today, but main is pre-GA: this PR is the right time to clean it up. Can we make the provider accept one client and derive the other internally?

KuberSethi: Fair call, this makes sense! Addressed to only accept one client and derive the other internally

nfdiop (dismissed):
Exciting! Nice work! Thanks for doing this!

bryantpark04 (approved):
stamping for merge conflicts

PR #3136, opened 2026-04-30
expose $includeAllBaseObjectProperties on useOsdkObjects

KuberSethi:
exposes \`\$includeAllBaseObjectProperties\` on \`useOsdkObjects\` and on the underlying \`ObservableClient.observeList\`

when set against an interface query, the server returns the underlying concrete object's full property set so \`obj.\$as(ConcreteType)\` yields a fully-populated concrete object

• overrides BaseListQuery's includeAllBaseObjectProperties getter on ListQuery to read from cache key
• partitions ListCacheKey by the flag, gated to interface queries in ListsHelper
• threads the flag through observeList signature and useOsdkObjects hook + tests
• wires the demo column in officenetwork EmployeePanel via \$as(Employee).jobTitle

### DIFF
--- a/packages/e2e.sandbox.officenetwork/src/components/EmployeePanel.tsx
+++ b/packages/e2e.sandbox.officenetwork/src/components/EmployeePanel.tsx
@@ -139,6 +139,7 @@ export function EmployeePanel({
     {
       where: { employeeNumber: employee.employeeNumber },
       pageSize: 1,
+      $includeAllBaseObjectProperties: true,
     },
   );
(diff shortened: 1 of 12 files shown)


packages/e2e.sandbox.officenetwork/src/components/EmployeePanel.tsx
nfdiop, line 1: nice updating!

nfdiop (approved):

nfdiop (approved):

PR #3139, opened 2026-04-30
add chat components to react-components experimental

KuberSethi:
ports four chat-essential components from the forge agent-engine demo into `@osdk/react-components/experimental` so OSDK consumers can build agent chat UIs without depending on forge internals

• `TypingIndicator`, `ContextItemPill`, `MessageList`, `MessageInput` plus a `Chat` composition storybook story; theming via `--osdk-*` CSS variables, English-only labels, no `useEffect`
• `MessageList` exposes an imperative `scrollToBottom()` ref + opt-in `autoScrollOnAppend` prop; `MessageInput` accepts an `attachmentTrigger` slot, optional `onConfigure`, and an optional auto-scroll toggle pair


packages/react-components/src/public/experimental.ts
```diff
@@ -15,8 +15,12 @@
  */
 
 export * from "./experimental/action-form.js";
```
ziyunp, line 17: Suggestion: On file structure, can we contain chat interfaces in a folder pls? Similarly for the tokens. We should have the same structure as all other components where we have the Osdk layer and base components layer

packages/react-components-storybook/src/stories/Chat/Chat.stories.tsx
```diff
@@ -0,0 +1,229 @@
...
+        height: "640px",
+        width: "100%",
+        display: "flex",
+        flexDirection: "column",
+        border: "1px solid var(--osdk-surface-border-color-default)",
+        borderRadius: 8,
+        overflow: "hidden",
+      }}
+    >
+      <MessageList
```
ziyunp, line 130: Suggestion: Can we have a single entry component like

<Chat /> 

which has both MessageList and MessageInput? Default the layout for now

PR #3148, opened 2026-04-30
Add SLS Structured Logging to OSDK Generation

jameszhang244:
## Motivation
The generator previously logged via `consola`, which produced human-readable output unsuitable for log forwarders. More importantly, several error paths embedded user-controlled data (API names, object type names) directly in error message strings, meaning that data would appear in stack traces logged as plain text, with no way to control where it ends up.

#### `Structured logging (SlsLogger, logDuration)`

`SlsLogger` emits newline-delimited JSON records to stdout. Each record separates log data into two explicit buckets:

- params: safe to log anywhere: RIDs, counts, durations, internal type identifiers
- unsafeParams: user-controlled data (API names, display names, etc.) that may require restricted handling

`logDuration` wraps an async operation and logs start, completion (with `durationMs`), and failure. The generator now uses it to instrument ontology metadata loading and package rendering as separate timed phases.

Both `JOB_ID` and `TRACE_ID` are read from the environment and threaded into every log record for cross-process correlation.

The previous `consola` output has been removed. The CLI now exits with a non-zero status code on failure instead of swallowing errors.

#### GeneratorError (@osdk/generator-converters)

A new `Error` subclass that carries structured log params alongside the error message:

- The message string is always safe to log
- `unsafeParams` holds user-controlled data that caused the error (e.g. the API name of the offending object type)
- `params` holds safe diagnostic context (e.g. constraint counts, internal type identifiers)

`logDuration` recognizes `GeneratorError` via `instanceof` and routes each bucket to the correct field in the failure log record, rather than letting that data fall through to the stack trace.

Throw sites throughout the generator that previously embedded API names directly in error message strings have been converted to use `GeneratorError`.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -47,9 +47,7 @@
     "@osdk/e2e.generated.*",
     "@osdk/e2e.sandbox.*",
     "@osdk/ontology-explorer-app",
-    "@osdk/ontology-explorer-server",
-    "@osdk/react-devtools",
-    "@osdk/e2e.sandbox.officenetwork"
+    "@osdk/ontology-explorer-server"
   ],
   "bumpVersionsWithWorkspaceProtocolOnly": true,
   "snapshot": {
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -48,7 +48,8 @@
     "@osdk/e2e.sandbox.*",
     "@osdk/ontology-explorer-app",
     "@osdk/ontology-explorer-server",
-    "@osdk/react-devtools"
+    "@osdk/react-devtools",
+    "@osdk/e2e.sandbox.officenetwork"
   ],
   "bumpVersionsWithWorkspaceProtocolOnly": true,
   "snapshot": {
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -47,7 +47,8 @@
     "@osdk/e2e.generated.*",
     "@osdk/e2e.sandbox.*",
     "@osdk/ontology-explorer-app",
-    "@osdk/ontology-explorer-server"
+    "@osdk/ontology-explorer-server",
+    "@osdk/react-devtools"
   ],
   "bumpVersionsWithWorkspaceProtocolOnly": true,
   "snapshot": {
--- a/.changeset/foundry-sdk-generator-sls-logging.md
+++ b/.changeset/foundry-sdk-generator-sls-logging.md
@@ -1,0 +1,5 @@
+---
+"@osdk/foundry-sdk-generator": minor
+---
+
+Emit structured `service.1` JSON logs to stdout for log forwarders, with timing for ontology metadata loading and OSDK package generation phases. Reads `JOB_ID` and `TRACE_ID` from env and threads them into every record's params for cross-process correlation. The previous human-readable console output (via consola) has been removed; the CLI now exits with a non-zero status code on failure.
--- a/.changeset/foundry-sdk-generator-sls-logging.md
+++ b/.changeset/foundry-sdk-generator-sls-logging.md
@@ -4,4 +4,4 @@
 "@osdk/generator-converters": patch
 ---
 
-Add structured SLS `service.1` JSON logging to stdout; add `GeneratorError` to `@osdk/generator-converters` for separating user-controlled data from safe log params.
+Emit structured `service.1` JSON logs to stdout for log forwarders, with timing for ontology metadata loading and OSDK package generation phases. Reads `JOB_ID` and `TRACE_ID` from env and threads them into every record's params for cross-process correlation. The previous human-readable console output (via consola) has been removed; the CLI now exits with a non-zero status code on failure.
--- a/.changeset/foundry-sdk-generator-sls-logging.md
+++ b/.changeset/foundry-sdk-generator-sls-logging.md
@@ -4,6 +4,4 @@
 "@osdk/generator-converters": patch
 ---
 
-Emit structured `service.1` JSON logs to stdout for log forwarders, with timing for ontology metadata loading and OSDK package generation phases. Reads `JOB_ID` and `TRACE_ID` from env and threads them into every record's params for cross-process correlation. The previous human-readable console output (via consola) has been removed; the CLI now exits with a non-zero status code on failure.
-
-Add `GeneratorError` class to `@osdk/generator-converters` that carries structured `unsafeParams` separately from the error message, keeping user-controlled data (API names, object type names) out of the SLS `stacktrace` field and routing it to the `unsafeParams` log field instead.
+Add structured SLS `service.1` JSON logging to stdout; add `GeneratorError` to `@osdk/generator-converters` for separating user-controlled data from safe log params.
--- a/.changeset/foundry-sdk-generator-sls-logging.md
+++ b/.changeset/foundry-sdk-generator-sls-logging.md
@@ -1,5 +1,9 @@
 ---
 "@osdk/foundry-sdk-generator": minor
+"@osdk/generator": patch
+"@osdk/generator-converters": patch
 ---
 
 Emit structured `service.1` JSON logs to stdout for log forwarders, with timing for ontology metadata loading and OSDK package generation phases. Reads `JOB_ID` and `TRACE_ID` from env and threads them into every record's params for cross-process correlation. The previous human-readable console output (via consola) has been removed; the CLI now exits with a non-zero status code on failure.
+
+Add `GeneratorError` class to `@osdk/generator-converters` that carries structured `unsafeParams` separately from the error message, keeping user-controlled data (API names, object type names) out of the SLS `stacktrace` field and routing it to the `unsafeParams` log field instead.
--- a/packages/foundry-sdk-generator/src/cli/foundrySdkGeneratorCli.ts
+++ b/packages/foundry-sdk-generator/src/cli/foundrySdkGeneratorCli.ts
@@ -19,13 +19,10 @@ import { hideBin } from "yargs/helpers";
 import { GeneratePackageCommand } from "../generate/index.js";
 import { SlsLogger } from "../logging/index.js";
 
-let crashHandlersInstalled = false;
-
 export async function cli(args: string[] = process.argv): Promise<
   Record<string, unknown> | undefined
 > {
   const logger = new SlsLogger();
-  installCrashHandlers(logger);
 
   const base = yargs(hideBin(args))
     .command(new GeneratePackageCommand())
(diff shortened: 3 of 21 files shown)


packages/foundry-sdk-generator/src/logging/SlsLogger.ts
```diff
@@ -0,0 +1,107 @@
...
+      params: { ...this.envParams, ...params?.params },
+    };
+    if (this.traceId) {
+      entry.traceId = this.traceId;
+    }
+    if (params?.unsafeParams && Object.keys(params.unsafeParams).length > 0) {
+      entry.unsafeParams = params.unsafeParams;
+    } else {
+      entry.safe = true;
+    }
```
jameszhang244, line 104: `safe: true` tells our logging service that this entry contains no PII/user-controlled data. We need it for logs to be visible/queryable.

jameszhang244, line 104: entries with user-controlled data go to unsafeParams, entries without get safe: true

packages/generator-converters/src/GeneratorError.ts
```diff
@@ -0,0 +1,36 @@
...
+    message: string,
+    unsafeParams?: Record<string, unknown>,
+    params?: Record<string, unknown>,
+  ) {
+    super(message);
+    this.name = "GeneratorError";
+    this.unsafeParams = unsafeParams;
+    this.params = params;
+  }
+}
```
jameszhang244, line 36: errors thrown during generation now throw GeneratorError

packages/foundry-sdk-generator/src/logging/SlsLogger.ts
```diff
@@ -0,0 +1,106 @@
...
+  private readonly traceId?: string;
+  private readonly stream: NodeJS.WritableStream;
+
+  constructor(stream: NodeJS.WritableStream = process.stdout) {
+    this.stream = stream;
+    this.envParams = {};
+    const jobId = process.env.JOB_ID;
+    if (jobId) this.envParams.jobId = jobId;
+    const traceId = process.env.TRACE_ID;
+    if (traceId) this.traceId = traceId;
```
jameszhang244, line 55: Attach a job id and trace to the logs. `traceId` is a top-level SLS field, while `jobId` goes into `params`

packages/foundry-sdk-generator/src/logging/SlsLogger.ts
jameszhang244, line 1: Based this off https://github.com/palantir/typescript-compute-module/blame/develop/typescript-compute-module/src/logging/SlsLogger.ts as there's no existing TS sls logger utility package. FLUP could be to make one, but not sure it's worth just for these two repos.

packages/foundry-sdk-generator/src/cli/foundrySdkGeneratorCli.ts
```diff
@@ -33,7 +36,11 @@ export async function cli(args: string[] = process.argv): Promise<
...
     return await base.parseAsync();
   } catch (e) {
-    // eslint-disable-next-line no-console
-    console.log(e);
+    logger.error(
+      "OSDK generation failed",
+      undefined,
+      e instanceof Error ? e : new Error(String(e)),
+    );
+    process.exit(1);
```
nihalbhatnagar, line 44: I cant think of a reason this would break someone but it is theoretically an invariant change, probably the right thing to do though

packages/maker/package.json
```diff
@@ -42,7 +42,7 @@
   "dependencies": {
     "@osdk/api": "workspace:~",
     "@osdk/generator-converters.ontologyir": "workspace:~",
-    "@osdk/typescript-sdk-docs": "^0.2.0",
+    "@osdk/typescript-sdk-docs": "^0.9.0",
```
nihalbhatnagar: Why?

turbo.json
```diff
@@ -234,7 +235,7 @@
     "test": {
       "outputLogs": "new-only",
       "env": ["FORCE_COLOR"],
-      "dependsOn": ["^transpileEsm"]
```
nihalbhatnagar, line 237: Why?

.changeset/config.json
```diff
@@ -47,7 +47,9 @@
     "@osdk/e2e.generated.*",
     "@osdk/e2e.sandbox.*",
     "@osdk/ontology-explorer-app",
-    "@osdk/ontology-explorer-server"
+    "@osdk/ontology-explorer-server",
+    "@osdk/react-devtools",
```
nihalbhatnagar: Why?

packages/foundry-sdk-generator/src/generate/GeneratePackageCommand.ts
```diff
@@ -184,67 +185,87 @@ export class GeneratePackageCommand
...
-    });
+        // Narrowing doesn't carry into the closure below, so extract here.
+        const ontologyInfo = wireOntologyDefinition.value;
 
-    const elapsedTime = Date.now() - timeStart;
-    consola.success(
-      `Finished generating package in ${(elapsedTime / 1000).toFixed(2)}s`,
+        await logDuration(
+          logger,
+          "Rendering OSDK",
```
nihalbhatnagar, line 254: What does rendering mean again, like constructing from templates?

packages/foundry-sdk-generator/src/logging/SlsLogger.ts
nihalbhatnagar: Agree

packages/foundry-sdk-generator/src/logging/SlsLogger.test.ts
```diff
@@ -0,0 +1,280 @@
...
+    const logger = new SlsLogger(stream);
+
+    logger.info("msg", { params: { count: 5 } });
+
+    const [entry] = parseEntries(writes);
+    expect(entry.params).toEqual({
+      jobId: "job-123",
+      count: 5,
+    });
+    expect(entry.traceId).toBe("trace-abc");
```
nihalbhatnagar: Lets use snapshot errors for this whole file

packages/foundry-sdk-generator/src/logging/SlsLogger.test.ts
```diff
@@ -0,0 +1,280 @@
...
+      message: "Loading data...",
+    });
+    expect(entries[1]).toMatchObject({
+      level: "INFO",
+      message: "Loading data complete.",
+      params: { durationMs: 150 },
+    });
+  });
+
+  it("propagates unsafeParams from errors that carry them", async () => {
```
nihalbhatnagar, line 305: Are both the message + cause  propagated from the original error?

packages/foundry-sdk-generator/src/logging/SlsLogger.test.ts
```diff
@@ -0,0 +1,280 @@
...
+  it("includes stacktrace on error() with an Error", () => {
+    const { stream, writes } = captureStream();
+    const logger = new SlsLogger(stream);
+    const err = new Error("boom");
+
+    logger.error("failure", undefined, err);
+
+    const [entry] = parseEntries(writes);
+    expect(entry.level).toBe("ERROR");
+    expect(entry.stacktrace).toContain("Error: boom");
```
nihalbhatnagar, line 242: Snapshot test here will be great

nihalbhatnagar (commented):
Mostly LGTM!

packages/foundry-sdk-generator/src/logging/SlsLogger.ts
nihalbhatnagar: I think a comment would be great!, and an explicit test (may already be there)

packages/generator-converters/src/GeneratorError.ts
nihalbhatnagar: So when being run in logDuration, what will the thrown error look like since its caught again?

packages/foundry-sdk-generator/src/generate/GeneratePackageCommand.ts
jameszhang244: yep exactly. I used the same log message as the other osdks

packages/generator-converters/src/GeneratorError.ts
jameszhang244: logDuration re-throws the original error after logging, so the caller gets the same `GeneratorError` object unchanged. Logging is just a side effect; the error propagation isn't modified

packages/foundry-sdk-generator/src/logging/SlsLogger.test.ts
jameszhang244: yep, the failure log entry captures both. So, both the formatted message in message and the original error's stack trace in stacktrace

packages/foundry-sdk-generator/src/logging/SlsLogger.test.ts
jameszhang244: yep, done in https://github.com/palantir/osdk-ts/pull/3148/changes/00bc1ecd66c978f6fcee35306b67f01d6e383c65

packages/foundry-sdk-generator/src/logging/SlsLogger.test.ts
jameszhang244: sg, done in https://github.com/palantir/osdk-ts/commit/00bc1ecd66c978f6fcee35306b67f01d6e383c65

packages/maker/package.json
jameszhang244: I think I did this bc ci was failing. Can't find the ci failure now. undid and things seem fine. Same w the other changes in turbo.json and config.json

packages/foundry-sdk-generator/src/logging/SlsLogger.ts
jameszhang244: added in https://github.com/palantir/osdk-ts/commit/00bc1ecd66c978f6fcee35306b67f01d6e383c65

There are three tests covering this in SlsLogger.test.ts:

- "keeps params and unsafeParams in separate fields and omits safe" (line 165): logs with unsafeParams present, snapshot shows unsafeParams field appears and safe is absent
- "marks records with no unsafeParams as safe:true" (line 192): logs with only params, snapshot shows `safe: true`
- "treats empty unsafeParams as no unsafe data and marks safe:true" (line 214): edge case: passing `unsafeParams: {}` still results in `safe: true`

.changeset/foundry-sdk-generator-sls-logging.md
```diff
@@ -0,0 +1,9 @@
+---
+"@osdk/foundry-sdk-generator": minor
+"@osdk/generator": patch
+"@osdk/generator-converters": patch
+---
+
+Emit structured `service.1` JSON logs to stdout for log forwarders, with timing for ontology metadata loading and OSDK package generation phases. Reads `JOB_ID` and `TRACE_ID` from env and threads them into every record's params for cross-process correlation. The previous human-readable console output (via consola) has been removed; the CLI now exits with a non-zero status code on failure.
```
nihalbhatnagar, line 7: Super nit but can we just make the changelog the first line here

nihalbhatnagar (dismissed):

.changeset/foundry-sdk-generator-sls-logging.md
jameszhang244: done!

nihalbhatnagar (approved):

PR #3151, opened 2026-05-01
Define AipAgentChat API surface

ziyunp:
## Summary

Lands the type contract for `AipAgentChat` in `@osdk/react-components` ahead of the implementation PR. Stacks on #3130 (which introduces `@osdk/aip-core` and `useChat`). Types-only — no runtime exports yet, so the package surface area is unchanged.

`AipAgentChatProps`:

- `model: LanguageModel` (required) — drives the `useChat` hook from `@osdk/react/experimental/aip`. Construct via `foundryModel()` from `@osdk/aip-core`.
- `system`, `initialMessages` — passthrough to `useChat`.
- `className`, `placeholder`, `enableAutoScroll` — UI knobs.
- `onError` / `onFinish` — non-controlling listeners forwarded to `useChat`'s callbacks.
- `renderEmptyState` / `renderMessage` — render slots; operate on `UIMessage` directly.

`@osdk/aip-core` is added as an optional peer dependency on `@osdk/react-components`.

The implementation PR will stack on this branch.

## Test plan

- [ ] `pnpm turbo typecheck --filter=@osdk/react-components` passes (verified locally).
- [ ] No public-API changes — the file is not exported, so the package surface is unchanged.

### DIFF
--- a/packages/react-components/src/aip-agent-chat/AipAgentChatApi.ts
+++ b/packages/react-components/src/aip-agent-chat/AipAgentChatApi.ts
@@ -33,6 +33,26 @@ export interface AipAgentChatProps {
   model: LanguageModel;
 
   /**
+   * When provided, the chat renders an in-header model picker populated
+   * with these options. Selecting a different option fires
+   * {@link AipAgentChatProps.onModelChange}; the consumer is expected to
+   * update {@link AipAgentChatProps.model} in response.
+   *
+   * If omitted, no picker is rendered.
+   */
+  availableModels?: ReadonlyArray<LanguageModel>;
+
+  /**
+   * Fires when the user selects a different model from the in-header
+   * picker. Has no effect unless {@link AipAgentChatProps.availableModels}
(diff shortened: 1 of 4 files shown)


packages/react-components/src/aip-agent-chat/AipAgentChatApi.ts
```diff
@@ -0,0 +1,135 @@
...
+ * `useChat` from `@osdk/react/experimental/aip` against a Foundry LMS
+ * model. Consumers do not need to import `useChat` or `foundryModel`
+ * themselves — passing the platform client and a model API name is
+ * enough to render a working chat.
+ *
+ * Default rendering is feature-complete with no overrides supplied;
+ * render slots and `on*` listeners are layered on top of the built-in
+ * behavior.
+ */
+export interface AipAgentChatProps {
```
KuberSethi, line 74: Should we call it Aip? Also i don't think OSDK aware is accurate it's not really interacting with OSDK

PR #3154, opened 2026-05-01
Drop redundant `--config $(find-up dprint.json)` from lint scripts

tdeitch:
## Summary

- dprint auto-discovers its config by walking up from cwd looking for `dprint.json` — same behavior we were getting from `find-up`. The flag was redundant.
- The `--config $(find-up dprint.json)` got carried over from the prettier era in commit 0f7e99745 ("Apply dprint", Sep 2023) via what looks like a mechanical rewrite from `prettier . --check --ignore-path $(find-up .prettierignore)`. The leftover double-space in `dprint check  --config` is a fingerprint of that copy-paste.
- Updated `standardPackageRules` in `.monorepolint.config.mjs` and ran `mrl check --fix` to propagate to 75 package.json files. Manually fixed 5 packages excluded from the standard rules (`e2e.generated.1.1.x`, `react-devtools`, `react-components-storybook`, `typescript-sdk-docs-examples`, `osdk-docs-context`).
- Dropped the now-unused `find-up-cli` devDependency from the root. The `find-up` *library* (separate npm package) is still used by `babel.config.mjs` and various source files — left untouched.

## Surprise: the `format` step had been silently broken

While dropping `--config $(find-up dprint.json)` from `typescript-sdk-docs-examples`'s `format` script, CI exposed a latent bug. The `find-up` npm package has no CLI (it's a library), so `$(find-up dprint.json)` always expanded to empty and `npx dprint fmt --config` errored out silently — meaning the `format` step in the codegen chain was a no-op on main. Turbo cache hits on main meant codegen rarely re-ran, so this was never noticed. My package.json edits invalidated the cache, codegen ran, and dprint actually formatted for the first time — surfacing two issues:

1. **`uploadMediaOntologyEdits` template typecheck failure**: the template had `// @ts-ignore` above a single-line `batch.create(...)` with `myMediaProperty`. After dprint splits the object literal across multiple lines (per `lineWidth: 80`, `indentWidth: 2`), the `// @ts-ignore` ends up two lines above the property — too far to suppress the error. Fixed by hoisting the props to a `const` so `// @ts-ignore` sits directly above the `batch.create(Employee, props)` statement.
2. **119 example files re-indented from 4-space to 2-space**: dprint was finally able to apply the repo's config to the generated examples.

## Verification

- `dprint --help` confirms auto-discovery: *"Defaults to dprint.json(c) … in current or ancestor directory when not provided"*
- `dprint -L debug check` from `packages/api/` shows `[DEBUG] Running for config: /Volumes/git/github-com/osdk-ts/dprint.json`
- `pnpm lint` exit 0 in `packages/api`, `packages/cli`, `packages/react`
- `pnpm turbo typecheck lint --filter=@osdk/typescript-sdk-docs-examples --filter=@osdk/osdk-docs-context --filter=@osdk/typescript-sdk-docs --filter=@osdk/react-sdk-docs` all pass
- `dprint check` clean on all changed files
- `mrl check` clean

## Test plan

- [ ] CI lint job passes across all packages
- [ ] CI typecheck passes for `@osdk/typescript-sdk-docs-examples` (the originally-failing job)
- [ ] Spot-check `pnpm lint` in a few packages locally still produces identical output

🤖 Generated with [Claude Code](https://claude.ai/code)


bryantpark04 (approved):
I think this might have also been causing flakes with the examples failing typechecking, great find!

PR #3155, opened 2026-05-01
Remove dead Constant<Q> type from DerivedProperty.ts

tdeitch:
## Summary

Removes the orphaned `Constant<Q>` helper type from `packages/api/src/derivedProperties/DerivedProperty.ts`.

`Constant<Q>` defined `.constant.{double,integer,long,datetime,timestamp}` accessors on the derived-property `Builder`. It was disconnected from the `Builder extends ...` clause in #2858 ("Remove RDP Literals"), and the same PR removed the runtime implementation in `createWithPropertiesObjectSet.ts` and the related tests in `ObjectSet.test.ts`. The type itself was left orphaned in the file.

Verified before deleting:
- Non-exported (declared with `type`, not `export type`)
- Zero references in the monorepo besides its own definition
- Not present in `etc/api.report.api.md`

So this is a pure cleanup with no public-API impact.

## Test plan

- [x] `pnpm turbo typecheck --filter=@osdk/api` passes
- [x] `pnpm turbo check-api --filter=@osdk/api` passes (no API report change)
- [x] `pnpm turbo transpile --filter=@osdk/client` passes (the closest published consumer of `@osdk/api`)
- [x] `npx dprint check` clean

### DIFF
--- a/.changeset/remove-dead-constant-type.md
+++ b/.changeset/remove-dead-constant-type.md
@@ -1,0 +1,5 @@
+---
+"@osdk/api": patch
+---
+
+Remove the unused internal `Constant<Q>` type from `DerivedProperty.ts`. Disconnected from the `Builder` chain in #2858 (Remove RDP Literals); never exported, no consumers. No public-API change.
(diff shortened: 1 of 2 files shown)


tdeitch:
Reopening from upstream branch so the Bundle Size workflow can post its comment (fork PRs only get a read-only token). Same commit, new PR.

PR #3157, opened 2026-05-01
Add docs for @osdk/functions-testing.experimental

nihalbhatnagar:
## Summary
- New "Functions Testing" navbar section on the Docusaurus site
- Pages: getting started, mock objects & links, stubbing client calls, platform APIs with MSW

## Test plan
- [x] `pnpm --filter @osdk/docs build` passes (no new broken anchors)

### DIFF
--- a/docs/functions-testing/getting-started.md
+++ b/docs/functions-testing/getting-started.md
@@ -1,0 +1,111 @@
+---
+sidebar_position: 1
+---
+
+# Getting Started
+
+`@osdk/functions-testing.experimental` lets you unit-test code that takes an OSDK `Client` — including Foundry Functions — without hitting a real Foundry. It gives you:
+
+- **`createMockClient`** — a `Client` you stub with fluent `.when` / `.whenObjectSet` / `.whenQuery` matchers.
+- **`createMockOsdkObject`** — construct fully-shaped `Osdk.Instance` values (`$primaryKey`, `$title`, `$rid`, `$link`, `$clone`).
+- **`createMockObjectSet`** — a standalone `ObjectSet` you can pass anywhere a real one would go (or use as a many-link target for aggregates).
+- **`createMockAttachment`** — a stand-in for attachment values.
+
+:::warning Experimental
+The package is published as `@osdk/functions-testing.experimental` and only exports through the `/experimental` subpath. The surface may evolve before promotion to a stable name.
+:::
+
+## Install
+
+```bash
+npm install --save-dev @osdk/functions-testing.experimental
+```
+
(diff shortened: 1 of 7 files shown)


KuberSethi (commented):
initial claude driven review
**critical**

1. `mock-objects.md` `createMockAttachment` example: `createMockAttachment` takes positional args `(rid, blob?, metadata?)`, not an options object, and there's no `filename` field on it. this snippet won't compile. `filename` lives on the third arg (the metadata).

2. `getting-started.md` experimental warning: the package also exports from the root, not only `/experimental`. either soften this to "we recommend `/experimental` because the surface may move" or drop the root export from the package.

**high**

3. `stubbing-client-calls.md` `thenReturnObjects` claim: stubs match by full call chain, so a `.fetchPage()` stub doesn't serve `.asyncIter()` calls. worth rewording so readers know they need to stub each terminal shape they actually call.

**important**

4. changeset: this PR doesn't touch any published source, so the changeset will write a release note for `@osdk/functions-testing.experimental` saying "documentation for functions testing library" without any actual code change. probably worth dropping (precedent: the recent docs-only react PR didn't have one either).

**simplifications (optional, ship now or follow up)**

5. `platform-apis-with-msw.md` example: the two `it` blocks are basically the same shape with different usernames. one happy path is enough to make the point, and trimming the fixture to lean on `satisfies User` makes the load bearing parts (mock url, `setupServer`, `onUnhandledRequest`) easier to spot.

6. `MockObjectSet` defined twice: the `whenObjectSet` example shows up in both `mock-objects.md` and `stubbing-client-calls.md`. probably cleaner to make the stubbing page canonical and have the links page just point at it, otherwise the two will drift.

docs/functions-testing/getting-started.md
```diff
@@ -0,0 +1,111 @@
+---
+sidebar_position: 1
+---
+
+# Getting Started
+
+`@osdk/functions-testing.experimental` lets you unit-test code that takes an OSDK `Client` — including Foundry Functions — without hitting a real Foundry. It gives you:
```
KuberSethi, line 7: nit: this is super LLM-y, up to you if you care though

docs/functions-testing/platform-apis-with-msw.md
```diff
@@ -0,0 +1,156 @@
...
+
+    const mockClient = createMockClient();
+    await expect(requireAdminUser(mockClient)).rejects.toThrow(
+      "User bob@example.com is not an admin",
+    );
+  });
+});
+```
+
+:::tip Type your fixtures against the real platform type
```
KuberSethi, line 128: nice this is actually helpful

PR #3158, opened 2026-05-01
Fill in missing JSDoc across @osdk/api and @osdk/client public surface

tdeitch:
## Summary

Fills in missing `@param`, `@example`, and `@returns` JSDoc across the public surface of `@osdk/api` and `@osdk/client`, so generated API docs and IDE hovers have something useful to show.

Touched surfaces:
- `ObjectSet` methods (`asyncIter`, `withProperties`, `pivotTo`, `fetchOne` / `fetchOneWithErrors`, `subscribe`, `narrowToType`, `experimental_asyncIterLinks`, `nearestNeighbors` indentation fix).
- Derived-property `Builder` chain (`where`, `pivotTo`, `aggregate`, `selectProperty`) and `NumericExpressions` / `DatetimeExpressions` chains.
- `Logger` interface (level fields, `isLevelEnabled`, `child`) and `Logger.LogFn`.
- `Attachment` and `Media` accessors.
- `TimeSeriesProperty.getFirstPoint` / `getLastPoint` and `GeotimeSeriesProperty.getLatestValue`.
- Top-level `Client` callable overloads and `Client.fetchMetadata`.
- `createClient` and `createClientWithTransaction` factories — the `createClient` example uses `createPublicOauthClient` from `@osdk/oauth`; `createClientWithTransaction` uses `createConfidentialOauthClient` since it's server-side only.

API report files (`etc/api.report.api.md`, `etc/client.report.api.md`) regenerated to reflect the new docs.

No runtime or type changes.

## Test plan

- [x] `pnpm turbo typecheck --filter=@osdk/api --filter=@osdk/client`
- [x] `pnpm turbo check-api --filter=@osdk/api --filter=@osdk/client`
- [x] `dprint check` (via lint-staged on commit)

### DIFF
--- a/packages/api/src/derivedProperties/DerivedProperty.ts
+++ b/packages/api/src/derivedProperties/DerivedProperty.ts
@@ -194,8 +194,8 @@ type Aggregatable<
    * Aggregates over the builder's object set to produce a derived property value.
    * @param aggregationSpecifier - Either `"$count"` or a `"<propertyName>:<aggregation>"` key
    *   (e.g., `"salary:max"`, `"salary:approximatePercentile"`, `"name:collectList"`)
-   * @param opts - Required for `collectList` / `collectSet` (`{ limit }`) and
-   *   `approximatePercentile` (`{ percentile }`); not accepted for other aggregations
+   * @param opts - For `collectList` / `collectSet` accepts `{ limit }`; for
+   *   `approximatePercentile` accepts `{ percentile }`. Not accepted for other aggregations.
    * @example
    * ```ts
    * client(Employee).withProperties({
--- a/packages/api/src/derivedProperties/DerivedProperty.ts
+++ b/packages/api/src/derivedProperties/DerivedProperty.ts
@@ -106,6 +106,20 @@ type Filterable<
   Q extends ObjectOrInterfaceDefinition,
   CONSTRAINED extends boolean,
 > = {
+  /**
+   * Filters the builder's object set with a where clause before further chaining.
+   * @param clause - A filter clause applied to the builder's current object set
+   * @example
+   * ```ts
+   * client(Employee).withProperties({
+   *   activeReportCount: (baseObjectSet) =>
+   *     baseObjectSet.pivotTo("reports")
+   *       .where({ status: { $eq: "active" } })
+   *       .aggregate("$count"),
+   * }).fetchPage();
+   * ```
+   * @returns the builder, narrowed to objects matching the clause
+   */
   readonly where: (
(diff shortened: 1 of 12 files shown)


packages/api/src/object/Media.ts
```diff
@@ -17,20 +17,49 @@
...
    * Returns the source location of this media (object type, primary key, property name).
    *
    * Optional because not all media has a source location (e.g., transient/uploaded media).
+   * @example
+   * ```ts
+   * const equipment = await client(Equipment).fetchOne(12345);
+   * const location = equipment.trainingMaterial?.getMediaSourceLocation?.();
+   * console.log(location?.objectType, location?.primaryKey, location?.propertyName);
+   * ```
+   * @returns the `MediaPropertyLocation` identifying which object property this media came from
```
ssanjay1: nit: can we leave docs out for this right now, or indicated its not implemented yet (we haven't piped this through yet was added by react folks)

packages/client/src/createClient.ts
```diff
@@ -373,6 +407,40 @@ export const createClient: (
   undefined,
 );
 
+/**
+ * Creates a {@link Client} that scopes all of its requests to a Foundry transaction. Used by transactional
```
ssanjay1: @nihalbhatnagar do you think documenting this would be confusing considering people can't make their own transactions yet outside of functions?

nihalbhatnagar: Yeah nobody should be using this in code, so maybe can hold off on documentation here

packages/api/src/derivedProperties/DerivedProperty.ts
```diff
@@ -106,6 +106,20 @@ type Filterable<
   Q extends ObjectOrInterfaceDefinition,
   CONSTRAINED extends boolean,
 > = {
+  /**
+   * Filters the builder's object set with a where clause before further chaining.
```
nihalbhatnagar: Should we explain what filtering is?

tdeitch: Updated, is this what you had in mind?

packages/api/src/object/Media.ts
tdeitch: Removed docs (there were existing docs but if this shouldn't be used by end users I'll remove them).

packages/client/src/createClient.ts
tdeitch: Updated.

nihalbhatnagar (approved):

PR #3159, opened 2026-05-04
Fix Action Form popovers in dialogs

nfdiop:
## Summary
- Thread Action Form portal containers through fields rendered inside Blueprint dialogs.
- Add a shared portal dismiss layer for date pickers and dropdowns so internal dialog clicks close popovers without reopening them.
- Extend the dialog Storybook story to cover date range picker behavior.

## Verification
- pnpm --filter @osdk/react-components test -- src/action-form/__tests__/DropdownField.test.tsx src/action-form/__tests__/DatetimePickerField.test.tsx src/action-form/__tests__/DateRangeInputField.test.tsx src/action-form/__tests__/BaseForm.test.tsx
- pnpm --filter @osdk/react-components lint
- pnpm --filter @osdk/react-components transpileBrowser
- Storybook and Playwright checks for dialog popup dismissal

### DIFF
--- a/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
@@ -1036,7 +1036,7 @@ const blueprintDialogFormContent: ReadonlyArray<FormContentItem> = [
 ];
 
 function BlueprintDialogBaseForm(): React.ReactElement {
-  const [isOpen, setIsOpen] = useState(true);
+  const [isOpen, setIsOpen] = useState(false);
   const portalContainerRef = useRef<HTMLDivElement>(null);
 
   const handleOpen = useCallback(() => {
--- a/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { Button, Dialog } from "@blueprintjs/core";
 import type { ObjectSet, ObjectTypeDefinition } from "@osdk/api";
 import type {
   FormContentItem,
@@ -22,7 +23,7 @@ import type {
 import { BaseForm } from "@osdk/react-components/experimental";
 import { useOsdkClient } from "@osdk/react/experimental";
 import type { Meta, StoryObj } from "@storybook/react-vite";
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useMemo, useRef, useState } from "react";
 import { fn } from "storybook/test";
 import { fauxFoundry } from "../../mocks/fauxFoundry.js";
 import { Employee } from "../../types/Employee.js";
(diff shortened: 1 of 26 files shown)


packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
nfdiop, line 1: On the fence about this. But I decided to add it so we can easily test subsequent PRs

packages/react-components/src/action-form/__tests__/DateRangeInputField.test.tsx
```diff
@@ -392,9 +434,34 @@ describe("DateRangeInputField", () => {
...
       fireEvent.focus(endSentinel, { relatedTarget: dialog });
 
-      expect(document.activeElement).not.toBe(startInput);
-      const endInput = screen.getByLabelText("End date");
-      expect(document.activeElement).not.toBe(endInput);
+      expect(document.activeElement).toBe(startInput);
+      expect(screen.queryByRole("dialog")).toBeNull();
+    });
+
+    it("does not reopen the popover on the next Tab after boundary exit", () => {
```
nfdiop, line 441: fix for a bug

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -26,6 +26,12 @@ import type {
 } from "@osdk/api";
 import type React from "react";
 
+export type PortalContainer =
+  | HTMLElement
+  | ShadowRoot
```
KuberSethi, line 31: Do we also want to export this in `action-form.ts`?

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
```diff
@@ -202,6 +223,20 @@ function renderFieldComponent(
   }
 }
 
+function resolvePortalContainer(
+  fieldComponentProps: { portalContainer?: PortalContainer },
+  formPortalContainer: PortalContainer | undefined,
+): PortalContainer | undefined {
+  // A field-level value, including null, is an explicit override. That lets a
+  // single field opt out of a form-level dialog portal when needed.
+  return Object.prototype.hasOwnProperty.call(
```
KuberSethi: Is the `hasOwnProperty` usage required / are there no other workarounds? Asking bc my understanding is this is generally an antipattern both in terms of safety and performance. I believe `hasOwn` is more recommended: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn#description

packages/react-components/src/action-form/fields/DropdownField.module.css
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.osdkSelectDismissLayer {
+  --osdk-portal-dismiss-layer-z-index: var(
+    --osdk-select-z-index,
```
KuberSethi, line 19: I'm not sure this is defined anywhere but perhaps I missed it, calling out in case

packages/react-components/src/action-form/fields/DropdownField.module.css
```diff
@@ -0,0 +1,29 @@
...
+    var(--osdk-surface-z-index-3)
+  );
+}
+
+.osdkComboboxDismissLayer {
+  --osdk-portal-dismiss-layer-z-index: var(
+    --osdk-combobox-z-index,
```
KuberSethi, line 26: I'm not sure this is defined anywhere but perhaps I missed it, calling out in case

packages/react-components/src/action-form/FormFieldApi.ts
KuberSethi, line 31: Not that familiar with Shadow DOM best practices - what's the reason for using this / what benefits do we derive?

KuberSethi:
Also flagging that search in the blueprint story doesn't seem to work, nothing gets filtered. Also I saw these errors when profiling w React Devtools, might be worth looking at

```
BaseForm.stories.tsx:1039 Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
BlueprintDialogBaseForm	@	BaseForm.stories.tsx:1039
DetermineComponentFrameRoot	@	installHook.js:1
describeNativeComponentFrame	@	installHook.js:1
describeFunctionComponentFrame	@	installHook.js:1
describeFiber	@	installHook.js:1
(anonymous)	@	installHook.js:1
(anonymous)	@	installHook.js:1
toggleProfilingStatus	@	installHook.js:1
stopProfiling	@	installHook.js:1
(anonymous)	@	react_devtools_backend_compact.js:1
emit	@	react_devtools_backend_compact.js:1
(anonymous)	@	react_devtools_backend_compact.js:1
listener	@	backendManager.js:1
```

KuberSethi (commented):
Mostly looks good, just a few things

packages/react-components/src/action-form/FormFieldApi.ts
nfdiop: Done — exported in the next push.

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
nfdiop: Nice catch, switched to `Object.hasOwn` in the next push.

packages/react-components/src/action-form/FormFieldApi.ts
nfdiop: ShadowRoot is in the union because when a form renders inside a Shadow DOM, the portal container must be the shadow root — document.body is outside the shadow boundary, so portaled content (popovers, dropdowns) would escape the encapsulation and miss scoped styles. Including it in the type makes the API ready for that use case without a separate prop.

packages/react-components/src/action-form/fields/DropdownField.module.css
nfdiop: These are intentional optional consumer-override tokens — they're not declared by us but can be set by a consumer on an ancestor element to override the z-index. The fallback `var(--osdk-surface-z-index-3)` provides the default. Same pattern already used in `Select.module.css` and `Combobox.module.css`.

packages/react-components/src/action-form/fields/DropdownField.module.css
nfdiop: Same as above — intentional optional override token with a fallback.

packages/react-components/src/action-form/FormFieldApi.ts
KuberSethi: This makes sense, thanks for explaining!

nfdiop:
> Also flagging that search in the blueprint story doesn't seem to work, nothing gets filtered. Also I saw these errors when profiling w React Devtools, might be worth looking at
> 
> ```
> BaseForm.stories.tsx:1039 Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app
> See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
> BlueprintDialogBaseForm	@	BaseForm.stories.tsx:1039
> DetermineComponentFrameRoot	@	installHook.js:1
> describeNativeComponentFrame	@	installHook.js:1
> describeFunctionComponentFrame	@	installHook.js:1
> describeFiber	@	installHook.js:1
> (anonymous)	@	installHook.js:1
> (anonymous)	@	installHook.js:1
> toggleProfilingStatus	@	installHook.js:1
> stopProfiling	@	installHook.js:1
> (anonymous)	@	react_devtools_backend_compact.js:1
> emit	@	react_devtools_backend_compact.js:1
> (anonymous)	@	react_devtools_backend_compact.js:1
> listener	@	backendManager.js:1
> ```

Thanks for flagging for the search: fixed! 
For the react warnings, I'm not able to repro

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -337,7 +361,9 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
                 <Combobox.Empty>No results</Combobox.Empty>
               )}
               <Combobox.List>
-                {items.map(renderItem)}
+                <Combobox.Collection>
```
nfdiop, line 364: This was the fix for search not working

KuberSethi (approved):

PR #3164, opened 2026-05-05
downgrade react-ga changeset from major to minor

KuberSethi:
react-ga changeset declared @osdk/react as major, which combined with @osdk/react being in the @osdk/client fixed group dragged the whole group to 3.0.0 for future releases

• change @osdk/react bump in .changeset/react-ga.md from major to minor so the fixed group lands at 2.15.0 instead of 3.0.0
• update stale version references in the changeset body (^1.0.0 / ^2.14.0 / @osdk/react@1.0) to ^2.15.0 / @osdk/react@2.15

we will unpublish the errant 3.0.0 as well

### DIFF
--- a/.changeset/react-ga.md
+++ b/.changeset/react-ga.md
@@ -1,5 +1,5 @@
 ---
-"@osdk/react": major
+"@osdk/react": minor
 "@osdk/client": minor
 "@osdk/react-components": patch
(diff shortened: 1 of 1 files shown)


bryantpark04 (approved):
by 
> we will unpublish the errant 3.0.0 as well

do you mean something got published at 3.0.0?

PR #3170, opened 2026-05-05
[Components] Add onFilterVisibilityChange

ziyunp:
add `onFilterVisibilityChange` callback to `FilterList` that fires when filters are reordered, added, or removed, returning all filters in display order with their visibility state.


packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -76,9 +78,35 @@ export function FilterList<Q extends ObjectTypeDefinition>(
...
+    (
+      visibleKeys: string[],
+      hiddenDefs: Array<FilterDefinitionUnion<Q>>,
+    ) => {
+      if (!onFilterVisibilityChange) {
+        return;
+      }
+      const states: Array<{ filterKey: FilterKey<Q>; isVisible: boolean }> = [];
+      for (const key of visibleKeys) {
+        states.push({ filterKey: key as FilterKey<Q>, isVisible: true });
```
KuberSethi: do we need the `as FilterKey` here and below?

packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -100,18 +128,60 @@ export function FilterList<Q extends ObjectTypeDefinition>(
...
+      fireOnFilterVisibilityChange,
+    ],
   );
 
   const handleFilterShown = useCallback(
     (filterKey: string) => {
       showFilter(filterKey);
+      const currentOrder = orderedVisibleKeysRef.current
+        ?? managedVisibleDefinitions.map(getFilterKey);
+      const nextOrder = [...currentOrder, filterKey];
```
KuberSethi: is this fine? we're always appending this to the end and i think this might cause jumps/jitter?

packages/react-components/src/filter-list/__tests__/FilterListDragAndDrop.test.tsx
```diff
@@ -197,4 +197,27 @@ describe("FilterList drag and drop", () => {
     const dragHandles = screen.queryAllByLabelText(/Reorder/);
     expect(dragHandles).toHaveLength(0);
   });
+
+  it("does not fire onOrderChange when no drag occurs", async () => {
```
KuberSethi, line 201: Could we add more tests to cover the behavior this PR adds?

packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -76,9 +78,35 @@ export function FilterList<Q extends ObjectTypeDefinition>(
...
+      }
+      onFilterVisibilityChange(states);
+    },
+    [onFilterVisibilityChange],
+  );
+
   const handleReset = useCallback(() => {
     reset();
     resetVisibility();
+    orderedVisibleKeysRef.current = null;
```
KuberSethi: Should we also reset the `dragOrder`?

packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -76,9 +78,35 @@ export function FilterList<Q extends ObjectTypeDefinition>(
     resetVisibility,
   } = useFilterVisibility(filterDefinitions, getFilterKey, getIsVisible);
 
+  const orderedVisibleKeysRef = useRef<string[] | null>(null);
```
KuberSethi: Hmm I'm not quite sure we really need a ref here, could we instead lift dragOrder from FilterListContent into FilterList or useFilterVisibility and consolidate things there? I think we'd avoid the need for the useRef and this would reduce the complexity of this code a lot / help us avoid the class of problems useRef tends to bring along

packages/react-components/src/filter-list/FilterList.tsx
ziyunp: This is on adding filter, and it's always added to the end

packages/react-components/src/filter-list/FilterList.tsx
ziyunp: It's needed because `getFilterKey` returns `string`.

packages/react-components/src/filter-list/__tests__/FilterListDragAndDrop.test.tsx
ziyunp: more tests added in useFilterVisibility

packages/react-components/src/filter-list/FilterList.tsx
ziyunp: I will push a commit to refactor this.

packages/react-components/src/filter-list/FilterList.tsx
ziyunp: Good catch, fixed it but will also refactor the handlers for filter visibility

packages/react-components/src/filter-list/FilterList.tsx
ziyunp: I noticed a bug (?) where the filter is added to its original position in the filter definition. I'm fixing it so that it's always added to the end, as that is more intuitive and also follows workshop Filter List behaviour

packages/react-components/src/filter-list/hooks/useFilterVisibility.ts
```diff
@@ -14,79 +14,161 @@
...
+        .map(getFilterKey) ?? [],
+    [filterDefinitions],
+  );
+
+  const [visibleKeyOrder, setVisibleKeyOrder] = useState<string[]>(
+    () => defaultVisibleKeyOrder,
+  );
+
+  // Sync state when filterDefinitions changes
+  useEffect(() => {
```
KuberSethi, line 56: I don't love this pattern here but I think it's fine for now, just calling out in case we decide to refactor later (I think we could probably derive everything from props + an override value)

packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -64,49 +64,42 @@ export function FilterList<Q extends ObjectTypeDefinition>(
...
-    (def: FilterDefinitionUnion<Q>) => def.isVisible !== false,
-    [],
+  const handleVisibilityChange = useCallback(
+    (visibleKeys: string[], hiddenKeys: string[]) => {
+      if (!onFilterVisibilityChange) {
+        return;
+      }
+      const states: Array<{ filterKey: FilterKey<Q>; isVisible: boolean }> = [
+        ...visibleKeys.map((key) => ({
+          filterKey: key as FilterKey<Q>,
```
KuberSethi, line 74: do we need the casts here and below? if the hook is parameterized to return FilterKey instead of string, then the casts go away. not blocking but worth considering.

packages/react-components/src/filter-list/hooks/useFilterVisibility.ts
```diff
@@ -14,79 +14,161 @@
...
+  }, [filterDefinitions, defByKey, visibleKeyOrder, allKeys]);
 
   const showFilter = useCallback((key: string) => {
-    setVisibilityOverrides((prev) => {
-      const next = new Map(prev);
-      next.set(key, true);
+    setVisibleKeyOrder((prev) => {
+      if (prev.includes(key)) return prev;
+      const next = [...prev, key];
+      fireVisibilityChange(next);
```
KuberSethi, line 121: hmm i think this technically violates the react rule that setStates updaters are supposed to be pure? i think this is fine though, just might result in some dev mode weirdness when things render twice

packages/react-components/src/filter-list/hooks/useFilterVisibility.ts
```diff
@@ -14,79 +14,161 @@
...
+    setVisibleKeyOrder(defaultVisibleKeyOrder);
+  }, [defaultVisibleKeyOrder]);
+
+  // Keep a ref to onVisibilityChange so callbacks don't need it as a dependency
+  const onVisibilityChangeRef = useRef(onVisibilityChange);
+  onVisibilityChangeRef.current = onVisibilityChange;
+
+  const fireVisibilityChange = useCallback(
+    (nextVisibleKeyOrder: string[]) => {
+      if (!onVisibilityChangeRef.current) return;
```
KuberSethi, line 66: nit: here and elsewhere i think we prefer if statements with braces

KuberSethi (approved):

packages/react-components/src/filter-list/hooks/useFilterVisibility.ts
ziyunp: I'll look into this and do a FLUP. Another way is to compute the next state outside of setState and call the callback with the next state.

PR #3173, opened 2026-05-05
make react-devtools package public

KuberSethi:
make @osdk/react-devtools publishable on npm so it can move through the release pipeline

the package's vite-based build did not fit any standard archetype because it imports SCSS modules with sass features (@use, @mixin, namespaced partials). this PR replaces the vite build with monorepo.tool.transpile + a SCSS-aware build-css.mjs that bundles styles into a runtime-injection module, so consumers get correct styling with no manual CSS imports

• drop private flag, add publishConfig, move to a new esmReactLibraryWithCss archetype
• new scripts/build-css.mjs compiles .module.scss via sass-embedded, scopes class names with postcss-modules, emits .module.scss.js proxies, bundles blueprint core+icons CSS, writes build/esm/styles.css and a runtime-injection build/esm/styles.js
• src/public/vite.ts re-exports the vite plugin so ./vite is auto-discovered; mrl ourExportsConvention now resolves cssExport under build/esm for ESM-only packages


packages/react-devtools/vitest.config.mts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ * Copyright 2023 Palantir Technologies, Inc. All rights reserved.
```
ssanjay1: nit: wrong year

ssanjay1 (dismissed):

ssanjay1 (approved):

PR #3176, opened 2026-05-06
filter-list: NoValueLabel + listogram dedupe

KuberSethi:
## Summary
• new `NoValueLabel` component + `isEmptyValue` helper, used by listogram, single/multi select, text tags, and null wrapper for a consistent italic-muted no-value placeholder
• legacy `--osdk-filter-listogram-empty-label-*` and `--osdk-filter-null-label-color` tokens forward onto the canonical `--osdk-filter-no-value-*` tokens for back-compat
• listogram dedupes empty-ish aggregation rows (e.g. `""` and `null`) into a single placeholder with summed count

## Test plan
- [ ] `pnpm turbo typecheck --filter=@osdk/react-components`
- [ ] `pnpm vitest run src/filter-list/` — 113 tests pass
- [ ] storybook → Experimental / FilterList / Default — empty department renders as a single italic muted "No value" row


packages/react-components/src/filter-list/base/inputs/MultiSelectInput.tsx
```diff
@@ -75,34 +77,44 @@ function MultiSelectInputInner({
   );
 
   const renderItem = useCallback(
```
ziyunp, line 79: Suggestion: As a FLUP, we can extract the same renderItem logic in SingleSelect and MultiSelect in one place

packages/react-components/src/filter-list/base/inputs/ListogramInput.tsx
```diff
@@ -64,7 +68,33 @@ function ListogramInputInner({
 }: ListogramInputProps): React.ReactElement {
   const [isExpanded, setIsExpanded] = useState(false);
 
-  const stableValues = useStableData(values, isLoading);
+  // Aggregations may return both "" and null as separate rows; merge them
+  // into a single "No value" placeholder at the position of the first match.
```
ziyunp, line 72: I'm concerned we will introduce bugs by merging them because technically empty string and null are different values. But let's get this in and address if it's actually causing issues

ziyunp (approved):

PR #3177, opened 2026-05-06
filter-list: stop refetching aggregations when any filter changes

KuberSethi:
selecting a value in one filter was refetching every other filter's aggregation because per-filter where-clause references churned on every render

• stabilize per-key where-clause refs in useFilterListState with a deep-equal cache
• add useStableMapEntries helper for per-entry structural memoization
• unchanged filters keep their previous clause reference and skip refetching


bryantpark04 (approved):
unblocking

PR #3178, opened 2026-05-06
combobox: cap popup max-height at 320px

KuberSethi:
## Summary
• long option lists were pushing other UI off-screen because the popup grew to fill available viewport space
• combobox popup max-height now resolves to \`min(var(--osdk-combobox-popup-max-height, 320px), var(--available-height))\` so it stays viewport-aware
• default cap is themable via the new \`--osdk-combobox-popup-max-height\` token in tokens/component-tokens/combobox.css

## Test plan
- [ ] \`pnpm turbo typecheck --filter=@osdk/react-components\`
- [ ] storybook → any combobox with > ~10 options — popup caps at 320px and scrolls inside
- [ ] computed style on the popup shows max-height: 320px (or smaller if the viewport is shorter)

### DIFF
--- a/packages/react-components/src/base-components/combobox/Combobox.module.css
+++ b/packages/react-components/src/base-components/combobox/Combobox.module.css
@@ -303,7 +303,10 @@
   display: flex;
   flex-direction: column;
   min-width: var(--anchor-width);
-  max-height: var(--available-height);
+  max-height: min(
+    var(--osdk-combobox-popup-max-height, 320px),
+    var(--available-height)
+  );
   overflow-y: auto;
   padding: calc(
     var(--osdk-combobox-spacing, var(--osdk-surface-spacing)) * 1.5
--- a/packages/react-components/src/base-components/combobox/Combobox.module.css
+++ b/packages/react-components/src/base-components/combobox/Combobox.module.css
@@ -304,7 +304,7 @@
   flex-direction: column;
   min-width: var(--anchor-width);
   max-height: min(
-    var(--osdk-combobox-popup-max-height, 320px),
+    var(--osdk-combobox-popup-max-height),
     var(--available-height)
   );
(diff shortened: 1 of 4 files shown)


packages/react-components/src/base-components/combobox/Combobox.module.css
```diff
@@ -303,7 +303,10 @@
   display: flex;
   flex-direction: column;
   min-width: var(--anchor-width);
-  max-height: var(--available-height);
+  max-height: min(
+    var(--osdk-combobox-popup-max-height, 320px),
```
ziyunp: Nit: Remove the default here since you defined it in combobox.css

packages/react-components/src/tokens/component-tokens/combobox.css
```diff
@@ -0,0 +1,8 @@
+:root {
+  /* Combobox popup: cap the dropdown's max-height so very long option lists
+     scroll inside the popup instead of forcing the popup to grow tall enough
+     to push other UI off-screen. The actual `max-height` rule uses
+     `min(this token, --available-height)` so the cap is automatically lowered
```
ziyunp: Where is --available-height defined?

ziyunp (dismissed):

packages/react-components/src/tokens/component-tokens/combobox.css
KuberSethi: `--available-height` is set automatically by base ui's `Combobox.Positioner`, we do the same for `--anchor-width` a few lines above for `min-width`.

packages/react-components/src/base-components/combobox/Combobox.module.css
KuberSethi: done

ziyunp (approved):

PR #3179, opened 2026-05-06
filter-list: shared date pickers, svg histogram, drag-to-select range

KuberSethi:
## Summary
extract action-form's date input fields to shared/calendar so filter-list, action-form, and object-table render the same popover calendar; replace div-based histograms with svg; add drag-to-select range with pointer events. supersedes #3196.

• extract DatetimePickerField/DateRangeInputField from action-form/fields to shared/calendar as DatePicker/DateRangePicker; filter-list date filters and the histogram From/To inputs use the shared popover so cross-locale viewers see the same YYYY-MM-DD instead of the browser-locale native picker
• replace div-based date and number range histograms with SVG (axis grid lines, count labels above each bar, ISO-consistent tick labels, period subtitle); new formatDate per-property callback overrides every display surface — input idle text, tooltip, subtitle, x-tick labels, and chip text — without changing the underlying ISO value
• new clickToFilter flag on DATE_RANGE / NUMBER_RANGE filters: pointerdown on a bar starts a drag, pointerup commits the bucket-aligned range; click inside an existing selection clears the filter; built on pointer events with setPointerCapture for touch + stylus, pointercancel for release-outside-window, and a click-vs-drag distance threshold so a micro-drag inside one bucket doesn't unintentionally clear

parseDate per-property callback removed; the shared pickers own input parsing.

## Test plan
- [ ] \`pnpm turbo typecheck --filter=@osdk/react-components\` passes
- [ ] \`pnpm turbo typecheck --filter=@osdk/react-components-storybook\` passes
- [ ] \`pnpm turbo test --filter=@osdk/react-components\` — 859 tests pass
- [ ] storybook → Experimental / FilterList / With All Filter Types — date histogram From/To open the shared popover calendar (no native browser picker), x-axis shows numeric months (01/02/...), subtitle shows ISO format (2020-05)
- [ ] storybook drag from bar A to bar D commits a range covering both; click inside selection clears
- [ ] storybook → Action Form / Datetime + Date Range fields still render correctly via the moved DatePicker / DateRangePicker


packages/react-components/src/filter-list/inputs/SingleDateFilterInput.tsx
```diff
@@ -21,6 +21,13 @@ import type { FilterState } from "../FilterListItemApi.js";
 interface SingleDateFilterInputProps {
   filterState: FilterState | undefined;
   onFilterStateChanged: (state: FilterState) => void;
+  /**
+   * Plumbed for API parity with the other date filter inputs. SingleDateInput
+   * renders only the HTML `<input type="date">` whose value attribute must be
+   * ISO `YYYY-MM-DD`, so `formatDate` has no display surface in v1.
+   */
+  formatDate?: (date: Date) => string;
+  parseDate?: (text: string) => Date | undefined;
```
ziyunp: issue (blocking): `formatDate` and `parseDate` are accepted as props here but never destructured or used in the body — the `SingleDateInput` below renders without them. The doc comment explains the intent ("plumbed for API parity"), but a caller passing `formatDate` to a `SINGLE_DATE` filter would reasonably expect *something* to use it. Either drop the props until v2 needs them, or actually plumb them through to `SingleDateInput` (e.g. for a custom-display label). Accepting unused props is a typecheck-clean footgun.

KuberSethi: Done, good callout

packages/react-components/src/filter-list/base/inputs/RangeInput.tsx
```diff
@@ -197,75 +623,364 @@ function RangeInputInner<T>({
...
-                  bucket.min,
-                  bucket.max,
-                  bucket.count,
-                )}
-              />
-            );
-          })}
-        </div>
+      {hasActiveFilter && (
+        <button
```
nfdiop: BaseUI button?

packages/react-components/src/filter-list/base/inputs/RangeInput.tsx
nfdiop, line 1: suggestion (FLUP): maybe this file could be shorter?

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -26,11 +26,13 @@ import type {
 } from "@osdk/api";
 import type React from "react";
 
-export type PortalContainer =
-  | HTMLElement
-  | ShadowRoot
-  | null
-  | React.RefObject<HTMLElement | ShadowRoot | null>;
+import type {
```
nfdiop, line 29: Can be a FLUP but this file only imports from types files not from files with components

packages/react-components/src/shared/calendar/DateRangePicker.tsx
```diff
@@ -24,16 +24,87 @@ import {
...
+  /** Parses a user-typed string back into a Date. */
+  parseDate?: (text: string) => Date | undefined;
+
+  /**
+   * Element that receives the date range picker portal. Use this when
+   * rendering inside modal dialogs so popovers stay in the dialog's
+   * stacking and focus context instead of being appended directly to
+   * document.body.
+   */
+  portalContainer?: PortalContainer;
```
nfdiop, line 105: question: do you need modal here too?

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
```diff
@@ -0,0 +1,141 @@
...
+  loFallback: string,
+  hiFallback: string,
+): string {
+  return `${min != null ? format(min) : loFallback} – ${
+    max != null ? format(max) : hiFallback
+  }`;
+}
+
+/** Short summary of a filter's current value for rendering inside a `FilterPopover` trigger. */
+export function summarizeFilterValue<Q extends ObjectTypeDefinition>(
```
nfdiop, line 74: suggestion (FLUP): maybe add tests

nfdiop (commented):
DatePicker in form seems to work like before. 
Some comments after testing:
- The inputs in the horizontal bar look very small. When the clear button is visible on the input, it creates some shift. Can we use the same input styles as in the rest of the codebase? 
- The range input for the date looks a bit off with the month as the x-axis. 
- Maybe also use compact numbers for the listogram

Image

Image

Image

packages/react-components/src/filter-list/base/inputs/RangeInput.tsx
KuberSethi: will address in a FLUP

packages/react-components/src/action-form/FormFieldApi.ts
KuberSethi: will address in a FLUP

packages/react-components/src/shared/calendar/DateRangePicker.tsx
KuberSethi: yeah good call, mirrored DatePicker's modal prop so callers can opt out of trap-focus when nesting in another popover

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
KuberSethi: will add in a FLUP!

nfdiop (dismissed):

nfdiop (approved):

PR #3186, opened 2026-05-06
Object Table FRs

ziyunp:
Feature 1: Make the editable column prop accept both boolean and function form (rowData: Osdk.Instance<...>) => boolean to enable conditional per-row editability
Feature 2: Add a getRowAttributes prop to ObjectTableProps that accepts (rowData) => Record<string, string | undefined> to allow setting arbitrary data attributes on table rows for conditional CSS styling
Feature 3: Convert editFieldConfig.fieldComponentProps (static object) to editFieldConfig.getFieldComponentProps(rowData) (function) so editor configuration can vary per row based on the row's data
Feature 4: Add a showEditFooter: boolean prop to hide the bottom edit footer UI 

https://github.com/user-attachments/assets/74450c37-05d3-40c9-977e-3a7fbe47851f

Screenshot 2026-05-06 at 18 48 16
Screenshot 2026-05-06 at 18 48 05
Screenshot 2026-05-06 at 18 47 07
Screenshot 2026-05-06 at 16 44 54


packages/react-components/docs/ObjectTable.md
```diff
@@ -137,20 +137,22 @@ Each column header has a menu with items for sorting, filtering, pinning, resizi
...
+| `editMode`           | `"always" \| "manual"`                        | Controls edit mode behavior. "always": Table is always in edit mode. "manual": User toggles edit mode on/off. Default: "manual"                                                                                        |
+| `onCellValueChanged` | `(info: CellEditInfo) => void`                | Called when a cell value is edited. The info object contains rowId, columnId, newValue, oldValue, and originalRowData                                                                                                  |
+| `onSubmitEdits`      | `(edits: CellEditInfo[]) => Promise<boolean>` | When provided, shows a "Submit Edits" button in the edit footer. When `editMode` is `"manual"`, the edit footer is automatically shown regardless of `showEditFooter`. Return true on success                          |
+| `showEditFooter`     | `boolean`                                     | Whether to render the bottom edit footer (Edit Table / Cancel / Submit Edits). Defaults to `true`. **Ignored when `editMode` is `"manual"` or `onSubmitEdits` is provided — the footer is always shown in that case.** |
```
KuberSethi: I thought `editMode === "manual" and onSubmitEdits != null` are what govern the footers rendering?

packages/react-components/src/object-table/utils/editableUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Whether a column declares any editing behavior.
+ *
+ * Returns `true` when `editable` is `true` or a predicate function. The
+ * predicate may still return `false` for individual rows — that decision is
+ * made per-cell via {@link isCellEditable}. This function answers the
+ * column-level question used to drive table-wide UI like the bottom edit bar.
+ */
+export function isColumnDeclaredEditable(
+  editable: boolean | ((rowData: never) => boolean) | undefined,
```
KuberSethi: hmm is there any way to improve the `rowData: never` typing?

packages/react-components/src/object-table/utils/types.ts
```diff
@@ -168,19 +168,22 @@ type EditFieldComponent = keyof EditFieldPropsByType;
...
- *     items: ["Active", "Inactive", "Pending"],
- *   },
+ *   getFieldComponentProps: (employee) => ({
+ *     items: getDepartmentOptions(employee),
+ *   }),
  * }
  * ```
  */
-export type EditFieldConfig = {
+export type EditFieldConfig<TData = unknown> = {
```
KuberSethi, line 174: why the change to unknown here?

KuberSethi (dismissed):
Mostly looks good just small optional things u can do in a FLUP

packages/react-components/docs/ObjectTable.md
ziyunp: Yes but it's also shown when both are false to show the number of modifications and errors, see the video above. User wants to have the option to hide it

packages/react-components/src/object-table/utils/editableUtils.ts
ziyunp: Yes this is wrong, I'll update

packages/react-components/src/object-table/utils/types.ts
ziyunp: TData wasn't passed to this type. The type of cell data is always default to unknown because there's no way to know beforehand. TData is added now because we want to pass the row object in getFieldComponentProps

packages/react-components/docs/ObjectTable.md
ziyunp: I kept the showEditFooter prop behaviour pure. If user turns it off, it should just be off, with no extra conditions. We'll see if it trips users up and review if we want to add some default conditions

packages/react-components/src/object-table/DefaultCellRenderer.tsx
```diff
@@ -44,7 +45,14 @@ export function renderDefaultCell<TData extends RowData>(
   const rowData = cellContext.row.original;
   const isEditable = isCellEditable(columnMeta?.editable, rowData);
 
-  if (!isEditable || !meta?.onCellEdit || !meta?.isInEditMode) {
+  if (
+    !meta?.onCellEdit // Type guard
```
KuberSethi, line 49: nit is this comment needed?

packages/react-components/src/object-table/utils/editableUtils.ts
```diff
@@ -28,22 +28,22 @@ export type EditablePredicate<TData extends RowData = RowData> =
...
 
 /**
  * Whether a specific cell is editable, given the column's `editable` value
- * and the row's data. Used at render time per cell.
+ * and the row's object. Used at render time per cell.
  */
 export function isCellEditable<TData extends RowData>(
   editable: EditablePredicate<TData> | undefined,
-  rowData: TData,
+  object: TData,
```
KuberSethi, line 43: hmm very minor but imo this isn't that descriptive, thoughts on a better one? can be in a FLUP

KuberSethi (dismissed):
looks good just some nits that can be in FLUPs

KuberSethi (approved):

PR #3192, opened 2026-05-07
filter-list: expose filterpopover, filterinput, usefilterliststate, summarizefiltervalue as composable primitives

KuberSethi:
Makes Filterlist component composable by updating and exporting many of the subcomponents that make up the overall component. This is to enable users to reuse and customize the look and feel of their FilterList. An example "horizontal" filter list Storybook story is given as an example.

• stable renderInput memoization: BaseFilterListProps + RenderFilterInput are now generic over a WhereClauseT param, so FilterList<Q> binds WhereClause<Q> end-to-end with no `unknown` casts. cross-filter narrowing still re-renders only the affected filter
• new composable primitives exported from experimental/filter-list: FilterPopover, FilterInput, useFilterListState, summarizeFilterValue. demo'd via a HorizontalToolbar recipe story
• fixes axis-line CSS variable mismatch (kebab-case typo broke histogram axis-line theming) and drops dead parseDate plumbing across 7 files


packages/react-components/src/filter-list/base/FilterPopoverContext.ts
```diff
@@ -0,0 +1,28 @@
...
+import { type Context, createContext } from "react";
+
+/**
+ * True when the surrounding tree is the body of an open `FilterPopover`. Inputs
+ * that wrap their value list in a portal-positioned popup (e.g. `MultiSelectInput`'s
+ * Combobox dropdown) read this to render in their always-open state so the user
+ * sees the value list immediately after opening the popover instead of having to
+ * click the inner trigger.
+ */
+export const FilterPopoverContext: Context<boolean> = createContext<boolean>(
```
ziyunp: question: this context flips `MultiSelectInput` into its inline-list rendering when it detects an enclosing `FilterPopover`. If a consumer wraps `FilterInput` inside *their own* popover (instead of `FilterPopover`) — which is the composability story this PR is selling — they'd lose the inline rendering. Did you consider exposing this as an explicit prop on `FilterInput`/`MultiSelectInput` (e.g. `layout: "inline" | "chips"`) so consumers can opt in without owning a `FilterPopover`? Context-as-mode-switch is fine for v1; just want to make sure it's intentional.

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
```diff
@@ -0,0 +1,133 @@
...
+  definition: FilterDefinitionUnion<Q>,
+  state: FilterState | undefined,
+): string {
+  if (state == null) {
+    return "";
+  }
+  const formatDate: (d: Date) => string =
+    "formatDate" in definition && definition.formatDate
+      ? definition.formatDate
+      : (d) => d.toLocaleDateString();
```
ziyunp: issue (blocking): the default formatter falls back to `d.toLocaleDateString()` (browser locale), which re-introduces the same locale inconsistency the date-picker PR (#3196) was solving for inputs. en-US viewers will see `1/15/2024` in the popover trigger, en-GB viewers will see `15/01/2024` — same dashboard, different display. Consider defaulting to ISO `YYYY-MM-DD` (matching the input format) or to a fixed locale, so triggers stay consistent across viewers when no `formatDate` is supplied.

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
```diff
@@ -0,0 +1,133 @@
...
+    case "TOGGLE":
+      return state.enabled ? "On" : "";
+    case "hasLink":
+      return state.hasLink ? "Has link" : "";
+    case "linkedProperty":
+      return summarizeFilterValue(definition, state.linkedFilterState);
+    case "keywordSearch":
+      return state.searchTerm ?? "";
+    case "custom":
+      return "Active";
```
ziyunp: nit: `"Active"` for `custom`, `"On"` for `TOGGLE`, and `"—"` (line 22) for no-value are short to the point of being cryptic in a popover trigger. `"Custom"` / `"Enabled"` / `"(No value)"` would be more discoverable. Up to you.

packages/react-components-storybook/src/stories/FilterList/Recipes/HorizontalToolbar.stories.tsx
```diff
@@ -0,0 +1,269 @@
...
+  const handleRemove = useCallback(
+    () => clearFilterState(filterKey),
+    [filterKey, clearFilterState],
+  );
+
+  // CONTAINS_TEXT renders inline — the trigger area IS the search input, so
+  // there is no nested input or popover indirection.
+  if (
+    definition.type === "PROPERTY"
+    && definition.filterComponent === "CONTAINS_TEXT"
```
ziyunp: suggestion (non-blocking): the `CONTAINS_TEXT` branch bypasses `FilterPopover` and re-implements the label + input layout inline (lines 149-156). That duplication is the kind of thing consumers will copy-paste into their own apps without understanding why this case is special. Either extract a small `InlineFilterField` helper used here, or add a code comment explaining that any non-popover-backed filter component (only `CONTAINS_TEXT` today) needs this treatment.

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
KuberSethi: done

packages/react-components/src/filter-list/base/FilterPopoverContext.ts
KuberSethi: done. removed the context and added an explicit layout prop on MultiSelectInput, plumbed through FilterInput. FilterPopover no longer dictates child layout, consumers pass layout="inline" themselves. story updated

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
KuberSethi: renamed Active to Custom and On to Enabled. for no value, split into two constants: "(No value)" for select triggers, kept "—" for range bounds where it reads naturally inline

packages/react-components-storybook/src/stories/FilterList/Recipes/HorizontalToolbar.stories.tsx
KuberSethi: expanded the comment to call out that any future single inline filter component would need the same treatment. didnt extract a helper since CONTAINS_TEXT is the only case today

ssanjay1 (dismissed):
Rubberstamping

PR #3193, opened 2026-05-08
fix function-backed columns and rdps not refreshing after action edits a related object

KuberSethi:
function-backed columns and lists with derived properties rendered stale values after an action mutated a related object until the page was reloaded

• action invalidation walks the cache once after the optimistic layer is removed; the walk skips ObjectQueries (handled per-PK) and primary-type lists (handled by Subject reactions), and dispatches per-edited-type to FunctionQueries with matching `dependsOn` and ListQueries with matching RDP traversals
• `Store.invalidateObject` now also fires `invalidateFunctionsByObject` so per-PK action edits propagate to function queries with explicit `dependsOnObjects`
• `FunctionColumnLocator` accepts an optional `dependsOn?: string[]`, and `useFunctionColumnsData` auto-passes the page's row instances as `dependsOnObjects` to each underlying `useOsdkFunctions` query
• `ListQuery` captures the `invalidationSet` from `getObjectTypesThatInvalidate` on first fetch; `revalidateObjectType` now also returns true when the edited type is in that set, so lists with `withProperties` traversing linked types refetch and recompute derived columns


ziyunp (dismissed):

ssanjay1 (dismissed):
Rubberstamping

nfdiop (approved):

PR #3194, opened 2026-05-08
stabilize filter-list per-filter where-clauses against transient renders

KuberSethi:
follow-up to #3177 — guards the per-filter where-clause cache against subtle transient re-renders that the original ship missed

• move the deep-equal cache into a reusable `useStableMapEntries` helper and swap lodash isEqual for fast-deep-equal
• tighten clause memoization so a filter's own change preserves its excluding-self reference
• add regression tests asserting sibling reference preservation when setFilterState writes deep-equal content and when filterDefinitions is a fresh-but-equal array

### DIFF
--- a/.changeset/filter-list-stable-render-input.md
+++ b/.changeset/filter-list-stable-render-input.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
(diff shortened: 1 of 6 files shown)


KuberSethi:
closing — branch was never re-parented after #3177 was squash-merged, so this PR was effectively reverting everything since that merge. reopening on a clean branch with just the new test.

PR #3195, opened 2026-05-08
dedupe empty aggregation rows across all filter-list dropdown surfaces

KuberSethi:
extends #3176's listogram-only dedupe to dropdown, multi-select, and any other filter-list consumer of usePropertyAggregation, fixing reports of two or more no-value rows in dropdowns when the backend returns separate null/empty/whitespace groupings

• extract the dedupe logic out of ListogramInput into a shared dedupeEmptyAggregationRows helper next to isEmptyValue
• apply it inside usePropertyAggregation so listogram, single-select, multi-select, text-tag, and any future consumer all collapse null/undefined/empty/whitespace rows into a single No value bucket automatically
• tighten the empty check from `value === "" || value == null` to isEmptyValue, which also catches whitespace-only strings


packages/react-components/src/filter-list/utils/__tests__/filterValues.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { dedupeEmptyAggregationRows, isEmptyValue } from "../filterValues.js";
+
+describe("filterValues", () => {
+  describe("isEmptyValue", () => {
+    it.each(
+      [
+        [null, true],
+        [undefined, true],
+        ["", true],
+        ["   ", true],
```
nfdiop: heads up: in Workshop (and the rest of tables) we don't show strings with whitespaces as "No value"

packages/react-components/src/filter-list/utils/__tests__/filterValues.test.ts
nfdiop, line 1: very cool series of tests!

packages/react-components/src/filter-list/utils/__tests__/filterValues.test.ts
KuberSethi: Agreed but unfortunately I think our users want this? Can revisit though

KuberSethi: actually went ahead and matched workshop here. isEmptyValue only treats null/undefined/empty string as no value now, whitespace strings render as is

nfdiop: How do they handle when you have two strings with different amount of whitespace character: 
like "    "  and "          "?

nfdiop (dismissed):

packages/react-components/src/filter-list/utils/__tests__/filterValues.test.ts
KuberSethi: Good call! Checked and it looks like in Workshop they preserve raw values " " and "  " as separate rows as they do a `uniqBy` on raw value without trimming, only null or undefined would get the "no value" label so I think we match them~!

ssanjay1 (approved):
Rubberstamping

PR #3196, opened 2026-05-08
swap filter-list date inputs to a use shared date picker 

KuberSethi:
the native <input type=\"date\"> renders the visible date in the browser's locale, so en-US viewers saw mm/dd/yyyy and en-GB viewers saw dd/mm/yyyy on the same dashboard. swap to a custom popover picker so every viewer sees the same YYYY-MM-DD format

• extract DateCalendar, DateRangeCalendar, calendarShared, the lazy variants, and DatePickerCommon.module.css from action-form/fields/ into shared/calendar/ so action-form and filter-list import from one place
• new FilterDatePicker pairs an ISO YYYY-MM-DD text input with a popover calendar built on the already-installed react-day-picker
• swap into SingleDateInput, MultiDateInput, and DateRangeInput; mobile viewers now see the popover calendar everywhere in place of the OS-native wheel picker
• extend formatDateForDisplay with an optional locale param for callers that want to pin a specific format


packages/react-components/src/filter-list/base/inputs/FilterDatePicker.tsx
nfdiop, line 1: Curious why not use DateTimePickerField directly?

packages/react-components/src/shared/dateUtils.ts
```diff
@@ -51,12 +51,19 @@ export function parseDateFromInput(
...
+ * format across viewers.
+ */
 export function formatDateForDisplay(
   date: Date | undefined | null,
   fallback: string = "",
+  locale?: string,
 ): string {
   if (date == null) return fallback;
-  return date.toLocaleDateString(undefined, {
+  return date.toLocaleDateString(locale, {
```
nfdiop: I probably missed it but when do allow users to pass a locale?

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.tsx
```diff
@@ -0,0 +1,182 @@
...
+          aria-label={ariaLabel}
+        />
+      </Popover.Trigger>
+      <Popover.Portal>
+        <Popover.Positioner className={styles.positioner} sideOffset={4}>
+          <Popover.Popup
+            className={styles.popup}
+            id={popoverId}
+            role="dialog"
+            aria-label="date picker"
```
ziyunp: nit: `aria-label="date picker"` — title-case it (`"Date picker"`) to match the rest of the codebase's aria-label convention.

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.module.css
```diff
@@ -0,0 +1,23 @@
...
+  min-width: 0;
+  width: 100%;
+}
+
+.positioner {
+  z-index: var(--osdk-filter-date-picker-popup-z-index, 9999);
+}
+
+.popup {
+  background: var(--osdk-filter-date-picker-popup-background, var(--bp-color-popover-background, #ffffff));
```
ziyunp: issue (non-blocking): falls back to Blueprint variables (`--bp-color-popover-background`, `--bp-color-border`, `--bp-border-radius`, `--bp-elevation-shadow-3`) — every other filter-list `.module.css` uses OSDK tokens consistently (e.g. `--osdk-surface-shadow-2`, `--osdk-background-primary`). Blueprint vars aren't guaranteed to be defined in every consumer app, so the visible fallback ends up being the hardcoded hex/rgba literal. Suggest registering `--osdk-filter-date-picker-popup-*` tokens in `tokens/component-tokens/filter-list.css` pointing at the OSDK surface tokens, the way `FilterPopover.module.css` does.

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.module.css
```diff
@@ -0,0 +1,23 @@
+.trigger {
```
ziyunp: nit: missing the Apache copyright header that every other `.module.css` in this folder has.

packages/react-components/src/filter-list/base/inputs/MultiDateInput.tsx
```diff
@@ -94,6 +83,7 @@ function MultiDateInputInner({
               {formatDateForDisplay(date)}
               <Button
                 className={sharedStyles.tagRemove}
+                // eslint-disable-next-line react/jsx-no-bind
```
ziyunp: useCallback

packages/react-components/src/shared/dateUtils.ts
KuberSethi: removed, think we dont need it

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.tsx
KuberSethi: Done

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.module.css
KuberSethi: Done

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.module.css
KuberSethi: Done

packages/react-components/src/filter-list/base/inputs/FilterDatePicker.tsx
KuberSethi: done. added an optional modal prop so we can disable the trap focus and dismiss layer when nesting inside another popover. filter-list now uses DatetimePickerField directly with that off, FilterDatePicker is gone

packages/react-components/src/filter-list/base/inputs/MultiDateInput.tsx
KuberSethi: extracted a DateTag component so the per row remove handler is wrapped properly

KuberSethi:
superseded by #3179 — that PR pulls in this branch's calendar extraction, lifts `DatetimePickerField` / `DateRangeInputField` out of action-form into `shared/calendar/` as `DatePicker` / `DateRangePicker`, and uses them everywhere (action-form, filter-list, object-table). closing in favor of the combined PR.

PR #3202, opened 2026-05-08
storybook: show single grey experimental badge at component level

KuberSethi:
the addon was spreading its purple `defaultConfig` on top of the custom grey "Experimental" badge config, so the addon's purple chip was winning across every story sidebar entry.

• drop `...defaultConfig` so the custom grey badge renders instead of the addon's purple default
• limit `display.sidebar` to component + group so the pill shows once at the component level (e.g. on FilterList) rather than on every story

### DIFF
--- a/.changeset/storybook-experimental-badge.md
+++ b/.changeset/storybook-experimental-badge.md
@@ -1,4 +1,4 @@
 ---
 ---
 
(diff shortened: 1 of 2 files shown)


ssanjay1 (approved):
Rubberstamping

PR #3203, opened 2026-05-08
Camel case faux

brandonwisnicki:
Adds support for Faux to camel case action type API names that are kebab cased. This reflects the behavior of api gateway


KuberSethi (approved):
🐫

PR #3208, opened 2026-05-09
Adds support for aggregations in faux foundry

brandonwisnicki:
Adds support for aggregations in faux foundry


packages/faux/src/FauxFoundry/aggregateObjects.ts
```diff
@@ -0,0 +1,388 @@
...
+
+function numericAgg(
+  objects: BaseServerObject[],
+  field: string,
+  name: string,
+  compute: (values: number[]) => number | undefined,
+): AggregationMetricResultV2 {
+  const values = getNumericValues(objects, field);
+  const value = values.length === 0 ? undefined : compute(values);
+  return { name, value: value as number };
```
KuberSethi: `value as number` isn't quite true when the empty-set path returns `undefined`. `AggregationMetricResultV2.value` is typed `any` anyway, so the cast does nothing, can we drop it?

packages/faux/src/FauxFoundry/aggregateObjects.ts
```diff
@@ -0,0 +1,388 @@
...
+      return { name: aggregation.name ?? "count", value: objects.length };
+    case "min": {
+      const field = getPropertyField(
+        aggregation.field,
+        aggregation.propertyIdentifier,
+      );
+      return numericAgg(
+        objects,
+        field,
+        aggregation.name ?? `${aggregation.field}.min`,
```
KuberSethi: this uses `aggregation.field` instead of the resolved `field` like every other case and then produces `"undefined.min"` when caller passes `propertyIdentifier`, i think this could be `` `${field}.min` ``

packages/faux/src/FauxFoundry/aggregateObjects.ts
```diff
@@ -0,0 +1,388 @@
...
+): number {
+  if (target === undefined) {
+    return 1;
+  }
+  if (typeof value === "number" && typeof target === "number") {
+    return value - target;
+  }
+  const strValue = String(value);
+  const strTarget = String(target);
+  return strValue.localeCompare(strTarget);
```
KuberSethi, line 244: `compareValue` falls through to `localeCompare` when one side is numeric and the other is a numeric string, can we coerce to number when either side is a number?

packages/faux/src/handlers/createObjectSetHandlers.ts
```diff
@@ -46,63 +47,11 @@ export const createObjectSetHandlers = (
       const ds = fauxFoundry.getDataStore(params.ontologyApiName);
       const objects = getObjectsFromSet(ds, body.objectSet, undefined);
 
-      const exactGroupBys = body.groupBy.filter(
-        (g): g is Extract<typeof g, { type: "exact" }> => g.type === "exact",
+      return aggregateObjects(
+        objects,
+        body.aggregation ?? [],
```
KuberSethi: i think `body.aggregation` and `body.groupBy` are required by `AggregateObjectSetRequestV2`, so the `?? []` fallbacks aren't required

packages/faux/src/FauxFoundry/aggregateObjects.ts
brandonwisnicki: good catch

packages/faux/src/FauxFoundry/aggregateObjects.ts
brandonwisnicki: yes, and i'll fall back to string if either side cannot be coereced

KuberSethi (approved):
LGTM, thanks for doing this!

PR #3209, opened 2026-05-11
Fix: dropdown validated on click

nfdiop:
## Summary

- **DropdownField** now owns its blur/touched signal instead of relying on the container div's `onBlur` handler
- When a dropdown popover opens in a portal, focus leaves the container div, causing `FieldBridge`'s `handleBlur` to fire prematurely and trigger RHF validation before the user interacts
- Removed `DROPDOWN` from `SELECT_LIKE_FIELDS` in `FieldBridge` — dropdown fields receive RHF's `onBlur` directly via a new `onFieldBlur` prop threaded through `FormFieldRenderer`
- `DropdownField` calls `onBlur` on popover close and on value change (selection, clear, chip removal)
- Non-dropdown fields are unchanged

## Notes

FLUP to come to pass the `onBlur` to all fields and delegate calling to them, so we avoid special cases inside `FieldBridge`.

## Test plan

- Added tests: no blur on open, blur on dismiss, blur on selection (single/multi), blur on searchable dismiss
- Existing BaseForm validation tests pass unchanged

### DIFF
--- a/packages/react-components/src/action-form/fields/DropdownField.tsx
+++ b/packages/react-components/src/action-form/fields/DropdownField.tsx
@@ -69,7 +69,6 @@ export const DropdownField: <V, Multiple extends boolean = false>(
   disableClientSideFiltering,
   popupStatus,
   trailingItem,
-  onBlur,
   ...rest
 }: DropdownFieldProps<V, Multiple> & {
   onBlur?: () => void;
@@ -103,7 +102,6 @@ export const DropdownField: <V, Multiple extends boolean = false>(
         disableClientSideFiltering={disableClientSideFiltering}
         popupStatus={popupStatus}
         trailingItem={trailingItem}
-        onBlur={onBlur}
       />
     );
   }
@@ -115,7 +113,6 @@ export const DropdownField: <V, Multiple extends boolean = false>(
       value={normalizedValue}
       itemToStringLabel={resolvedItemToStringLabel}
       getKey={getKey}
-      onBlur={onBlur}
     />
   );
 });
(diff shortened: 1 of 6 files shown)


packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -128,28 +134,46 @@ const SelectDropdown = typedReactMemo(function SelectDropdownFn<
...
   const [open, setOpen] = useState(false);
 
   const hasValue = value != null;
 
+  const handleOpenChange = useCallback((nextOpen: boolean) => {
+    setOpen(nextOpen);
+    // Mark the field as touched when the popover closes so RHF validates.
+    // Opening the popover does not trigger validation.
+    if (!nextOpen) {
+      onBlur?.();
```
nfdiop: I was hesitating to rename this onBlur to `onTouched` but wasn't sure it was worth it

packages/react-components/src/action-form/fields/FieldBridge.tsx
```diff
@@ -94,7 +94,11 @@ export const FieldBridge: React.FC<FieldBridgeProps> = memo(
         value={value}
         fieldDefinition={fieldDef}
         onFieldValueChange={handleChange}
-        onBlur={handleBlur}
+        // Dropdown fields own their blur signal because their popover renders
+        // in a portal outside the container div — the container-level blur
+        // would fire prematurely when the popover opens.
+        onBlur={isDropdown ? undefined : handleBlur}
```
nfdiop, line 100: FLUP to completely delete this and let every field handle how they call onBlur

olliejenningsPT (commented):
**SelectDropdown / ComboboxDropdown: `onBlur` fires twice on single-select value selection**

When a user picks an item in single-select mode, two separate code paths both call `onBlur`:

1. `handleValueChange` calls `onBlur?.()` directly
2. The popover closes → `handleOpenChange(false)` calls `onBlur?.()` again

Same double-fire happens in `ComboboxDropdown` single-select, and also in `handleClear` for single-select:

```ts
// handleClear — single-select path
onBlur?.();                    // ← first
handleOpenChange(false);       // ← calls onBlur again inside
```

This is functionally harmless (RHF's `onBlur` is idempotent for marking touched), but it's unnecessary work and a potential source of confusion.

Consider removing the direct `onBlur()` call from `handleValueChange` in `SelectDropdown` — since single-select always closes the popover on selection, `handleOpenChange(false)` already covers it. For `ComboboxDropdown`, the `onBlur` in `handleValueChange` is needed for multi-select (where the popover stays open), but you could guard it with `if (isMultiple)` to avoid the double-fire in single-select mode.

olliejenningsPT (approved):

PR #3211, opened 2026-05-11
Add DocumentViewer with media-type routing and viewer components

olliejenningsPT:
## Summary

- Adds a unified **DocumentViewer** component that accepts an OSDK `Media` object and automatically renders the appropriate viewer based on MIME type
- Introduces a shared **`useMediaContents`** hook for fetching and transforming media contents with cleanup support
- Adds OSDK Media wrappers for existing **TiffRenderer** and **MarkdownRenderer**
- Opt-in **multi-page TIFF to PDF conversion** via the experimental MIO `imageToDocument.createPdf` transform API

### New viewer components

Each viewer follows the two-layer pattern: a base OSDK-agnostic component and a Media wrapper.

| Component | Format | Dependency |
|-----------|--------|------------|
| **ImageViewer** | PNG, JPEG, GIF, SVG, WebP, BMP | Native `` |
| **VideoViewer** | video/* | Native `<video>` |
| **DocxViewer** | .docx | `docx-preview` |
| **ExcelViewer** | .xlsx (multi-sheet with tabs) | `xlsx` (SheetJS) |
| **EmailViewer** | .eml (sandboxed HTML body) | `postal-mime` |
| **XmlViewer** | .xml | None (styled `<pre>`) |

### DocumentViewer MIME routing

Uses a `ViewerType` enum for routing. TIFF detection also checks `.tif`/`.tiff` file extensions via the `fileName` prop.

| MIME type | Viewer |
|-----------|--------|
| `application/pdf` | PdfViewer |
| `image/tiff` (or `.tif`/`.tiff` extension) | TiffRenderer (or BasePdfViewer when `enableTiffToPdf` detects multi-page) |
| `image/png`, `image/jpeg`, `image/gif`, `image/svg+xml`, `image/webp`, `image/bmp` | ImageViewer |
| `video/*` | VideoViewer |
| `text/markdown`, `text/x-markdown` | MarkdownRenderer |
| `application/vnd.openxmlformats-officedocument.wordprocessingml.document` | DocxViewer |
| `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` | ExcelViewer |
| `message/rfc822` | EmailViewer |
| `application/xml`, `text/xml` | XmlViewer |

### Multi-page TIFF to PDF conversion

When `enableTiffToPdf` is true:
1. Fetches the TIFF and counts pages using `utif` (IFD count)
2. Single-page TIFFs render with TiffRenderer as normal
3. Multi-page TIFFs are converted to PDF via the experimental `__EXPERIMENTAL__NOT_SUPPORTED_YET__transformAndWait` API (`imageToDocument.createPdf`)
4. The resulting PDF is rendered in `BasePdfViewer`
5. On any error, falls back to TiffRenderer

### DocumentViewer (MIME routing)

https://github.com/user-attachments/assets/b6d42836-da03-458e-95bf-1bf81d0ea321

### ImageViewer

https://github.com/user-attachments/assets/dfdadbf8-f914-435a-8705-5b1bb893c461

### VideoViewer

https://github.com/user-attachments/assets/fc8dd138-6e60-49f1-a03e-05ef5d54b18e

### DocxViewer

https://github.com/user-attachments/assets/1a1706e7-fb0d-48d0-bccb-d43521e44b9c

### ExcelViewer (multi-sheet tabs)

https://github.com/user-attachments/assets/b280ce5a-04fe-4766-83b3-99283399a206

### EmailViewer (sandboxed HTML)

https://github.com/user-attachments/assets/d4893e31-f850-4872-923f-ff103c9a4633

### XmlViewer

https://github.com/user-attachments/assets/951ced56-c0aa-43aa-ad92-a1fb25ba9c2a

## Test plan

- [x] Storybook: verify all viewer stories render correctly
- [x] Storybook: verify DocumentViewer routes to correct viewer for each MIME type
- [x] Storybook: verify ExcelViewer sheet tab switching
- [x] Storybook: verify EmailViewer HTML body renders in sandboxed iframe
- [x] Storybook: verify DocxViewer with real .docx file
- [x] Storybook: verify ExcelViewer WithRealFile story with provided .xlsx
- [x] Storybook: verify multi-page TIFF to PDF conversion story
- [x] Typecheck: `pnpm turbo typecheck --filter=@osdk/react-components`
- [x] Build: `pnpm turbo transpile`


.github/workflows/ci.yml
```diff
@@ -253,7 +253,7 @@ jobs:
           git checkout ${{ github.event.pull_request.base.sha }}
 
       - name: Install and build (base)
-        run: pnpm install && pnpm exec turbo transpile
+        run: pnpm exec turbo clean && pnpm install && pnpm exec turbo transpile
```
nfdiop, line 256: curious why this change?

packages/react-components-storybook/src/stories/DocumentViewer/DocumentViewer.stories.tsx
```diff
@@ -0,0 +1,429 @@
...
+const SAMPLE_DOCX_URL = `${import.meta.env.BASE_URL}notional-word-example.docx`;
+
+const SAMPLE_VIDEO_URL = `${import.meta.env.BASE_URL}example.mp4`;
+
+const SAMPLE_TIFF_URL = `${import.meta.env.BASE_URL}multi-page-tiff.tiff`;
+
+/**
+ * Creates a sample PNG image as a Blob.
+ */
+function createSampleImageBlob(): Blob {
```
nfdiop, line 38: curious why not add the image in the folders?

packages/react-components-storybook/src/stories/DocumentViewer/DocumentViewer.stories.tsx
```diff
@@ -0,0 +1,429 @@
...
+  </book>
+</catalog>`;
+
+const mockXmlMedia = createMockMedia(
+  "application/xml",
+  () => Promise.resolve(new Response(SAMPLE_XML)),
+  "data.xml",
+);
+
+function createMockExcelMedia(): Media {
```
nfdiop, line 163: like for png why not have a excel file which could make it easier to debug?

packages/react-components/docs/EmailViewer.md
```diff
@@ -0,0 +1,88 @@
...
+<EmailViewer media={ticket.emailAttachment} />;
+```
+
+### With parsed email data
+
+```tsx
+import { BaseEmailViewer } from "@osdk/react-components/experimental/email-viewer";
+
+<BaseEmailViewer
+  email={{
```
nfdiop, line 35: could we parse for the user?

packages/react-components/src/document-viewer/DocumentViewer.tsx
```diff
@@ -0,0 +1,224 @@
...
+  );
+  const rootClassName = classnames(styles.container, className);
+
+  switch (viewerType) {
+    case ViewerType.Pdf:
+      return (
+        <PdfViewer
+          media={media}
+          className={rootClassName}
+          {...pdfViewerProps}
```
nfdiop: will pdfViewerProps have a classname that can override rootClassName?

packages/react-components/src/document-viewer/DocumentViewer.tsx
```diff
@@ -0,0 +1,224 @@
...
+  const viewerType = useMemo(
+    () => getViewerType(mimeType, fileName),
+    [mimeType, fileName],
+  );
+  const rootClassName = classnames(styles.container, className);
+
+  switch (viewerType) {
+    case ViewerType.Pdf:
+      return (
+        <PdfViewer
```
nfdiop, line 134: Are we lazy loading all fields so we only load them when needed?

packages/react-components/src/document-viewer/DocumentViewerApi.ts
```diff
@@ -0,0 +1,71 @@
...
+  Pdf = "pdf",
+  Tiff = "tiff",
+  Image = "image",
+  Video = "video",
+  Markdown = "markdown",
+  Docx = "docx",
+  Excel = "excel",
+  Email = "email",
+  Xml = "xml",
+  Unsupported = "unsupported",
```
nfdiop, line 38: nice!

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
```diff
@@ -0,0 +1,76 @@
...
+  src,
+  className,
+  onError,
+}: BaseDocxViewerProps): React.ReactElement {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [renderError, setRenderError] = useState<string | undefined>(undefined);
+  const onErrorRef = useRef(onError);
+  onErrorRef.current = onError;
+
+  useEffect(() => {
```
nfdiop, line 36: nit: could you use a ref callback?

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
```diff
@@ -0,0 +1,76 @@
...
+      renderEndnotes: true,
+    }).catch(() => {
+      if (!cancelled) {
+        setRenderError("Failed to render document");
+        onErrorRef.current?.();
+      }
+    });
+
+    return () => {
+      cancelled = true;
```
nfdiop, line 62: do we need any other clean up on the container?

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export function BaseDocxViewer({
+  src,
+  className,
+  onError,
+}: BaseDocxViewerProps): React.ReactElement {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [renderError, setRenderError] = useState<string | undefined>(undefined);
+  const onErrorRef = useRef(onError);
+  onErrorRef.current = onError;
```
nfdiop: nit: you can use `useEventCallback`

packages/react-components/src/docx-viewer/DocxViewer.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export function DocxViewer({
+  media,
+  className,
+  ...docxViewerProps
+}: DocxViewerMediaProps): React.ReactElement {
+  const { data: src, loading, error } = useMediaContents(
+    media,
+    useCallback(transformToArrayBuffer, []),
```
nfdiop: can we drop useCallback here?

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return () => {
+      cancelled = true;
+    };
+  }, [src]);
+
+  const rootClassName = classnames(styles.container, className);
+
+  return (
+    <div className={rootClassName}>
+      {renderError != null && (
```
nfdiop: I was thrown off by the name: can we rename to `error`?

packages/react-components/src/email-viewer/BaseEmailViewer.tsx
```diff
@@ -0,0 +1,114 @@
...
+  email,
+  className,
+}: BaseEmailViewerProps): React.ReactElement {
+  const rootClassName = classnames(styles.container, className);
+
+  const bodyContent = useMemo(() => {
+    if (email.html != null) {
+      return (
+        <div className={styles.bodyContainer}>
+          {
```
nfdiop, line 43: I wonder how this will work in custom widgets?

packages/react-components/src/email-viewer/parseEmail.ts
nfdiop, line 1: could we add tests?

packages/react-components/src/document-viewer/DocumentViewer.module.css
```diff
@@ -0,0 +1,52 @@
...
+  justify-content: center;
+  gap: calc(var(--osdk-surface-spacing) * 2);
+  width: 100%;
+  height: 100%;
+  font-family: var(--osdk-typography-family-default);
+  font-size: var(--osdk-typography-size-body-medium);
+  color: var(--osdk-typography-color-muted);
+}
+
+@keyframes spin {
```
nfdiop: **suggestion (non-blocking):** The `@keyframes spin`, `.loadingContainer`, `.errorContainer`, `.spinnerIcon`, `.errorIcon` blocks are copy-pasted identically across 9+ CSS module files (DocumentViewer, BaseDocxViewer, BaseEmailViewer, BaseExcelViewer, BaseImageViewer, BaseVideoViewer, BaseXmlViewer, MarkdownRenderer, TiffRenderer).

Consider extracting these into a shared `shared/loading.module.css` and using `composes:` from each viewer's CSS module. This would eliminate ~200 lines of duplicated CSS and create a single source of truth for the loading/error pattern.

packages/react-components/src/document-viewer/DocumentViewer.tsx
```diff
@@ -0,0 +1,224 @@
...
+  excelViewerProps,
+  emailViewerProps,
+  xmlViewerProps,
+  fileName,
+  enableTiffToPdf = false,
+}: DocumentViewerProps): React.ReactElement {
+  const mimeType = mimeTypeOverride ?? media.getMediaReference().mimeType;
+  const viewerType = useMemo(
+    () => getViewerType(mimeType, fileName),
+    [mimeType, fileName],
```
nfdiop, line 127: **issue (blocking):** The spread `{...pdfViewerProps}` comes *after* `className={rootClassName}`, so a user passing `className` inside `pdfViewerProps` will silently override the container class and lose all container styling.

Since `pdfViewerProps` is typed as `Partial<Omit<PdfViewerProps, "src">>` which includes `className`, this is a real footgun. Same pattern in every viewer branch below.

Fix options:
- Omit `className` from the forwarded props types, or
- Merge classnames: `className={classnames(rootClassName, pdfViewerProps?.className)}`

packages/react-components/src/docx-viewer/DocxViewer.tsx
nfdiop: **suggestion (non-blocking):** `useCallback(transformToArrayBuffer, [])` wrapping a module-level constant is unnecessary — the function reference is already stable across renders. Just pass `transformToArrayBuffer` directly to `useMediaContents`.

Same pattern appears in EmailViewer, ExcelViewer, XmlViewer, ImageViewer, VideoViewer, TiffViewerMedia, and MarkdownViewerMedia.

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  useEffect(() => {
+    let cancelled = false;
+    const container = containerRef.current;
+    if (container == null) {
+      return;
+    }
+    setRenderError(undefined);
+
+    renderAsync(src, container, undefined, {
```
nfdiop, line 45: **issue (non-blocking):** When `src` changes, `renderAsync` is called on the same container div without clearing previous content first. If `docx-preview` appends rather than replaces, you'll get stacked document renderings.

Consider adding `container.innerHTML = "";` before the `renderAsync` call.

packages/react-components/src/excel-viewer/BaseExcelViewer.tsx
```diff
@@ -0,0 +1,127 @@
...
+  let num = index;
+  do {
+    const remainder = num % 26;
+    letter = String.fromCharCode(65 + remainder) + letter;
+    num = Math.floor(num / 26) - 1;
+  } while (num >= 0);
+  return letter;
+}
+
+const SheetTable: React.FunctionComponent<{ sheet: SheetData }> = React.memo(
```
nfdiop, line 36: **nitpick (non-blocking):** `SheetTable` is a `const` wrapping `React.memo`, so it shows as "Anonymous" in React DevTools. Consider adding:
```tsx
SheetTable.displayName = "SheetTable";
```

packages/react-components/package.json
```diff
@@ -132,14 +195,17 @@
...
     "lodash-es": "^4.17.21",
     "pdfjs-dist": "~4.8.69",
+    "postal-mime": "^2.7.4",
     "react-day-picker": "^8.10.0",
     "react-hook-form": "^7.54.0",
     "react-markdown": "^9.0.3",
     "remark-gfm": "^4.0.1",
-    "utif": "^3.1.0"
+    "utif": "^3.1.0",
+    "xlsx": "^0.18.5"
```
nfdiop, line 208: **question (non-blocking):** The `xlsx` package (SheetJS) v0.18.5 uses a custom license with restrictions on certain commercial uses. Has this been reviewed for compliance with Palantir's licensing requirements?

packages/react-components/src/document-viewer/hooks/useTiffToPdf.ts
```diff
@@ -0,0 +1,140 @@
...
+          loading: false,
+          error: err instanceof Error ? err : new Error(String(err)),
+        });
+      }
+    });
+
+    return () => {
+      cancelled = true;
+    };
+  }, [media, enabled, client]);
```
nfdiop, line 141: **suggestion (non-blocking):** The effect depends on `media` object identity. If the parent re-renders and creates a new `media` object (common with OSDK query results), this triggers a full re-fetch + TIFF decode + MIO transform cycle even when the underlying media item hasn't changed.

Consider stabilizing on a key derived from the media reference:
```ts
const mediaKey = JSON.stringify(media.getMediaReference());
```
and using `mediaKey` in the deps array instead of `media`.

packages/react-components/src/email-viewer/BaseEmailViewer.tsx
```diff
@@ -0,0 +1,114 @@
...
+            /*
+             * Security: allow-same-origin is needed so CSS within the email HTML
+             * resolves correctly. Scripts are blocked because allow-scripts is NOT
+             * set. Never add allow-scripts here — the combination of
+             * allow-same-origin + allow-scripts bypasses the sandbox entirely.
+             */
+          }
+          <iframe
+            className={styles.bodyIframe}
```
nfdiop, line 57: **suggestion (non-blocking):** The security comment explaining why `allow-scripts` must never be added is excellent. One additional note worth documenting: `allow-same-origin` means the email HTML shares the parent page's origin — it can read CSS custom properties and potentially exfiltrate data via CSS `url()` in `background-image`. Acceptable for an experimental component, but worth noting as a known limitation if this graduates to stable.
Also curious how this will work inside a custom widget

nfdiop (commented):
This is exciting! Thanks for doing this! 
Left some comments and could we add tests?

.github/workflows/ci.yml
olliejenningsPT: The bundle analyzer breaks when new packages are introduced as it can't compare and so everything including old packages become new, this prevents that.

packages/react-components-storybook/src/stories/DocumentViewer/DocumentViewer.stories.tsx
olliejenningsPT: just easy to create one programmatically

packages/react-components-storybook/src/stories/DocumentViewer/DocumentViewer.stories.tsx
olliejenningsPT: we have both

packages/react-components/src/document-viewer/DocumentViewer.tsx
olliejenningsPT: So we can add in react lazy, but the bundler needs to be the one in charge of this

packages/react-components/docs/EmailViewer.md
olliejenningsPT: @nfdiop how do you mean?

packages/react-components/package.json
olliejenningsPT: So it uses the Apache 2 license that is almost identical to MIT in terms of permissions, and Palantir allows those

packages/react-components/src/document-viewer/DocumentViewer.tsx
olliejenningsPT: Fixed — added `| "className"` to the `Omit` on all 9 forwarded prop types so `className` can no longer leak through the spread and override the container class.

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
olliejenningsPT: Done — switched to `useEventCallback`.

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
olliejenningsPT: Done — renamed to `error`.

packages/react-components/src/docx-viewer/DocxViewer.tsx
olliejenningsPT: Done — removed unnecessary `useCallback` wrappers here and in all other viewers (XmlViewer, ImageViewer, VideoViewer, TiffViewerMedia, MarkdownViewerMedia, EmailViewer). The transform functions are module-level constants so they're already stable references.

olliejenningsPT: Fixed across all viewers — passing the stable module-level functions directly now.

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
olliejenningsPT: Good catch — added `container.innerHTML = ""` before the `renderAsync` call.

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
olliejenningsPT: Added `container.innerHTML = ""` at the start of the effect to clear previous content. `docx-preview` doesn't expose a teardown API so this plus the `cancelled` flag should be sufficient.

packages/react-components/src/excel-viewer/BaseExcelViewer.tsx
olliejenningsPT: Done — added `SheetTable.displayName = "SheetTable"`.

packages/react-components/src/email-viewer/parseEmail.ts
olliejenningsPT: Added tests covering: complete email parsing, missing optional fields, group addresses, multiple recipients, and group-as-from edge case.

packages/react-components/src/document-viewer/DocumentViewer.module.css
olliejenningsPT: Done — extracted shared `loadingContainer`, `errorContainer`, `spinnerIcon`, `errorIcon`, and `@keyframes spin` into `src/shared/LoadingError.module.css`. All 10 viewer CSS modules now use `composes:` from the shared file.

packages/react-components/src/docx-viewer/BaseDocxViewer.tsx
olliejenningsPT: Going to leave this as `useRef` — the container div is unconditionally rendered and never remounts, so a ref callback wouldn't add anything here. A ref callback is more useful when the target element is conditionally rendered or you need to react to mount/unmount.

packages/react-components/src/document-viewer/hooks/useTiffToPdf.ts
olliejenningsPT: OSDK `Media` objects from `@osdk/react` hooks are referentially stable across re-renders (they come from the observable cache), so this shouldn't cause spurious re-fetches in practice. Switching to `JSON.stringify(media.getMediaReference())` would add serialization cost every render and could mask legitimate changes if the reference fields aren't sufficient. Happy to revisit if someone hits unnecessary re-fetches.

packages/react-components/src/email-viewer/BaseEmailViewer.tsx
olliejenningsPT: Good call — added the CSS exfiltration limitation to the existing security comment. Re: custom widgets, the iframe approach works fine nested inside a widget iframe. The sandbox restrictions still apply, and the main practical concern is just ensuring the parent container provides a defined height so the iframe doesn't collapse.

packages/react-components/src/email-viewer/BaseEmailViewer.tsx
olliejenningsPT: Nested iframes work fine in browsers — the sandbox attribute still restricts the inner iframe. The email HTML can't execute scripts regardless of the hosting context. The main thing to watch is that the widget layout gives the email viewer a defined height, otherwise the iframe may collapse to zero (same constraint as all other viewers).

packages/react-components/src/document-viewer/hooks/useTiffToPdf.ts
```diff
@@ -0,0 +1,140 @@
...
+      if (!cancelled) {
+        // On any error, fall back to TiffRenderer
+        setResult({
+          viewerType: ViewerType.Tiff,
+          pdfData: undefined,
+          loading: false,
+          error: err instanceof Error ? err : new Error(String(err)),
+        });
+      }
+    });
```
ziyunp, line 136: **suggestion (non-blocking):** When the MIO transform fails, we silently render the raw TIFF and the user has no signal that the requested PDF conversion didn't happen. Maybe at least `console.warn` on failure

packages/react-components/src/document-viewer/hooks/useTiffToPdf.ts
```diff
@@ -0,0 +1,140 @@
...
+    async function detectAndTransform() {
+      const currentMedia = mediaRef.current;
+
+      // Step 1: Fetch TIFF bytes and count pages
+      const response = await currentMedia.fetchContents();
+      const buffer = await response.arrayBuffer();
+
+      if (cancelled) return;
+
+      const ifds = UTIF.decode(buffer);
```
ziyunp, line 81: **thought:** `UTIF.decode(buffer)` is synchronous on the main thread. Fine for "count IFDs" on small files, but a big multi-page TIFF could stall the UI. Possible follow-up to offload via a tiny worker if you ever see jank in real data.

packages/react-components/src/document-viewer/DocumentViewer.tsx
```diff
@@ -0,0 +1,224 @@
...
+    case ViewerType.Unsupported:
+      return (
+        <div className={rootClassName}>
+          <div className={styles.unsupportedContainer}>
+            <DocumentIcon className={styles.unsupportedIcon} />
+            Unsupported file type: {mimeType}
+          </div>
+        </div>
+      );
+  }
```
ziyunp, line 226: **suggestion (non-blocking):** The package already has `assertUnreachable` (`src/shared/assertUnreachable.ts`) used in `filter-list` and `action-form` for this exact pattern. Can we reuse that?

packages/react-components/src/document-viewer/DocumentViewer.tsx
```diff
@@ -0,0 +1,224 @@
...
+  emailViewerProps,
+  xmlViewerProps,
+  fileName,
+  enableTiffToPdf = false,
+}: DocumentViewerProps): React.ReactElement {
+  const mimeType = mimeTypeOverride ?? media.getMediaReference().mimeType;
+  const viewerType = useMemo(
+    () => getViewerType(mimeType, fileName),
+    [mimeType, fileName],
+  );
```
ziyunp, line 128: **nit:** `getViewerType` is a handful of string compares, the `useMemo` adds more overhead than it saves. Maybe return a string union type from getViewerType instead of an enum and we won't need useMemo

packages/react-components/src/email-viewer/BaseEmailViewer.tsx
```diff
@@ -0,0 +1,119 @@
...
+ * limitations under the License.
+ */
+
+import classnames from "classnames";
+import React, { useMemo } from "react";
+import styles from "./BaseEmailViewer.module.css";
+import type { BaseEmailViewerProps, EmailAddress } from "./types.js";
+
+function formatAddress(address: EmailAddress): string {
+  if (address.name !== "") {
```
ziyunp: **nit:** Use `!= null` or `!address.name` guard to handle undefined and null too

ziyunp (commented):
Did a review against Contributing.md on top of code review, not blocking, can address in FLUP

packages/react-components/src/document-viewer/DocumentViewerApi.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export interface DocumentViewerProps {
```
ziyunp, line 41: **issue (non-blocking):** CONTRIBUTING.md says "Define the API in its own file: `<Name>Api.ts` co-located with the component." All 8 new viewers use `types.ts` instead. Would it make sense to rename these to `DocumentViewerApi.ts`, `DocxViewerApi.ts`, etc. to match the convention? The `types.ts` naming also mixes internal types (like `ViewerType` enum) with public API surface.

packages/react-components/src/document-viewer/DocumentViewerApi.ts
```diff
@@ -0,0 +1,77 @@
...
+  emailViewerProps?: Partial<Omit<BaseEmailViewerProps, "email" | "className">>;
+  /** Props forwarded to BaseXmlViewer when rendering XML */
+  xmlViewerProps?: Partial<
+    Omit<BaseXmlViewerProps, "content" | "className">
+  >;
+  /** File name hint for MIME type detection (e.g. "scan.tif"). Used to detect
+   *  TIFF files when the MIME type is ambiguous (e.g. application/octet-stream). */
+  fileName?: string;
+  /** When true, multi-page TIFF files are converted to PDF via MIO transform API */
+  enableTiffToPdf?: boolean;
```
ziyunp, line 89: **suggestion (non-blocking):** CONTRIBUTING.md says "Document defaults inline with `@default` JSDoc tags on every optional prop." None of the new props types have `@default` annotations — e.g. `enableTiffToPdf` defaults to `false`, `alt` defaults to `""`, `mimeType` is optional. Worth a pass to add them.

packages/react-components/src/tokens/component-tokens/document-viewer.css
```diff
@@ -0,0 +1,5 @@
+:root {
+  /* Document Viewer - Container */
+  --osdk-document-viewer-bg: var(--osdk-background-primary);
+  --osdk-document-viewer-border: var(--osdk-surface-border);
+}
```
ziyunp, line 5: **issue (non-blocking):** CONTRIBUTING.md says "If you added CSS variables, update `docs/CSSVariables.md`." The 7 new token files add ~60+ CSS variables but `CSSVariables.md` wasn't updated.

packages/react-components/docs/DocumentViewer.md
```diff
@@ -0,0 +1,93 @@
+<!-- cspell:words openxmlformats officedocument wordprocessingml spreadsheetml -->
```
ziyunp, line 1: **issue (non-blocking):** CONTRIBUTING.md step 9 asks for three more doc updates that are missing:
- Add a one-line entry to the components table in `AGENTS.md` and `README.md`
- Register each new doc in `docs/sidebarsReactComponents.ts` so they appear in the Docusaurus site

ziyunp (dismissed):

packages/react-components/src/document-viewer/DocumentViewer.tsx
olliejenningsPT: Prefer keeping the enum — we converted to an enum intentionally in an earlier commit for type safety in the exhaustive switch (just added `assertUnreachable` in the default case). The `useMemo` is lightweight but can stay for consistency with how we wrap derivations elsewhere. Not worth changing the type representation for this.

packages/react-components/src/document-viewer/hooks/useTiffToPdf.ts
olliejenningsPT: Good thought. `UTIF.decode` only parses IFD headers (not pixel data), so it should be fast even on large files. Happy to revisit with a worker if we see jank on real data.

packages/react-components/src/document-viewer/DocumentViewer.tsx
olliejenningsPT: Done — added `assertUnreachable(viewerType)` in the default case.

packages/react-components/src/document-viewer/hooks/useTiffToPdf.ts
olliejenningsPT: Done — added `console.warn` on transform failure before the fallback.

packages/react-components/src/email-viewer/BaseEmailViewer.tsx
olliejenningsPT: Done — switched to `if (address.name)` which handles empty string, null, and undefined.

packages/react-components/src/document-viewer/DocumentViewerApi.ts
olliejenningsPT: Done — renamed all `types.ts` to `<Name>Api.ts` (DocumentViewerApi.ts, DocxViewerApi.ts, etc.) across all 7 viewers. Import paths updated throughout.

packages/react-components/src/document-viewer/DocumentViewerApi.ts
olliejenningsPT: Done — added `@default` JSDoc tags to all optional props across all viewer API files.

packages/react-components/src/tokens/component-tokens/document-viewer.css
olliejenningsPT: Done — added all 30 new CSS variables to `docs/CSSVariables.md` (DocumentViewer, DocxViewer, EmailViewer, ExcelViewer, ImageViewer, VideoViewer, XmlViewer sections).

packages/react-components/docs/DocumentViewer.md
olliejenningsPT: Done — added entries to AGENTS.md (entry points + components table), README.md components table, and `docs/sidebarsReactComponents.ts`.

ziyunp (approved):

PR #3212, opened 2026-05-11
Fix boolean properties rendering as empty in ObjectTable

olliejenningsPT:
## Summary

- Boolean property columns in `ObjectTable` rendered as empty because React does not render boolean values (`true`/`false`) as JSX children
- Added a `toDisplayValue` helper in `DefaultCellRenderer.tsx` that converts booleans to their string representation before passing to JSX
- Affects both normal read-only cells and non-editable cells in edit mode

## Root Cause

`DefaultCellRenderer` was returning `<>{cellValue}</>` directly. When `cellValue` is a boolean, React silently ignores it (this is by design — it's how `{condition && <Component />}` avoids rendering "false"). The editable cell path already handled this correctly via its `valueToString` helper, but the read-only path did not.

## Test plan

- [x] Added `DefaultCellRenderer.test.tsx` covering boolean true/false, string, number, null, and undefined values
- [x] `pnpm turbo typecheck --filter=@osdk/react-components` passes
- [x] `pnpm turbo test --filter=@osdk/react-components` — all tests pass
- [x] Manual verification: render an `ObjectTable` with a boolean property column and confirm values display as "true"/"false"

### DIFF
--- a/packages/react-components/src/object-table/__tests__/DefaultCellRenderer.test.tsx
+++ b/packages/react-components/src/object-table/__tests__/DefaultCellRenderer.test.tsx
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


packages/react-components/src/object-table/__tests__/DefaultCellRenderer.test.tsx
nfdiop, line 1: nice set of tests!

nfdiop (approved):

PR #3216, opened 2026-05-11
Show selected filter values even when they have zero matching rows

nfdiop:
## Summary

- When `initialFilterStates` contains a value absent from aggregation results (e.g. a saved filter for a specialty with zero admissions today), the filter was active but invisible — users couldn't see or remove the selection
- Adds `mergeAggregationValues` utility that appends missing selected values with `count: 0` to aggregation data
- Wired into `ListogramFilterInput`, `MultiSelectFilterInput`, and `SingleSelectFilterInput`
- Adds "With initial filter states" storybook story demonstrating saved state hydration

## Test plan

- [x] Unit tests for `mergeAggregationValues` (4 cases)
- [ ] Component integration tests for all three filter input types
- [x] Visual verification in Storybook "With initial filter states" story 
- [x] `pnpm turbo typecheck --filter=@osdk/react-components` passes
- [x] `pnpm turbo test --filter=@osdk/react-components` passes


packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1999,3 +2009,164 @@ export const NoValueRendering: Story = {
...
+          `// "Research", "Chief Scientist", and "Berlin" are not in the current
+// dataset — they represent saved selections with zero matching rows
+// today. The filter list still shows them so users can see and clear them.
+const savedStates = new Map([
+  // Direct property filters
+  ["department", { type: "EXACT_MATCH", values: ["Marketing", "Research"] }],
+  ["jobTitle-multi", { type: "SELECT", selectedValues: ["Marketing Manager", "Chief Scientist"] }],
+  ["locationCity-single", { type: "SELECT", selectedValues: ["Berlin"] }],
+  // Linked property filters
+  ["linkedProperty:lead:department", {
```
nfdiop, line 2152: It's surprising: I don't see these in the active clause

packages/react-components/src/filter-list/hooks/usePropertyAggregation.ts
```diff
@@ -125,8 +131,14 @@ export function usePropertyAggregation<
     [countData, propertyKey, options?.limit, options?.sortBy],
   );
 
+  const activeValues = options?.activeValues ?? EMPTY_ACTIVE_VALUES;
+  const mergedData = useMemo(
```
KuberSethi: I believe we might be bypassing the sorting above if we do this here, meaning these would always end up at the bottom. That might be intentional (I think zeros always being at the bottom is fine) but just calling out as some users may expect to see things properly sort.

packages/react-components/src/filter-list/utils/mergeAggregationValues.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import type { PropertyAggregationValue } from "../types/AggregationTypes.js";
+
+/**
+ * Ensures every selected value appears in the aggregation data. Values present
+ * in `selectedValues` but absent from `data` (e.g. a saved filter for a value
+ * that currently has zero matching rows) are appended with `count: 0` so the
+ * UI always renders them as visible, checkable items.
+ */
+export function mergeAggregationValues(
```
KuberSethi: Question: do we always want to do this? I think it might make sense to not do this for empty string values or null values but up to you, just flagging as we handle the `isNull` explicitly in the other logic for this component.

KuberSethi (dismissed):
LGTM, just two concerns which may or may not be applicable and could also be addressed in a FLUP

KuberSethi (approved):

packages/react-components/src/filter-list/hooks/usePropertyAggregation.ts
nfdiop: Good flag!

packages/react-components/src/filter-list/utils/mergeAggregationValues.ts
nfdiop: Thanks for flagging! Handling the empty strings like the rest now

PR #3217, opened 2026-05-11
Add auto-size (fit to width) toggle to PDF viewer

olliejenningsPT:
## Summary

- Adds a **fit-to-width** toggle button (ZoomToFit icon) to the PDF viewer toolbar
- When active, the PDF scales to fill the container width and re-fits automatically on window/container resize via `ResizeObserver`
- Manual zoom in/out automatically disables auto-size
- Exposed as `initialAutoSize` prop on `BasePdfViewer`, `PdfViewer`, and `usePdfViewerInstance`

## Demo

https://github.com/user-attachments/assets/e385e0dc-7ff3-43b0-a5b2-4947716f292a

## Test plan

- [x] Toggle auto-size button in toolbar — PDF should scale to fit container width
- [x] Resize browser window while auto-size is active — PDF should re-fit
- [x] Zoom in/out while auto-size is active — auto-size should deactivate and button should un-highlight
- [x] Toggle auto-size back on after manual zoom — should re-fit to width
- [x] Open sidebar while auto-size is active — PDF should re-fit to narrower container
- [x] `initialAutoSize={true}` prop — should start in fit-to-width mode
- [ ] Existing tests pass (`pnpm turbo test --filter=@osdk/react-components`)


packages/react-components/src/pdf-viewer/hooks/usePdfViewerState.ts
```diff
@@ -102,6 +115,21 @@ export function usePdfViewerState({
     }
   }, [core.pdfViewerRef, rotation]);
 
+  // Re-apply page-width after rotation changes while auto-size is active.
+  // Rotation changes the effective page width, so auto-size must re-fit.
+  const prevRotationRef = useRef(rotation);
+  useEffect(function reapplyAutoSizeAfterRotation() {
```
nfdiop, line 121: nice comment!

packages/react-components/src/pdf-viewer/hooks/usePdfViewerSync.ts
```diff
@@ -18,40 +18,96 @@ import type { PDFDocumentProxy } from "pdfjs-dist";
...
+
+    const observer = new ResizeObserver(() => {
+      pdfViewer.currentScaleValue = PAGE_WIDTH_SCALE_VALUE;
+    });
+    observer.observe(container);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, [containerRef, pdfViewerRef, autoSize, document]);
```
nfdiop, line 110: why pass the document here?

packages/react-components/src/pdf-viewer/hooks/usePdfViewerState.ts
```diff
@@ -117,12 +145,18 @@ export function usePdfViewerState({
...
+  }, [core.scale, core.setScale, core.setAutoSize]);
 
   const zoomOut = useCallback(() => {
+    core.setAutoSize(false);
     core.setScale(Math.max(core.scale - SCALE_STEP, MIN_SCALE));
-  }, [core.scale, core.setScale]);
+  }, [core.scale, core.setScale, core.setAutoSize]);
+
+  const toggleAutoSize = useCallback(() => {
+    core.setAutoSize(!core.autoSize);
```
nfdiop, line 158: Is it possible to do core.setAutoSize(prev => !prev);

packages/react-components/src/pdf-viewer/hooks/usePdfViewerSync.ts
```diff
@@ -18,40 +18,96 @@ import type { PDFDocumentProxy } from "pdfjs-dist";
...
+
+  // ResizeObserver: re-apply page-width on container resize when auto-size is active
+  useEffect(function observeContainerResize() {
+    const container = containerRef.current;
+    const pdfViewer = pdfViewerRef.current;
+    if (container == null || pdfViewer == null || !autoSize) {
+      return;
+    }
+
+    const observer = new ResizeObserver(() => {
```
nfdiop, line 102: question: should we debounce or use requestAnimationFrame?

nfdiop (approved):

packages/react-components/src/pdf-viewer/hooks/usePdfViewerSync.ts
olliejenningsPT: ResizeObserver is specifically designed for reacting to change of sizes of DOM and is more efficient than rAF

packages/react-components/src/pdf-viewer/hooks/usePdfViewerSync.ts
olliejenningsPT: The document is in the dependency array so the effects re-run when the pdfjs viewer is initialized (since pdfViewerRef.current goes from null to non-null when document loads, but refs don't trigger re-renders).

packages/react-components/src/pdf-viewer/hooks/usePdfViewerState.ts
olliejenningsPT: Yes, but the interface types setAutoSize as (autoSize: boolean) => void rather than Dispatch<SetStateAction<boolean>>. We'd need to widen the type to support the functional updater form

PR #3220, opened 2026-05-11
Fix filter list dropdown positioning and visibility

olliejenningsPT:
## Summary

- **Flip dropdowns above trigger** when there isn't enough space below within the filter sidebar container. Previously, portaled popups used the viewport as their collision boundary, so they'd overflow the sidebar without flipping.
- **Hide dropdowns when anchor scrolls out of view** — when the filter input scrolls off-screen inside the sidebar's scrollable area, the dropdown disappears via `visibility: hidden` (leveraging `@base-ui/react`'s built-in `data-anchor-hidden` attribute).

### How it works

**Flipping:** A `FilterListBoundaryContext` provides the sidebar's `.expandedContent` element as a custom `collisionBoundary` to all dropdown positioners within the filter list. This tells `@base-ui/react`'s positioning engine to flip based on the sidebar boundary instead of the viewport.

**Hiding:** Added `&[data-anchor-hidden] { visibility: hidden }` to positioner CSS for Combobox, Select, SearchableMenu, and ExcludeDropdown. `@base-ui/react` already detects when the anchor scrolls out of its clipping ancestors and sets this attribute — we just needed to style it.

## Demo

https://github.com/user-attachments/assets/1dde52af-53cd-44a2-aa71-9971b1f13151

## Test plan

- [x] Open filter sidebar with several filters visible
- [x] Expand a filter near the bottom of the sidebar — dropdown should flip above the trigger
- [x] Scroll the sidebar while a dropdown is open — dropdown should disappear when the input scrolls out of view
- [x] Verify filters near the top still open downward normally
- [x] Verify the "Add filter" popover and Exclude/Keep dropdown also flip and hide correctly

### DIFF
--- a/packages/react-components/src/base-components/combobox/Combobox.module.css
+++ b/packages/react-components/src/base-components/combobox/Combobox.module.css
@@ -312,6 +312,13 @@
      navigation. Keep visible focus on the trigger/input or highlighted option
      instead of drawing a browser-default ring around this wrapper. */
   outline: none;
+
+  /* Hide the popup when its anchor scrolls out of view (e.g. inside a
+     scrollable sidebar). @base-ui/react sets this attribute automatically
+     via floating-ui's hide middleware. */
+  &[data-anchor-hidden] {
+    visibility: hidden;
+  }
 }
 
 .osdkComboboxPopup {
--- a/packages/react-components/src/filter-list/base/ExcludeDropdown.module.css
+++ b/packages/react-components/src/filter-list/base/ExcludeDropdown.module.css
@@ -50,6 +50,10 @@
 
 .positioner {
   z-index: var(--osdk-surface-z-index-3);
+
+  &[data-anchor-hidden] {
+    visibility: hidden;
+  }
 }
 
 .popup {
--- a/packages/react-components/src/filter-list/base/FilterListBoundaryContext.ts
+++ b/packages/react-components/src/filter-list/base/FilterListBoundaryContext.ts
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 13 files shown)


packages/react-components/src/base-components/combobox/Combobox.module.css
```diff
@@ -312,6 +312,13 @@
      navigation. Keep visible focus on the trigger/input or highlighted option
      instead of drawing a browser-default ring around this wrapper. */
   outline: none;
+
+  /* Hide the popup when its anchor scrolls out of view (e.g. inside a
+     scrollable sidebar). @base-ui/react sets this attribute automatically
+     via floating-ui's hide middleware. */
+  &[data-anchor-hidden] {
+    visibility: hidden;
```
KuberSethi, line 320: Normally I'd say this breaks accessibility but are we good because of "@base-ui/react sets this attribute automatically"? Just double checking

packages/react-components/src/filter-list/base/FilterListBoundaryContext.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import type React from "react";
+import { createContext, useContext } from "react";
+
+/**
+ * Provides the filter list container element so that dropdown positioners
+ * inside the filter list can use it as their collision boundary, flipping
+ * above the trigger when space below is insufficient within the sidebar.
+ */
+const FilterListBoundaryContext = createContext<Element | null>(null);
```
KuberSethi, line 25: Do we really need a context for this?

KuberSethi:
Hmm it's a little jarring on the first scroll (0:02 - 0:04 of your video) but I think this is definitely better than what we have today!  

I'd also recommend adding some tests in a FLUP

packages/react-components/src/filter-list/base/ExcludeDropdown.module.css
```diff
@@ -50,6 +50,10 @@
 
 .positioner {
   z-index: var(--osdk-surface-z-index-3);
+
+  &[data-anchor-hidden] {
+    visibility: hidden;
```
KuberSethi, line 55: Is this the same as before where we're good on accessibility concerns bc of base-ui?

KuberSethi (approved):

packages/react-components/src/filter-list/base/ExcludeDropdown.module.css
olliejenningsPT: yes exactly

PR #3222, opened 2026-05-11
Fix broken design tokens and improve dark theme compatibility

Leolide:
## Summary
Fixes that don't overlap with #3188 (which covers PDF viewer and datetime-picker day-color):

- Replace hardcoded gray colors (`custom-color-light-gray-2`, `custom-color-gray-*`) with semantic surface tokens in table header menu
- Fix table row hover/active using `--osdk-intent-primary-rest` instead of `--osdk-intent-primary-hover`
- Change calendar "today" indicator from border to bold font-weight for better theme adaptability
- Extract `getLocalToday()` helper for consistent midnight-normalized dates in DateCalendar
- Remove `min-height: 1lh` from editable cell that caused layout issues

## Test plan
- [ ] Verify table header menu backgrounds and borders work in both light and dark themes
- [ ] Verify table row hover/active states render correctly
- [ ] Verify datetime picker calendar "today" shows bold text instead of border
- [ ] Verify editable cells don't have extra vertical space


packages/react-components/src/action-form/fields/DateCalendar.module.css
```diff
@@ -81,7 +81,7 @@
 }
 
 .calendarDayButton.calendarToday {
-  border: var(--osdk-datetime-calendar-today-border);
+  font-weight: var(--osdk-datetime-calendar-today-font-weight);
```
nfdiop: So I reverted this because the user mentioned they wanted a border similar to Blueprint

packages/react-components/src/action-form/fields/DateCalendar.tsx
```diff
@@ -163,3 +160,11 @@ export default function DateCalendar({
     />
   );
 }
+
+function getLocalToday(): Date {
```
nfdiop: Removed this earlier to address an user issue that they wanted the time to be selected

packages/react-components/src/action-form/fields/DateCalendar.tsx
```diff
@@ -101,20 +101,13 @@ export default function DateCalendar({
...
-  );
-  const isTodaySelectable = isDateInRange(todayMidnight, min, max);
+  const today = getLocalToday();
+  const isTodaySelectable = isDateInRange(today, min, max);
   const handleTodayClick = useCallback(() => {
     if (!isTodaySelectable) {
       return;
     }
-    onSelect(new Date());
+    onSelect(getLocalToday());
```
ziyunp: Nit: Can just use `today` variable directly?

packages/react-components/src/action-form/fields/DateCalendar.tsx
ziyunp: @nfdiop is it ok to put this back then?

packages/react-components/src/object-table/EditableCell.module.css
```diff
@@ -137,7 +137,6 @@
 
 .readonlyDisplayCell {
   padding: calc(var(--osdk-surface-spacing) * 0.5) 0;
-  min-height: 1lh;
```
ziyunp, line 147: Issue (blocking): This is needed for a styling bug where the cell collapses when there's no value

packages/react-components/src/action-form/fields/DateCalendar.module.css
ziyunp: @nfdiop just to confirm, we want the border instead of font-weight right?

nfdiop: yes

packages/react-components/src/action-form/fields/DateCalendar.tsx
nfdiop: it's not ok: should be dropped

nfdiop (approved):

PR #3234, opened 2026-05-12
Mikhailprice/tagging interaction fix

mikhailprice:
### Summary                                                         
                                                                                                                                                                                                
  Fixes a NotFoundError: Failed to execute 'removeChild' on 'Node' crash in PdfViewer when zooming or switching documents. The root cause was annotation overlays being portaled into           
  pdfjs-owned DOM nodes — when pdfjs re-rendered or tore down a page, React's portal cleanup raced against pdfjs's own DOM mutations.
                                                                                                                                                                                                
 ### Changes                                                         

  - Render annotations as siblings, not portals. PdfAnnotationOverlay now positions itself absolutely over each page using measured page geometry, instead of being portaled into the pdfjs page
   div. React owns the overlay subtree end-to-end, so pdfjs page tear-downs no longer trigger removeChild failures.
  - AnnotationPortalTarget shape change. Replaces container: HTMLDivElement with left / top / width / height / transform. Consumers position via CSS rather than reading from a pdfjs DOM ref.  
  - Coalesced remeasure. Annotation remeasures triggered by scalechanging, rotationchanging, and container ResizeObserver events are batched into a single requestAnimationFrame tick, avoiding 
  O(annotated pages) getBoundingClientRect reads on every pinch-zoom frame.                                                                                                                     
                                                                                                                                                                                                
### Test plan                                                                                                                                                                                     
                                                                  
  - Pinch-zoom a document with annotations — no console errors, overlays track page geometry                                                                                                    
  - Switch between documents rapidly — no removeChild crash
  - Rotate document — overlays reposition correctly                                                                                                                                             
  - Resize the viewer container — overlays follow  



https://github.com/user-attachments/assets/f649fce3-6be3-42db-a06b-004e73647ce2


palantirtech:
Thanks for your interest in [palantir/osdk-ts](https://github.com/palantir/osdk-ts), @mikhailprice! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/react-components/src/pdf-viewer/hooks/usePdfAnnotationPortals.ts
```diff
@@ -36,38 +40,117 @@ export function usePdfAnnotationPortals(
...
-        return;
+        return null;
+      }
+      const div = pageView.div as HTMLDivElement;
+      const divRect = div.getBoundingClientRect();
+      if (divRect.width <= 0 || divRect.height <= 0) {
+        return null;
       }
+      return {
+        pageNumber,
```
olliejenningsPT, line 65: **Potential transient mismatch during zoom/rotation**

`remeasureAll` is subscribed to `scalechanging` and `rotationchanging`, which fire *before* pdfjs re-renders pages. At that point, `viewport.transform` may already reflect the new scale/rotation, but `div.getBoundingClientRect()` still reflects the old layout. This means the overlay container is sized/positioned from old measurements, while the `transform` inside is for the new scale — briefly misplacing annotations within the overlay.

`pagerendered` fires after and corrects everything, so this is transient. It's probably imperceptible during a zoom animation, but worth being aware of. If it does cause a flash, one option would be to skip the remeasure on these events and rely solely on `pagerendered` + `pagecleanup`.

packages/react-components/src/pdf-viewer/hooks/usePdfAnnotationPortals.ts
```diff
@@ -36,38 +40,117 @@ export function usePdfAnnotationPortals(
...
+    const readContainer = (): {
+      rect: DOMRect;
+      el: HTMLElement;
+    } | null => {
+      const scrollContainer = pdfViewer.container as HTMLElement | null;
+      if (scrollContainer == null) {
+        return null;
+      }
+      return {
+        rect: scrollContainer.getBoundingClientRect(),
```
olliejenningsPT, line 87: Nit: the coordinate math `divRect.left - containerRect.left + scrollContainer.scrollLeft` is correct only if the scroll container has zero `border` and zero `padding` (since `position: absolute` positions relative to the padding edge, but `getBoundingClientRect` returns the border box). The current CSS has neither, so this works — but a brief comment here noting that assumption would prevent a subtle positioning bug if someone later adds padding/border to `.scrollContainer`.

packages/react-components/src/pdf-viewer/components/__tests__/PdfViewerAnnotationLayer.test.tsx
```diff
@@ -35,6 +35,12 @@ function createAnnotation(
   } as PdfAnnotation;
 }
 
+// pdfjs's viewport.transform for a non-rotated page: scales and flips y.
+// Applied to (x, y): (scale*x, scale*(pageHeight - y)).
+function pageTransform(scale: number, pageHeight: number): number[] {
```
olliejenningsPT, line 40: The `pageTransform` helper only models the non-rotated case `[scale, 0, 0, -scale, 0, pageHeight*scale]`. Since the switch to `applyTransform` is specifically motivated by handling rotation correctly, it would be valuable to add at least one test with a 90° rotation transform (e.g. `[0, scale, -scale, 0, pageWidth*scale, 0]`) to verify the `Math.min`/`Math.abs` logic in `computeRectStyle` when axes swap.

olliejenningsPT (commented):
Overall this is a well-motivated fix — the portal-into-pdfjs-DOM approach was fundamentally fragile. A few observations below.

packages/react-components/src/pdf-viewer/hooks/usePdfAnnotationPortals.ts
```diff
@@ -36,38 +40,110 @@ export function usePdfAnnotationPortals(
...
+          if (remeasured != null) {
+            next.push(remeasured);
+          }
+        }
+        return next;
       });
     };
 
     eventBus.on(PAGE_RENDERED_EVENT, handlePageRendered);
+    eventBus.on("scalechanging", remeasureAll);
```
olliejenningsPT, line 140: Each `scalechanging` runs `remeasureAll` synchronously and calls `getBoundingClientRect` on every annotated page. Pinch zoom fires the event repeatedly, so on long docs we end up doing N layout reads per tick. Worth wrapping `remeasureAll` in `requestAnimationFrame` to coalesce.

olliejenningsPT (approved):

PR #3236, opened 2026-05-12
Add Storybook PR previews

nfdiop:
## Summary
- Add a Storybook preview workflow for same-repo PRs
- Deploy previews to `gh-pages` under `/storybook/pr-<number>/`
- Upsert a PR comment with preview status and clean up previews when PRs close
- Preserve PR previews during main Storybook deploys

## Why gh-pages instead of CI artifacts?

Blueprint uses CircleCI's `store_artifacts`, which serves static files at browsable URLs. GitHub Actions' `upload-artifact` only produces zip downloads — there's no equivalent browsable URL. To get a clickable Storybook preview link on PRs, we deploy to `gh-pages` under `storybook/pr-<number>/`.

An alternative would be `actions/deploy-pages` with named environments per PR, but that's significantly more complex for the same result. The gh-pages approach is simple and matches our existing Storybook and docs deploy pattern.

**Known limitation:** until this PR merges, main's `deploy-storybook.yml` doesn't have the preview-preservation logic yet, so a main Storybook deploy will wipe active PR previews. Once merged, main deploys will preserve `pr-*` directories.

## Testing
- Pushed a temporary red background on BaseForm to verify the preview deployed correctly and reflected component changes — confirmed the preview URL rendered the change, then reverted
- `pnpm turbo typecheck --filter=@osdk/react-components-storybook`
- `pnpm --filter "@osdk/*" transpileBrowser`
- `STORYBOOK_BASE_PATH=/osdk-ts/storybook/pr-1234/ pnpm --filter @osdk/react-components-storybook build`
- Parsed workflow YAML with Ruby
- `git diff --check`


.github/workflows/deploy-storybook.yml
```diff
@@ -15,6 +15,9 @@ permissions:
 jobs:
   build-and-deploy:
     runs-on: ubuntu-latest
+    concurrency:
+      group: storybook-gh-pages
```
nfdiop, line 21: This is to avoid a case where two jobs could be working on gh-pages and one could overwrite the changes made by the other.

olliejenningsPT (approved):

PR #3237, opened 2026-05-12
Fix dropdown scroll blocking in editable ObjectTable

olliejenningsPT:
## Summary

- Base UI's `Select.Root` defaults `modal=true`, which locks document scroll when the select popup is open
- Added a `modal` prop to `DropdownField` (defaulting `true` to preserve action-form behavior)
- Pass `modal={false}` from `DropdownCellField` in the object-table so dropdowns don't block page scrolling
- The `PortalDismissLayer` overlay is also gated on `modal` to avoid rendering a full-viewport overlay unnecessarily

## Demo

https://github.com/user-attachments/assets/6a926d38-e3dd-447b-8a8e-29046de3adc5

## Test plan

- [x] Open a dropdown in an editable ObjectTable row — verify the page/table scrolls while the dropdown is open
- [x] Verify clicking outside the dropdown still dismisses it
- [x] Verify action-form dropdowns (inside `<label>`s) still work correctly with scroll lock and dismiss layer

### DIFF
--- a/packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
+++ b/packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
@@ -130,6 +130,22 @@ describe("DropdownField", () => {
         expect(screen.queryByRole("option", { name: "Alice" })).toBeNull();
       });
     });
+
+    it("does not render the portal dismiss layer when modal is false", async () => {
+      render(
+        <DropdownField value={null} items={STRING_ITEMS} modal={false} />,
+      );
+
+      fireEvent.click(screen.getByRole("combobox"));
+
+      await vi.waitFor(() => {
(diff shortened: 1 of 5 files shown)


packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -249,6 +265,27 @@ describe("DropdownField", () => {
       });
     });
 
+    it("does not render the portal dismiss layer when modal is false", async () => {
```
nfdiop: should we just keep one test?

nfdiop (dismissed):

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
olliejenningsPT: sorted

nfdiop (approved):

PR #3238, opened 2026-05-12
Add scoped object select

nfdiop:
## Summary
- add `objectSet` as an alternative data source for `OBJECT_SELECT`
- make `objectType` and `objectSet` mutually exclusive in the public props and form definitions
- add runtime coverage, type coverage, docs, Storybook example, sandbox demo, and changeset

## Verification
- `pnpm vitest run src/action-form/__tests__/ObjectSelectField.test.tsx`
- `pnpm turbo typecheck --filter=@osdk/react-components --filter=@osdk/e2e.sandbox.peopleapp --filter=@osdk/react-components-storybook`
- `pnpm turbo check-api --filter=@osdk/react-components`
- `DPRINT_CACHE_DIR=/private/tmp/dprint-cache pnpm exec dprint check ...`

Screenshot 2026-05-12 at 15 54 00

### DIFF
--- a/packages/react-components/docs/ActionForm.md
+++ b/packages/react-components/docs/ActionForm.md
@@ -97,7 +97,7 @@ const fields = [
 
 ### Scoped object select fields
 
-`OBJECT_SELECT` can load options from either an object type or a pre-scoped object set. Pass `objectType` for an unfiltered selector, or pass `objectSet` to limit selectable options. Do not pass both. Search text is applied within the object set, and the current value is not automatically cleared when it is outside that set.
+`OBJECT_SELECT` can load options from either an object type or a pre-scoped object set. Pass `objectType` for an unfiltered selector, or pass `objectSet` to limit selectable options. The two are mutually exclusive. Search text is applied within the object set, and the current value is not automatically cleared when it is outside that set.
 
 ```tsx
 import { $, Employee, updateEmployee } from "@YourApp/sdk";
--- a/packages/react-components/docs/ActionForm.md
+++ b/packages/react-components/docs/ActionForm.md
@@ -95,6 +95,41 @@ const fields = [
 <ActionForm actionDefinition={updateEmployee} formFieldDefinitions={fields} />;
 ```
 
+### Scoped object select fields
+
+`OBJECT_SELECT` can load options from either an object type or a pre-scoped object set. Pass `objectType` for an unfiltered selector, or pass `objectSet` to limit selectable options. Do not pass both. Search text is applied within the object set, and the current value is not automatically cleared when it is outside that set.
+
+```tsx
+import { $, Employee, updateEmployee } from "@YourApp/sdk";
+import { useMemo } from "react";
+
+function UpdateEmployeeForm() {
+  const marketingEmployees = useMemo(
+    () => $(Employee).where({ department: "Marketing" }),
+    [],
+  );
+
+  const fields = [
+    {
+      fieldKey: "manager",
+      label: "Marketing manager",
+      fieldComponent: "OBJECT_SELECT",
+      fieldComponentProps: {
+        objectSet: marketingEmployees,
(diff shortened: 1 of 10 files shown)


packages/react-components/src/action-form/__tests__/ObjectSelectField.test.tsx
```diff
@@ -104,16 +109,145 @@ function mockLoadedState(
...
+    renderObjectSelect({
+      objectSet: createMockObjectSet(),
+      value: makeMockObject(999, "Outside Scope") as never,
+      onChange,
+    });
+
+    expect(onChange).not.toHaveBeenCalled();
+  });
+
+  it("prefers the object set when malformed props include both sources", () => {
```
nfdiop: since it's more specific

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -403,17 +394,36 @@ export interface ObjectSetFieldProps<T extends ObjectTypeDefinition>
...
+
 /**
  * Object select field props for selecting object instances from the ontology.
  * Used for action parameters that accept a single object or multiple objects.
  *
  * Extends DropdownFieldProps with props that ObjectSelectField
  * manages internally (items, search, filtering) omitted from the public surface.
  */
-export interface ObjectSelectFieldProps<
+export type ObjectSelectFieldProps<
```
nfdiop, line 423: break but this is experimental

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
```diff
@@ -243,6 +237,49 @@ function renderFieldComponent(
...
+  );
+  const sharedProps = {
+    id: fieldDefinition.fieldKey,
+    value: narrowToOsdkObject(value),
+    onChange,
+    placeholder: fieldDefinition.placeholder,
+    error,
+  };
+
+  // Keep the two source branches separate so TypeScript preserves the
```
KuberSethi: Hmm is this really how it works even though the JSX here is identical? This seems a big brittle, is there any way to improve things here?

packages/react-components/docs/ActionForm.md
```diff
@@ -95,6 +95,41 @@ const fields = [
 <ActionForm actionDefinition={updateEmployee} formFieldDefinitions={fields} />;
 ```
 
+### Scoped object select fields
+
+`OBJECT_SELECT` can load options from either an object type or a pre-scoped object set. Pass `objectType` for an unfiltered selector, or pass `objectSet` to limit selectable options. Do not pass both. Search text is applied within the object set, and the current value is not automatically cleared when it is outside that set.
```
KuberSethi: Can we enforce the "Do not pass both" in our typing at all?

KuberSethi (commented):
Generally looks good, just two questions!

packages/react-components/docs/ActionForm.md
nfdiop: So we enforce in our typing already by using a discriminated union. Let me update this comment

packages/react-components/src/action-form/fields/FormFieldRenderer.tsx
nfdiop: Good call! Dropped!

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -403,17 +394,36 @@ export interface ObjectSetFieldProps<T extends ObjectTypeDefinition>
   emptyMessage?: string;
 }
 
+type ObjectSelectDataSource<Q extends ObjectTypeDefinition> =
```
nfdiop, line 397: mutually exclusive types since it makes more sense to me that it should be one or the other

KuberSethi (approved):

PR #3239, opened 2026-05-12
Block PRs that add never-published @osdk packages

ericanderson:
## Summary

- The Release workflow publishes from GitHub Actions but cannot create brand-new npm package names on its own — npm requires a human with publish rights to claim a scoped name the first time. Without this guard, the first release after a PR that introduces a new package fails on main.
- This adds a check workflow that runs on PRs to main and blocks merge until every publishable workspace package has at least a placeholder version on npm.
- A companion `scripts/publish-placeholder.sh` lets an authorized human claim a new `@osdk/*` name with a `0.0.0` empty tarball so the check passes.

## What's in here

- `scripts/check-packages-published.sh` — walks the pnpm workspace, filters out packages with `"private": true`, and sends one HEAD per name to `registry.npmjs.org` through a single keep-alive curl invocation. 404 → fail with an actionable message pointing at the placeholder script. Bash 3.2-compatible (`jq` + `curl` only, both preinstalled on `ubuntu-latest`).
- `scripts/publish-placeholder.sh @osdk/<name>` — validates the `@osdk/` scope and segment, refuses to clobber an existing name, then publishes `{"name":"...","version":"0.0.0"}` from a `mktemp -d` with `npm publish --access public`.
- `.github/workflows/check-new-packages.yml` — wires the check into PRs targeting `main`. Mark required in branch protection after merge.

## Test plan

- [x] `./scripts/check-packages-published.sh` passes against current main (45 packages, all known to npm).
- [x] Failure path tested by adding a temporary `packages/__fake/package.json` with a never-claimed `@osdk/__fake-not-published-zzzzzzz` name — script exits 1 and names the missing package.
- [x] All four reject paths of `publish-placeholder.sh` exercised: no arg, non-`@osdk/` prefix, invalid name characters, name already exists on npm.
- [x] Both scripts run cleanly under stock macOS `/bin/bash 3.2.57`.
- [ ] CI run on this PR shows the new `Check new packages` job passing.


ericanderson (commented):
## Code review

### Findings — medium

**1. Curl has no timeout.** Both `scripts/check-packages-published.sh:84` and `scripts/publish-placeholder.sh:37` invoke `curl -sS …` without `--connect-timeout` / `--max-time`. A slow or hung npm response will block the CI job until the workflow's 5-minute `timeout-minutes` ceiling, producing a flaky-feeling failure instead of fast feedback. Suggest `--connect-timeout 10 --max-time 30` on both. Cheap.

**2. `WORKSPACE_DIRS` is a hand-mirrored copy of `pnpm-workspace.yaml`.** The comment at `check-packages-published.sh:18` flags this, and right now the script and pnpm agree exactly (`pnpm m ls --depth -1 --json` → 45 publishable packages, script → 45 publishable packages, zero diff). But the mirror is enforced by convention only — if someone later adds e.g. `apps/*` to `pnpm-workspace.yaml` without touching the script, new publishable packages there silently skip the guard, which is the exact failure mode this PR exists to prevent. Two options:

   - In CI, after `pnpm install`, diff `pnpm m ls --depth -1 --json | jq …` against the script's CHECKED list and fail loudly on mismatch.
   - Drop the hand-mirror entirely and have the script call `pnpm m ls` (requires `pnpm install` in the workflow, ~30s slower).

   The first is the cheaper safety net.

### Findings — minor / nice-to-have

**3. Workflow has no explicit `permissions:` block.** Defaults to whatever the repo default is — for a read-only registry probe, `permissions: { contents: read }` is the principle-of-least-privilege baseline and matches modern GH guidance.

**4. No `concurrency:` group on the new workflow.** Rapid PR pushes will run overlapping ~5-second checks. Standard pattern (already used in `release.yml`) is `concurrency: ${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true`.

**5. `publish-placeholder.sh` publishes immediately with no `Confirm? [y/N]` or `--dry-run`.** The operator is presumed to know what they're doing, but a dry-run flag would make the script safer to test and easier to demo.

**6. Stray stdout `echo` interleaves with stderr error block.** `check-packages-published.sh:108` writes a blank line to stdout right before the stderr error block — minor display oddity in CI logs. Change to `echo >&2`.

### Things checked and OK

- Bash 3.2 portability verified by running the script under stock macOS `/bin/bash 3.2.57` (both success and failure paths).
- Single curl invocation with `next` separators reuses one TLS session — drops ~15s of sequential handshakes to ~5s total.
- Status-count guard at line 86 catches truncated curl output (e.g., mid-stream connection drop) instead of silently miscounting.
- Workflow scoped to PRs whose **base** is `main` — feature-to-feature PRs don't trigger it.
- Executable bit (`100755`) committed in the tree for both scripts.
- `find -mindepth 2 -maxdepth 3` only picks up one real depth-3 file (`benchmarks/tests/primary/package.json`), which is itself `private: true` and filtered out anyway.
- `jq '.private // false'` correctly handles missing, null, false, and true.
- Placeholder script's regex `^[a-z0-9][a-z0-9._-]*$` matches npm's name rules (can't start with `_` or `.`, lowercase only).
- All four rejection paths of the placeholder script exercised locally.
- `scripts/publish-placeholder.sh` no-ops on existing names (exit 0 with "Nothing to do.") rather than 403-ing on `npm publish`.

### Verdict

Ship after addressing **(1)** at minimum. **(2)** is a real correctness gap worth one extra CI step. The rest are polish.

ericanderson (commented):
..

ericanderson:
## Verification

End-to-end exercise of the new guard on this PR itself:

1. **Failure path.** Added `@osdk/z-test-package` as a bare, non-private workspace package (commit `e628018b6`) — a name that had never been claimed on npm. On the next CI run, the `Check new packages / Verify publishable packages exist on npm` job [failed](https://github.com/palantir/osdk-ts/actions/runs/25740067501/job/75587657387) with exit 1 and the annotation `The following publishable packages have never been published to npm:` listing `@osdk/z-test-package`.

2. **Bootstrap.** Ran `./scripts/publish-placeholder.sh @osdk/z-test-package` from a shell with `@osdk` publish rights. The script validated the `@osdk/` scope, probed the registry and confirmed the name was free, wrote `{"name":"@osdk/z-test-package","version":"0.0.0"}` plus a copy of the repo's `.npmrc` into a `mktemp -d`, and ran `npm publish --access public`. `@osdk/z-test-package@0.0.0` is now reserved on npm.

3. **Happy path.** Re-ran the same failed CI run (same commit `e628018b6`, identical workflow code and package set, no code change). The job [succeeded](https://github.com/palantir/osdk-ts/actions/runs/25740067501/job/75591426535) — the registry probe returned 200 for `@osdk/z-test-package` instead of 404. This isolates the behavior change to npm's registry state, not to anything in the repo.

4. **Cleanup.** Removed the test package and its monorepolint archetype (commit `5ccda14ba`). The `@osdk/z-test-package@0.0.0` placeholder is intentionally left on npm — leaving a claimed name costs nothing, and means any future re-run of this exact validation would no-op rather than re-trip the guard.

Net result: the guard correctly blocks merge when a PR introduces an unclaimed `@osdk/*` name, and unblocks immediately once a human runs the placeholder publish. Both halves of the loop observed on real CI runs in this PR.

scripts/publish-placeholder.sh
```diff
@@ -0,0 +1,76 @@
...
+    ;;
+esac
+
+tmp_dir=$(mktemp -d)
+trap 'rm -rf "${tmp_dir}"' EXIT
+
+cat > "${tmp_dir}/package.json" <<EOF
+{
+  "name": "${PACKAGE_NAME}",
+  "version": "0.0.0"
```
ssanjay1, line 60: nit: could we use the version number from the code as well? Thinking of cases where people rename packages e.g. experimental/my-package -> myPackage

ericanderson: Are you okay with people needing to bump a release for first ship then?

ssanjay1: Mm I see what you're saying, let's keep at 0.0 for now I do think its nice to use that to indicate this is like a throwaway version. I don't know that the case i mentioned will be super prevalent

ssanjay1 (approved):

PR #3242, opened 2026-05-12
allow filter list renderValue to return a reactnode

KuberSethi:
widen `renderValue` return type on filter list property and static-values filter definitions from `string` to `ReactNode` so callers can render rich content (e.g. `<SmartUserAnchor>`) for filter values

• update `renderValue` type on `PropertyFilterDefinition` and `StaticValuesFilterDefinition`, plus propagation through `ListogramInput`, `MultiSelectInput`, `SingleSelectInput` and their wrappers
• search matching uses the rendered value when it's a string (existing behavior) and falls back to the raw value when renderValue returns non-string jsx

### DIFF
--- a/.changeset/filter-list-render-value-react-node.md
+++ b/.changeset/filter-list-render-value-react-node.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": minor
+---
+
+Widen `renderValue` return type on FilterList property and static-values filter definitions from `string` to `ReactNode` so callers can render custom React components (e.g. avatars, anchors) for filter values. When `renderValue` returns a non-string `ReactNode`, search matching falls back to the raw value.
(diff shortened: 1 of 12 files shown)


olliejenningsPT:
**Missing ReactNode tests for MultiSelectInput and SingleSelectInput**

The PR adds a `ListogramInput renderValue (ReactNode)` test block that validates both JSX rendering and search fallback with a ReactNode-returning `renderValue`. However, `MultiSelectInput` and `SingleSelectInput` received the same `comboboxFilter` logic changes but have no corresponding ReactNode tests. The existing tests for these components only cover a string-returning `renderValue`.

The `SingleSelectInput` test comment notes jsdom limitations for dropdown items, but the search/filter logic change in `comboboxFilter` should be testable without opening a dropdown (e.g. verifying the filter function itself handles non-string returns correctly).

KuberSethi:
good call. extracted the inline comboboxFilter out and added unit tests too

olliejenningsPT (approved):

PR #3243, opened 2026-05-12
fix $as on objects fetched with $loadPropertySecurityMetadata

KuberSethi:
calling $as on an object loaded with $loadPropertySecurityMetadata: true threw at createOsdkInterface because parseWhenSecuritiesLoaded returned a shallow copy, leaving the wire object the holder symbols still pointed at via [UnderlyingOsdkObject] without ObjectDefRef / ClientRef / PropertySecuritiesRef

• #2352 introduced a `{ ...rawObject }` shallow copy in parseWhenSecuritiesLoaded which broke the pre-existing invariant that holder[UnderlyingOsdkObject] === holder; downstream consumers (getDollarAs, get$linkForInterface, createOsdkInterface, $clone) silently saw undefined symbols
• drop the copy and mutate the input in place, matching the convention already established by fixObjectPropertiesInPlace; one less allocation per fetch and the divergent-state class of bug is gone

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import type { InterfaceMetadata } from "@osdk/api";
+import type { InterfaceMetadata, PropertySecurity } from "@osdk/api";
 import { extractNamespace } from "../../internal/conversions/extractNamespace.js";
 import type { FetchedObjectTypeDefinition } from "../../ontology/OntologyProvider.js";
 import { get$linkForInterface } from "./getDollarLink.js";
@@ -25,6 +25,10 @@ import {
   UnderlyingOsdkObject,
 } from "./InternalSymbols.js";
 import type { ObjectHolder } from "./ObjectHolder.js";
+
+type PropertySecuritiesMap = {
+  [propName: string]: PropertySecurity[] | PropertySecurity[][];
+};
 
 /** @internal */
 export function createOsdkInterface<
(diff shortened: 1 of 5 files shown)


packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -351,7 +351,7 @@ function parseWhenSecuritiesLoaded(
     return { parsedObject: rawObject, clientPropertySecurities: undefined };
   }
 
-  const parsedObject: SimpleOsdkProperties = { ...rawObject };
```
KuberSethi, line 354: This is the actual fix.

ziyunp (dismissed):

packages/client/src/object/convertWireToOsdkObjects/createOsdkObject.ts
```diff
@@ -62,17 +62,17 @@ const specialPropertyTypes = new Set(
...
     },
   },
   "$link": {
-    get: function(this: ObjectHolder) {
+    get(this: ObjectHolder) {
       return get$link(this);
     },
   },
   "$clone": {
-    value: function(
```
nihalbhatnagar, line 75: whyd we do all this btw

KuberSethi: eslint

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -70,7 +74,14 @@ export function createOsdkInterface<
         enumerable: false,
       },
       "$propertySecurities": {
-        value: underlying.$propertySecurities,
+        value: remapPropertySecuritiesForInterface(
```
bryantpark04, line 77: so the spread issue was the first fix, and we also weren't remapping the prop securities before?

bryantpark04 (approved):

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
nihalbhatnagar: yes

PR #3245, opened 2026-05-12
Support value types in maker-experimental

ethanai22:


packages/client.unstable/src/generated/type-registry-marketplace/index.ts
ethanai22, line 1: idk why the diff renders like this, I just deleted the old file and the new one comes from the local conjure script

aep000 (approved):

PR #3248, opened 2026-05-12
fix rdp staleness on cross-variant base-field write

KuberSethi:
when two queries hold the same object and one fetches with withProperties while the other fetches without, the no-rdp variant's fresh base fields were getting stitched onto the rdp variant's old derived value, leaving the derived field inconsistent with the inputs sitting next to it

• detect base-field change during cross-variant propagation and refetch the rdp variant instead of merging stale derived values forward
• add baseFieldsChanged helper in rdpFieldOperations that ignores rdp fields on both sides
• cross-variant only in v1; self-write merge path unchanged

### DIFF
--- a/.changeset/rdp-staleness-fix.md
+++ b/.changeset/rdp-staleness-fix.md
@@ -1,0 +1,5 @@
+---
+"@osdk/client": patch
+---
+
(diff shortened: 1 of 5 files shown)


KuberSethi:
Needs repro

PR #3252, opened 2026-05-13
fix filter list crash when object type is undefined

KuberSethi:
filter list with a listogram input crashed with "cannot read properties of undefined (reading 'apiName')" when objectType arrived as undefined (e.g. async-loaded or unresolved type defs). Specifically we're seeing users suppress our types and wire in nullish inputs (especially LLMs) so this PR aims to make that less painful

• widen `useOsdkAggregation`, `useOsdkObjects`, `useOsdkFunction`, and `useOsdkMetadata` to accept an undefined type/queryDef and short-circuit to a loading state instead of crashing
• widen `FilterListProps.objectType` to optional and propagate the change through all property/input prop interfaces so the chain compiles when the source type is unresolved
• inputs continue to render their existing loading skeletons via the underlying hooks' isLoading path; no runtime behavior changes when objectType is provided


packages/react-components/src/filter-list/FilterInput.tsx
```diff
@@ -95,6 +95,9 @@ function FilterInputInner<Q extends ObjectTypeDefinition>({
           <div data-unsupported="true">Custom filter missing renderInput</div>
         );
       }
+      if (objectType === undefined) {
```
ziyunp, line 108: question: The CUSTOM case returns `<></>` when objectType is undefined while the other branches propagate undefined down to inputs that render their loading skeletons via usePropertyAggregation. Would it make sense to widen `CustomFilterInputRendererProps.objectType` to `Q | undefined` and let the custom renderer decide what to show during the async type-load window? Returning an empty fragment means custom filters blink out while everything else shows a skeleton.

ziyunp:
Q (blocking): Ignoring the rendering of loading skeleton, shouldn't the hook be disabled if objectType is undefined?

PR #3261, opened 2026-05-13
fix number range filter emitting decimals for integer properties

KuberSethi:
clicking or dragging a histogram bucket on an integer property like length of stay was committing fractional min/max values (e.g. 13.4, 14.45) that broke downstream filters

• plumb the property type via useOsdkMetadata in NumberRangeFilterInput and pass isInteger to the base layer
• add a clampCommitRange hook on RangeInputConfig that snaps bucket boundaries to integer half-open semantics, uses Math.floor on the closed last bucket, and returns undefined to cancel commits on buckets that contain zero integers
• gate clickToFilter on metadata being loaded so the aggregation/metadata race can't commit a fractional value before the property type is known


olliejenningsPT (commented):
Thanks for the fix! Left two comments on potential issues.

packages/react-components/src/filter-list/inputs/NumberRangeFilterInput.tsx
```diff
@@ -50,6 +50,13 @@ function NumberRangeFilterInputInner<Q extends ObjectTypeDefinition>({
     : undefined;
   const includeNull = filterState?.includeNull;
 
+  const { metadata } = useOsdkMetadata(objectType);
+  const propertyType = metadata?.properties?.[propertyKey]?.type;
+  const isInteger = propertyType === "integer"
```
olliejenningsPT, line 55: The PR description mentions gating `clickToFilter` on metadata being loaded, but this isn't implemented. If the aggregation data loads before `useOsdkMetadata` resolves, the histogram is clickable while `isInteger` is still `false` (because `propertyType` is `undefined`), so a click during that window will commit fractional values — the exact bug this PR fixes.

Something like passing `clickToFilter={clickToFilter && metadata != null}` down to `NumberRangeInput` would close the race.

packages/react-components/src/filter-list/inputs/NumberRangeFilterInput.tsx
```diff
@@ -68,14 +75,16 @@ function NumberRangeFilterInputInner<Q extends ObjectTypeDefinition>({
 
   const handleRangeChange = useCallback(
     (minValue: number | undefined, maxValue: number | undefined) => {
+      const round = (v: number | undefined): number | undefined =>
+        isInteger && v !== undefined ? Math.round(v) : v;
```
olliejenningsPT: `Math.round` can include integers from adjacent buckets. Bucket boundaries are `minNum + i * (range / 20)`, so for e.g. a bucket `[9.7, 14.55)`:

- `Math.round(14.55)` → 15, but 15 is **not** in `[9.7, 14.55)` — the last integer in the bucket is 14.

Similarly for min, `Math.round` can round *down* past the bucket start (e.g. `Math.round(14.3)` → 14, but 14 < 14.3 so it's not in `[14.3, ...)`).

`Math.ceil` for min and `Math.floor` for max would be more correct:
```ts
const round = (v: number | undefined, mode: 'min' | 'max'): number | undefined =>
  isInteger && v !== undefined
    ? (mode === 'min' ? Math.ceil(v) : Math.floor(v))
    : v;
```

This also has a nice semantic for manual text input — typing `14.55` as a max on an integer property gives 14 (the last integer ≤ 14.55).

packages/react-components/src/filter-list/inputs/NumberRangeFilterInput.tsx
KuberSethi: good catch, fixed, gated clickToFilter so the histogram isn't clickable until the property type is known

packages/react-components/src/filter-list/inputs/NumberRangeFilterInput.tsx
KuberSethi: yeah ceil/floor is right, switched

olliejenningsPT (approved):

PR #3266, opened 2026-05-14
[Component] Pass row's pending edits to getFieldComponentProps

ziyunp:
EditFieldConfig.getFieldComponentProps now receives a second optional argument: the row's pending cell edits keyed by columnId. Editor configuration can react to other in-progress edits within the same row (e.g. dropdown items that change once another column is edited).

### DIFF
--- a/packages/react-components/src/object-table/DefaultCellRenderer.tsx
+++ b/packages/react-components/src/object-table/DefaultCellRenderer.tsx
@@ -23,12 +23,32 @@ import { isAsyncCellData } from "./utils/AsyncCellData.js";
 import { isCellEditable } from "./utils/editableUtils.js";
 import { getCellId } from "./utils/getCellId.js";
 import { shouldShowEditableCell } from "./utils/shouldShowEditableCell.js";
+import type { CellEditInfo } from "./utils/types.js";
 
 function toDisplayValue(value: unknown): React.ReactNode {
   if (typeof value === "boolean") {
     return String(value);
   }
   return value as React.ReactNode;
+}
+
+// Returns the subset of `cellEdits` belonging to `rowId`, re-keyed by columnId.
+// Returns `undefined` (stable reference) when the row has no pending edits, so
+// that `React.memo` on `EditableCell` can skip re-renders of unedited rows when
+// edits change elsewhere in the table.
+function filterCellEditsToRow<TData extends RowData>(
+  cellEdits: Record<string, CellEditInfo<TData, unknown>> | undefined,
+  rowId: string,
+): Record<string, CellEditInfo<TData, unknown>> | undefined {
+  if (!cellEdits) return undefined;
+  let result: Record<string, CellEditInfo<TData, unknown>> | undefined;
+  for (const edit of Object.values(cellEdits)) {
+    if (edit.rowId === rowId) {
+      result ??= {};
+      result[edit.columnId] = edit;
+    }
+  }
+  return result;
 }
 
 export function renderDefaultCell<TData extends RowData>(
@@ -84,6 +104,7 @@ export function renderDefaultCell<TData extends RowData>(
     : editedValue?.newValue;
   const validationError = meta.validationErrors?.get(cellId);
   const isRowFocused = meta.focusedRowId === rowId;
+  const rowCellEdits = filterCellEditsToRow(cellEdits, rowId);
 
   return (
     <EditableCell<TData>
@@ -92,7 +113,7 @@ export function renderDefaultCell<TData extends RowData>(
       cellId={cellId}
       dataType={columnMeta?.dataType}
       editFieldConfig={columnMeta?.editFieldConfig}
-      cellEdits={cellEdits}
+      rowCellEdits={rowCellEdits}
       onCellEdit={meta.onCellEdit}
       onCellValidationError={meta.onCellValidationError}
       clearCellValidationError={meta.clearCellValidationError}
--- a/packages/react-components/src/object-table/DefaultCellRenderer.tsx
+++ b/packages/react-components/src/object-table/DefaultCellRenderer.tsx
@@ -92,6 +92,7 @@ export function renderDefaultCell<TData extends RowData>(
       cellId={cellId}
       dataType={columnMeta?.dataType}
       editFieldConfig={columnMeta?.editFieldConfig}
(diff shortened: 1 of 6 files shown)


olliejenningsPT (commented):
### Performance: passing the full `cellEdits` map defeats `React.memo` on `EditableCell`

`EditableCell` is wrapped in `React.memo`, but now every instance receives the entire table-wide `cellEdits` map, which gets a new object reference on every cell edit anywhere in the table. This means `React.memo` can no longer prevent re-renders — every `EditableCell` re-renders on every single edit, even if nothing changed for its row.

The `rowCellEdits` memo inside correctly returns `undefined` for rows with no edits (referentially stable), preventing downstream `dropdownFieldProps`/`datePickerFieldProps` recomputation. But the component render itself (including JSX reconciliation) still runs.

**Suggestion:** compute the row-scoped edits here in `DefaultCellRenderer` (which already has `rowId` and `cellEdits`) and pass only the filtered map as the prop. Then:
- Rows with **no** edits receive `undefined` → stable reference → `React.memo` skips the render entirely
- Rows **with** edits still re-render, but only those rows

This also lets you remove the `rowCellEdits` `useMemo` inside `EditableCellInner`, simplifying that component.

---

### Missing tests

This PR adds a new user-facing API surface (`edits` argument to `getFieldComponentProps`) but no tests. At minimum, a test verifying that `getFieldComponentProps` receives the correct row-scoped edits when another cell in the same row has been edited would be valuable — especially since the filtering logic (iterating `Object.values(cellEdits)` and matching by `rowId`) is non-trivial enough to warrant coverage.

packages/react-components/src/object-table/DefaultCellRenderer.tsx
```diff
@@ -92,6 +92,7 @@ export function renderDefaultCell<TData extends RowData>(
       cellId={cellId}
       dataType={columnMeta?.dataType}
       editFieldConfig={columnMeta?.editFieldConfig}
+      cellEdits={cellEdits}
```
olliejenningsPT: **Performance: passing the full `cellEdits` map defeats `React.memo` on `EditableCell`**

`EditableCell` is wrapped in `React.memo`, but now every instance receives the entire table-wide `cellEdits` map, which gets a new object reference on every cell edit anywhere in the table. This means `React.memo` can no longer prevent re-renders — every `EditableCell` re-renders on every single edit, even if nothing changed for its row.

The `rowCellEdits` memo inside correctly returns `undefined` for rows with no edits (referentially stable), preventing downstream memo recomputation. But the component render itself (including JSX reconciliation) still runs unnecessarily.

**Suggestion:** compute the row-scoped edits here in `DefaultCellRenderer` (which already has `rowId` and `cellEdits`) and pass only the filtered map as the prop. Then:
- Rows with **no** edits receive `undefined` → stable reference → `React.memo` skips the render entirely
- Rows **with** edits still re-render, but only those rows

This also lets you remove the `rowCellEdits` `useMemo` inside `EditableCellInner`, simplifying that component.

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -64,6 +64,12 @@ export interface EditableCellProps<TData extends RowData, CellValue = unknown> {
...
   validateEdit?: (value: unknown) => Promise<string | undefined>;
   editFieldConfig?: EditFieldConfig<TData>;
+  /**
+   * Full table cellEdits map (keyed by `rowId-columnId`). Filtered to the
+   * current row internally and forwarded to
+   * `EditFieldConfig#getFieldComponentProps`.
+   */
+  cellEdits?: Record<string, CellEditInfo<TData, unknown>>;
   isRowFocused?: boolean;
 }
```
olliejenningsPT, line 76: **Missing tests**

This PR adds a new user-facing API surface (`edits` argument to `getFieldComponentProps`) but no tests. At minimum, a test verifying that `getFieldComponentProps` receives the correct row-scoped edits when another cell in the same row has been edited would be valuable — especially since the filtering logic (iterating `Object.values(cellEdits)` and matching by `rowId`) is non-trivial enough to warrant coverage.

packages/react-components/src/object-table/EditableCell.tsx
```diff
@@ -258,22 +265,36 @@ function EditableCellInner<TData extends RowData, CellValue = unknown>({
...
+    if (!cellEdits) return undefined;
+    let result: Record<string, CellEditInfo<TData, unknown>> | undefined;
+    for (const edit of Object.values(cellEdits)) {
+      if (edit.rowId === rowId) {
+        result ??= {};
+        result[edit.columnId] = edit;
+      }
+    }
+    return result;
+  }, [cellEdits, rowId]);
```
olliejenningsPT: **Minor: iterates all edits in the table per cell**

`Object.values(cellEdits)` iterates every pending edit across all rows to find ones matching this cell's `rowId`. This is O(total edits) per cell per render. In practice, the number of pending edits is usually small so this is fine — but if you move the filtering to `DefaultCellRenderer` as suggested on the other comment, you'd centralize this work per cell-render call rather than per component render, and you could drop this memo entirely.

packages/react-components/src/object-table/DefaultCellRenderer.tsx
ziyunp: Good catch, moved the row-scoped filtering up into `DefaultCellRenderer` (new `filterCellEditsToRow` helper). Rows with no edits now receive a stable `undefined`, so `React.memo` on `EditableCell` can skip them. Dropped the inner `useMemo` in `EditableCell` accordingly.

packages/react-components/src/object-table/EditableCell.tsx
ziyunp: Added a test in `DefaultCellRenderer.test.tsx` that verifies `getFieldComponentProps` receives only the current row's pending edits when other cells in different rows have been edited.

packages/react-components/src/object-table/EditableCell.tsx
ziyunp: Done, filtering is now centralized in `DefaultCellRenderer` and the inner `useMemo` has been removed, so the iteration happens once per cell render in the renderer rather than inside the memoized component.

olliejenningsPT (approved):

PR #3268, opened 2026-05-14
Fix object set invalidation when the derived property changes

ziyunp:
## Root cause

`ObjectSetQuery.invalidateObjectType` only consulted `#objectTypes` (base + union/intersect/subtract types). When an action edited a type that the query's RDP traversed (e.g. `Office` reached via `pivotTo('officeLink')` from an `Employee` set), the query didn't recognize itself as affected and stayed stale. `ListQuery` got the equivalent fix in #3193 but the sibling `ObjectSetQuery` was overlooked.

## Fix

Mirror `ListQuery`'s pattern: lazily populate `#rdpInvalidationSet` on first fetch via `getObjectTypesThatInvalidate`, and consult it in `invalidateObjectType` alongside `#objectTypes`. Regression test in `ActionApplication.test.ts` asserts the post-action RDP value, not just that a notification fired.


packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -216,14 +220,37 @@ export class ObjectSetQuery extends BaseListQuery<
...
+          wireObjectSet,
+        );
       this.sortingStrategy = new OrderBySortingStrategy(
         resultType.apiName,
         this.#operations.orderBy,
       );
+      this.#rdpInvalidationSet = invalidationSet;
+    }
+
+    // When the orderBy branch above didn't run but the query has RDPs, we
```
KuberSethi: nit: this comment is likely to go stale (I typically don't like comments referring to the line placement of other code/logic)

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -216,14 +220,37 @@ export class ObjectSetQuery extends BaseListQuery<
...
+    ) {
+      try {
+        const wireObjectSet = getWireObjectSet(this.#composedObjectSet);
+        const { invalidationSet } = await getObjectTypesThatInvalidate(
+          this.store.client[additionalContext],
+          wireObjectSet,
+        );
+        this.#rdpInvalidationSet = invalidationSet;
+      } catch {
+        // Some ObjectSet shapes (static, reference) aren't walkable. Fall
```
KuberSethi: Hmm I don't like the idea of try-catching to handle this, could we try to visit all types and handle their behavior explicitly?

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -461,7 +488,10 @@ export class ObjectSetQuery extends BaseListQuery<
     objectType: string,
     changes: Changes | undefined,
   ): Promise<void> => {
-    if (this.#objectTypes.has(objectType)) {
+    if (
+      this.#objectTypes.has(objectType)
+      || !!this.#rdpInvalidationSet?.has(objectType)
```
KuberSethi: nit: we should be consistent in terms of `!!` / how we evaluate truthiness

KuberSethi (commented):
Thanks for doing this!! Generally looks good just a few small comments

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
ziyunp: Do you prefer `?? false` ?

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
ziyunp: This is the same pattern in aggregate and list query. I moved the try catch into a private method to followe aggregate query pattern

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
ziyunp: Switched to `?? false` to match `ListQuery`'s `revalidateObjectType` at line 381 (the parallel codepath we're mirroring).

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
ziyunp: removed

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
KuberSethi: I'm not that opinionated just that we should be consistent where possible!

KuberSethi (approved):

PR #3269, opened 2026-05-14
[Component] Add renderEmptyState

ziyunp:
Updated default empty and error states:

Screenshot 2026-05-14 at 23 20 35

Screenshot 2026-05-14 at 23 20 39

Example of custom empty state:
Screenshot 2026-05-14 at 22 54 19


KuberSethi (approved):
nice!

PR #3271, opened 2026-05-14
Add override options to createClientFromWriteableClient and guard batch actions on transactional clients

nihalbhatnagar:
## Summary
- `createClientFromWriteableClient` now accepts an optional `options` argument to override `transactionId`, `baseUrl`, `ontologyRid`, or `tokenProvider` from the source client, so callers can derive a related client without inheriting every field.
- `applyAction` throws an invariant when batch actions are invoked on a client with an active `transactionId`, since batch actions are not supported for staged edit functions.

## Test plan
- [ ] CI passes (typecheck, lint, tests, changeset check)
- [ ] Manual: confirm batch action call on a transactional client now throws with the expected message
- [ ] Manual: confirm `createClientFromWriteableClient(writeable, { baseUrl })` honors the override

### DIFF
--- a/packages/client/src/createClientFromWriteableClient.ts
+++ b/packages/client/src/createClientFromWriteableClient.ts
@@ -17,8 +17,23 @@
 import { additionalContext, type Client } from "./Client.js";
 import { createClientWithTransaction } from "./createClient.js";
 
+/**
+ * Experimental helper method to create a new client instantiated with a transaction id
+ * Will extract the transactionId, URL, Ontology RID, and token from the original client
+ * unless overridden.
+ *
+ * @param writeableClient A client initiated with a transaction id, such as a WriteableClient
+ * @param options A set of options to override from the provided client.
+ * @returns Client instantiated on a transaction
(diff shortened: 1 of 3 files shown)


packages/client/src/createClientFromWriteableClient.ts
```diff
@@ -17,8 +17,23 @@
...
+ * @param writeableClient A client initiated with a transaction id, such as a WriteableClient
+ * @param options A set of options to override from the provided client.
+ * @returns Client instantiated on a transaction
+ */
 export function createClientFromWriteableClient(
   writeableClient: Client,
+  options?: {
+    transactionId?: string;
+    baseUrl?: string;
+    ontologyRid?: string | Promise<string>;
```
KuberSethi, line 34: why this typing?

nihalbhatnagar: Thats how its typed in createClient

KuberSethi (approved):

PR #3276, opened 2026-05-15
[Component] Fix custom display of dropdown cell in ObjectTable

ziyunp:
Fix ObjectTable dropdown field custom rendering

Before:

When custom render cell is passed, in edit mode, the dropdown options differ from the custom renderer.
 
https://github.com/user-attachments/assets/310a82e6-5e78-411a-b7f6-3cdfdfe1fe74

After: 

Apply itemToStringLabel to dropdown cell renderer

https://github.com/user-attachments/assets/026ef528-73e1-4dc4-8a42-0742893ae994


packages/react-components/src/object-table/components/DropdownCellField.tsx
```diff
@@ -35,16 +42,20 @@ function DropdownCellFieldInner({
...
 }: DropdownCellFieldProps): React.ReactElement {
   const portalRef = useRegisterPortal();
+  const { itemToStringLabel } = fieldComponentProps;
 
   if (!isRowFocused) {
     return (
       <ReadonlyDisplayCell
-        inputValue={inputValue}
+        inputValue={itemToStringLabel
+          ? itemToStringLabel(value)
```
olliejenningsPT, line 57: Concern: `itemToStringLabel` is now invoked with `value` directly, but `value` can be `null`/`undefined` when the cell is unset. The public contract `(item: V) => string` suggests V ranges over `items`, so most users won't write a null-safe formatter (e.g. `(item: boolean) => item ? "Yes" : "No"` would return `"No"` for `null`). Before this PR the readonly cell rendered `valueToString(null) === ""`; now it renders whatever the user's formatter produces for `null`, which is a regression for existing callers.

Suggestion:
```ts
inputValue={value != null && itemToStringLabel
  ? itemToStringLabel(value)
  : inputValue}
```

Separate but related: in the `isMultiple` case `value` is an array, and calling `itemToStringLabel(arrayValue)` formats the array as if it were a single item. Not used anywhere yet, but the PR is the natural place to guard it (`Array.isArray(value) ? value.map(itemToStringLabel).join(", ") : itemToStringLabel(value)`).

packages/react-components/src/object-table/__tests__/EditableCell.test.tsx
```diff
@@ -150,6 +148,111 @@ describe("EditableCell", () => {
...
+        screen.getByRole("option", { name: "true" }).getAttribute(
+          "aria-selected",
+        ),
+      ).toBe("true");
+      expect(
+        screen.getByRole("option", { name: "false" }).getAttribute(
+          "aria-selected",
+        ),
+      ).toBe("false");
+    });
```
olliejenningsPT, line 219: Test coverage gap: the new tests all run with `isRowFocused: true`, exercising only the `DropdownField` branch. The other half of the fix — `itemToStringLabel(value)` in the `ReadonlyDisplayCell` branch — has no tests. In particular a case with `currentValue: null` would lock down the behavior raised in the `DropdownCellField.tsx` comment.

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -250,6 +249,100 @@ const defaultEmployeeColumns: ColumnDefinition<Employee>[] = [
...
+      fieldComponent: "DATE_PICKER",
+      getFieldComponentProps: () => ({
+        showTime: false,
+        placeholder: "Select date...",
+      }),
+    },
+  },
+  {
+    locator: { type: "property", id: "isRemote" },
+    renderCell: (object: Osdk.Instance<Employee>) => {
```
olliejenningsPT, line 326: This `renderCell` is dead while the row is in edit mode — `useColumnDefs.ts` only calls `renderCell` when `!isEditable`, so once editing begins the cell text is driven entirely by `itemToStringLabel`. Since this is the canonical example users will copy, consider dropping the `renderCell` (the new `itemToStringLabel` covers display) or adding a comment explaining the split, otherwise readers will likely assume `renderCell` formats both states.

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -1520,20 +1562,7 @@ return (
...
-      {
-        locator: { type: "property", id: "emailPrimaryWork" },
-        editable: true,
-      },
-      {
-        locator: { type: "property", id: "jobTitle" },
-        editable: true,
-      },
-    ],
+    columnDefinitions: editableColumnDefinitions,
```
olliejenningsPT, line 1567: Scope creep: previously this story had 3 editable columns (`fullName`, `emailPrimaryWork`, `jobTitle`). Switching to `editableColumnDefinitions` makes it a 7-column story including date pickers and the new boolean. If that's intentional cleanup, fine — but the `source: code` block below was only partially updated to match (no `firstInternStartDate`/`firstFullTimeStartDate`), so the docs panel and the actual rendered story are now inconsistent.

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -1969,3 +2050,5 @@ return (
     );
   },
 };
+
+export default meta;
```
olliejenningsPT: Nit: was this intentional? Storybook works with the default export anywhere in the file, but moving it ~1800 lines away from the `meta` declaration looks like a refactor artifact. Easier for future readers if it sits next to `meta`.

packages/react-components/src/object-table/__tests__/EditableCell.test.tsx
```diff
@@ -21,16 +21,14 @@ import {
...
+import type { EditFieldConfig } from "../utils/types.js";
 
 describe("EditableCell", () => {
   afterEach(() => {
     cleanup();
     vi.clearAllMocks();
-  });
-
-  afterAll(() => {
     vi.useRealTimers();
```
olliejenningsPT, line 32: This is a real latent fix — without it the `unmount` test below leaks fake timers into any subsequent test — but it's unrelated to the boolean dropdown bug and not called out in the PR description. Worth a sentence in the description or splitting into its own commit.

.changeset/ten-humans-film.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/react-components-storybook": patch
+"@osdk/react-components": patch
+---
+
+Fix boolean dropdown field in ObjectTable
```
olliejenningsPT: The fix is broader than booleans: the prior code passed a stringified value to the dropdown, which would also break number-typed item lists (`Object.is("1", 1) === false`) and any other non-string item type. Consider rewording so callers understand the scope, e.g. "Fix custom value rendering in ObjectTable dropdown cells for non-string item types (booleans, numbers, etc.)".

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
ziyunp: This is intentional, renderCell is used when the cell is not in edit mode and we fallback to use the default cell editor in edit mode. Using itemToLabel allow user to configure the options in dropdown to match their custom renderer if they change the displayed value

packages/react-components/src/object-table/__tests__/EditableCell.test.tsx
ziyunp: Added this while debugging the new tests because the `unmount` test below leaks fake timers into subsequent tests.

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
ziyunp: No, reverting.

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
ziyunp: Updated

.changeset/ten-humans-film.md
ziyunp: Updated

packages/react-components/src/object-table/__tests__/EditableCell.test.tsx
ziyunp: Added tests

packages/react-components/src/object-table/components/DropdownCellField.tsx
ziyunp: Not adding value != null guard because I want to allow user to customise the rendering of null value. Widened the type to V | undefined instead.

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -1453,6 +1477,25 @@ export const EditableTable: Story = {
...
+        return "No Value";
+      }
+      return object.isRemote ? "Yes" : "No";
+    },
+    editable: true,
+    editFieldConfig: {
+      fieldComponent: "DROPDOWN",
+      getFieldComponentProps: () => ({
+        items: [true, false],
+        itemToStringLabel: (item) =>
```
olliejenningsPT: this `item` can now be undefined, we this should be item | undefined. My worry is if consumers expecting it to be never undefined.

olliejenningsPT (approved):

PR #3277, opened 2026-05-15
[Component] Fix select all bugs

ziyunp:
- Fix "select all" + scroll: newly-loaded rows are now checked and the header checkbox stays in sync. `onRowSelection` refires in uncontrolled mode with the expanded id list as new pages load.
- Add `onRowSelectionChanged(change)` callback that delivers a `RowSelectionChange` payload with `selectedRowIds`, `selectedRows`, `isSelectAll`, and a derived `objectSet` (full underlying set on "select all", otherwise narrowed by `$primaryKey`).
- Deprecate `onRowSelection` in favor of `onRowSelectionChanged`. The legacy callback continues to fire for backwards compatibility.

Bugs:

1. onRowSelection called with isAllSelected when select all is toggled off
 
https://github.com/user-attachments/assets/ad50776c-983b-4e0e-bf0f-15f40ae2f032

Fixed:

https://github.com/user-attachments/assets/908ce668-79a8-4dd4-9181-2fc1bc5c0541

2. When all rows selected and scroll, the new rows are not selected.

https://github.com/user-attachments/assets/59511b68-04ce-489f-a76b-598c7ad114c8

Fixed, also added a new onRowSelectionChanged and deprecate the old callback. The new callback passes the selected object set so the client app can use it in other parts of the app, e.g. Action Form

https://github.com/user-attachments/assets/5e6af7c4-1f9e-4f84-9e2e-f908fe3b357d


packages/react-components/docs/ObjectTable.md
```diff
@@ -653,6 +654,50 @@ function EmployeesTable() {
...
+  objectType={Employee}
+  selectionMode="multiple"
+  onRowSelectionChanged={({
+    selectedRowIds,
+    selectedRows,
+    isSelectAll,
+    objectSet,
+  }) => {
+    // selectedRowIds: PrimaryKeyType<Employee>[] — current selection
+    // selectedRows:   loaded row instances matching selectedRowIds
```
nfdiop, line 674: Curious why pass also `selectedRowIds` when already passing `selectedRows`?

packages/react-components/src/object-table/hooks/useRowSelection.ts
```diff
@@ -155,21 +219,38 @@ export function useRowSelection<
...
       isControlled,
-      onRowSelection,
+      fireSelectionCallbacks,
       rowSelectionState,
       lastSelectedRowIndex,
     ],
   );
 
+  // Refire callbacks when uncontrolled "select all" is active and new rows arrive.
+  useEffect(() => {
```
nfdiop, line 240: nit:   useEffect(function syncSelectAllOnDataChange()

packages/react-components/src/object-table/hooks/useRowSelection.ts
```diff
@@ -88,42 +112,87 @@ export function useRowSelection<
...
+  // Dedupes refires when "select all" is active and data grows.
+  const lastFiredAllSelectedIdsRef = useRef<string | null>(null);
+
+  const fireSelectionCallbacks = useCallback(
+    (ids: PrimaryKeyType<Q>[], isSelectAll: boolean) => {
+      onRowSelection?.(ids, isSelectAll);
+      if (onRowSelectionChanged) {
+        const currentData = data ?? [];
+        const selectedKeySet = new Set(ids.map(id => String(id)));
+        const instances = currentData.filter(item =>
```
nfdiop, line 154: nit: maybe 
```
const instances = isSelectAll
    ? currentData
    : currentData.filter(item => selectedKeySet.has(String(item.$primaryKey)));
```

packages/react-components/src/object-table/hooks/useRowSelection.ts
```diff
@@ -88,42 +112,87 @@ export function useRowSelection<
...
     if (!isControlled) {
+      setInternalIsAllSelected(newIsAllSelected);
       setInternalRowSelection(getRowSelectionState(newSelectedRows));
     }
-    onRowSelection?.(newSelectedRows, true);
-  }, [enableRowSelection, data, isAllSelected, isControlled, onRowSelection]);
+    lastFiredAllSelectedIdsRef.current = newIsAllSelected
+      ? JSON.stringify(newSelectedRows)
+      : null;
+    fireSelectionCallbacks(newSelectedRows, newIsAllSelected);
```
nfdiop, line 182: If fireSelectionCallbacks is not stable it could cause issues. Maybe try using useEventCallback on the deps
```
 const onRowSelectionRef = useRef(onRowSelection);
  onRowSelectionRef.current = onRowSelection;
```

nfdiop (approved):

packages/react-components/docs/ObjectTable.md
ziyunp: yeah, wasn't sure if we should still pass selectedRowIds, main reason was for backward compatibility with the deprecated callback. I can remove it in a flup

PR #3283, opened 2026-05-15
Fix maker experimental imports

aep000:


packages/maker-experimental/src/cli/marketplaceSerialization/semver.ts
ethanai22, line 1: rename this file?

packages/maker-import/src/generate/writeImportedOntology.ts
ethanai22: we can delete this type

aep000: Done

ethanai22 (approved):

PR #3285, opened 2026-05-15
fix storybook missing docx-preview and postal-mime deps

KuberSethi:
storybook dev fails to resolve `docx-preview` and `postal-mime` because they were added as deps of `@osdk/react-components` (in #3211, for the new docx/email viewers) but never added to the storybook package. `storybook build` in CI gets away with transitive resolution; `storybook dev` doesn't.

• declare `docx-preview` and `postal-mime` as devDependencies of `@osdk/react-components-storybook`
• lockfile updated via `pnpm install`


olliejenningsPT:
`docx-preview` should be dropped from this PR. PR #3281 (`6400c8bad` — "Remove DocxViewer") landed on main after this branch's base and removed `docx-preview` from `@osdk/react-components`'s `package.json` along with the `DocxViewer`/`./experimental/docx-viewer` entrypoint. Current main's `packages/react-components/package.json` no longer lists it, and `grep -r docx-preview packages/` (excluding CHANGELOG/lockfile) returns no results — nothing in `react-components` or `react-components-storybook` imports it anymore.

Once rebased onto current main, adding `docx-preview` here would introduce a devDependency with zero consumers. Suggest:

- removing the `docx-preview` line from `packages/react-components-storybook/package.json`
- re-running `pnpm install` to drop it from the lockfile
- updating the changeset summary to mention only the email viewer (currently says "docx/email viewer stories")

The `postal-mime` addition is correct and should stay.

ssanjay1 (dismissed):

olliejenningsPT (approved):

PR #3287, opened 2026-05-15
filter-list: add baseObjectSet and applyLinkedFilters for linked filter narrowing

KuberSethi:
direct property facets dual-aggregate against a base scope so values present in base but absent under narrowing render as count=0 greyed-out ghost rows

• new optional `baseObjectSet` prop on `FilterList` paired with a `useDualScopeAggregation` hook that owns the dual-scope decision; linked-property facets pivot from `baseObjectSet` so their value lists don't collapse
• new `applyLinkedFilters(base, definitions, filterStates)` helper builds the narrowed `ObjectSet` via forward+reverse pivot, plus a new optional `reverseLinkName` field on `LinkedPropertyFilterDefinition`
• new `onFilterStatesChanged` callback emits the full filter state map so callers can use it as a stable cache key — `filterClause` never reflects linked-filter selections


packages/react-components/src/filter-list/utils/filterStateToWhereClause.ts
```diff
@@ -309,11 +311,31 @@ export function buildWhereClause<Q extends ObjectTypeDefinition>(
       }
 
       case "LINKED_PROPERTY": {
```
KuberSethi: This is not an exhaustive all-encompassing implementation but this is a poor mans version of the real thing (which is quite complex) and solves out immediate problems

packages/react-components/src/filter-list/hooks/__tests__/useFilterListState.test.tsx
```diff
@@ -135,7 +136,7 @@ describe("useFilterListState", () => {
     expect(result.current.whereClause).toEqual({ name: "John" });
   });
 
-  it("calls onFilterClauseChanged on mount with initial where clause", () => {
+  it("does not fire onFilterClauseChanged on mount", () => {
```
nfdiop: Hm, this is a breaking behaviour

packages/react-components/src/filter-list/FilterListApi.ts
```diff
@@ -91,9 +91,8 @@ export interface FilterListProps<Q extends ObjectTypeDefinition> {
   objectType: Q;
 
   /**
```
nfdiop, line 93: Ok my main take is that we can simplify the code by making a breaking change and dropping the whereClause so we only have the objectSet like in [here](https://github.com/palantir/osdk-ts/blob/1aee8e77e4d0651175bd3a6d45bc89dd0f2fde5b/packages/react-components/src/action-form/FormFieldApi.ts#L445).

packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -195,12 +198,13 @@ export function FilterList<Q extends ObjectTypeDefinition>(
         filterState={filterState}
         onFilterStateChanged={onFilterStateChanged}
         whereClause={perFilterWhereClauses.get(filterKey)
-          ?? ({} as WhereClause<Q>)}
+          ?? (EMPTY_WHERE as WhereClause<Q>)}
```
nfdiop: nice fix!

packages/react-components/src/filter-list/utils/filterStateToWhereClause.ts
```diff
@@ -253,14 +240,19 @@ export interface PropertyTypeInfo {
...
+ * Linked-property filters produce `$reverseLink` link entries; direct filters
+ * produce plain SDK-shaped property clauses.
+ */
 export function buildWhereClause<Q extends ObjectTypeDefinition>(
   definitions: Array<FilterDefinitionUnion<Q>> | undefined,
   filterStates: Map<string, FilterState>,
   propertyTypes?: Map<string, PropertyTypeInfo>,
   excludeFilterKey?: string,
-): WhereClause<Q> {
+): Record<string, unknown> {
```
nfdiop: Hm, I don't think we should do this. We're getting a weaker type and creating a fake whereClause. Instead, we should have another function that just handle the linked properties then combine the result

packages/react-components/src/filter-list/FilterListApi.ts
KuberSethi: Great suggestion! Let me look into doing this

ziyunp: In this case, it will also be a breaking change to ObjectTable. We have to make objectSet the required prop and not allow the use case with just objectType prop.

ziyunp: Thinking through this - it's not exactly a breaking change but maybe we need some docs to say you have to pass in an objectSet to the ObjectTable to use it with FilterList

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1856,6 +1856,128 @@ const filterDefinitions = [
...
+    [argsOnFilterClauseChanged],
+  );
+
+  return (
+    <div style={COMBINED_LAYOUT_STYLE}>
+      <div style={SIDEBAR_FIXED_STYLE}>
+        <FilterList
+          {...args}
+          objectType={Employee}
+          objectSet={baseObjectSet}
```
nfdiop, line 1930: why not pass effectiveObjectSet?

packages/react-components/docs/FilterList.md
```diff
@@ -131,6 +132,51 @@ When using `type: "PROPERTY"`, the definition supports:
...
+  linkName: "manager",          // Employee → Manager
+  reverseLinkName: "directReports", // Manager → Employee (back-link)
+  linkedPropertyKey: "fullName",
+  linkedFilterComponent: "MULTI_SELECT",
+  linkedFilterState: { type: "SELECT", selectedValues: [] },
+  filterState: { type: "linkedProperty", linkedFilterState: { type: "SELECT", selectedValues: [] } },
+}
+```
+
+**UI-only** — omit `reverseLinkName`. The filter still renders and fires `onFilterStateChanged`; downstream narrowing is up to the consumer:
```
nfdiop, line 166: is this to avoid breaking change?

packages/react-components/src/filter-list/__tests__/testUtils.ts
```diff
@@ -85,10 +85,15 @@ export function createHasLinkFilterDef(
 export function createLinkedPropertyFilterDef(
   linkName: string,
   linkedPropertyKey: string,
+  options: { reverseLinkName?: string; omitReverseLinkName?: boolean } = {},
```
nfdiop: Curious why not consider that omitReverseLinkName = reverseLinkName == null

packages/react-components/src/filter-list/hooks/useDualScopeAggregation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const EMPTY_VALUES: string[] = [];
+
+export interface UseDualScopeAggregationOptions {
+  limit?: number;
+  sortBy?: "count" | "value";
+  selectedValues?: string[];
+}
+
+export function useDualScopeAggregation<
```
nfdiop, line 53: maybe we can add doc string here

packages/react-components/src/filter-list/types/LinkedFilterTypes.ts
```diff
@@ -28,6 +29,22 @@ import type {
   ValidComponentsForPropertyType,
 } from "../FilterListItemApi.js";
 
+/**
+ * Runtime representation of an active linked-property filter. Each entry
+ * binds a `linkName` to its `reverseLinkName` and an `innerWhere` typed
+ * against the linked object type, so `narrowObjectSet` can pivot type-safely.
+ */
+export type LinkedFilter<Q extends ObjectTypeDefinition> = {
```
nfdiop, line 37: great docs! thanks for the type safety!

packages/react-components/src/filter-list/FilterListApi.ts
nfdiop: I don't want to block but let's discuss in a FLUP :)

nfdiop (dismissed):
Thanks for handling this complex problem. Left some comments

packages/react-components/src/filter-list/hooks/__tests__/useFilterListState.test.tsx
KuberSethi: reverted it! i don't think we have breaks

packages/react-components/docs/FilterList.md
KuberSethi: Yup

packages/react-components/src/filter-list/__tests__/testUtils.ts
KuberSethi: just reduces testing boilerplate but am not opinionated here can fix

bryantpark04 (approved):
stamping

PR #3288, opened 2026-05-15
Interface action type constraint in ontology as code

anhhoang14022001:
### Summary 
This PR adds support for interface action type constraints with parameter constraints in OAC    
  - Define the syntax to create interface action type constraints with parameter constraints 
  - Add validation for duplicate IATC and IPC's API names                                                                                                                                   

Also will track implementing IATC support in marker-experimental as a FLUP.


packages/maker/src/api/defineInterfaceActionTypeConstraint.ts
```diff
@@ -0,0 +1,105 @@
...
+
+interface ParameterConstraintDefinition {
+  apiName?: string;
+  displayName: string;
+  type: OntologyIrBaseParameterConstraintType;
+  requireImplementation?: boolean;
+}
+
+interface InterfaceActionTypeConstraintDefinition {
+  from: InterfaceType;
```
ethanai22: does `from` make sense here? Something like `interfaceType` seems more semantically correct

packages/maker/src/api/defineInterfaceActionTypeConstraint.ts
```diff
@@ -0,0 +1,105 @@
...
+import { OntologyEntityTypeEnum } from "./common/OntologyEntityTypeEnum.js";
+import {
+  importedTypes,
+  namespace,
+  withoutNamespace,
+} from "./defineOntology.js";
+import type { InterfaceType } from "./interface/InterfaceType.js";
+import { combineApiNamespaceIfMissing } from "./namespace/combineApiNamespaceIfMissing.js";
+
+interface ParameterConstraintDefinition {
```
ethanai22: can we move these types to their own file? Probably just a InterfaceActionTypeConstraintDefinition.ts with both of them is fine

packages/maker/src/api/defineInterfaceActionTypeConstraint.ts
```diff
@@ -0,0 +1,105 @@
...
+  apiName?: string;
+  displayName: string;
```
ethanai22: curious why displayName is required but not apiName?

packages/maker/src/api/defineInterfaceActionTypeConstraint.ts
```diff
@@ -0,0 +1,105 @@
...
+    ) == null,
+    `Action type constraint with apiName ${apiNameWithNamespace} already exists on interface ${def.from.apiName}`,
+  );
+
+  const parameters: Record<
+    InterfaceParameterConstraintApiName,
+    OntologyIrInterfaceParameterConstraint
+  > = {};
+  const seenParamApiNames = new Set<string>();
+  for (const param of def.parameters ?? []) {
```
ethanai22: do you have a sense of how most people want to use these constraints? Does it make sense to provide a default that includes all the parameters of the interface instead of defaulting to an empty list? Asking since I could see these definitions becoming tedious for large interfaces. For actions, for example, we default to adding parameters for all properties, and add an optional field for excludes. I'm sure you have better sense of how impl plans on using these though,

packages/maker/src/conversion/toMarketplace/convertObject.ts
```diff
@@ -116,6 +116,7 @@ export function convertObject(
             }],
           )),
         properties: {},
+        actionTypes: {},
```
ethanai22, line 120: ~~we probably want to add configuration for this in `InterfaceImplementation`~~

we'll handle this later

packages/maker/src/api/defineInterfaceActionTypeConstraint.ts
anhhoang14022001: ah yeah I was trying to mirror OMS API and `apiName` was added to IPC later so we have to add it as optional to avoid breaking change. But I think we enforced on the BE that all IPC will need to have `apiName` so I revised the syntax to require not just apiName but all the fields available on IPC

packages/maker/src/api/defineInterfaceActionTypeConstraint.ts
anhhoang14022001: Tbh, from what I have seen in Athena workflows, the parameter constraints on an IATC are usually free-from and independent of the interface's properties. If there is a request to provide defaults in the future, we can definitely support that.

ethanai22 (dismissed):

ethanai22 (approved):

PR #3289, opened 2026-05-15
fix misleading patterns in @osdk/react and react-components docs

KuberSethi:
There were some inaccuracies our snippets re what package manager to use, what `createClient` actually takes, some code examples, etc. This PR attempts to correct them.

• rewrite `import { $ } from "@my/osdk"` to `client(Type)` — `$` is never exported from the generated SDK, it's a local alias users sometimes create; replaced ~30 sites with the `client(Type)` pattern already used elsewhere in the docs
• standardize SDK placeholder on `@my/osdk` across docs (was a mix of `@my/osdk`, `@YourApp/sdk`, `@your-app/sdk`) and add an `About @my/osdk` callout to each react-components doc
• fix `createClient`'s 2nd arg in react-sdk-docs snippets — was rendering the npm package name, now uses an ontology rid placeholder; fix several other broken snippets with missing imports (cache-management setup block, advanced-queries derived-property fragments) and one duplicate `const { data }` that wouldn't compile
• install commands now show npm/pnpm/yarn alternatives with a skip-if-installed tip — addresses pilot running pnpm in npm projects and the install-race-with-harness issue

### DIFF
--- a/docs/react/cache-management.md
+++ b/docs/react/cache-management.md
@@ -85,38 +85,6 @@ Automatic updates don't cover:
 
 The `ObservableClient` provides methods to manually invalidate cached data.
 
-### Setup
-
-`OsdkProvider` creates its own `ObservableClient` internally — you only need to construct one yourself if you want to call invalidation methods (`invalidateObjects`, `invalidateObjectType`, `invalidateFunction`, …) from outside the React tree (for example, from a WebSocket handler in `client.ts`). In that case, create one and pass it explicitly so React and your handler share the same cache:
-
-```tsx
-// client.ts
-import { createClient } from "@osdk/client";
-import { createObservableClient } from "@osdk/client/observable";
-
-const client = createClient(
-  "https://your-stack.palantirfoundry.com",
-  "ri.ontology.main.ontology.00000000-0000-0000-0000-000000000000",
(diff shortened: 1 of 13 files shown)


bryantpark04 (dismissed):

docs/react/cache-management.md
```diff
@@ -90,13 +90,14 @@ The `ObservableClient` provides methods to manually invalidate cached data.
...
-import { useObservableClient } from "@osdk/react";
+// client.ts
+import { createClient } from "@osdk/client";
+import { createObservableClient } from "@osdk/client/observable";
 
 const client = createClient(
   "https://your-stack.palantirfoundry.com",
-  "your-ontology-rid",
-  authProvider,
+  "ri.ontology.main.ontology.00000000-0000-0000-0000-000000000000",
```
etherandrius: `your-ontology-rid` is probably better for models. 

If you specific a rid, models will copy the rid exactly including the 0000 UUID

As an alternative we can use `ri.ontology.main.ontology.{UUID}`

KuberSethi: Thanks for the suggestion - implementing

olliejenningsPT (approved):

PR #3291, opened 2026-05-18
add link traversal to whereclause and remove filterlist applylinkedfilters workaround

KuberSethi:
extends whereclause so the filter list can express linked-property filters directly as where conditions; OSDK expands link-keyed entries into pivot+intersect ObjectSet compositions at where() time

• new `LinkWhereClause<T>` arm on `WhereClause<T, RDPs>` in `@osdk/api`; `.where()` runtime detects link entries via the `$reverseLink` sentinel and composes `searchAround → filter → searchAround → intersect`; `$and`/`$or`/`$not` compose via intersect/union/subtract
• filter-list `LINKED_PROPERTY` case emits `{ [linkName]: { $reverseLink, ...inner } }` from a new `reverseLinkName` field on `LinkedPropertyFilterDefinition`
• `perFilterWhereClauses` now naturally carry linked-filter conditions so direct property facet counts narrow correctly when a linked filter is active

supersedes #3287, which was building the ghost-row/applyLinkedFilters workaround for the same problem

### DIFF
--- a/.changeset/where-clause-link-traversal.md
+++ b/.changeset/where-clause-link-traversal.md
@@ -1,0 +1,14 @@
+---
+"@osdk/api": minor
+"@osdk/client": minor
+"@osdk/react-components": minor
+---
+
+add link traversal support to WhereClause
+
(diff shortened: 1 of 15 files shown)


KuberSethi:
No longer pursuing this path of addressing linked property filtering

PR #3295, opened 2026-05-18
Add ActionForm edited tags

nfdiop:
## Summary
- add an Edited tag to ActionForm fields after users edit them
- keep edited state tied to user changes rather than RHF dirty/touched semantics
- add coverage for restored values staying marked as edited

## Verification
- pnpm -C packages/react-components exec vitest run src/action-form/__tests__/BaseForm.test.tsx src/action-form/__tests__/FormField.test.tsx
- pnpm turbo typecheck --filter=@osdk/react-components
- git diff --check

Screenshot 2026-05-18 at 13 57 29

### DIFF
--- a/packages/react-components/src/action-form/fields/FieldBridge.tsx
+++ b/packages/react-components/src/action-form/fields/FieldBridge.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { memo, useCallback, useMemo } from "react";
+import React, { memo, useCallback, useMemo, useState } from "react";
 import type { Control } from "react-hook-form";
 import { useController } from "react-hook-form";
 import type {
@@ -32,10 +32,7 @@ export interface FieldBridgeProps {
   portalContainer?: PortalContainer;
 }
 const SELECT_LIKE_FIELDS: ReadonlySet<FieldComponent> = new Set<FieldComponent>(
-  [
-    "RADIO_BUTTONS",
-    "SWITCH",
-  ],
+  ["RADIO_BUTTONS", "SWITCH"],
 );
 
 export const FieldBridge: React.FC<FieldBridgeProps> = memo(
@@ -45,10 +42,7 @@ export const FieldBridge: React.FC<FieldBridgeProps> = memo(
     onExternalChange,
     portalContainer,
   }: FieldBridgeProps): React.ReactElement {
-    const rules = useMemo(
-      () => extractValidationRules(fieldDef),
-      [fieldDef],
-    );
+    const rules = useMemo(() => extractValidationRules(fieldDef), [fieldDef]);
 
     const {
       field: { onChange, onBlur, value },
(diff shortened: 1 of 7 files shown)


packages/react-components/src/action-form/fields/FieldBridge.tsx
```diff
@@ -61,10 +55,15 @@ export const FieldBridge: React.FC<FieldBridgeProps> = memo(
 
     const isSelectLike = SELECT_LIKE_FIELDS.has(fieldDef.fieldComponent);
     const isDropdown = fieldDef.fieldComponent === "DROPDOWN";
+    // RHF's touched state is blur-based and dirty state is value-comparison
+    // based. The tag represents an actual user edit, so set it from the
+    // field change path and keep it visible even if the value is restored.
+    const [isEdited, setIsEdited] = useState(false);
```
nfdiop, line 61: Matching the existing behaviour in Workshop vs comparing the current value with the initial one. 

RHF is exposing `isDirty` but they recommend having defaultValue set: which is not always the case here

packages/react-components/src/action-form/fields/FieldBridge.tsx
```diff
@@ -61,10 +55,15 @@ export const FieldBridge: React.FC<FieldBridgeProps> = memo(
...
 
     const handleChange = useCallback(
       (newValue: unknown) => {
         onChange(newValue);
+        setIsEdited(true);
```
ziyunp, line 66: Do we need to check if the newValue is not the same as oldValue?

ziyunp:
Styling - Are we not using the same outlined field when a field has been edited? Like in the ObjectTable

packages/react-components/src/action-form/fields/FieldBridge.tsx
nfdiop: Answered here https://github.com/palantir/osdk-ts/pull/3295#discussion_r3258665427
Essentially keeping the same behaviour as in Workshop where Edited means the user touched it.

nfdiop:
> Styling - Are we not using the same outlined field when a field has been edited? Like in the ObjectTable

Using the same styling as in Workshop. Happy to change it

packages/react-components/src/action-form/fields/FieldBridge.tsx
ziyunp: Just wondering about the bug we've seen in editable table, where user clicks in and out of a null field and the field is "edited". If you have tested it, then it's okay. 

Is it also showing "Edited" when user changed the value from A --> B --> A?

ziyunp (approved):

packages/react-components/src/action-form/fields/FieldBridge.tsx
nfdiop: Yes. So I tested and that was the behaviour in Workshop so I kept it

PR #3297, opened 2026-05-18
[maker-experimental] imported interface fixes

ethanai22:
misc interface fixes that came up during defense ontology testing

- Match imported interface input shapes to match java - only populate `properties` and use correct internal ids
- Fix imported interface link shapes
- Fix spt shape generation for imported interfaces
- Extract value type shapes from spts on all interfaces
- More objectTypeReference fixes on actions

### DIFF
--- a/packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ImportedShapeExtractor.ts
+++ b/packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ImportedShapeExtractor.ts
@@ -319,25 +319,16 @@ function extractImportedInterfaceTypes(
       interfaceType.apiName,
     );
 
-    // Build properties list (SPT references)
-    const properties: string[] = (interfaceType.properties ?? []).map(
-      (spt: any) => {
-        const sptId = knownIdentifiers.sharedPropertyTypes?.[spt.rid];
-        return sptId ?? spt.rid;
-      },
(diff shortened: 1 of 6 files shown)


packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ImportedShapeExtractor.ts
```diff
@@ -406,12 +374,48 @@ function extractImportedInterfaceTypes(
...
+    for (
+      const [propertyRid, property] of Object.entries(
+        interfaceType.propertiesV3 ?? {},
+      )
+    ) {
+      if (property.type !== "interfaceDefinedPropertyType") continue;
+
+      const propReadableId = ridGenerator.getInterfacePropertyTypeRids()
+        .inverse().get(propertyRid);
+      if (!propReadableId) continue;
```
aep000: Do we want to silently skip here? Or do we want explicitly error when there is no readable id allocated?

aep000 (commented):
one comment

packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ImportedShapeExtractor.ts
ethanai22: yea throwing makes sense

aep000 (approved):

PR #3298, opened 2026-05-18
Add title and pk filtering support

ssanjay1:
You can now filter using title and primary key identifiers

E.g.
`  await client(BoundariesUsState).where({"$title": {$in: ["a"]}}).fetchPage();
`


packages/api/src/aggregate/WhereClause.ts
```diff
@@ -178,6 +178,22 @@ type WhereClauseNumberPropertyTypes =
...
+  | DatetimeFilter;
+
+type PrimaryKeyFilterFor<T extends ObjectOrInterfaceDefinition> =
+  CompileTimeMetadata<T> extends { primaryKeyType: infer PKT extends string }
+    ? BaseFilterFor<PKT>
+    : SpecialPropertyFilter;
+
+export type SpecialPropertyWhereClause<T extends ObjectOrInterfaceDefinition> =
+  {
+    $title?: SpecialPropertyFilter;
```
ssanjay1, line 193: Title was deliberately left as a general filter for now, may need to play with it a bit more though but my first couple of attempts the compiler was complaining about infinitely deep types

bryantpark04: Surely titles are always strings 🤣

bryantpark04 (dismissed):

packages/client/src/internal/conversions/modernToLegacyWhereClause.ts
```diff
@@ -146,9 +146,18 @@ function handleWherePair(
...
-  const isRdp = !structFieldSelector && rdpNames?.has(fieldName);
+  const isTitleProperty = fieldName === "$title";
+  const isPrimaryKeyProperty = fieldName === "$primaryKey";
+  const isSpecialProperty = isTitleProperty || isPrimaryKeyProperty;
 
-  const propertyIdentifier: PropertyIdentifier | undefined = isRdp
+  const isRdp = !structFieldSelector && !isSpecialProperty
+    && rdpNames?.has(fieldName);
+
+  const propertyIdentifier: PropertyIdentifier | undefined = isSpecialProperty
```
nihalbhatnagar: can we make these actual if statments and a helper method, this ternary is kind of hard to read now

nihalbhatnagar (dismissed):

nihalbhatnagar (approved):

PR #3312, opened 2026-05-19
[maker-experimental] write static object files

ethanai22:


aep000 (approved):
Approving but I think we may want to consider a re-factor of this entire system so we don't have to do code-gen. Probably a maker-v2 thing

PR #3320, opened 2026-05-20
Fix ObjectTable overlays inside drawers and dialogs

nfdiop:
## Summary

- Adds an ObjectTable-local portal container for overlays.
- Renders ObjectTable header menus, column config dialogs, and searchable menus inside that local portal boundary.
- Keeps overlays inside Drawer/Dialog focus boundaries while avoiding clipping by the table scroll container.
- Adds Storybook overlay scenarios for ObjectTable and BaseForm.
- Adds tests covering contextual portal rendering for dialogs and header menus.

## Why

ObjectTable header menus could fail or render incorrectly when ObjectTable was inside a Blueprint Drawer/Dialog or Base UI Dialog. The column config dialog also had focus issues inside Blueprint Drawer because it portaled to `document.body`, outside the drawer’s focus trap.

## Verification

- `pnpm --dir=packages/react-components typecheck`
- `pnpm --dir=packages/react-components vitest run src/object-table/__tests__/TableHeaderWithPopover.test.tsx src/base-components/dialog/__tests__/Dialog.test.tsx`
- `pnpm --dir=packages/react-components-storybook typecheck`
- Storybook manual check:
  - ObjectTable header menu opens inside Blueprint Drawer/Base UI Dialog
  - Configure Columns search input remains focusable and responsive inside Blueprint Drawer

### DIFF
--- a/packages/react-components/src/object-table/Table.module.css
+++ b/packages/react-components/src/object-table/Table.module.css
@@ -26,6 +26,11 @@
   min-height: 0;
   overflow: auto;
   position: relative;
+  /* Sticky headers, pinned columns, and resize handles use z-index only within
+     the scroll container. Isolating that stacking context lets the sibling
+     header-menu portal paint above them by DOM order instead of competing with
+     table internals on z-index values. */
+  isolation: isolate;
   background-color: var(--osdk-background-primary);
 }
 
--- a/packages/react-components/src/object-table/Table.tsx
+++ b/packages/react-components/src/object-table/Table.tsx
@@ -23,6 +23,7 @@ import React, {
   useRef,
   useState,
 } from "react";
+import { PortalContainerProvider } from "../shared/PortalContainerContext.js";
 import { LoadingStateTable } from "./LoadingStateTable.js";
 import { NonIdealState } from "./NonIdealState.js";
 import styles from "./Table.module.css";
@@ -134,6 +135,7 @@ function BaseTableInner<
   }: BaseTableProps<TData>,
 ): ReactElement {
   const tableContainerRef = useRef<HTMLDivElement>(null);
+  const objectTablePortalRef = useRef<HTMLDivElement>(null);
   const [isLoadingMore, setIsLoadingMore] = useState(false);
   const [focusedRowId, setFocusedRowId] = useState<string | null>(null);
   const portalTracker = usePortalTracker();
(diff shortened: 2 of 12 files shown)


packages/react-components/src/object-table/Table.tsx
```diff
@@ -220,60 +222,67 @@ function BaseTableInner<
...
+          <TableEditContainer
+            editableConfig={editableConfig}
+            focusedRowId={focusedRowId}
+          />
         )}
+        {
+          /* Keep ObjectTable overlays inside drawer/dialog focus boundaries without
+            putting them in the scroll container where they can be clipped. */
+        }
+        <div ref={objectTablePortalRef} />
```
ziyunp: Q: Is it intended to use a dummy div here? Should we assign this to the components that are actually a popover ?

packages/react-components/src/object-table/Table.module.css
```diff
@@ -26,6 +26,11 @@
   min-height: 0;
   overflow: auto;
   position: relative;
+  /* Sticky headers, pinned columns, and resize handles use z-index only within
+     the scroll container. Isolating that stacking context lets the sibling
+     header-menu portal paint above them by DOM order instead of competing with
+     table internals on z-index values. */
+  isolation: isolate;
```
ziyunp, line 33: 👍 didn't know we also need this within the components

packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
```diff
@@ -1555,6 +1559,67 @@ export const InsideBlueprintDialog: Story = {
   },
 };
 
+export const InsideBaseUIDialog: Story = {
```
ziyunp: Nit: Don't think we need BaseUIDialog but no harm having it

ziyunp (dismissed):

packages/react-components/src/object-table/Table.tsx
nfdiop: good flag!

packages/react-components/src/object-table/Table.module.css
```diff
@@ -26,6 +26,11 @@
   min-height: 0;
   overflow: auto;
   position: relative;
+  /* Sticky headers, pinned columns, and resize handles use z-index only within
+     the scroll container. Isolating that stacking context lets the sibling
+     header-menu portal paint above them by DOM order instead of competing with
+     table internals on z-index values. */
```
nfdiop, line 32: otherwise the table menu shows below the header

olliejenningsPT (approved):

PR #3322, opened 2026-05-20
Pin all github-actions to specific commit shas

asvoboda:
https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/


asvoboda:
obe

PR #3330, opened 2026-05-20
Bump @osdk/foundry.* catalog 2.57.0 -> 2.61.0

nihalbhatnagar:
## Summary

Bumps the `foundry-platform-typescript` catalog from 2.57.0 to 2.61.0 so `@osdk/foundry.ontologies` exposes the new `OntologyScenarios` namespace (`createScenario`, `listScenarioEditedEntityTypes`, `listScenarioEditedObjects`, `listScenarioEditedLinkTypes`, `listScenarioEditedLinks`) and the `scenarioRid` query parameter on `Actions.apply` / `OntologyObjectSets.*` / `Queries.execute`. Fixes the breakages this introduces across the monorepo: new `applyScenario` variant in `LogicRule`, new `scenarioReference` variant in `ActionParameterType`, and `QueryParameterV2.required` becoming a required field on test stubs.

## Usage

No new user-facing API in this PR — it's a catalog bump and prep for the scenarios series. Consumers of the bumped SDK can now import:

\`\`\`ts
import { OntologyScenarios } from "@osdk/foundry.ontologies";

await OntologyScenarios.createScenario(ctx, ontologyRid, {});

await OntologyScenarios.listScenarioEditedEntityTypes(ctx, ontologyRid, scenarioRid);
\`\`\`

## Test plan

- [x] \`pnpm turbo typecheck\` — all 202 tasks pass
- [x] \`pnpm turbo test --filter=@osdk/client\` — 795/795 tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3331, opened 2026-05-20
Add withScenario / createScenario helpers (beta)

nihalbhatnagar:
## Summary

Introduces a first-class entry point for ontology scenarios in \`@osdk/client/unstable-do-not-use\`. \`withScenario\` attaches an existing scenario to a base client synchronously; \`createScenario\` mints a fresh one via \`OntologyScenarios.createScenario\`. The returned \`ScenarioClient\` extends \`Client\`, so it is usable anywhere a \`Client\` is accepted. \`scenarioRid\` is plumbed through \`MinimalClient\` and injected as a query parameter on \`applyAction\`, \`fetchPage\`, \`aggregate\`, and \`applyQuery\`. A compile-time brand prevents nested scenarios; a runtime guard rejects scenario-on-transaction.

## Usage

```ts
import { withScenario, createScenario } from "@osdk/client/unstable-do-not-use";
import { Doctor, Schedule } from "@my/sdk";

// Attach to an existing scenario — sync, no network call
const scenario = withScenario(client, "ri.actions..scenario.abc");

// Or mint a fresh scenario from the current branch context
const fresh = await createScenario(client);

// All regular operations scope to the scenario automatically
const page = await scenario(Doctor).fetchPage({ \$pageSize: 50 });
await scenario(Schedule).applyAction({ doctor: page.data[0], shift: "AM" });

// Read the rid back at any time
const { scenarioRid } = scenario.getScenarioReference();
```

> Stacked on https://github.com/palantir/osdk-ts/pull/3330


packages/client/src/scenarios/ScenarioClient.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface ScenarioClient extends Client {
+  /**
+   * Returns the scenario reference for this client. Pass directly to an action whose parameter is typed as
+   * `scenarioReference`, or use to inspect the underlying scenario RID.
+   */
+  getScenarioReference(): { scenarioRid: string };
+}
+
+/**
+ * Shared internal builder used by both {@link withScenario} and {@link createScenario}. Validates the parent client
```
nihalbhatnagar, line 35: Im in favor of these claude generated jsdocs for internal methods

bryantpark04:
What's the utility of the `getScenarioReference` method being public? I see in #3334 you can pass it in to the action, but you can also just pass the scenario client in. Is it to be easily serializable/persistable?

packages/client/src/scenarios/createScenario.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @example
+ * ```ts
+ * import { createScenario } from "@osdk/client/unstable-do-not-use";
+ *
+ * const scenario = await createScenario(client);
+ * const { scenarioRid } = scenario.getScenarioReference();
+ * ```
+ */
+export async function createScenario(
```
ssanjay1, line 43: Does this and withScenario really need to be two separate things? Can't we just have this optionally take a scenario RID and branch logic from there?

ssanjay1 (approved):

PR #3332, opened 2026-05-20
Scenarios: getEditedEntities + editedEntitiesAsyncIter (beta)

nihalbhatnagar:
## Summary

Extends \`ScenarioClient\` with three methods backed by \`OntologyScenarios.listScenarioEditedEntityTypes\` and \`listScenarioEditedObjects\`. \`getEditedEntityTypes\` returns the union of edited object and link types on the scenario. \`getEditedEntities\` paginates edited objects by type (sparse instances — only \`\$primaryKey\` and \`\$apiName\` populated). \`editedEntitiesAsyncIter\` auto-paginates and dedupes by \`\$primaryKey\` across pages. Together they're the building blocks for N-way scenario diff workflows.

## Usage

```ts
import { withScenario } from "@osdk/client/unstable-do-not-use";
import { Doctor } from "@my/sdk";

const scenario = withScenario(client, "ri.actions..scenario.abc");

// 1. Discover what changed
const { objectTypes, linkTypes } = await scenario.getEditedEntityTypes();

// 2. Paginate edited objects (page form, explicit pageToken loop)
let pageToken: string | undefined;
const keys: unknown[] = [];
do {
  const page = await scenario.getEditedEntities(Doctor, { pageSize: 500, pageToken });
  keys.push(...page.data.map(o => o.\$primaryKey));
  pageToken = page.nextPageToken;
} while (pageToken);

// 3. Or stream and dedupe via the async iterator
for await (const obj of scenario.editedEntitiesAsyncIter(Doctor, { pageSize: 500 })) {
  console.log(obj.\$apiName, obj.\$primaryKey);
}

// 4. Re-fetch under the scenario context for full property values
const full = await scenario(Doctor)
  .where({ \$primaryKey: { \$in: keys } })
  .fetchPage();
```

## Test plan

- [x] \`pnpm turbo typecheck --filter=@osdk/client\`
- [x] \`pnpm turbo test --filter=@osdk/client\` — 799/799 tests pass, 4 new tests in \`scenarioClient.test.ts\`
- [ ] Manual: run the N-way diff workflow against a stack with 2+ scenarios sharing edits

> Stacked on https://github.com/palantir/osdk-ts/pull/3331

### DIFF
--- a/packages/client/src/scenarios/ScenarioClient.test.ts
+++ b/packages/client/src/scenarios/ScenarioClient.test.ts
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Employee } from "@osdk/client.test.ontology";
+import type { MockedFunction } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { Client } from "../Client.js";
+import { createClient } from "../createClient.js";
+import { mockFetchResponse } from "../createClient.test.js";
+import { withScenario } from "./withScenario.js";
+
+describe("ScenarioClient methods", () => {
+  const ontologyRid = "ri.not.important";
+  let fetchFunction: MockedFunction<typeof globalThis.fetch>;
+  let client: Client;
+
+  beforeEach(() => {
+    fetchFunction = vi.fn();
+    client = createClient(
+      "https://mock.com",
+      ontologyRid,
+      async () => "Token",
+      undefined,
+      fetchFunction,
+    );
+  });
+
+  describe("getEditedEntityTypes", () => {
+    it("calls the editedEntityTypes endpoint and returns the response shape", async () => {
+      const scenario = withScenario(client, "ri.actions..scenario.abc");
+      mockFetchResponse(fetchFunction, {
+        objectTypes: ["Employee", "Office"],
+        linkTypes: [{ objectType: "Employee", linkType: "manages" }],
+      });
+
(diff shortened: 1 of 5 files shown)


packages/client/src/scenarios/ScenarioClient.test.ts
```diff
@@ -0,0 +1,138 @@
...
+        /\/scenarios\/ri\.actions\.\.scenario\.abc\/objects\/Employee\/edited$/,
+      );
+      expect(url.searchParams.get("pageSize")).toBe("100");
+      expect(url.searchParams.get("pageToken")).toBe("tok-1");
+      expect(result.nextPageToken).toBe("tok-2");
+      expect(result.data).toEqual([]);
+    });
+  });
+
+  describe("editedEntitiesAsyncIter", () => {
```
nihalbhatnagar, line 108: put async iter at the end because its better for intellisense

packages/client/src/scenarios/ScenarioClient.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      ]);
+    });
+  });
+
+  describe("getEditedEntities (page form)", () => {
+    it("calls the listScenarioEditedObjects endpoint with pageSize/pageToken", async () => {
+      const scenario = withScenario(client, "ri.actions..scenario.abc");
+      const mock: ListScenarioEditedObjectsResponse = {
+        data: [
+          { __apiName: "Employee", __primaryKey: 50030 },
```
nihalbhatnagar, line 78: This response type doesn't really make sense, as I realize now scenarios will never return more than the locators here.  I'll flup on making this strongly typed in gateway

packages/client/src/scenarios/ScenarioClient.ts
```diff
@@ -14,10 +14,46 @@
...
+  linkTypes: string[];
+}
+
+/**
+ * The object types and link types that have been modified within a scenario. Returned by
+ * {@link ScenarioClient.getEditedEntityTypes}. Only many-to-many link type edits are surfaced via `linkTypes`;
+ * one-to-many edits surface as object edits on the object type that owns the foreign key property.
+ */
+export interface ScenarioEditedEntityTypes {
+  objectTypes: string[];
```
nihalbhatnagar, line 42: The reason these are separated and not `{objectType: "Employee", linkTypes: ["peeps,", "lead"]}` is because you can edit the link and not a property on the OT, which is what we decided for the gateway API. I'm against an enum that describes the type of edit that was applied as I feel this complicates the API

I'm not really sure how to feel about this though, as again `linkTypes` only corresponds to many to many links for now.

ssanjay1 (dismissed):

bryantpark04 (approved):

PR #3336, opened 2026-05-20
filter-list: configurable actions and overflow menu

KuberSethi:
add configurable per-filter actions and surface an overflow menu on every filter

• every filter row renders an overflow menu including linked, range, and timeline filters
• the standalone remove (X) button moves into the overflow menu by default
• new `actions` field on `FilterDefinition` controls search, overflow, and remove visibility


KuberSethi:
superseded by #3342 which has the polished hand-written implementation with team-review iterations

PR #3338, opened 2026-05-21
filter-list: accept formatDate on all date inputs

KuberSethi:
forward formatDate to single-date, multi-date, and timeline inputs

• all date-typed filters now respect a `formatDate` function on `PropertyFilterDefinition`
• keeps date formatting consistent across single, multi, range, and timeline inputs

### DIFF
--- a/packages/react-components/src/filter-list/inputs/SingleDateFilterInput.tsx
+++ b/packages/react-components/src/filter-list/inputs/SingleDateFilterInput.tsx
@@ -21,11 +21,18 @@ import type { FilterState } from "../FilterListItemApi.js";
 interface SingleDateFilterInputProps {
   filterState: FilterState | undefined;
   onFilterStateChanged: (state: FilterState) => void;
+  /**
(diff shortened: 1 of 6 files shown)


packages/react-components/src/filter-list/inputs/SingleDateFilterInput.tsx
```diff
@@ -21,11 +21,18 @@ import type { FilterState } from "../FilterListItemApi.js";
 interface SingleDateFilterInputProps {
   filterState: FilterState | undefined;
   onFilterStateChanged: (state: FilterState) => void;
+  /**
+   * Consumer-provided display formatter forwarded to the underlying
+   * `SingleDateInput` / `DatePicker`. Falls back to ISO `YYYY-MM-DD` when
+   * omitted.
+   */
+  formatDate?: (date: Date) => string;
```
nfdiop, line 29: one thing to double check: that we don't need to pass a `parseDate` too

nfdiop (approved):

PR #3340, opened 2026-05-21
filter-list: thread renderValue through linked inputs

KuberSethi:
thread renderValue through LinkedPropertyFilter inputs

• linked property inputs now accept and forward the `renderValue` prop
• matches the existing behavior of non-linked filter inputs

### DIFF
--- a/packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
+++ b/packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
@@ -14,14 +14,23 @@
  * limitations under the License.
  */
 
+import type { ObjectSet, ObjectTypeDefinition, PropertyKeys } from "@osdk/api";
+import { useOsdkAggregation } from "@osdk/react";
 import { cleanup, render, screen } from "@testing-library/react";
 import React from "react";
 import { afterEach, describe, expect, it, vi } from "vitest";
+import { LinkedPropertyInput } from "../../../inputs/LinkedPropertyInput.js";
 import type { PropertyAggregationValue } from "../../../types/AggregationTypes.js";
+import type { LinkedPropertyFilterDefinition } from "../../../types/LinkedFilterTypes.js";
 import { createRenderValueFilter } from "../comboboxFilter.js";
 import { ListogramInput } from "../ListogramInput.js";
 import { MultiSelectInput } from "../MultiSelectInput.js";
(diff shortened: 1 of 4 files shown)


packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
```diff
@@ -288,3 +297,164 @@ describe("SingleSelectInput renderValue", () => {
...
+      $group: { name: g.name },
+      $count: g.count,
+    })),
+    isLoading: false,
+    error: null,
+    refetch: vi.fn(),
+  } as unknown as ReturnType<typeof useOsdkAggregation>);
+}
+
+describe("LinkedPropertyInput renderValue", () => {
```
olliejenningsPT, line 377: nit: could we add a single-select case here too? we wire renderValue through SINGLE_SELECT in LinkedPropertyInput but it's the only one of the three without a test, and it'd be nice to keep parity with the base SingleSelectInput tests above.

packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
```diff
@@ -14,15 +14,24 @@
...
 import type { PropertyAggregationValue } from "../../../types/AggregationTypes.js";
+import type { LinkedPropertyFilterDefinition } from "../../../types/LinkedFilterTypes.js";
 import { createRenderValueFilter } from "../comboboxFilter.js";
 import { ListogramInput } from "../ListogramInput.js";
 import { MultiSelectInput } from "../MultiSelectInput.js";
 import { SingleSelectInput } from "../SingleSelectInput.js";
 
+vi.mock("@osdk/react", () => ({
+  useOsdkAggregation: vi.fn(),
+  useRegisterUserAgent: vi.fn(),
```
olliejenningsPT: neither this file nor usePropertyAggregation actually call useRegisterUserAgent, can we drop this from the mock?

KuberSethi: good catch, dropped it

packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
KuberSethi: added a smoke test!

bryantpark04 (approved):
stamping

PR #3341, opened 2026-05-21
filter-list: disable reset until state diverges from initial

KuberSethi:
disable filter list reset until state diverges from initial definitions

• snapshot the initial filter state on mount and compare against current state
• reset stays disabled when current state matches the initial snapshot
• re-enables as soon as any filter value or toggle differs


bryantpark04 (dismissed):
stamping

packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -50,9 +50,92 @@ export interface UseFilterListStateResult<Q extends ObjectTypeDefinition> {
...
+/**
+ * Recursive structural equality for the JSON-shaped FilterState union.
+ * Avoids JSON.stringify so key ordering doesn't produce false negatives.
+ *
+ * FilterState must remain plain JSON-shaped (primitives, Date, arrays, plain
+ * records). Class instances like Map/Set fall through to `Object.is` and will
+ * always compare unequal — if a future filter variant needs to carry one,
+ * extend the cases here.
+ */
+function areValuesEqual(a: unknown, b: unknown): boolean {
```
ziyunp: Can we use lodash isEqual?

packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -142,18 +225,22 @@ export function useFilterListState<Q extends ObjectTypeDefinition>(
     return map;
   }, [metadata?.properties]);
 
-  const [filterStates, setFilterStates] = useState<Map<string, FilterState>>(
+  const [initialFilterStatesSnapshot] = useState<Map<string, FilterState>>(
```
ziyunp, line 150: Why not do useMemo instead?

packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
```diff
@@ -406,7 +406,6 @@ describe("LinkedPropertyInput renderValue", () => {
           },
         }}
         onFilterStateChanged={vi.fn()}
-        whereClause={{} as WhereClause<ObjectTypeDefinition>}
```
nfdiop: do we need to merge main again?

packages/react-components/src/filter-list/base/BaseFilterListApi.ts
```diff
@@ -49,6 +49,7 @@ export interface BaseFilterListProps<D extends FilterDefinitionControls> {
   showResetButton?: boolean;
   showActiveFilterCount?: boolean;
   hasVisibilityChanges?: boolean;
+  hasChangesFromInitial?: boolean;
```
nfdiop: Looking at the API, can we deprecate hasVisibilityChanges and have a single `canReset` boolean to keep the API simple. Otherwise we are leaking the hook’s “initial snapshot” implementation detail into BaseFilterList.

packages/react-components/src/filter-list/base/FilterListHeader.tsx
```diff
@@ -48,6 +50,10 @@ function FilterListHeaderInner({
     onCollapsedChange?.(!collapsed);
   }, [onCollapsedChange, collapsed]);
 
+  const resetDisabled = canReset !== undefined
```
ziyunp: Isn't the convention to do `canReset != null`

packages/react-components/src/filter-list/base/BaseFilterListApi.ts
```diff
@@ -48,6 +48,16 @@ export interface BaseFilterListProps<D extends FilterDefinitionControls> {
   titleIcon?: React.ReactNode;
   showResetButton?: boolean;
   showActiveFilterCount?: boolean;
+  /**
+   * Whether the reset button is enabled. Hosts compute this from whether the
+   * filter and/or visibility state has diverged from its initial snapshot.
+   */
+  canReset?: boolean;
+  /**
+   * @deprecated Use {@link canReset} instead. When `canReset` is provided it
```
ziyunp, line 57: Less worried about marking base component api deprecated as it's unlikely to be used yet

ziyunp (dismissed):

packages/react-components/src/filter-list/base/FilterListHeader.tsx
KuberSethi: good catch, switched to != null to match the convention right above it

packages/react-components/src/filter-list/hooks/useFilterListState.ts
KuberSethi: useMemo can get discarded by react so it won't reliably pin the baseline across renders. went with useState + a lazy initializer instead and left a comment explaining why

packages/react-components/src/filter-list/base/BaseFilterListApi.ts
KuberSethi: done, added canReset as the primary prop and marked hasVisibilityChanges deprecated. left it as a fallback for now since it already shipped, but canReset takes precedence

packages/react-components/src/filter-list/base/inputs/__tests__/renderValue.test.tsx
KuberSethi: we're current with main now, no conflicts

packages/react-components/src/filter-list/base/BaseFilterListApi.ts
KuberSethi: yeah unlikely to be hit yet, kept the deprecation just to nudge people toward canReset

bryantpark04 (approved):

PR #3342, opened 2026-05-21
filter-list: extend per-item exclude row to every filter type

KuberSethi:
extend the inline Keep/Exclude row + Remove + Clear all to every filter type that supports excluding, and fix the linkedProperty exclude toggle

• filters wrapping linkedProperty (and other excluding-capable shapes) now show the inline Keeping/Excluding dropdown + count + Clear all that was previously SELECT/EXACT_MATCH only; toggle flips isExcluding on the inner wrapped state
• new `searchField` flag on `FilterDefinition` hides the header monocle for filters that already have their own inline search (e.g. `MULTI_SELECT`)
• extracted reusable helpers (`toggleIsExcluding`, `clearFilterState`, `filterHasActiveState`, `getEffectiveFilterState`, `getSelectedCount`) from inline component logic into `filterValues.ts`


bryantpark04 (approved):
stamping

PR #3343, opened 2026-05-21
filter-list: opt-in relative-range shortcuts in date picker

KuberSethi:
add opt-in relative-range shortcut rail to filter list date pickers

• new `dateShortcuts` property on date-typed `PropertyFilterDefinition`s toggles the shortcut bar
• applies to DATE_RANGE and SINGLE_DATE filters
• shortcuts cover relative ranges like last 7 days, this month, year to date


bryantpark04 (dismissed):
stamping

packages/react-components/src/filter-list/base/inputs/SingleDateInput.module.css
```diff
@@ -23,6 +23,8 @@
   display: flex;
   align-items: center;
   gap: var(--osdk-filter-select-container-gap);
+  flex: 1 1 auto;
+  min-width: 0;
```
olliejenningsPT, line 27: These flex props apply to every SingleDateInput consumer, not just the shortcuts case. Worth a quick visual check of existing non-shortcut date filters to confirm no layout regression.

packages/react-components/src/filter-list/FilterListItemApi.ts
```diff
@@ -243,14 +249,33 @@ export interface DateFormattingProps {
...
+   * renders exactly those periods in order. Stored `FilterState`
+   * remains absolute.
+   */
+  dateShortcuts?: RelativeDatePeriod[] | boolean;
+}
+
+/** Filter components that support the shortcuts rail. */
+export type FilterComponentsSupportingDateShortcuts =
+  | "DATE_RANGE"
+  | "SINGLE_DATE";
```
olliejenningsPT: PR description says this applies to TIMELINE, but TIMELINE is excluded here and asserted as a type error in the test-d file. Update the PR body.

packages/react-components/src/filter-list/base/inputs/DateRangeHistogramInput.tsx
```diff
@@ -98,12 +107,36 @@ function DateRangeHistogramInputInner({
...
+  );
+
+  const renderLeftRail = useCallback(
+    (closePopover: () => void) => (
+      <ShortcutBar
+        periods={shortcutPeriods ?? []}
+        onSelect={(range) => handleShortcutSelect(range, closePopover)}
+      />
+    ),
+    [shortcutPeriods, handleShortcutSelect],
```
olliejenningsPT: The inline `(range) => handleShortcutSelect(range, closePopover)` creates a new function each render, so the `memo()` on ShortcutBar and ShortcutButton never skips. Either pass closePopover through the bar separately or drop the memos.

packages/react-components/src/shared/calendar/DatePicker.tsx
```diff
@@ -128,6 +128,15 @@ export interface DatePickerProps {
...
   modal?: "trap-focus" | false;
+
+  /**
+   * Optional content rendered as a left rail in the popover. Receives a
+   * `closePopover` callback so left-rail interactions (e.g. shortcut buttons
+   * committing a new value) can exit the picker's edit mode before firing
+   * `onChange`. Without that, the value-sync inside `useDateEditState` skips
+   * the update to preserve any in-progress typing.
+   */
+  leftRail?: (closePopover: () => void) => React.ReactNode;
```
nfdiop: could we instead add shortcuts directly to the datepicker like in here so it also benefits the form? https://blueprintjs.com/docs/#datetime/date-picker

packages/react-components/src/filter-list/base/inputs/SingleDateInput.module.css
KuberSethi: thanks for flagging, I confirmed and I think things look fine unless I misunderstood

packages/react-components/src/filter-list/FilterListItemApi.ts
KuberSethi: Good catch, timeline was dropped when shortcuts moved inside the picker popover. Updated the PR body accordingly

packages/react-components/src/filter-list/base/inputs/DateRangeHistogramInput.tsx
KuberSethi: nice catch, done

packages/react-components/src/shared/calendar/DatePicker.tsx
KuberSethi: good idea, however thats probably a decent amount of changes so prefer to do that in a FLUP if that's ok with you!

packages/react-components/src/filter-list/__tests__/PropertyFilterDateExtras.test-d.ts
nfdiop, line 1: thanks for adding type tests!

packages/react-components/src/filter-list/__tests__/SingleDateFilterInput.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    vi.useFakeTimers({ now });
+    const onFilterStateChanged = vi.fn();
+    render(
+      <SingleDateFilterInput
+        filterState={undefined}
+        onFilterStateChanged={onFilterStateChanged}
+        dateShortcuts={["past-week"]}
+      />,
+    );
+    fireEvent.focus(screen.getByRole("combobox"));
```
nfdiop: It looks a bit odd to show a "Past week" text for selecting a single date. Could we have something like? 

Image

packages/react-components/src/shared/calendar/DatePickerCommon.module.css
```diff
@@ -42,13 +42,33 @@
 }
 
 .osdkDatePickerPopover {
+  display: flex;
+  flex-direction: row;
+  align-items: stretch;
+  gap: var(--osdk-datetime-popover-gap, 12px);
```
nfdiop: should it be defined in tokens?

packages/react-components/src/filter-list/base/inputs/SingleDateInput.tsx
```diff
@@ -55,6 +61,36 @@ function SingleDateInputInner({
...
+      closePopover();
+      onChange(day);
+    },
+    [onChange],
+  );
+
+  const renderLeftRail = useCallback(
+    (closePopover: () => void) => (
+      <ShortcutBar
+        periods={shortcutPeriods ?? []}
```
nfdiop: note: here and in DateRangeHistogramInput.tsx, we need a constant empty array to avoid re-render

packages/react-components/src/shared/dateUtils.ts
```diff
@@ -123,3 +125,66 @@ export function parseTimeString(
 export function getTimeValue(date: Date | null): string {
   return date != null ? formatTime(date) : "00:00";
 }
+
+/** Relative date range identifier; resolve via {@link getRelativeDateRange}. */
+export type RelativeDatePeriod = keyof typeof RELATIVE_DATE_PERIODS;
```
nfdiop: I am not sure this is the correct API. If someone wants something like the last 12 hours, or the last 6 hours, we will have to come here and update this list. Users may also want Instead, can we either:
- Option 1 (fastest to implement): expose `shortcuts : boolean` and ship some defaults
- Option 2: expose boolean | [DatePickerShortcut](https://blueprintjs.com/docs/#api/DatePickerShortcut) so the user fully owns the shortcuts (see also https://mantine.dev/dates/date-picker-input/#presets)

packages/react-components/src/filter-list/base/inputs/DateRangeHistogramInput.tsx
```diff
@@ -98,12 +107,37 @@ function DateRangeHistogramInputInner({
...
+  const dateLeftRail = shortcutPeriods != null ? renderLeftRail : undefined;
+
   return (
     <RangeInput
       {...rest}
+      onChange={onChange}
       valueCountPairs={valueCountPairs}
       config={config}
       histogramData={histogramData}
+      dateLeftRail={dateLeftRail}
```
nfdiop: Could we pass the shortcuts directly instead and let the DatePicker handle it?

packages/react-components/src/filter-list/base/inputs/ShortcutBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+   * can pass a stable `onSelect` and let the `memo` on the bar/buttons skip.
+   */
+  closePopover: () => void;
+  className?: string;
+}
+
+interface ShortcutButtonProps {
+  period: RelativeDatePeriod;
+  onSelect: ShortcutBarProps["onSelect"];
+  closePopover: () => void;
```
nfdiop: I don't think the shortcuts should know about the concept of a closePopover. Instead it should just take `onSelect`

packages/react-components/src/filter-list/base/inputs/ShortcutBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+      appearance="minimal"
+      className={styles.shortcut}
+      onClick={handleClick}
+    >
+      {getRelativeDatePeriodLabel(period)}
+    </ActionButton>
+  );
+}
+
+const ShortcutButton = memo(ShortcutButtonInner);
```
nfdiop: extra nit: I find it nice to use the `memo` inline directly. Otherwise you have to jump through ShortcutButton to see the real code.

packages/react-components/src/shared/__tests__/dateUtils.test.ts
```diff
@@ -171,3 +175,60 @@ describe("getTimeValue", () => {
...
+    expect(DEFAULT_RELATIVE_DATE_PERIODS).toEqual(
+      Object.keys(labels) as RelativeDatePeriod[],
+    );
+    for (const period of DEFAULT_RELATIVE_DATE_PERIODS) {
+      expect(getRelativeDatePeriodLabel(period)).toBe(labels[period]);
+    }
+  });
+
+  it("clamps past-month from a 31st to the last day of the prior shorter month", () => {
+    // Mar 31 - 1 month should land on Feb 29 in a leap year (date-fns clamps).
```
nfdiop, line 214: nice test for the edge case!

packages/react-components/src/shared/calendar/DatePicker.tsx
nfdiop: Hum, let's instead do it before we merge otherwise we will have a breaking API change

packages/react-components/src/filter-list/__tests__/SingleDateFilterInput.test.tsx
KuberSethi: kept the current wording labels as that is what workshop uses but addressed your comment on where it lives / the API so if people want their own they can supply them

nfdiop: If I am not mistaken we show this wording for ranges but not for single dates. But happy with this wording on this test since the label is supplied

Image

Image

packages/react-components/src/shared/calendar/ShortcutBar.module.css
```diff
@@ -0,0 +1,29 @@
...
+  flex-direction: column;
+  align-items: stretch;
+  gap: var(--osdk-filter-input-gap);
+}
+
+.shortcut {
+  justify-content: flex-start;
+  text-align: left;
+  white-space: nowrap;
+  font-size: var(--osdk-typography-size-body-small);
```
nfdiop: Can we use normal font size? It looks a bit off to have the shortcuts so small. We can polish the rest of it in a FLUP PR with @Leolide

packages/react-components/src/shared/dateUtils.ts
```diff
@@ -123,3 +125,106 @@ export function parseTimeString(
...
+  "past-2-years",
+];
+
+/**
+ * A user-defined date-picker shortcut. The picker renders a button labeled
+ * {@link label}; clicking it computes a relative `{ min, max }` range from the
+ * current time. Range pickers apply both bounds; single-date pickers commit
+ * `min` (with wall-clock time stripped).
+ */
+export interface DatePickerShortcut {
```
nfdiop: Can we have?
- DatePickerShortcut
- DateRangePickerShortcut (which takes a range).
Essentially the API looks off that we would take a range shortcut when the user only wants to select one day (see https://github.com/palantir/blueprint/blob/a2cd8f9ff584e3ae9d4b6ead852cf9c6f56a77da/packages/datetime/src/components/shortcuts/shortcuts.tsx#L39: you can also use the existing DateRange type vs min/max. We need to make easy when we migrate to blueprint)

PR #3348, opened 2026-05-21
Fix empty-AND filter emission from `.where({})`

schmelczer:
`modernToLegacyWhereClause({}, ...)` falls through to `{ type: "and", value: [] }`, which the
Foundry wire contract rejects as `Core:InvalidAndFilter` (an AND filter must have at least one
sub-filter). The observable layer's `ListsHelper`, `ObjectListQuery`, `InterfaceListQuery`, and
`ObjectAggregationQuery` default missing `where` to `{}` and unconditionally call
`.where(canonicalWhere)`, so any `useOsdkObjects` / `useOsdkAggregation` without a `where` clause
ships the malformed payload.

### Failure

```
ERROR Service error: Core:InvalidAndFilter
   4: lohi_lib::store::tables::filters::FilterCreator::create_filter
  10: <SqliteLohiStore as LohiStore>::aggregate_object_set
  11: <OntologyResource as AsyncOntologiesService>::aggregate_object_set
```

### Fix

Treat `.where({})` as a no-op at the boundary in `createObjectSet.ts` and
`createWithPropertiesObjectSet.ts`: when the clause has no keys, return an `ObjectSet` on the same
wire object set without a filter wrapper.

### Why client-side

The contract is enforced at the API gateway.
Loosening the client and asking the server to accept empty AND would diverge from the wire contract
production Foundry enforces.


.changeset/sour-plums-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/client": minor
```
nfdiop: nit: should it be patch?

nfdiop (dismissed):

.changeset/sour-plums-pull.md
schmelczer: Sounds good, changed it!

nfdiop (approved):

PR #3350, opened 2026-05-21
Codegen interface namespace-stripped -> namespaced property mappings and add namespaces back to properties in edits

bryantpark04:


bryantpark04:
Made the fix elsewhere

PR #3351, opened 2026-05-21
fix tailwind preflight overriding osdk.components in officenetwork

KuberSethi:
## Summary

filter-list per-item action icons (search, overflow menu) and drag handles were permanently visible in the officenetwork sandbox because tailwind's `@layer base` preflight (`button { opacity: 1 }`) was overriding the hover-only `.headerActionButton { opacity: 0 }` rule from `@osdk/react-components`. cascade layer order ended up `osdk.tokens, osdk.components, theme, base, components, utilities` — so tailwind's base layer won.

separate but related: storybook's vite dep optimizer was hoisting date-fns v2 (pulled in by react-day-picker) instead of v4 (pulled in by @osdk/faux for \`constructNow\`), crashing the preview on \`filterTimeSeriesData.ts\` import.

• declare cascade layer order in officenetwork \`index.css\` so \`osdk.components\` comes last in the layer order and wins over tailwind's base/components/utilities
• alias \`date-fns\` to the v4 install in \`react-components-storybook/.storybook/main.ts\` via \`createRequire\` resolved through \`@osdk/faux\`, plus \`resolve.dedupe: ["date-fns"]\`

## Test plan

- [ ] run officenetwork locally (\`pnpm --filter @osdk/e2e.sandbox.officenetwork dev\`) and confirm filter-list per-item icons are hidden until you hover the row
- [ ] run storybook locally (\`pnpm --filter @osdk/react-components-storybook dev\`) and confirm any date-fns-dependent story (e.g. filter-list with timeline / date-range) loads without \`constructNow\` errors

### DIFF
--- a/packages/e2e.sandbox.officenetwork/src/index.css
+++ b/packages/e2e.sandbox.officenetwork/src/index.css
@@ -1,3 +1,8 @@
+/* Cascade layer order: tailwind layers first, osdk last so osdk.components
+   wins over tailwind's preflight (e.g. button { opacity: 1 } would otherwise
+   override .headerActionButton { opacity: 0 } from react-components). */
+@layer theme, base, components, utilities, osdk.tokens, osdk.components;
+
 @import "@osdk/react-components/styles.css";
 @import "tailwindcss";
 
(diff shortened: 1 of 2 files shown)


KuberSethi:
splitting — officenetwork fix only at #PLACEHOLDER, storybook date-fns dropped from this PR

PR #3359, opened 2026-05-22
Use Source Sans 3 as default font family

nfdiop:
## Summary
- Replace `Source Sans Pro` with `"Source Sans 3", "Source Sans Pro"` as the default font-family across all templates and examples
- Keeps `Source Sans Pro` as a fallback for systems that haven't updated the font name yet

## Test plan
- [x] Verify templates render with Source Sans 3 where installed, falling back to Source Sans Pro or system fonts
- [x] Spot-check a few templates in the browser to confirm no layout regressions from the font metric change

Screenshot 2026-05-22 at 14 53 24

### DIFF
--- a/.changeset/gentle-rivers-shine.md
+++ b/.changeset/gentle-rivers-shine.md
@@ -1,0 +1,12 @@
+---
+"@osdk/create-app.template.react": patch
+"@osdk/create-app.template.react.beta": patch
+"@osdk/create-app.template.vue": patch
+"@osdk/create-app.template.vue.v2": patch
+"@osdk/create-app.template.tutorial-todo-app": patch
+"@osdk/create-app.template.tutorial-todo-app.beta": patch
+"@osdk/create-app.template.tutorial-todo-aip-app": patch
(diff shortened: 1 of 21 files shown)


examples/example-react-sdk-2.x-no-osdk/src/index.css
```diff
@@ -4,8 +4,6 @@
 @import "@blueprintjs/icons/lib/css/blueprint-icons.css";
 
 :root {
-  font-family:
```
nfdiop, line 7: If we keep this, Blueprint font is overriding the font. Not sure that is what we want since we have this code present.

.changeset/gentle-rivers-shine.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/create-app.template.react.beta": patch
+"@osdk/create-app.template.vue.v2": patch
+---
+
+Use Source Sans 3 as the default font family with Source Sans Pro fallback
```
nfdiop, line 6: Source Sans 3 is the modern version of Source Sans Pro

samialfhaily (approved):

PR #3360, opened 2026-05-22
Scenarios runtime test: create scenario, apply action under it, read edits

nihalbhatnagar:
## Summary

Chained on #3334. Adds a runtime test under `packages/e2e.sandbox.catchall/src/` that exercises the scenarios surface end-to-end against a real Foundry stack:

1. `createScenario(client)` — mints a fresh scenario via `OntologyScenarios.createScenario`.
2. `scenario(createScenarioTestOsdk).applyAction({ string, scenarioRidCreatedOn })` — apply an action *under* the scenario context (`?scenarioRid=...` query param plumbed through `Actions.apply`).
3. `scenario.getEditedEntityTypes()` — verify the scenario sees the edit.
4. `scenario.getEditedEntities(ScenarioTestOsdk)` — read back edited primary keys.

Gated behind `testScenarios = false` in `index.ts` (mirroring `testSubscriptions`) so the default catchall run is unaffected. Flip the flag locally to opt in.

The catchall SDK is regenerated from the updated `ontology.json` which now carries:
- `ScenarioTestOsdk` object type
- `scenarioOsdkTestAction` (the scenario-merge action — has a `scenarioReference` parameter, exercises PR4's codegen end-to-end)
- `createScenarioTestOsdk` / `deleteScenarioTestOsdk` companions

## Test plan

- [x] `pnpm turbo typecheck --filter=@osdk/e2e.sandbox.catchall` clean
- [x] `pnpm turbo transpile --filter=@osdk/e2e.sandbox.catchall` clean
- [x] `pnpm turbo typecheck --filter=@osdk/e2e.generated.catchall` clean
- [ ] Manual run against a scenarios-enabled Foundry stack with `testScenarios = true` flipped in `index.ts` — observe `POST /v2/ontologies/{ontology}/scenarios/create`, `?scenarioRid=...` on `applyAction`, then `GET …/scenarios/{rid}/editedEntityTypes` + `…/objectTypes/ScenarioTestOsdk/objects`.


packages/e2e.sandbox.catchall/src/loggingFetch.ts
```diff
@@ -32,5 +32,5 @@ export async function loggingFetch(
   const cleaned = url.pathname.replace(/ri.ontology..*?\//, "{rid}/");
 
   logger.debug(`fetch(${chalk.blue(cleaned)})`);
-  return await fetch(input, init);
+  return await fetch(url, init);
```
ssanjay1, line 35: was this always broken?

ssanjay1 (approved):

packages/e2e.sandbox.catchall/src/loggingFetch.ts
nihalbhatnagar: I think its a no - op, theres a diff here because I was overriding the fetch fn and reset incorrectly

PR #3367, opened 2026-05-26
Fix useOsdkFunction objectSet param not serialized correctly

ziyunp:
Fix useOsdkFunction objectSet param not serialized with object set params, e.g. the where clause. This bug was previously seen in useOsdkFunctions. Extracted the common serializer for objectSet and updated useStableObjectSet and useOsdkFunctions to use the same stableSerialize function.


packages/react/src/new/__tests__/useOsdkFunctions.test.tsx
```diff
@@ -25,6 +25,17 @@ import type { UseOsdkFunctionOptions } from "../useOsdkFunction.js";
 import type { UseOsdkFunctionsProps } from "../useOsdkFunctions.js";
 import { useOsdkFunctions } from "../useOsdkFunctions.js";
 
+const MOCK_WIRE_FORM = Symbol.for("test.mockWireForm");
```
KuberSethi, line 28: can we match folder structure? apologies if it was inconsistent before but we should try and make it consistent!

packages/react/src/new/core/stableSerialize.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { getWireObjectSet, isObjectSet } from "@osdk/client";
+
+/**
+ * Serialize a value to a stable string for React.useMemo dependency arrays.
+ * Uses {@link stableSerializeReplacer} so OSDK `ObjectSet`s and
+ * `Osdk.Instance`s normalize to a discriminative wire form before
+ * stringification.
+ */
+export function stableSerialize(value: unknown): string {
```
KuberSethi: does this provide much value?

KuberSethi: shouldn't this return string or undefined? I don't think we're guaranteed a string out here

packages/react/src/new/core/stableSerialize.ts
```diff
@@ -0,0 +1,58 @@
...
+  _key: string,
+  value: unknown,
+): unknown {
+  if (value != null && typeof value === "object") {
+    if (isObjectSet(value)) {
+      return { __objectSet: getWireObjectSet(value) };
+    }
+    if ("$apiName" in value && "$primaryKey" in value) {
+      return {
+        $apiName: (value as { $apiName: unknown }).$apiName,
```
KuberSethi: after the if above can't we destructure safely without these casts?

packages/react/src/new/core/useStableObjectSet.ts
KuberSethi: should we refactor out the `getWireObjectSet` mention?

packages/react/src/new/__tests__/useOsdkFunctions.test.tsx
```diff
@@ -584,6 +595,72 @@ describe("useOsdkFunctions", () => {
     );
   });
 
+  describe("params memoization", () => {
+    // const MOCK_WIRE_FORM = Symbol.for("test.mockWireForm");
```
KuberSethi: delete?

.changeset/lazy-llamas-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/react": minor
+---
+
+Fix useOsdkFunction objectSet param not serialized correctly. Updated useStableObjectSet and useOsdkFunctions to use the same stableSerialize function
```
KuberSethi, line 5: should this be a patch if there's no user-facing change?

packages/react/src/new/core/stableSerialize.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * `JSON.stringify` replacer that normalizes OSDK runtime values to a stable
+ * wire form, suitable for use as a React.useMemo dependency-array key.
+ *
+ * - `ObjectSet` instances are replaced with their wire-form definition
+ *   (via `getWireObjectSet`) wrapped in `{ __objectSet: ... }`, so composed
+ *   operations (`.where`, `.union`, `.intersect`, ...) participate in the
+ *   key.
+ * - `Osdk.Instance`-shaped values (any object with both `$apiName` and
```
KuberSethi: Hmm this is a large behavioral / caching correctness change, we should think this through carefully. Does this fix / change need to include this change? Why do you consider them "unrelated" properties / are you confident they won't have any impact on cache correctness? Curious for your reasoning!

packages/react/src/new/core/stableSerialize.ts
ziyunp: This allows us to reuse the same logic anywhere where we do JSON.stringify. I'm also using this in useStableObjectSet hook which can be exposed to users

packages/react/src/new/core/stableSerialize.ts
ziyunp: This was taken from the original stableDependsOnObjects implementation. I can leave this out and keep just the objectSet replacer if you wanna reduce the risk

packages/react/src/new/__tests__/useOsdkFunctions.test.tsx
ziyunp: Moved `stableSerialize.test.ts` and `useStableObjectSet.test.ts` from `src/new/core/__tests__/` to `src/new/__tests__/` to match the established pattern (all existing tests live in `src/new/__tests__/`).

packages/react/src/new/core/stableSerialize.ts
ziyunp: Good catch — `JSON.stringify(undefined)` returns `undefined`, so the return type is now `string | undefined`. All callers use the result in a deps array where `undefined` is fine.

packages/react/src/new/core/stableSerialize.ts
ziyunp: Done — TS narrows correctly after the `in` checks, so destructuring works without casts.

packages/react/src/new/core/useStableObjectSet.ts
ziyunp: Updated the JSDoc to reference `stableSerialize` (which is what the hook actually calls) instead of `getWireObjectSet`.

packages/react/src/new/__tests__/useOsdkFunctions.test.tsx
ziyunp: Deleted.

.changeset/lazy-llamas-own.md
ziyunp: downgraded to `patch` since this is a bug fix with no API surface change.

packages/react/src/new/core/stableSerialize.ts
ziyunp: Removed it to reduce the scope of this PR

KuberSethi (approved):

PR #3369, opened 2026-05-26
Fix late registration of vite proxy

schmelczer:
The proxy server reloaded when new discovery files appeared. If 2 appeared close to the same time, the 2nd one got skipped.

Fix this by reloading as long as there are unseen discovery files.


packages/monorepo.cspell/dict.normal-dev-words.txt
```diff
@@ -1,3 +1,4 @@
+ACMR
```
schmelczer, line 1: alphabetise

ethanai22 (approved):

PR #3373, opened 2026-05-26
reject numeric literals on derived-property numeric expressions

KuberSethi:
## Summary
literal numbers like `.max(1)` typed as valid but threw "literals for derived properties are not yet supported" at runtime; tighten the @osdk/api types so the same call now fails at compile time

• narrow `NumericExpressionArg` so add/subtract/multiply/divide/max/min only accept another numeric derived property
• delete the unused, never-implemented `Constant<Q>` type from the builder surface
• prune the dead `extends number ? "double"` branch from `ExtractWirePropertyTypeFromNumericArg`

### DIFF
--- a/packages/api/src/derivedProperties/Expressions.ts
+++ b/packages/api/src/derivedProperties/Expressions.ts
@@ -46,8 +46,7 @@ export type DefinitionForType<
   : DerivedProperty.Definition<T, Q>;
 
 type NumericExpressionArg<Q extends ObjectOrInterfaceDefinition> =
-  | number
-  | DerivedProperty.NumericPropertyDefinition<any, Q>;
+  DerivedProperty.NumericPropertyDefinition<any, Q>;
 
 type ReturnTypeForNumericMethod<
   Q extends ObjectOrInterfaceDefinition,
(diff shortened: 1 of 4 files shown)


packages/api/src/derivedProperties/Expressions.ts
```diff
@@ -46,8 +46,7 @@ export type DefinitionForType<
   : DerivedProperty.Definition<T, Q>;
 
 type NumericExpressionArg<Q extends ObjectOrInterfaceDefinition> =
-  | number
```
bryantpark04, line 49: are there some contexts in which this doesn't throw? surely it was added for a reason

ssanjay1 (changes requested):
Would rather not do this, we kept these types in here because we will be introducing literals support soon. Is this causing a lot of issues for users?

KuberSethi:
> Would rather not do this, we kept these types in here because we will be introducing literals support soon. Is this causing a lot of issues for users?

It's more of the severity of errors. This is a really bad confusing error that leaves users confused and with no action (they literally posted '????' in our internal channel). Would you be amenable to more detailed error messages? Also when you say we're adding support soon, is there a timeline?

KuberSethi:
Closing for now since we're going to address the root cause, if this reemerges we will look into improving error handling here.

PR #3375, opened 2026-05-26
Add transitive OAC imports to SDK generation

ethanai22:
aka OAC grandparent problem.

Transitive interfaces on OAC imports that are not directly used do not show up in our block data/IR, so we were unable to generate SDK bindings for them. Our internal types encoded the info we need for transitive extended interfaces, but we didn't use them before this PR.

This PR adds a `transitiveImportedOntology` blockdata to the `OntologyIrV2` that parses the imported internal types for transitive entities that are not directly imported. We also add a new `linkedInterfaces` field to the internal InterfaceType that functions similarly to the `extendedInterfaces` but with the added complexity of handling cycles

This PR also writes the OAC import metadata to oac-imported-metadata.json so that it can be consumed by the CLI. A flup PR to the CLI will automatically look for this file


packages/maker/src/index.ts
```diff
@@ -129,4 +129,5 @@ export {
 } from "./api/propertyConversionUtils.js";
 export type { ValueTypeDefinitionVersion } from "./api/values/ValueTypeDefinitionVersion.js";
 export { wrapWithProxy } from "./api/wrapWithProxy.js";
+export { default } from "./cli/main.js";
```
ethanai22, line 132: this got accidentally removed previously

packages/maker-experimental/src/conversion/toMarketplace/convertOntologyDefinition.ts
```diff
@@ -46,9 +46,58 @@ export function convertOntologyDefinition(
...
+  const transitiveOntology: OntologyDefinition = {
+    INTERFACE_TYPE: transitiveInterfaces,
+    OBJECT_TYPE: {},
+    ACTION_TYPE: {},
+    LINK_TYPE: {},
+    SHARED_PROPERTY_TYPE: {},
+    VALUE_TYPE: {},
+  };
+
+  const throwawayRidGenerator = new OntologyRidGeneratorImpl(
```
ethanai22, line 88: throwaway generator so we don't pollute the main one

aep000 (approved):

PR #3377, opened 2026-05-27
Enable streaming in ObjectTable

ziyunp:
https://github.com/user-attachments/assets/ee7a6398-c2ce-497f-8e12-19b4c86a3b2a

### DIFF
--- a/packages/react-components/docs/ObjectTable.md
+++ b/packages/react-components/docs/ObjectTable.md
@@ -72,11 +72,12 @@ Add selection mode to enable row selection:
 
 ### Core Props
 
-| Prop         | Type     | Required | Default | Description                     |
-| ------------ | -------- | -------- | ------- | ------------------------------- |
-| `objectType` | `Q`      | ✅       | -       | The OSDK object type to display |
-| `className`  | `string` | ❌       | -       | CSS class for custom styling    |
-| `rowHeight`  | `number` | ❌       | `40`    | Height of each row in pixels    |
+| Prop            | Type      | Required | Default | Description                                                                                                                                                                                                                                                                                                                                                                         |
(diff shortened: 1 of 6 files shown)


packages/react-components/docs/ObjectTable.md
```diff
@@ -72,11 +72,12 @@ Add selection mode to enable row selection:
...
+| `rowHeight`     | `number`  | ❌       | `40`    | Height of each row in pixels                                                                                                                                                                                                                                                                                                                                                        |
+| `streamUpdates` | `boolean` | ❌       | `false` | When `true`, subscribes via websocket so the table automatically updates as matching objects are added, updated, or removed in Foundry. Cannot be used together with `pivotTo` or `withProperties` — the server does not support websocket subscriptions for link-traversal or derived-property queries, so those queries still fetch normally but won't receive real-time updates. |
```
nfdiop, line 80: nice comment

packages/react-components/src/object-table/hooks/useObjectTableData.ts
```diff
@@ -77,6 +77,7 @@ export function useObjectTableData<
   objectSetOptions?: ObjectSetOptions<Q>,
   dedupeIntervalMs: number = DEFAULT_OBJECT_TABLE_DEDUPE_INTERVAL_MS,
   pageSize: number = DEFAULT_PAGE_SIZE,
+  streamUpdates?: boolean,
```
nfdiop, line 80: nit: we can FLUP and have this take an object to avoid too many arguments

nfdiop (approved):

packages/react-components/src/object-table/hooks/useObjectTableData.ts
ziyunp: Sure we can review this later. I got a comment previously when I tried to bundle the network-related props together, like dedupeIntervalMs, pageSize, streamUpdates, but at that time the preference was to keep props at top-level.

PR #3379, opened 2026-05-27
Fix datetime picker time reset

nfdiop:
## Summary
- reset Today selection to the current wall-clock time in datetime picker
- preserve visible edited time when selecting calendar days
- derive TimePicker display from controlled value with only draft segment text stored locally

## Tests
- Added more tests
- Storybook

## Before
before_bug

## After
after


packages/react-components/src/shared/calendar/TimePicker.tsx
```diff
@@ -45,68 +41,49 @@ export const TimePicker: React.NamedExoticComponent<TimePickerProps> = React
     label = "Time",
   }: TimePickerProps): React.ReactElement {
     const valueTimestamp = value?.getTime() ?? null;
-    // Keep typed text locally so users can enter temporarily invalid values
-    // like "99" or "" without immediately mutating/clamping the parent Date.
-    const [internalSegments, setInternalSegments] = useState<InternalSegments>(
```
nfdiop, line 50: This was causing the bug: valueTimestamp could diverge from segments. The fix is to store just prevSegments then compare to new one

olliejenningsPT (approved):

PR #3392, opened 2026-05-28
Drop the `@osdk/aip-core` peer dependency

ziyunp:


ziyunp:
Fixed by https://github.com/palantir/osdk-ts/pull/3406/changes

PR #3394, opened 2026-05-28
show pressed and toggled states on filter list header buttons

KuberSethi:
clicking the search / × / ⋯ buttons on a FilterList row had no visible click feedback, and the toggled-open state on the search and overflow buttons was only signalled by full opacity — same color, no background — so users couldn't tell anything was selected

• add `:active` (mouse-down) styling on `.headerActionButton` using a new `--osdk-filter-item-header-action-bg-active` token (gray-2)
• give `aria-pressed="true"` / `data-popup-open` buttons a gray-1 background and the default-rest text color so the toggled state reads as "selected"
• document the two new tokens (`--osdk-filter-item-header-action-bg-active`, `--osdk-filter-item-header-action-color-active`) in both CSSVariables docs


nfdiop (commented):
Tested and noted that there is a missing padding when there is a background

Image

ziyunp:
Note that in dark mode the active button is not visible

Screenshot 2026-06-02 at 13 47 11

ziyunp:
> Tested and noted that there is a missing padding when there is a background
> 
> Image

Another example - it looks like keeping and excluding have different styling

Screenshot 2026-06-02 at 13 48 19

PR #3398, opened 2026-05-29
right-align clear all in filter-list exclude row via flex leading wrapper

KuberSethi:
the previous right-align fix (#3337) used `margin-left: auto` on the clear all button, which collapses when a consumer app's css reset zeroes out button margins

• wrap the keeping dropdown and count label in a leading div with `flex: 1 1 auto; min-width: 0` so it consumes remaining row space
• drop `margin-left: auto` on the clear all button; use `flex: 0 0 auto` and move the inter-item spacing onto the row gap and leading-group gap
• no api change, no consumer-visible behavior change in well-behaved css contexts

### DIFF
--- a/.changeset/filter-list-clear-all-flex-leading.md
+++ b/.changeset/filter-list-clear-all-flex-leading.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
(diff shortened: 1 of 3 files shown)


nfdiop (approved):
Would be nice to have a storybook repro

PR #3399, opened 2026-05-29
Sort attached MDX docs entries first within component folders

Leolide:
## Summary
- Add a `storySort` function to `preview.tsx` that puts docs entries before stories within the same component in the Storybook sidebar
- Preserves the existing top-level Docs category ordering (Welcome, Installation, Changelog, etc.)
- Falls back to alphabetical sorting by id for everything else

## Context
Storybook 10 renders attached MDX docs pages as the last entry inside a component folder by default. This change puts them first, matching the expected `autodocs` behavior.

## Test plan
- [ ] Verify "Docs" appears as the first entry inside DocumentViewer folder
- [ ] Verify top-level Docs category order is preserved (Welcome > Installation > Changelog > ...)
- [ ] Verify Components are still sorted correctly

### DIFF
--- a/packages/react-components-storybook/.storybook/preview.tsx
+++ b/packages/react-components-storybook/.storybook/preview.tsx
@@ -50,7 +50,10 @@ const mockClient = createClient(
 const preview: Preview = {
   parameters: {
     options: {
-      storySort: (a, b) => {
+      storySort: (
+        a: { title: string; type: string; id: string },
+        b: { title: string; type: string; id: string },
+      ) => {
         // Top-level category ordering
         const categoryOrder = ["Docs", "Components"];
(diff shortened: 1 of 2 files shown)


packages/react-components-storybook/.storybook/preview.tsx
```diff
@@ -49,26 +49,47 @@ const mockClient = createClient(
 
 const preview: Preview = {
   parameters: {
-    // Storybook 10 reads storySort from parameters.options (statically parsed)
     options: {
-      storySort: {
```
nfdiop, line 54: Could we try using `includeNames: true`? The TS expect error and the imperative code feels a bit off

nfdiop (approved):

PR #3404, opened 2026-05-29
Improve action form inputs

nfdiop:
## Summary
- match Blueprint input styling by rendering input borders/focus states with layered `box-shadow` instead of real borders
- add normal and danger focus shadow tokens, plus error shadow tokens for inputs
- keep normal input/file text hover neutral while preserving visible button-style hover affordances for interactive controls
- update file picker behavior so hovering the trigger highlights the Browse button, while the text region stays input-like
- update number input to match Blueprint
- keep deprecated public focus tokens for compatibility and update CSS variable docs to match the source tokens

## Before/after

Screenshot 2026-05-29 at 13 02 32

Screenshot 2026-05-29 at 13 01 05

Screenshot 2026-06-11 at 12 10 17

Screenshot 2026-06-11 at 12 10 11


packages/react-components-styles/CSS_VARIABLES.md
```diff
@@ -478,9 +478,9 @@ Styling for date and time picker components.
 | `--osdk-datetime-input-border-width`        | `var(--osdk-input-border-width)`                                            | Input border width         |
 | `--osdk-datetime-input-border-color`        | `var(--osdk-input-border-color)`                                            | Input border color         |
 | `--osdk-datetime-input-border-color-focus`  | `var(--osdk-input-border-color-focus)`                                      | Input focused border color |
-| `--osdk-datetime-input-focus-width`         | `var(--osdk-input-focus-width)`                                             | Input focus ring width     |
-| `--osdk-datetime-input-focus-color`         | `var(--osdk-input-focus-color)`                                             | Input focus ring color     |
-| `--osdk-datetime-input-focus-offset`        | `var(--osdk-input-focus-offset)`                                            | Input focus ring offset    |
+| `--osdk-datetime-input-focus-width`         | `var(--osdk-input-focus-width)`                                             | Deprecated                 |
```
nfdiop: Can track a deletion later on

packages/react-components/src/action-form/fields/FilePickerField.module.css
```diff
@@ -152,6 +149,10 @@
   }
 }
 
+.osdkFilePickerBrowse.osdkFilePickerBrowse:focus-visible {
+  outline: none;
```
nfdiop: otherwise there is an outline inside the FilerPicker browse button and it looks off. 

## Before
Image

## After
Image

Leolide: Nice here

Leolide:
I think in general it looks good to me

packages/react-components-styles/src/tokens/datetime-picker.css
```diff
@@ -118,9 +118,14 @@
   --osdk-datetime-input-border-width: var(--osdk-input-border-width);
   --osdk-datetime-input-border-color: var(--osdk-input-border-color);
   --osdk-datetime-input-border-color-focus: var(--osdk-input-border-color-focus);
+
+  --osdk-datetime-input-focus-outline: var(--osdk-input-focus-outline);
```
nfdiop, line 118: Introducing a single variable for the outline vs before where we had one for the different parts. We still need this token because we need to set it to none to avoid the browser default

packages/react-components-styles/src/tokens/input.css
```diff
@@ -25,9 +26,13 @@
...
 
   /* Input focus ring */
+  --osdk-input-focus-outline: none;
   --osdk-input-focus-width: var(--osdk-emphasis-focus-width);
   --osdk-input-focus-color: var(--osdk-emphasis-focus-color);
-  --osdk-input-focus-offset: var(--osdk-emphasis-focus-offset);
+  --osdk-input-focus-offset: 0px;
+  --osdk-input-focus-shadow-error: var(--osdk-input-shadow), 0 0 0
+    var(--osdk-input-focus-width)
+    color-mix(in srgb, var(--osdk-input-border-color-error) 75%, transparent);
```
nfdiop: Inspired from Blueprint
cc @Leolide

Leolide:
A tiny nit:
Screenshot 2026-06-02 at 9 16 32 AM

Screenshot 2026-06-02 at 9 16 55 AM

The number selector one I kinda want the button to be outside the input box

packages/react-components-styles/src/tokens/input.css
```diff
@@ -25,9 +26,13 @@
   --osdk-input-transition-ease: var(--osdk-emphasis-ease-default);
 
   /* Input focus ring */
+  --osdk-input-focus-outline: none;
   --osdk-input-focus-width: var(--osdk-emphasis-focus-width);
   --osdk-input-focus-color: var(--osdk-emphasis-focus-color);
-  --osdk-input-focus-offset: var(--osdk-emphasis-focus-offset);
+  --osdk-input-focus-offset: 0px;
+  --osdk-input-focus-shadow-error: var(--osdk-input-shadow), 0 0 0
```
olliejenningsPT: `--osdk-input-focus-shadow-error` references `--osdk-input-shadow`. When a consumer sets `--osdk-input-shadow: none` (e.g. EditableCell.module.css:106), this expands to `none, 0 0 0 ...` which is invalid `box-shadow` syntax and the entire declaration is discarded — the danger ring silently disappears. Consider decoupling: stack the shadows at the rule site instead (`box-shadow: var(--osdk-input-shadow), var(--osdk-input-focus-shadow-error);`) so `none` collapses cleanly via the cascade.

packages/react-components/src/action-form/fields/FilePickerField.module.css
```diff
@@ -152,6 +155,10 @@
   }
 }
 
+.osdkFilePickerBrowse.osdkFilePickerBrowse:focus-visible {
```
olliejenningsPT: Browse now has no visible focus indication: the outline is stripped here, and the trigger's `:has(...)` focus rules only key off `.osdkFilePickerText` and `.osdkFilePickerClear` (not `.osdkFilePickerBrowse`). Unless the underlying Button renders its own focus ring, a keyboard user tabbing onto Browse gets nothing. Also worth understanding what required the doubled `.osdkFilePickerBrowse.osdkFilePickerBrowse` selector — usually a sign of fighting specificity instead of fixing the source.

packages/react-components/src/shared/calendar/TimePicker.module.css
```diff
@@ -31,6 +31,10 @@
   &[aria-invalid] {
     border-color: var(--osdk-input-border-color-error);
   }
+
+  &[aria-invalid]:focus {
```
olliejenningsPT: Uses `:focus` while the equivalent new rules in BaseInput / NumberInputField / DatePickerCommon use `:focus-visible` (or `:focus-within`). With `:focus`, mouse clicks on an invalid TimePicker will paint the danger ring even though the keyboard wasn't used. The base rule above already uses `:focus` so this mirrors the file, but the cross-component inconsistency is worth either fixing here or addressing as a follow-up.

packages/react-components/src/action-form/fields/FilePickerField.module.css
nfdiop: let me look on the first and comment on .osdkFilePickerBrowse.osdkFilePickerBrowse. It's required for the internal browse button

packages/react-components-styles/src/tokens/file-picker.css
```diff
@@ -21,19 +21,31 @@
   --osdk-file-picker-transition-duration: var(--osdk-input-transition-duration);
   --osdk-file-picker-transition-ease: var(--osdk-input-transition-ease);
 
-  /* Focus ring */
+  /* Clear button focus ring */
+  --osdk-file-picker-clear-outline: var(--osdk-file-picker-trigger-focus-width)
```
nfdiop: This is a button different from the file input so adding outline to it

olliejenningsPT:
A few tokens look orphaned after this change removed their consumers — defined but no longer read anywhere (in both `react-components` and `react-components-styles`):

- `--osdk-datetime-input-error-border-color` (`datetime-picker.css:4`) — previously consumed by the `DatePickerCommon` error selector, now replaced by `--osdk-datetime-input-shadow-error`.
- `--osdk-datetime-input-border-color-focus` (`datetime-picker.css:119`) — was used in the wrapper's `:focus-within` border-color; that rule is gone (only re-declared as `transparent` in `EditableCell`).
- `--osdk-time-picker-input-border-color-focus` (`timepicker.css:14`) — was used in TimePicker's `:focus` border-color, now removed.
- `EditableCell.module.css` still overrides `--osdk-datetime-input-border-width` / `-border-color` / `-border-color-focus`, but the datetime wrapper is now `border: none`, so those overrides no longer do anything.

Since the branch already has dedicated cleanup commits ("Remove orphaned file picker tokens", "Remove legacy focus tokens"), these look like stragglers that slipped past the same pass rather than tokens intentionally kept.

olliejenningsPT (approved):

PR #3405, opened 2026-05-29
Add ObjectTable snapshot API

nfdiop:
## Summary
- add an ObjectTable `tableRef` snapshot API for currently loaded visible data
- expose pagination state and a safe `fetchNextPage` hook through snapshots
- add Storybook recipe showing loaded-row CSV download and consumer-owned paginated download-all flow 
- `?path=/story/components-objecttable-recipes--with-loaded-data-download` 

## Testing
- pnpm turbo typecheck --filter=@osdk/react-components
- pnpm turbo check-api --filter=@osdk/react-components
- pnpm --dir=packages/react-components exec vitest run src/object-table/utils/createObjectTableSnapshot.test.ts src/object-table/__tests__/ObjectTable.test.tsx src/public/experimental/__tests__/object-table.test.tsx
- pnpm turbo typecheck --filter=@osdk/react-components-storybook

### DIFF
--- a/packages/e2e.sandbox.peopleapp/src/app/employees/DownloadEmployeesButton.tsx
+++ b/packages/e2e.sandbox.peopleapp/src/app/employees/DownloadEmployeesButton.tsx
@@ -12,8 +12,9 @@ import type { Employee } from "../../generatedNoCheck2/index.js";
 // lines up.
 type RDPs = { managerName: "string" };
 
-// Cap the export so a large object set can't pull unbounded pages into the
-// client. getSnapshot pages until it has this many rows or the set is exhausted.
+// Bound pagination work so a large object set can't pull unbounded pages into
+// the client. This is a fetch threshold, not a CSV row cap: getSnapshot keeps
+// already-loaded rows and fetches full pages, so the CSV can contain more rows.
 const MAX_DOWNLOAD_ROWS = 10_000;
 
 export function DownloadEmployeesButton(
--- a/packages/e2e.sandbox.peopleapp/src/app/employees/DownloadEmployeesButton.tsx
+++ b/packages/e2e.sandbox.peopleapp/src/app/employees/DownloadEmployeesButton.tsx
@@ -1,0 +1,111 @@
+import type {
+  ObjectTableDataCell,
+  ObjectTableHandle,
+  ObjectTableSnapshot,
+} from "@osdk/react-components/experimental/object-table";
+import type { RefObject } from "react";
+import React, { useCallback, useState } from "react";
+import { Button } from "../../components/Button.js";
+import type { Employee } from "../../generatedNoCheck2/index.js";
+
+// Mirrors the RDP set EmployeesTable passes to ObjectTable, so the handle type
+// lines up.
+type RDPs = { managerName: "string" };
+
+// Cap the export so a large object set can't pull unbounded pages into the
+// client. getSnapshot pages until it has this many rows or the set is exhausted.
+const MAX_DOWNLOAD_ROWS = 10_000;
+
+export function DownloadEmployeesButton(
+  { tableRef }: {
+    tableRef: RefObject<ObjectTableHandle<Employee, RDPs> | null>;
+  },
+): React.ReactElement {
+  const [isDownloading, setIsDownloading] = useState(false);
+
+  const handleDownload = useCallback(async () => {
+    setIsDownloading(true);
+    try {
+      const snapshot = await tableRef.current?.getSnapshot({
+        rowLimit: MAX_DOWNLOAD_ROWS,
+      });
+      if (!snapshot) {
+        return;
+      }
+
+      await downloadCsv(snapshotToCsv(snapshot), "employees.csv");
+    } finally {
+      setIsDownloading(false);
+    }
+  }, [tableRef]);
+
+  return (
+    <Button type="button" onClick={handleDownload} disabled={isDownloading}>
+      {isDownloading ? "Downloading…" : "Download"}
+    </Button>
+  );
+}
+
+function snapshotToCsv(snapshot: ObjectTableSnapshot<Employee, RDPs>): string {
+  const { columns, rows } = snapshot;
+  return [
+    columns.map((column) => escapeCsvCell(column.name)).join(","),
+    ...rows.map((row) =>
+      columns
+        .map((column) =>
+          escapeCsvCell(formatCellValue(row.getValue(column.id)))
+        )
+        .join(",")
+    ),
+  ].join("\n");
+}
+
+function formatCellValue(cell: ObjectTableDataCell | undefined): string {
+  if (cell == null) {
+    return "";
+  }
+  if (cell.status === "error" && cell.value == null) {
+    return "";
+  }
+  return formatUnknownValue(cell.value);
+}
+
+function formatUnknownValue(value: unknown): string {
+  if (value == null) {
+    return "";
+  }
+  if (typeof value === "string") {
+    return value;
+  }
+  if (typeof value === "number" || typeof value === "boolean") {
+    return String(value);
+  }
+  try {
+    return JSON.stringify(value) ?? "";
+  } catch {
+    return String(value);
+  }
+}
+
+function escapeCsvCell(value: string): string {
+  if (!/[",\n\r]/.test(value)) {
+    return value;
+  }
+  return `"${value.replaceAll("\"", "\"\"")}"`;
+}
+
+async function downloadCsv(csv: string, fileName: string): Promise<void> {
+  const blob = new Blob([csv], { type: "text/csv;charset=utf-8" });
+  const url = URL.createObjectURL(blob);
+  const link = document.createElement("a");
+  link.href = url;
+  link.download = fileName;
+  document.body.append(link);
+  link.click();
+  link.remove();
+  // click() only queues the download; the browser reads the blob from the
+  // object URL on a later tick. Wait a tick before revoking so the download has
+  // started and the awaiting caller keeps its loading state until then.
+  await new Promise((resolve) => setTimeout(resolve, 0));
+  URL.revokeObjectURL(url);
+}
--- a/packages/react-components-storybook/src/stories/ObjectTable/Recipes/WithLoadedDataDownload.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ObjectTable/Recipes/WithLoadedDataDownload.stories.tsx
@@ -87,17 +87,6 @@ const handleDownload = async () => {
   downloadCsv(csv, "employees-loaded-rows.csv");
 };
 
-const handleLoadMoreRows = async () => {
-  const snapshot = await tableRef.current?.getSnapshot();
-  if (!snapshot?.hasNextPage) {
-    return;
-  }
-
-  await tableRef.current?.getSnapshot({
-    rowLimit: snapshot.rows.length + PAGE_SIZE,
-  });
-};
-
 const handleDownloadUpToLimit = async () => {
   const snapshot = await tableRef.current?.getSnapshot({
     rowLimit: ${MAX_DOWNLOAD_ROWS},
@@ -116,7 +105,6 @@ const handleDownloadUpToLimit = async () => {
 return (
   <>
     <button onClick={handleDownload}>Download loaded rows as CSV</button>
-    <button onClick={handleLoadMoreRows}>Load ${PAGE_SIZE} more rows</button>
     <button onClick={handleDownloadUpToLimit}>
       Download up to ${MAX_DOWNLOAD_ROWS} rows as CSV
     </button>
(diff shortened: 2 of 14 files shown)


packages/react-components/src/object-table/__tests__/ObjectTable.test.tsx
```diff
@@ -0,0 +1,290 @@
...
+  });
+
+  it("exposes table-level error and total count metadata", () => {
+    const error = new Error("Failed to load rows");
+    mockTableData({
+      data: [],
+      hasMore: false,
+      isLoading: false,
+      error,
+      totalCount: "42",
```
nfdiop: Our current API return these as strings so kept it that way.

packages/react-components/src/object-table/ObjectTable.tsx
```diff
@@ -296,6 +306,20 @@ export function ObjectTable<
 
   const isTableLoading = isLoading || isColumnsLoading;
 
+  useImperativeHandle(tableRef, () => ({
+    getLoadedData: () =>
```
nfdiop: not 100% sure of the API name here

packages/react-components/src/object-table/ObjectTable.tsx
```diff
@@ -296,6 +306,20 @@ export function ObjectTable<
...
+      createObjectTableLoadedData({
+        table,
+        hasNextPage: hasMore,
+        isLoading: isTableLoading,
+        error,
+        totalCount,
+      }),
+    loadNextPage: hasMore && fetchMore != null
```
nfdiop: separated from getLoadedData since I expect this one to change less than the data

nfdiop: Also, I was hesitating to type this more strongly and have this only defined if hasMore = true but decided that could be overkill

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -32,6 +32,62 @@ import type { CellEditInfo, EditFieldConfig } from "./utils/types.js";
...
+  /**
+   * Returns a fresh read of ObjectTable's currently loaded data and pagination
+   * state.
+   *
+   * The result includes visible data columns only, excludes internal control
+   * columns such as row selection, and does not fetch additional pages.
+   * Cell values come from the table's accessor values, not from rendered React
+   * content supplied by `renderCell`.
+   */
+  getLoadedData: () => ObjectTableLoadedData<Q, RDPs>;
```
nfdiop: want the review on this API. My idea is to have this simple function and later add options (for example allow users to see more than the visible columns)

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -32,6 +32,62 @@ import type { CellEditInfo, EditFieldConfig } from "./utils/types.js";
...
+  Q extends ObjectOrInterfaceDefinition,
+  RDPs extends Record<string, SimplePropertyDef> = Record<string, never>,
+> {
+  id: string;
+  object: Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>;
+  getValue: (columnId: string) => ObjectTableDataCell | undefined;
+}
+
+export type ObjectTableDataCell =
+  | { status: "ready"; value: unknown }
```
nfdiop, line 101: Adding the status to give information to the user

packages/react-components-storybook/src/stories/ObjectTable/Recipes/WithLoadedDataDownload.stories.tsx
```diff
@@ -0,0 +1,321 @@
...
+  reachedLimit: boolean;
+}> {
+  let snapshot = tableRef.current?.getSnapshot();
+  while (
+    snapshot?.hasNextPage
+    && snapshot.rows.length < MAX_DOWNLOAD_ROWS
+  ) {
+    await tableRef.current?.loadNextPage();
+    snapshot = tableRef.current?.getSnapshot();
+  }
```
olliejenningsPT: I think there's a timing thing worth calling out here. After `await tableRef.current.loadNextPage()`, the continuation runs as a microtask, but React may not have committed a render yet, so the next `getSnapshot()` can still see the previous `hasMore` / `totalCount` / row count.

In practice the await usually yields enough for the render to flush, but if fetchMore ever resolves without state advancing (a transient failure, or hasMore staying true while rows.length doesn't grow), this loop spins forever. Since this is the documented way to use the API, could we guard against no-progress iterations here, or add a JSDoc note on loadNextPage explaining that callers need to let React re-render before snapshotting again?

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -32,6 +32,62 @@ import type { CellEditInfo, EditFieldConfig } from "./utils/types.js";
...
+  getSnapshot: () => ObjectTableSnapshot<Q, RDPs>;
+
+  /**
+   * Fetches the next page using ObjectTable's existing pagination mechanism.
+   *
+   * This function is always safe to call. It resolves immediately when no next
+   * page is available. Use `getSnapshot().hasNextPage` to decide whether
+   * calling it can load additional rows.
+   */
+  loadNextPage: () => Promise<void>;
```
olliejenningsPT: Question on the contract here: this resolves to a noop whenever `hasMore` is false or `fetchMore` is null, but it doesn't account for an in-flight fetch. So if a user hits "Load next page" while a previous load is still resolving (isTableLoading is true but hasMore hasn't refreshed yet), the call silently does nothing rather than queueing or awaiting the in-flight load.

That's probably fine and consumers can gate on isLoading themselves, but I wanted to double check it's the intended behavior. If so, worth a line in the JSDoc so callers know they need to check isLoading too.

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -295,6 +351,15 @@ export interface ObjectTableProps<
    */
   objectType: Q;
 
+  /**
+   * Ref-like handle for reading a point-in-time snapshot of the currently
```
ziyunp, line 369: Nit - maybe remove "for reading a point-in-time snapshot", we may want to add more handlers here, e.g. there are asks to clearRowSelections

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -32,6 +32,62 @@ import type { CellEditInfo, EditFieldConfig } from "./utils/types.js";
...
+export interface ObjectTableHandle<
+  Q extends ObjectOrInterfaceDefinition,
+  RDPs extends Record<string, SimplePropertyDef> = Record<string, never>,
+> {
+  /**
+   * Returns a point-in-time snapshot of ObjectTable's currently loaded data
+   * and pagination state.
+   *
+   * The snapshot includes visible data columns only, excludes internal control
+   * columns such as row selection, and does not fetch additional pages.
```
ziyunp: Issue (blocking) - I think it should fetch all pages of objects, and we don't need loadNextPage function

packages/react-components/src/object-table/utils/createObjectTableSnapshot.ts
```diff
@@ -0,0 +1,124 @@
...
+): string {
+  const columnName = column.columnDef.meta?.columnName;
+  if (columnName != null) {
+    return columnName;
+  }
+
+  const header = column.columnDef.header;
+  if (typeof header === "string") {
+    return header;
+  }
```
ziyunp: Nit - in ObjectTableApi, I stated that renderHeader always wins over columnName. Can you reverse the order? 
```suggestion
  const header = column.columnDef.header;
  if (typeof header === "string") {
    return header;
  }
  
    const columnName = column.columnDef.meta?.columnName;
  if (columnName != null) {
    return columnName;
  }
```

packages/react-components/src/object-table/ObjectTableApi.ts
nfdiop: So I kept it because that is the current behaviour and we can remove when that changes

packages/react-components/src/object-table/ObjectTableApi.ts
nfdiop: Hum, not sure to follow. Can you elaborate what you want to see here?

packages/react-components/src/object-table/ObjectTableApi.ts
nfdiop: Let's talk async but I am hesitant to allow users to load all pages and potentially OOM the browser. Yes it can happen with the current API but users have to explicitly opt-in to the behaviour

packages/e2e.sandbox.peopleapp/src/app/employees/DownloadEmployeesButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+        .join(",")
+    ),
+  ].join("\n");
+}
+
+function formatCellValue(cell: ObjectTableDataCell | undefined): string {
+  if (cell == null) {
+    return "";
+  }
+  if (cell.status === "error" && cell.value == null) {
```
nfdiop, line 68: we probably need to handle better later when there is an error

packages/e2e.sandbox.peopleapp/src/app/employees/DownloadEmployeesButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const link = document.createElement("a");
+  link.href = url;
+  link.download = fileName;
+  document.body.append(link);
+  link.click();
+  link.remove();
+  // click() only queues the download; the browser reads the blob from the
+  // object URL on a later tick. Wait a tick before revoking so the download has
+  // started and the awaiting caller keeps its loading state until then.
+  await new Promise((resolve) => setTimeout(resolve, 0));
```
nfdiop, line 110: with this, you still have to wait between the button no longer being disabled and the file downloaded but the waiting time looks shorter

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
nfdiop, line 1: This is most of the complex logic. Essentially since we read the data from the tanstack table, we need to make sure we wait till we get the new data (after we loaded more rows)

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
```diff
@@ -0,0 +1,245 @@
...
+  > {
+    if (!mountedRef.current) {
+      return Promise.resolve();
+    }
+    return new Promise<void>((resolve) => {
+      commitWaitersRef.current.push(resolve);
+      // fetchMore resolves once the store holds the next page, but its commit
+      // may have landed before we parked above (or the page was identical and
+      // produced none). Force a commit so this resolves either way.
+      forceCommit();
```
nfdiop, line 144: The idea is to force a new render so that we get the latest data on the  ObjectTable. Will fLUP to think more about the logic

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
```diff
@@ -0,0 +1,245 @@
...
+          fetched = await Promise.race([
+            currentFetchMore().then(() => true),
+            unmountedRef.current.promise.then(() => false),
+          ]);
+        } catch {
+          // The underlying table hook publishes fetch failures through its
+          // committed error state. Resolve with the latest committed snapshot so
+          // imperative callers can inspect the same shape as other terminal
+          // pagination paths instead of handling a rejected getSnapshot promise.
+          return readSnapshot();
```
olliejenningsPT: One concern here: the comment says the underlying hook publishes fetch failures through its committed error state, but the test `resolves with the latest snapshot when fetching the next page fails` mocks fetchMore to throw while leaving error undefined, and asserts the resulting snapshot.error is undefined. So a caller doing `getSnapshot({ rowLimit: 10_000 })` for a CSV export can get back 200 rows with no indication anything failed.

Could we either surface the rejection on the snapshot, or call out in the JSDoc that callers can't rely on snapshot.error to detect a failed paginated fetch?

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
```diff
@@ -0,0 +1,245 @@
...
+  hasNextPage,
+  isLoading,
+  error,
+  totalCount,
+  fetchMore,
+}: UseObjectTableSnapshotHandleParams<Q, RDPs>): void {
+  // Bumped to force a render so the layout effect below republishes the latest
+  // inputs. Used as a fallback after a fetch whose committed page arrived before
+  // the loop parked its waiter, guaranteeing the parked waiter still resolves.
+  const [, forceCommit] = useReducer((version: number) => version + 1, 0);
```
ziyunp, line 86: Is this how the other table libraries do it?

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
ziyunp: Checked in workshop, they actually fetch the data separate from the data displayed in the table

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
nfdiop: It's a bit easier for them. In the example here https://www.mantine-react-table.com/docs/examples/export-csv,  they already have all the data. The main trouble here is that we can fetch more data but then there is a delay before tanstack table updates

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
nfdiop: changed. thanks!

packages/react-components/src/object-table/hooks/useObjectTableSnapshotHandle.ts
nfdiop: Do we want to do that? My approach here was to reuse the table data for efficiency and also not use the data directly since it could be different from the table.

ziyunp: No, as we discussed we will keep this approach now. Just a comment on the FLUP

ziyunp (approved):

PR #3408, opened 2026-05-31
use $title/$primaryKey special filters in object components

KuberSethi:
use the new $title/$primaryKey special filters to simplify object components and harden the observable matcher

PR #3298 added generic $title/$primaryKey where-clause filtering to the client; this adopts it in @osdk/react-components and removes a latent crash in the observable matcher.

• ObjectSelectField searches via the $title filter instead of resolving the title property from object metadata, dropping the metadata fetch and its loading gate
• ObjectTable derives its row-selection ObjectSet via $primaryKey, so interface-typed tables now produce a derived set on partial selection instead of undefined
• the observable where-clause matcher no longer throws when $startsWith is evaluated against a null $title


packages/react-components/src/object-table/__tests__/deriveSelectionObjectSet.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    expect(result).toBe(FILTERED);
+  });
+
+  it("returns undefined when there is no underlying object set", () => {
+    const result = deriveSelectionObjectSet(undefined, makeChange([1], false));
+
+    expect(result).toBeUndefined();
+  });
+
+  it("derives a set for interface-typed selections without a primary key property name", () => {
```
ziyunp: Which part of this makes it a test for interface type?

packages/react-components/src/action-form/fields/ObjectSelectField.tsx
```diff
@@ -80,24 +80,18 @@ const ObjectSelectInner: React.NamedExoticComponent<
     [onChange],
   );
 
-  const { metadata } = useOsdkMetadata(source.objectType);
-  const titleProperty = typeof metadata?.titleProperty === "string"
```
ziyunp, line 84: Nice!

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.titlePk.stories.tsx
ziyunp, line 1: Do we need a new story?

packages/react-components-storybook/src/stories/ActionForm/ActionForm.stories.tsx
```diff
@@ -949,3 +949,81 @@ function ControlledActionFormStory(): React.ReactElement {
     </FormStoryLayout>
   );
 }
+
+// ---------------------------------------------------------------------------
+// PR #3408 — ObjectSelectField search-by-$title coverage.
```
ziyunp: I also don't think we need extra stories here, we should just check that we don't break existing stories

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.titlePk.stories.tsx
KuberSethi: nope, removed it

packages/react-components-storybook/src/stories/ActionForm/ActionForm.stories.tsx
KuberSethi: agreed, dropped these. the existing stories cover the path and are untouched

packages/react-components/src/object-table/__tests__/deriveSelectionObjectSet.test.ts
KuberSethi: fair, not really a great test for interfaces, removed

ziyunp (approved):

PR #3410, opened 2026-06-01
Fix static FilterList value colors

nfdiop:
## Summary
- prevent static FilterList values from being treated as filtered-out count=0 aggregation rows
- keep filtered-out rows/options readable by removing row-level opacity while preserving muted text color
- update filtered-out token docs and regression coverage

## Storybook stories to test
- `Components/FilterList / With static values`
  - Static listogram values should render at normal contrast (not greyed out).
  - Open the Team multi-select dropdown; options should render normally.
- `Components/FilterList / Combined linked + direct filters (zero-count filtered-out rows)`
  - Select a Manager Name such as Margaret Jackson.
  - Department / Location City count=0 filtered-out values should remain readable and pass color-contrast checks.

## Before
Screenshot 2026-06-01 at 11 48 54

## After
Screenshot 2026-06-01 at 11 48 16

## Verification
- `pnpm --dir=packages/react-components exec vitest run src/filter-list/inputs/__tests__/filteredOutValues.test.tsx`
- Storybook visual check for the stories above
- axe color-contrast check in Storybook after selecting a Manager Name

### DIFF
--- a/packages/react-components/docs/CSSVariables.md
+++ b/packages/react-components/docs/CSSVariables.md
@@ -960,10 +960,9 @@ Canonical "No value" styling used by the shared `NoValueLabel` component across 
 
 Styling for count=0 muted rows surfaced by `showFilteredOutValues`. Applied to listogram rows via `[data-filtered-out]` and to multi-select items via the `filteredOutItem` class.
 
-| Variable                                  | Default Value                        | Description                         |
-| ----------------------------------------- | ------------------------------------ | ----------------------------------- |
-| `--osdk-filter-filtered-out-item-opacity` | `0.5`                                | Opacity for filtered-out facet rows |
-| `--osdk-filter-filtered-out-item-color`   | `var(--osdk-typography-color-muted)` | Text color for filtered-out rows    |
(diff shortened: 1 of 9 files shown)


packages/react-components/docs/CSSVariables.md
```diff
@@ -960,10 +960,9 @@ Canonical "No value" styling used by the shared `NoValueLabel` component across
 
 Styling for count=0 muted rows surfaced by `showFilteredOutValues`. Applied to listogram rows via `[data-filtered-out]` and to multi-select items via the `filteredOutItem` class.
 
-| Variable                                  | Default Value                        | Description                         |
-| ----------------------------------------- | ------------------------------------ | ----------------------------------- |
-| `--osdk-filter-filtered-out-item-opacity` | `0.5`                                | Opacity for filtered-out facet rows |
```
nfdiop, line 965: Dropping opacity for accessibility purposes

olliejenningsPT (dismissed):

olliejenningsPT (approved):

PR #3412, opened 2026-06-01
add officeassignment e2e sandbox

KuberSethi:
a sandbox app reproducing the "latest value of a linked status type" filtering pattern over a one-to-many link, dogfooding @osdk/react and @osdk/react-components.

• adds a Workforce-domain sandbox (Assignment → statusUpdate one-to-many) that builds the latest-value derived-property filter via targetMax/diff timestamp aggregations and narrows the object set
• wires FilterList facets, ObjectTable, a metrics header, and a detail drawer with the status timeline and the four actions
• registers the package in the viteSandboxes monorepolint archetype

PR #3415, opened 2026-06-01
add toDerivedNarrowing to filterlist for rdp-backed filters

KuberSethi:
lets a FilterList custom filter narrow the object set by runtime derived properties (RDPs), so predicates that need an aggregation over a link — e.g. "the latest linked row of a given kind has value X" — can live inside the filter list and compose with the other facets, instead of forcing a bespoke companion control + client-side key materialization.

• adds `toDerivedNarrowing` to custom filter definitions (returns `{ withProperties, where }`); `toWhereClause` is now optional so a filter supplies one or the other
• threads the active narrowings through `useFilterListState` and `narrowObjectSet`, applying them as `objectSet.withProperties(...).where(...)`, including the per-facet leave-one-out scopes so facet counts stay correct
• exposes the `DerivedNarrowing` type from the experimental filter-list entrypoint

stacked on #3412. next commit on this branch migrates the officeassignment sandbox to consume this (drops the matched-keys narrowing, adds the latest-status filter as a custom toDerivedNarrowing filter, plus a LINKED_PROPERTY any-match facet for contrast).


KuberSethi:

- #3412
    - #3415 ◀

PR #3416, opened 2026-06-01
fix broken guide links in react-components readme

KuberSethi:
the `Guide` backlinks in the components table (and a few other doc links) pointed at relative paths like `./docs/PdfViewer.md`, which npm resolves against the repo root instead of the package directory, so they 404 on npmjs.com

• switch every relative file link in the readme to absolute github urls under `packages/react-components/`
• covers the components table guides plus the contributing, css variables, add-new-component, and @osdk/react links
• patch changeset so the corrected readme actually republishes to npm

### DIFF
--- a/.changeset/fix-react-components-readme-links.md
+++ b/.changeset/fix-react-components-readme-links.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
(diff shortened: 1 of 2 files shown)


nfdiop (approved):
Thanks for the fix!

PR #3417, opened 2026-06-01
add fetchFullMetadata to Media for python-osdk parity

lucy-k-wang:
## BLUF
Typing in client code:
image

Adds `Media.fetchFullMetadata()` on TSV2 OSDK for parity with python-osdk's `get_media_full_metadata`. Returns a `MediaFullMetadata` wrapper around a `MediaItemMetadata` discriminated union, enabling type-narrowing in functions per the SDK Parity & Media GA workstream.

## Changes in this PR

- New optional `fetchFullMetadata?()` method on the `Media` interface, implemented in both ontology-backed and reference-backed `Media` impls via `MediaSets.metadata` (no Ontologies-backed full-metadata endpoint exists, mirrors python-osdk behavior).
- Discriminator-trimmed `MediaItemMetadata` union (9 variants) mirrored in `@osdk/api` to preserve the `@osdk/foundry.*` decoupling rule. Per-variant fields (page counts, dimensions, audio specs, etc.) remain on `@osdk/foundry.mediasets`; callers wanting variant-specific shape narrow + cast as documented.
- tsv1-style `is*MediaItemMetadata` type guards in `@osdk/api`, re-exported from `@osdk/client`.
- Unit tests cover both `Media` impls; e2e exercise in `runMediaTest.ts` validates a real PNG returns the `imagery` variant.

## Open questions

- Method on `Media` vs experimental helper (per #2748 precedent for `createMediaFromReference`). Landing as a method since this is a read-only sibling of the existing stable `fetchMetadata` / `fetchContents`, and matches python's `media.get_media_full_metadata()` shape. Push back if you'd prefer the experimental gating.

## Out of scope

- `streamingVideo` variant: mio supports it internally but api-gateway doesn't expose it on the Platform SDK yet. Tracked separately for GA.
- `originallyUploadedFileMimeType` exposure on slim metadata: a separate gap surfaced during analysis.


packages/api/src/object/mediaTypeGuards.ts
```diff
@@ -0,0 +1,78 @@
...
+import type { MediaItemMetadata } from "./Media.js";
+
+/**
+ * Type-guard helpers for narrowing `MediaItemMetadata`. Mirrors the tsv1 `is*` guards
+ * (`isDocumentMetadata`, `isImageryMetadata`, etc.) from the functions-typescript-runtime
+ * `MediaReference` family. After narrowing, all variant-specific fields are accessible with
+ * full IntelliSense (no cast required).
+ */
+
+export function isDocumentMediaItemMetadata(
```
nihalbhatnagar: Do we need these type guards? Can we just rely on narrowing with `.type` for now

packages/api/src/object/Media.ts
```diff
@@ -95,3 +98,439 @@ export interface MediaMetadata {
   sizeBytes: number;
   mediaType: string;
 }
+
+/**
+ * Wrapper around the type-specific `MediaItemMetadata` discriminated union returned by
+ * `Media.fetchFullMetadata`. The wrapper exists so optional sibling fields (e.g. attribution,
```
nihalbhatnagar: Second sentence is an implementation detail, and can we reword the first to be more generic (This i the metdata for media, different from `MediaMetadata`

packages/api/src/object/Media.ts
```diff
@@ -95,3 +98,439 @@ export interface MediaMetadata {
...
+   * Type-specific metadata. Narrow on `itemMetadata.type` (or use the `is*MediaItemMetadata`
+   * helpers) to access variant-specific fields.
+   */
+  itemMetadata: MediaItemMetadata;
+}
+
+/**
+ * Type-specific metadata for a media item, discriminated by `type`.
+ *
+ * **Full structural mirror of `MediaItemMetadata` from `@osdk/foundry.mediasets`.** Every
```
nihalbhatnagar: Can we remove most of this JS doc, its implementation focused

packages/api/src/object/Media.ts
```diff
@@ -95,3 +98,439 @@ export interface MediaMetadata {
...
+
+export type PaletteInterpretation = "GRAY" | "RGB" | "RGBA" | "CMYK" | "HLS";
+
+export interface UnitInterpretation {
+  unit?: string;
+  scale?: number;
+  offset?: number;
+}
+
+export type ImageAttributeDomain = LooselyBrandedString<"ImageAttributeDomain">;
```
nihalbhatnagar: Use the direct string not a loosely branded string

packages/api/src/object/Media.ts
```diff
@@ -95,3 +98,439 @@ export interface MediaMetadata {
...
+  mimeType: string;
+}
+
+// ─── Brand utility ───────────────────────────────────────────────────────────
+
+/**
+ * Loose brand: structurally a string, with an optional phantom field for IntelliSense.
+ * Matches the `@osdk/foundry.mediasets` definition so brand-bearing types unify across packages.
+ */
+export type LooselyBrandedString<T extends string> = string & {
```
nihalbhatnagar: Remove

packages/client/src/object/mediaItemMetadataMirror.test.ts
```diff
@@ -0,0 +1,331 @@
...
+ * Pins the `@osdk/api` `MediaItemMetadata` mirror to the platform shape exposed by
+ * `@osdk/foundry.mediasets`. If the platform schema changes (a field added, removed, made
+ * optional, an enum value added, a nested type restructured), these assertions fail at
+ * `typecheck` time and the mirror must be updated.
+ *
+ * The mirror exists so `@osdk/api` doesn't take a runtime / declaration dep on
+ * `@osdk/foundry.mediasets`. The verify lives here in `@osdk/client` (which already depends on
+ * both) and turns the maintenance burden from "silent drift over time" into "loud build break".
+ */
+describe("MediaItemMetadata mirror", () => {
```
nihalbhatnagar, line 125: I do kind of like these tests, how often are we expecting changes to this union here?

packages/e2e.sandbox.catchall/src/runMediaTest.ts
```diff
@@ -252,6 +424,10 @@ export async function runMediaTest(): Promise<void> {
   const testImage: Blob = await runReadMediaTest(result.mediaReference);
   console.log("SUCCESS: Reading Media Reference");
 
+  console.log("Reading Full Media Metadata");
+  await runReadMediaFullMetadataTest(result.mediaReference);
```
nihalbhatnagar, line 433: Did this actually work for the media object

packages/client/src/createMediaReferenceProperty.ts
```diff
@@ -70,6 +77,24 @@ export class MediaReferencePropertyImpl implements Media {
...
 
+  /**
+   * Fetches type-specific metadata via `MediaSets.metadata`. There is no Ontologies-backed
+   * full-metadata endpoint, so this routes through the media set view using the embedded media
+   * reference.
+   */
+  public async fetchFullMetadata(): Promise<MediaFullMetadata> {
+    const { mediaSetRid, mediaItemRid, token } =
+      this.#mediaReference.reference.mediaSetViewItem;
+    const itemMetadata: MediaItemMetadata = await MediaSets.metadata(
```
nihalbhatnagar: We should have a runtime guard against new union types and return an unknown type

nihalbhatnagar (commented):
Few nits, but overall lgtm

packages/api/src/object/Media.ts
lucy-k-wang: `@osdk/foundry.mediasets` ships `LooselyBrandedString` for `ImageAttributeDomain`, `ImageAttributeKey`, and `DicomDataElementKey`. The mirror test in `[mediaItemMetadataMirror](https://github.com/palantir/osdk-ts/pull/3417/files/4841d30c5afb07d61063209907ca11cc646afa61#diff-4c35c29e6f8e5f83fb370a4b785129e3e354e2c4f3ab2f15647dcd10a9494a85).test.ts` asserts type-equality with `toEqualTypeOf`, so plain `string` fails.

We can downgrade the asserts from `toEqualTypeOf` to compatibility checks, and use plain `string` here. Should we go with that instead?

packages/client/src/object/mediaItemMetadataMirror.test.ts
lucy-k-wang: SHouldn't be very often. variant additions are major events (new schema), nested field tweaks are minor and infrequent. 

The asserts are typecheck-only and feel worthwhile to get a loud build break

packages/e2e.sandbox.catchall/src/runMediaTest.ts
lucy-k-wang: what do you mean?

packages/client/src/createMediaReferenceProperty.ts
lucy-k-wang: is this to guard against silent narrowing failures when a stale sdk hits a new server variant (newer osdk.mediasets version)?

nihalbhatnagar: Yeah I feel like we should be opinionated about providing runtime safety for the end user

packages/e2e.sandbox.catchall/src/runMediaTest.ts
nihalbhatnagar: Sorry I meant were you able to run this and have it work

packages/api/src/object/Media.ts
nihalbhatnagar: Yeah that sounds good to me

packages/e2e.sandbox.catchall/src/runMediaTest.ts
lucy-k-wang: yes, ran the ete on an image and got:
```
Full metadata variant: imagery
imagery: {
  format: 'PNG',
  dimensions: { width: 4032, height: 3024 },
  bandCount: 3,
  iccProfile: '...',
  sizeBytes: 18484484
}
SUCCESS: Reading Full Media Metadata
```
The imagery branch of the `switch` fires and per-variant fields are accessible after narrowing on `metadata.type` = "imagery" switch case.

packages/client/src/createMediaReferenceProperty.ts
lucy-k-wang: what do you think about: [packages/client/src/object/validateMediaItemMetadata.ts](https://github.com/palantir/osdk-ts/pull/3417/changes/b71fadb769223b91da06f163051489f477ec3aa9#diff-f1c2099eb2b9cf3a6af0578062921f3adaa36c57d1f1b18ad42298091c9b8d6dR51)

packages/api/src/object/Media.ts
```diff
@@ -95,3 +88,436 @@ export interface MediaMetadata {
   sizeBytes: number;
   mediaType: string;
 }
+
+/**
+ * Metadata for a media item, returned by `Media.fetchFullMetadata`. Distinct from
+ * the generic `MediaMetadata`.
```
nihalbhatnagar: This comment doesnt make sense, its already distinct right

packages/api/src/object/Media.ts
```diff
@@ -95,3 +88,436 @@ export interface MediaMetadata {
...
+  | ({ type: "dicom" } & DicomMediaItemMetadata)
+  | ({ type: "email" } & EmailMediaItemMetadata)
+  | ({ type: "model3d" } & Model3dMediaItemMetadata)
+  | ({ type: "spreadsheet" } & SpreadsheetMediaItemMetadata)
+  | ({ type: "untyped" } & UntypedMediaItemMetadata);
+
+/**
+ * Fallback variant emitted by the SDK when the platform returns a `MediaItemMetadata.type`
+ * value not modeled by this version of `@osdk/foundry.mediasets`. The raw wire payload is
+ * preserved on `raw` so callers can inspect it. Bump the platform SDK to promote the new
```
nihalbhatnagar: say bump client, not platform SDK

packages/client/src/object/mediaItemMetadataMirror.test.ts
nihalbhatnagar: Well any type changes would be a break, we're specifically testing union extensions right and new fields right?

nihalbhatnagar (commented):
lgtm, some nits then will +1 since itll invalidate

packages/client/src/object/mediaItemMetadataMirror.test.ts
lucy-k-wang: yep, additions to the union and new fields on existing variants

nihalbhatnagar (approved):

PR #3419, opened 2026-06-01
make react-devtools package public again

KuberSethi:
reverses the react-devtools-specific parts of #3210 so the package publishes to npm again

• drop `"private": true` from `packages/react-devtools/package.json`
• restore the real `files` allowlist (was blanked to `[]`)
• remove `private: true` from the react-devtools archetype in `.monorepolint.config.mjs`

### DIFF
--- a/.changeset/react-devtools-public-again.md
+++ b/.changeset/react-devtools-public-again.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-devtools": patch
+---
+
+Make react-devtools package public again
--- a/.monorepolint.config.mjs
+++ b/.monorepolint.config.mjs
@@ -421,7 +421,6 @@ const archetypeRules = archetypes(
       react: true,
       output: OUTPUT_ESM_ONLY,
       cssExport: ["styles.css"],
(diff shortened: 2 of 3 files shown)


nfdiop (commented):
What's changed in this approach compared to last time? Want to make sure we're covering the gaps.

KuberSethi:
> What's changed in this approach compared to last time? Want to make sure we're covering the gaps.

I guess I'm not sure why we removed publishing considering there are releases for this https://www.npmjs.com/package/@osdk/react-devtools?activeTab=versions, perhaps trusted publishing wasn't enabled before? It is now though I believe cc @ssanjay1

nfdiop:
> > What's changed in this approach compared to last time? Want to make sure we're covering the gaps.
> 
> I guess I'm not sure why we removed publishing considering there are releases for this https://www.npmjs.com/package/@osdk/react-devtools?activeTab=versions, perhaps trusted publishing wasn't enabled before? It is now though I believe cc @ssanjay1

amazing!

nfdiop (approved):

PR #3420, opened 2026-06-02
add chromatic visual regression testing for storybook

KuberSethi:
adds chromatic to catch visual regressions against the react-components storybook

it's not super smart in running only on PRs where it makes sense (uses the same paths filtering the storybook preview jobs do) but if this turns out to be useful we should look into integrating https://www.chromatic.com/docs/turbosnap/ for more performant / smarter running

• new `.github/workflows/chromatic.yml` runs on prs and pushes to main, skips fork prs, and is non-blocking (`exitZeroOnChanges`); not wired into the required ci-build gate.
• adds the chromatic dep and a local `pnpm chromatic` script that builds storybook with a root base path so stories render correctly in chromatic instead of under the github pages path
• token is read from the `CHROMATIC_PROJECT_TOKEN` repo secret which has not been configured and i think we'll need devtools to help us configure? unsure but will figure out the process here  

~~note: the chromatic check will be red until the pre-existing PdfViewerAnnotationLayer story crashes are fixed in https://github.com/palantir/osdk-ts/pull/3421~~

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -19,6 +19,10 @@ on:
 
 permissions: {}
 
+concurrency:
+  group: chromatic-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   chromatic:
     name: Run Chromatic
@@ -30,8 +34,9 @@ jobs:
       contents: read
     steps:
       - name: Check out code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
+          ref: ${{ github.event.pull_request.head.sha }}
           # Chromatic needs full git history to track baselines across commits.
           fetch-depth: 0
           persist-credentials: false
@@ -39,7 +44,7 @@ jobs:
       - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4.3.0
 
       - name: Setup Node.js environment
-        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: 24
           cache: "pnpm"
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -6,8 +6,16 @@ env:
 on:
   push:
     branches: [main]
+    paths:
+      - 'packages/react-components/**'
+      - 'packages/react-components-storybook/**'
+      - '.github/workflows/chromatic.yml'
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - 'packages/react-components/**'
+      - 'packages/react-components-storybook/**'
+      - '.github/workflows/chromatic.yml'
 
 permissions: {}
 
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,0 +1,60 @@
+name: Chromatic
+
+env:
+  DO_NOT_TRACK: 1
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize, reopened]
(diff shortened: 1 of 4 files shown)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,68 @@
...
+      - '.github/workflows/chromatic.yml'
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'packages/react-components/**'
+      - 'packages/react-components-storybook/**'
+      - '.github/workflows/chromatic.yml'
+
+permissions: {}
+
```
olliejenningsPT, line 21: Consider adding a `concurrency` group so rapid pushes to a PR cancel in-progress runs instead of burning Chromatic snapshot quota:

```yaml
concurrency:
  group: chromatic-${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,68 @@
...
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Check out code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          # Chromatic needs full git history to track baselines across commits.
```
olliejenningsPT, line 40: `storybook-preview.yml` explicitly pins `ref: ${{ github.event.pull_request.head.sha }}` here; without it, `pull_request` events check out the synthetic `refs/pull/N/merge` commit. Worth confirming Chromatic associates the build with the PR head SHA rather than the merge SHA.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,68 @@
...
+          fetch-depth: 0
+          persist-credentials: false
+
```
olliejenningsPT, line 43: Version drift vs. `storybook-preview.yml`, which uses `actions/checkout@v4.3.1` and `actions/setup-node@v4.4.0`. Worth aligning the two workflows on the same major versions.

ziyunp:
Question about what we get when this PR is merged - my understanding is the chromatic job runs on main and open PRs. Does it mean that the run on main takes visual snapshots and published somewhere (?) and PR runs are compared against the main snapshots? Can you add examples of failed and successful visual tests?

.github/workflows/chromatic.yml
KuberSethi: done, added it

.github/workflows/chromatic.yml
KuberSethi: done, pinned to the pr head sha to match storybook-preview

.github/workflows/chromatic.yml
KuberSethi: aligned to storybook-preview's pins

KuberSethi:
> Question about what we get when this PR is merged - my understanding is the chromatic job runs on main and open PRs. Does it mean that the run on main takes visual snapshots and published somewhere (?) and PR runs are compared against the main snapshots? Can you add examples of failed and successful visual tests?


Yep that's right. There's no real pass/fail example to add in code though, a fail is just a story that differs from its baseline (which is in Chromatic), and once a story is committed it becomes the baseline so it passes. you'd see the actual pass/fail in the Chromatic app once the token's wired up and this is all working.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,73 @@
...
+    paths:
+      - 'packages/react-components/**'
+      - 'packages/react-components-storybook/**'
+      - '.github/workflows/chromatic.yml'
+
+permissions: {}
+
+concurrency:
+  group: chromatic-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
olliejenningsPT, line 24: cancel-in-progress is the right call for PR refs but a bit risky for pushes to main. If two commits land on main close together, the in-progress run gets cancelled and the baseline for that commit never gets recorded. Could gate this on the event type, something like cancel-in-progress: ${{ github.event_name != 'push' }}, or use a separate concurrency group for main.

ziyunp:
Created a new PR https://github.com/palantir/osdk-ts/pull/3979

PR #3425, opened 2026-06-02
Surface CBAC/MANDATORY marking subtype on ObjectMetadata.Property

olliejenningsPT:
## Summary

Surfaces the CBAC/MANDATORY marking subtype from the v2 ontology metadata API on `ObjectMetadata.Property`. Step one of a two-step change that ends with `@osdk/react-components`' `ObjectTable` rendering marking columns with CBAC banners and MANDATORY pills, matching Workshop.

The wire shape this consumes is the new `markingType?: \"CBAC\" | \"MANDATORY\"` field on `Core.MarkingType` shipped in api-gateway 1.1623.0 ([foundry/api-gateway#15407](https://github.palantir.build/foundry/api-gateway/pull/15407)) and surfaced in `@osdk/foundry.core@2.63.0`. Main is already on that catalog version, so this PR is purely additive on top.

## Changes

- `packages/api/src/ontology/ObjectTypeDefinition.ts`: new optional `markingType?: \"CBAC\" | \"MANDATORY\"` field on `ObjectMetadata.Property`. Absent for non-marking properties and for marking properties whose subtype is not exposed by the platform.
- `packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts`: reads `markingType` from the wire `Core.MarkingType` schema and forwards it through for both single `marking` and array-of-marking property variants.
- `packages/generator-converters/src/wireObjectTypeFullMetadataToSdkObjectMetadata.test.ts`: new test covering CBAC, MANDATORY, unspecified, and array-of-marking cases.
- `etc/api.report.api.md`: regenerated to capture the new optional field.

Incidental (not on the marking-type critical path): `packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts` extends the `IDiscoveredFunction.inputs` element with an optional `required?: boolean` so TypeScript function discovery can forward per-parameter requiredness instead of defaulting everything to `true`. Picked up during conflict resolution; happy to split out if reviewers prefer.

## Follow-up (PR B)

`@osdk/react-components` will plumb `markingType` through column meta, render CBAC columns via `@osdk/cbac-components`' `CbacBanner`, and dispatch MANDATORY through a per-marking pill list. Out of scope here.

### DIFF
--- a/packages/api/src/ontology/ObjectTypeDefinition.ts
+++ b/packages/api/src/ontology/ObjectTypeDefinition.ts
@@ -105,13 +105,29 @@ export namespace ObjectMetadata {
     };
     hasReducers?: boolean;
     /**
-     * For properties of type `"marking"`, indicates whether the marking is
-     * classification-based access control (`"CBAC"`) or mandatory (`"MANDATORY"`).
-     * Absent for non-marking properties and for marking properties whose subtype
-     * is not exposed by the platform.
+     * Per-`type` metadata, discriminated on the wire property `type`.
+     *
+     * New per-type fields should be added as variants of {@link PropertyTypeMetadata}
+     * rather than as new top-level optionals on `Property`, so that the shape
+     * stays narrow and illegal combinations (e.g. a marking subtype on a `double`)
+     * are not representable.
      */
-    markingType?: "CBAC" | "MANDATORY";
+    typeMetadata?: PropertyTypeMetadata;
   }
+
+  /**
+   * Discriminated union of per-`type` property metadata. Narrow on
+   * `typeMetadata.type` to access the variant-specific fields.
+   */
+  export type PropertyTypeMetadata = {
+    /**
+     * Marking subtype: `"CBAC"` for classification-based access control,
+     * `"MANDATORY"` for mandatory markings. Absent for marking properties
+     * whose subtype is not exposed by the platform.
+     */
+    type: "marking";
+    subtype?: "CBAC" | "MANDATORY";
+  };
 
   export interface Link<
     Q extends ObjectTypeDefinition,
--- a/packages/api/src/ontology/ObjectTypeDefinition.ts
+++ b/packages/api/src/ontology/ObjectTypeDefinition.ts
@@ -104,6 +104,13 @@ export namespace ObjectMetadata {
       fields: readonly string[];
     };
     hasReducers?: boolean;
+    /**
+     * For properties of type `"marking"`, indicates whether the marking is
+     * classification-based access control (`"CBAC"`) or mandatory (`"MANDATORY"`).
+     * Absent for non-marking properties and for marking properties whose subtype
+     * is not exposed by the platform.
+     */
+    markingType?: "CBAC" | "MANDATORY";
   }
 
   export interface Link<
--- a/packages/api/src/ontology/ObjectTypeDefinition.ts
+++ b/packages/api/src/ontology/ObjectTypeDefinition.ts
@@ -126,7 +126,7 @@ export namespace ObjectMetadata {
      * whose subtype is not exposed by the platform.
      */
     type: "marking";
-    subtype?: "CBAC" | "MANDATORY";
+    markingType?: "CBAC" | "MANDATORY";
   };
 
   export interface Link<
--- a/packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts
+++ b/packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts
@@ -464,7 +464,7 @@ export class OntologyIrToFullMetadataConverter {
         >((acc, input) => {
           acc[input.name] = {
             dataType: convertDataType(input.dataType, func.customTypes),
-            required: input.required ?? false,
+            required: input.required ?? true,
           };
           return acc;
         }, {}),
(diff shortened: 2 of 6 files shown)


packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts
```diff
@@ -464,7 +464,7 @@ export class OntologyIrToFullMetadataConverter {
         >((acc, input) => {
           acc[input.name] = {
             dataType: convertDataType(input.dataType, func.customTypes),
-            required: true,
```
ziyunp, line 467: Why was it default to true? What's the impact of this change?

packages/generator-converters/src/wireObjectTypeFullMetadataToSdkObjectMetadata.test.ts
```diff
@@ -242,6 +242,64 @@ describe(wireObjectTypeFullMetadataToSdkObjectMetadata, () => {
...
+            typeClasses: [],
+          },
+          unspecifiedMarking: {
+            dataType: { type: "marking" },
+            rid: "rid",
+            typeClasses: [],
+          },
+          arrayOfMandatoryMarkings: {
+            dataType: {
+              type: "array",
```
ziyunp, line 280: Can array type have a list of different markings?

ziyunp (dismissed):

packages/generator-converters/src/wireObjectTypeFullMetadataToSdkObjectMetadata.test.ts
olliejenningsPT: you can have an array of markings

packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts
olliejenningsPT: because they now have the required propertyy, and so we are readying from that instead

packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
```diff
@@ -98,6 +97,21 @@ export function wirePropertyV2ToSdkPropertyDefinition(
...
+        multiplicity: false,
+        description: input.description,
+        type: sdkPropDefinition,
+        nullable: input.nullable == null ? isNullable : input.nullable,
+        valueTypeApiName: input.valueTypeApiName,
+        valueFormatting: input.valueFormatting != null
+          ? wirePropertyFormattingToSdkFormatting(input.valueFormatting, log)
+          : undefined,
+        ...(input.dataType.markingType != null
+          ? { markingType: input.dataType.markingType }
```
nihalbhatnagar: Maybe we should make a typed metadata field here with the wire property type as a discriminator so this doesn't get bloated? Arguably should have done for reducer/main value

olliejenningsPT: We can do, but if we do this for just markingType it might look confusing if only markingType is populated. But if we are happy with that approach, then lets do it

packages/api/src/ontology/ObjectTypeDefinition.ts
```diff
@@ -104,8 +104,31 @@ export namespace ObjectMetadata {
...
+   * `typeMetadata.type` to access the variant-specific fields.
+   */
+  export type PropertyTypeMetadata = {
+    /**
+     * Marking subtype: `"CBAC"` for classification-based access control,
+     * `"MANDATORY"` for mandatory markings. Absent for marking properties
+     * whose subtype is not exposed by the platform.
+     */
+    type: "marking";
+    subtype?: "CBAC" | "MANDATORY";
```
nihalbhatnagar: markingType ?

packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts
nihalbhatnagar: Should we default to true to keep behavior from before in the missing case and match this https://github.com/palantir/osdk-ts/blob/5b4ca298628e89093848b1229ac9cc529b5df7be/packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts#L610-L611

packages/generator-converters/src/wirePropertyV2ToSdkPropertyDefinition.ts
```diff
@@ -53,6 +53,25 @@ function hasReducers(
...
+ * added here so each call site stays a one-liner.
+ */
+function extractTypeMetadata(
+  dataType: ObjectPropertyType,
+): ObjectMetadata.PropertyTypeMetadata | undefined {
+  const inner = dataType.type === "array" ? dataType.subType : dataType;
+  if (inner.type === "marking") {
+    return {
+      type: "marking",
+      ...(inner.markingType != null ? { subtype: inner.markingType } : {}),
```
nihalbhatnagar: Can we add similar handling as we did here https://github.com/palantir/osdk-ts/blob/675317e703ddfcc017c064ad08c9be4b4e95d117/packages/generator-converters/src/wireObjectTypeFullMetadataToSdkObjectMetadata.ts#L158, theoretically we dont generate this metadata so it won't throw but its probably better to be safe

nihalbhatnagar (dismissed):

nihalbhatnagar (approved):

PR #3427, opened 2026-06-02
Document array element-subtype filters via $contains

nihalbhatnagar:
## Summary

Adds an optional render-time `arrayContains` variable to the where-clause filter snippet templates so the dev console can document filtering ARRAY properties by their element subtype. When `arrayContains` is set, the chosen filter is wrapped inside `$contains: { ... }`, demonstrating subtype filtering — e.g. `myStringArray: { $contains: { $startsWith: "foo" } }`, `myNumberArray: { $contains: { $gt: 5 } }`.

This mirrors the existing `structSubPropertyApiName` mechanism. `arrayContains` is a render-time Mustache optional variable (provided by the dev console), NOT a `computedVariable`.

## Templates wrapped (2.0.0 SDK block — where struct conditionals live)

- `equalityTemplate` → `{ $contains: { $eq: ... } }`
- `inFilterTemplate` → `{ $contains: { $in: [...] } }` (valid: element StringFilter/NumberFilter both include `$in` via `BaseFilterOptions`)
- `rangeTemplate` → `{ $contains: { $<operation>: ... } }`
- `stringStartsWithTemplate` → `{ $contains: { $startsWith: "foo" } }`
- `containsAllTermsTemplate`, `containsAnyTermTemplate`, `containsAllTermsInOrderTemplate` → wrapped existing term filter object in `$contains`

Each adds an `{{#arrayContains}}` branch that takes precedence over the struct/plain branches via an `{{^arrayContains}}` fallback wrapping the unchanged struct + plain branches.

## Left unchanged

- `containsTemplate` (simple array membership `{ $contains: value }`) — kept for membership as instructed
- `nullTemplate` — array null is an array-level filter, not a subtype filter
- Geo templates (`within*`, `intersects*`) — not part of the element-subtype filter set

## Dependency

Depends on `@osdk/docs-spec-sdk@0.19.0` (palantir/foundry-platform-typescript#314), which adds `arrayContains: "optional"` to `ObjectTypeWithPropertyTemplateStrings`. The catalog pin in `pnpm-workspace.yaml` is bumped to `0.19.0`. Since 0.19.0 is not yet published, the lockfile is intentionally not updated here — CI will regenerate it once the spec publishes.

## Verification

Verified locally against the published `@osdk/docs-spec-sdk@0.18.0` (which already contains `arrayContains: "optional"`): `pnpm run codegen` + `pnpm turbo typecheck transpile check-spelling --filter=@osdk/typescript-sdk-docs` all pass. The template body is an opaque string to the type system, so the change compiles regardless of spec version. Attempting a local `file:` link to the built spec failed because the built package carries a `workspace:*` dependency on `@osdk/docs-spec-core` that does not exist in the osdk-ts workspace — so verification used the published 0.18.0 instead.


packages/typescript-docs-example-generator/src/generateExamples.ts
```diff
@@ -319,7 +319,9 @@ function generateAllExamplesForVersion(
     }
 
     // Extract blocks using TemplateAnalyzer
-    const blocks = analysis.value.blocks;
+    const blocks = analysis.value.blocks.filter(
+      block => block.name.replace(/^[#^]/, "") !== "arrayContains",
```
ssanjay1, line 323: What's going on here

ssanjay1 (approved):
Stamping, contigent on us doing wider review on the docs generator package later

PR #3431, opened 2026-06-02
Refactor React (beta) template onto a shared `react.beta.common` package

beloiual:
## What & why
The authed React template and the upcoming React (Public) template share ~80% of their files (UI components, styling, configs, assets). Rather than duplicate all of that, this PR extracts the framework-agnostic, auth-agnostic files into a new shared package and teaches the template packager to merge it in at codegen time. No behavior change to the generated React app.

## Changes
- **New `@osdk/create-app.react.beta.common`** — carrier package holding the shared `templates/` (Home, Header, ErrorBoundary, Loading, NextSteps, Osdk, `main.tsx`, `index.css`, eslint/prettier/tsconfig, `.vscode`, `_gitignore`, public svgs, `vite-env.d.ts`). Exports nothing
at runtime.
- **`@osdk/create-app.template-packager`** — new `--shared <pkg>` flag: resolves a sibling package's `templates/` and merges it with the template's own `templates/` (the template's own files win on conflict).
- **`@osdk/create-app.template.react.beta`** — shared files removed (now sourced from common); `package.json` gains the `react.beta.common` dependency; `codegen` script becomes `create-app.template-packager --shared @osdk/create-app.react.beta.common`. The template keeps only
its auth-specific files (`client.ts.hbs`, `index.html.hbs`, `router.tsx.hbs`, `env.test.ts.hbs`, `vite.config.ts.hbs`, `AuthCallback.tsx`, `package.json.*`).
- `.monorepolint.config.mjs` — registers the common package under the internal-templates archetype.
- Regenerated the react example apps (trailing-whitespace only).

### 1. Process Files into a Map<String, Entry> instead of a List<String>
We want to make sure that templates that share a common core can de-dup and override common files. We do this by changing `result += `["${destPath}", ${output}],\n`;` to now to collect the sharedFiles and then overwrites it with any local files.

```typescript
const entries = new Map<string, Entry>();
  for (const pkgName of sharedPackages) {
    const sharedDir = await resolveSharedTemplatesDir(pkgName, sourcePackageJsonPath);
    await collectFiles(sharedDir, sharedDir, entries, sourcePackageJson);   // shared first
  } 
  await collectFiles(templatesDir, templatesDir, entries, sourcePackageJson); // local LAST
```

### 2. Add a --shared flag
We want to make this modular so we are adding a `--shared` flag that allows you specify which packages are a shared directory.

`parseSharedFlags` returns the list of shared package names (throwing if --shared is given with no value)

### 3. resolveSharedTemplatesDir to locate the shared template package
``` typescript
  async function resolveSharedTemplatesDir(pkgName, sourcePackageJsonPath) {
    let dir = path.dirname(sourcePackageJsonPath);
    while (true) {
      const candidate = path.join(dir, "node_modules", pkgName, "package.json");
      try { await fs.access(candidate); return path.join(path.dirname(candidate), "templates"); }
      catch { /* keep walking up */ }
      const parent = path.dirname(dir);
      if (parent === dir) throw new Error(`Could not resolve shared template package "${pkgName}" …`);
      dir = parent;
    }
  }
```
It walks up the node_modules chain from the consuming package to find node_modules/<pkgName>/package.json, then returns that package's templates/ dir. It deliberately reads node_modules directly rather than using Node's module resolution. This avoids needing the shared package to export ./package.json, and avoids depending on the shared package being built (codegen runs before builds).

#### Before - stripped only by key prefix:
```typescript
  for (const key of Object.keys(packageJson[d])) {
    if (key.startsWith("@osdk/monorepo.")) delete packageJson[d][key];
  }
```
####  After - also strips anything with a `workspace:` version:
```typescript
  for (const [key, value] of Object.entries(packageJson[d])) {
    if (key.startsWith("@osdk/monorepo.")
        || (typeof value === "string" && value.startsWith("workspace:"))) {
      delete packageJson[d][key];
    } 
```

The refactor adds `@osdk/create-app.react.beta.common: workspace:~ ` to the React template packages' deps. Those deps get merged into the generated end-user app. workspace: is a pnpm monorepo-internal protocol, meaning if it leaked into a real user's package.json, their pnpm install would fail. The generalized value-based filter removes any workspace:* dep (catching the new common package and any future one), while still removing the @osdk/monorepo.* keys and the hardcoded @osdk/create-app.template-packager delete just
above it.

## Why the file count looks large
Most entries are **pure file moves** (react.beta `templates/*` → `react.beta.common/templates/*`) that GitHub renders as renames with no content change.

## No changeset
All touched publishable packages (`react.beta`, `react.beta.common`, `template-packager`) are `private`; `@osdk/create-app`'s generated output is unchanged.

## Verification
- `example-generator` check: react examples regenerate identically
- `create-app` `check-bundle`, `mrl check`, transpile — all green

### Linked PRs
1. https://github.com/palantir/osdk-ts/pull/3430
2. **(this PR)**
3. https://github.com/palantir/osdk-ts/pull/3432


olliejenningsPT:
@beloiual I think we need add a `@osdk/create-app.react.beta.common#codegen` to the `dependsOn` for the codegen tasks for `packages/create-app.template.react.beta/turbo.json`. We might need a per-package override in monorepolint to avoid this impacting all turbo.json fiels

beloiual:
@olliejenningsPT I added additional rules for the templates that depend on the shared common to have an explicit transpile dependency. Additionally I added monorepolint rules to prevent this from applying to other packages

See: cbe36dcc25fc85859df58690783b793011c3ea96

olliejenningsPT (approved):

PR #3432, opened 2026-06-02
Add React (Public) template for authless public apps

beloiual:
## What & why
Adds **`@osdk/create-app.template.react-public.beta`**, a React template for Self-Service Public Apps that talk to Foundry via `createAuthlessClient` + a Vite dev proxy instead of an OAuth login flow. Selectable from the `create-app` framework prompt (2.x only). It reuses the
shared `react.beta.common` package from the previous PR, so it ships only its auth-specific files.

## CLI behavior (authless gating)
For templates flagged `authless`, the CLI (`cli.ts`):
- **skips** the Client ID and Scopes prompts
- **forces** the CORS/dev proxy on
- **restricts** the template to SDK **2.x**

## Public vs. authed
| File | Authed | Public |
|---|---|---|
| `src/client.ts` | `createPublicOauthClient(...)`, reads `osdk-clientId`/`osdk-redirectUrl` meta tags | `createAuthlessClient()`, reads `VITE_FOUNDRY_API_URL`/`VITE_FOUNDRY_ONTOLOGY_RID` |
| `index.html` | has `osdk-clientId`/`osdk-redirectUrl` meta tags | removed |
| `src/router.tsx` | `/auth/callback` route + `AuthCallback` | none |
| `vite.config.ts` | optional `corsProxy` → `foundryUrl` | always-on `/api-proxy` → `applicationUrl` (rewrites to `/proxy`) |
| `src/env.test.ts` | asserts client-id/redirect-url env | removed |

Everything else comes from `react.beta.common`, so it's identical to the authed template.

### 1. Build Generator
```bash
pnpm install
pnpm turbo build
```

### 2. Run Generator
```bash
cd /tmp
node <REPO-PATH>/packages/create-app/bin/createOsdkApp.mjs
```

### 3. Deploy and Test
```bash
cd /tmp/my-public-app
npm install
npm run dev
```

## Changes
- New `packages/create-app.template.react-public.beta/**` (template + `.hbs` files)
- `create-app`: `codegen.mjs` registers `react-public` (`authless: true`); `cli.ts`/`run.ts`/`generateEnv.ts`/`templates.ts` implement authless gating + env generation; `package.json` adds the template devDependency; unit tests updated
- `example-generator`: passes `clientId/scopes = undefined`, `corsProxy = true` for authless templates
- `monorepo.tool.transpile`: adds `react-public.beta` to `noExternalList` so it bundles into the `create-app` binary
- `cspell`: adds `authless`/`Authless`
- New generated example apps `example-react-public-sdk-2.x{,-no-osdk}` (required by the example-generator CI check)
- Changeset: `@osdk/create-app` **minor**

## Verification
- `example-generator` check (incl. the new react-public examples)
- `create-app` unit tests + `check-bundle`, `mrl check`, `pnpm install --frozen-lockfile` — all green

### Linked PRs
1. https://github.com/palantir/osdk-ts/pull/3430
2. https://github.com/palantir/osdk-ts/pull/3431
3. **this pr**)


packages/create-app.template.react-public.beta/templates/vite.config.ts.hbs
```diff
@@ -0,0 +1,35 @@
...
+    resolve: {
+      alias: {
+        "@": path.resolve(__dirname, "./src"),
+      },
+    },
+    server: {
+      port: 8080,
+      proxy: {
+        "^/.*api-proxy.*": {
+          target: "{{applicationUrl}}",
```
olliejenningsPT, line 18: if `applicationUrl` is not provided, this will break `npm run dev`, do you potentially want to require this of have some conditional around this? Or fine as is?

packages/create-app.template.react-public.beta/templates/package.json.osdk.hbs
```diff
@@ -0,0 +1,29 @@
...
+    "test": "vitest run",
+    "check": "eslint . --report-unused-disable-directives --max-warnings 0 && tsc --noEmit && vitest run",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "{{osdkPackage}}": "latest",
+    "@osdk/client": "{{clientVersion}}",
+    "@osdk/foundry": "latest",
+    "@osdk/oauth": "^1.10.0",
+    "@osdk/react": "^0.13.0",
```
olliejenningsPT: Do we want to pin the react version if the client version is not? Since the other templates that use react have it has `clientVersion`

packages/create-app.template.react-public.beta/templates/package.json.osdk.hbs
```diff
@@ -0,0 +1,29 @@
...
+    "lint:fix": "eslint --report-unused-disable-directives --max-warnings 0 --fix",
+    "test": "vitest run",
+    "check": "eslint . --report-unused-disable-directives --max-warnings 0 && tsc --noEmit && vitest run",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "{{osdkPackage}}": "latest",
+    "@osdk/client": "{{clientVersion}}",
+    "@osdk/foundry": "latest",
+    "@osdk/oauth": "^1.10.0",
```
olliejenningsPT, line 20: This version is `1.10.0` whilst the version in the `package.json.psdk.hbs` is `1.1.0`, is that expected?

packages/monorepo.cspell/cspell.config.js
olliejenningsPT: The `authless` already exists here, so might want to add to global list instead

packages/create-app.template.react-public.beta/templates/package.json.psdk.hbs
```diff
@@ -0,0 +1,27 @@
...
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "format": "prettier --write src/",
+    "lint": "eslint . --report-unused-disable-directives --max-warnings 0",
+    "lint:fix": "eslint --report-unused-disable-directives --max-warnings 0 --fix",
+    "test": "vitest run",
+    "check": "eslint . --report-unused-disable-directives --max-warnings 0 && tsc --noEmit && vitest run",
```
olliejenningsPT: I know we have `check` but the other templates also have `typecheck` as a script, do we want to keep `typecheck` to align with them?

packages/create-app.template.react-public.beta/templates/package.json.osdk.hbs
beloiual: Good catch, lets make it `"@osdk/react": "{{clientVersion}}",`

packages/create-app.template.react-public.beta/templates/vite.config.ts.hbs
beloiual: I think this is fine as is. When we are in developer console, we have the capability to bootstrap an application without knowing the domain. However, when we are actually developing, to access the authless endpoints, you need to have the `applicationUrl` as our endpoints are served on this route.

We want to make it clear that if we bootstrap this, we'll need to specify this application URL for the application to properly load data and I think the tradeoff of this failing to build seems reasonable for not fully set-up applications.
Screenshot 2026-08-19 at 1 33 57 PM

packages/create-app.template.react-public.beta/templates/package.json.osdk.hbs
beloiual: This is expected as the `createAuthlessClient` is only present in the `1.10.0` version and above

packages/create-app/src/cli.ts
```diff
@@ -160,17 +160,35 @@ export async function cli(args: string[] = process.argv): Promise<void> {
     ...parsed,
     template,
   });
+  const authless: boolean = template.authless ?? false;
+  if (authless && sdkVersion !== "2.x") {
```
Animesh-Saxena: Why not make a big if (authless) block and then have these if statements inside it

packages/create-app/src/cli.ts
```diff
@@ -160,17 +160,35 @@ export async function cli(args: string[] = process.argv): Promise<void> {
     ...parsed,
     template,
   });
+  const authless: boolean = template.authless ?? false;
```
Animesh-Saxena: nit: isAuthless

packages/create-app/src/generate/generateEnv.ts
```diff
@@ -14,32 +14,77 @@
...
   ontology: string | undefined;
-}): string {
-  const foundryApiUrl = corsProxy ? "http://localhost:8080" : foundryUrl;
-  const applicationUrl = "http://localhost:8080";
-  const ontologyEnvSection =
-    ontology != null
-      ? `
+}
+
+interface AuthlessDevelopmentArgs {
```
Animesh-Saxena, line 27: Nit: There is something weird in the typing system here. Hardcoding the authless as true/false is weird smell. You shouldnt need both descriptive name and a boolean

packages/create-app/src/generate/generateEnv.ts
```diff
@@ -14,32 +14,77 @@
...
+  args: AuthProductionArgs | AuthlessProductionArgs,
+): string {
+  return args.authless ? authlessEnvProduction(args) : authEnvProduction(args);
+}
+
+function ontologySection(
+  envPrefix: string,
+  ontology: string | undefined,
+): string {
+  if (ontology == null) {
```
Animesh-Saxena: Why did we change from a ternary to this? Im okay with it, just making sure that no logical change happened

packages/create-app/src/generate/generateEnv.ts
```diff
@@ -14,32 +14,77 @@
...
+}
+
+function authEnvDevelopment({
+  envPrefix,
+  foundryUrl,
+  clientId,
+  corsProxy,
+  ontology,
+}: AuthDevelopmentArgs): string {
+  const foundryApiUrl = corsProxy ? "http://localhost:8080" : foundryUrl;
```
Animesh-Saxena: It's weird that we're hardcoding http://localhost:8080 instead of making it a const somewhere

.changeset/create-app-react-public.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/create-app": minor
+---
+
+Add React (Public) template for authless, public-facing applications. The previous `--authless` flag has been removed; select the React (Public) template from the framework prompt instead.
```
Animesh-Saxena: Did the flag ever except on main?

packages/create-app/src/generate/generateEnv.ts
beloiual: Refactored this!

packages/create-app/src/generate/generateEnv.ts
beloiual: I reverted this back to the ternary since there was no logical change

Animesh-Saxena (approved):
LGTM

Animesh-Saxena (approved):

Animesh-Saxena (approved):

PR #3436, opened 2026-06-03
[Components] Separate empty string from no value filter

ziyunp:
FilterList now separates "No value" (null/undefined) from a literal empty string. null and undefined are grouped under a single "No value" option, while an empty string is its own option rendered as "(empty string)". Selecting "No value" is now represented by the exported `NO_VALUE` sentinel in `selectedValues` (mapping to `$isNull`), whereas a literal empty string maps to an equality filter.


packages/react-components/src/filter-list/base/inputs/EmptyStringLabel.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import classnames from "classnames";
+import React, { memo } from "react";
+// Shares the muted/italic styling used by NoValueLabel so the two special
```
KuberSethi, line 19: nit: formatting

packages/react-components/src/filter-list/base/inputs/OptionLabel.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { memo } from "react";
+import { isNoValue } from "../../utils/filterValues.js";
+import { EmptyStringLabel } from "./EmptyStringLabel.js";
+import { NoValueLabel } from "./NoValueLabel.js";
+
+/** Plain-text label for the screen-reader / aria-label equivalent of an option. */
```
KuberSethi, line 22: nice!

KuberSethi (approved):

PR #3438, opened 2026-06-03
Added FunctionConfig and Scope to @osdk/functions

oleh308:
### Overview
Added support for FunctionConfig and Scope

Scope is a new declarative configuration that allows the function to declare what it needs access to during execution

```tsx
  import { ScopeAuthorization, ScopeResources } from "@osdk/functions/experimental";
  import { Employee } from "ontology-sdk";

  // 1. Manual declaration — mix of OSDK type references and string aliases
    scope: {
      resources: { objects: [Employee, "objectA"], links: ["reportsTo"], queries: ["myQuery"] },
      authorization: { read: ["marking1"], write: ["marking2"] },
    },
  };
  export function withManualDeclaration(): void {}

  // 2. Defaults — populated from resources.json and functions.json at discovery time
    scope: {
      resources: ScopeResources.defaultScopeResources,
      authorization: ScopeAuthorization.defaultRead,
    },
  };
  export function withDefaults(): void {}

```


packages/functions/src/scope/types.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export namespace ScopeResources {
+  /**
+   * Sentinel value: discovery populates scope from `defaultResources` declared in functions.json.
+   */
+  export const defaultResources: unique symbol = Symbol(
+    "ScopeResources.defaultResources",
+  );
+
+  export function of(opts: ScopeResources = {}): ScopeResources {
```
tmisirpash: Since we have the `FunctionConfig` type to validate that the scope is declared correctly, I don't feel super strongly that we need these constructors. The `.of()` wrappers in Example 3 in the PR description don't feel like they improve the ergonomics in a meaningful way, since the correct typing should already be enforced by the `FunctionConfig`, no?

packages/functions/src/scope/types.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Or use these helpers:
+ *   `resources: ScopeResources.defaultScopeResources`  — use the repo's default scope from resources.json
+ *   `resources: ScopeResources.of({ objects: ["myAlias"] })` — equivalent to a plain object literal
+ */
+export namespace ScopeResources {
+  /**
+   * Sentinel value: discovery populates scope from `defaultResources` declared in functions.json.
+   */
+  export const defaultResources: unique symbol = Symbol(
+    "ScopeResources.defaultResources",
```
tmisirpash, line 51: nit: `ScopeResources.defaultResources` feels a bit repetitive. Maybe `ScopeResources.default`?

oleh308: was trying to avoid using anything that is called "default", to avoid potential usage of reserved keyword

packages/functions/src/scope/types.ts
oleh308: can do!

packages/functions/src/scope/types.ts
```diff
@@ -0,0 +1,103 @@
...
+ * - A string alias from resources.json (e.g. `"myObject"`)
+ * - An OSDK type reference imported from your generated ontology SDK (e.g. `Employee`)
+ *
+ * For `links`, `queries`, and `models`, only string aliases are supported.
+ */
+export interface ScopeResources {
+  objects?: Array<string | ObjectTypeDefinition>;
+  interfaces?: Array<string | InterfaceDefinition>;
+  links?: string[];
+  queries?: string[];
```
tmisirpash: Should we support `QueryDefinition` here?

packages/functions/src/public/experimental.ts
```diff
@@ -14,4 +14,10 @@
  * limitations under the License.
  */
 
+export { ScopeAuthorization, ScopeResources } from "../scope/types.js";
```
ssanjay1, line 17: do we want functionconfig in experimental as well?

ssanjay1: either that or, if users don't need this, should we throw in internal?

packages/functions/src/scope/types.ts
oleh308: done

packages/functions/src/public/experimental.ts
oleh308: discussed in slack, FunctionConfig is a pre-existing concept but it was never reflected in this package, now it is

tmisirpash (dismissed):

ssanjay1 (approved):

PR #3441, opened 2026-06-04
Support interface instance invalidation

nfdiop:
## Summary
- allow observable cache invalidation to accept interface instances
- invalidate the backing concrete object for interface-origin instances
- add coverage for interface-origin invalidation

## Test plan
- pnpm --dir=packages/client vitest run src/observable/internal/Store.invalidation.test.ts
- pnpm turbo typecheck --filter=@osdk/client
- pnpm turbo check-api --filter=@osdk/client

### DIFF
--- a/packages/client/src/observable/internal/Store.invalidation.test.ts
+++ b/packages/client/src/observable/internal/Store.invalidation.test.ts
@@ -681,9 +681,11 @@ describe("Store Invalidation Type Isolation", () => {
 
       await cache.invalidateObjects(emp1AsFoo);
 
-      await new Promise(resolve => setTimeout(resolve, 500));
-
-      expect(empSubFn.next).toHaveBeenCalled();
+      await vi.waitFor(() => {
+        expect(empSubFn.next).toHaveBeenCalled();
+      });
+      // The two objects are invalidated in the same batch, so once emp1's
+      // subscription has fired the decision for emp2 has already been made.
       expect(emp2SubFn.next).not.toHaveBeenCalled();
     });
   });
--- a/packages/client/src/observable/internal/Store.invalidation.test.ts
+++ b/packages/client/src/observable/internal/Store.invalidation.test.ts
@@ -14,7 +14,12 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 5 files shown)


packages/client/src/observable/internal/Store.invalidation.test.ts
```diff
@@ -646,6 +651,41 @@ describe("Store Invalidation Type Isolation", () => {
...
+        type: Employee,
+        primaryKey: EMPLOYEE_2_ID,
+      });
+
+      empSubFn.next.mockClear();
+      emp2SubFn.next.mockClear();
+
+      // An interface-origin instance is one concrete object viewed through the
+      // interface lens; it carries the backing object's $objectType/$primaryKey.
+      const { data: foos } = await client(FooInterface).fetchPage();
```
nfdiop, line 672: Explicitly fetching the interface to make the example prove the point about interfaces

packages/client/src/observable/internal/Store.invalidation.test.ts
```diff
@@ -646,6 +651,41 @@ describe("Store Invalidation Type Isolation", () => {
...
+        && foo.$primaryKey === EMPLOYEE_1_ID
+      );
+      invariant(emp1AsFoo);
+      expect(emp1AsFoo.$apiName).toBe(FooInterface.apiName);
+      expect(emp1AsFoo.$objectType).toBe(Employee.apiName);
+      expect(emp1AsFoo.$primaryKey).toBe(EMPLOYEE_1_ID);
+
+      await cache.invalidateObjects(emp1AsFoo);
+
+      await new Promise(resolve => setTimeout(resolve, 500));
```
ziyunp: Issue (non-blocking): Don't think we should add timeout in a test, is there a better way to await this? Maybe something similar to act()?

ziyunp (dismissed):

packages/client/src/observable/internal/Store.invalidation.test.ts
nfdiop: chnaged

ziyunp (approved):

PR #3442, opened 2026-06-04
fix(react-components): switch xlsx to xlsx-republish to address CVE

olliejenningsPT:
## Summary

The maintainer-published `xlsx` on npm is pinned at `0.18.5`, which has a known prototype-pollution CVE. SheetJS has stopped publishing fixed releases to the npm registry, distributing them only via their own CDN, so npm consumers cannot get a patched version through the registry.

This PR replaces the `xlsx` dependency with [`xlsx-republish`](https://www.npmjs.com/package/xlsx-republish), a well-regarded community republish of the latest SheetJS CDN release (`0.20.3`) back to npm. This gives downstream consumers (e.g. `foundry-sdk-asset-bundle`) a patched version through normal registry channels, without needing to pull tarballs from the SheetJS CDN.

## Changes

- `@osdk/react-components`: `xlsx@^0.18.5` → `xlsx-republish@^0.20.3`
- `@osdk/react-components-storybook`: same swap (storybook fixtures synthesize test `.xlsx` files via `xlsx`)
- Updated three import sites (one in `parseSpreadsheet.ts`, two in stories)
- Regenerated `pnpm-lock.yaml`

API usage is unchanged. `xlsx-republish` re-exports the same `read` / `utils` / `write` surface.

### DIFF
--- a/packages/react-components/package.json
+++ b/packages/react-components/package.json
@@ -195,7 +195,7 @@
     "react-markdown": "^9.0.3",
     "remark-gfm": "^4.0.1",
     "utif": "^3.1.0",
-    "xlsx": "^0.18.5"
+    "xlsx-republish": "0.20.3"
   },
   "peerDependencies": {
     "@osdk/aip-core": "workspace:^",
--- a/.changeset/fix-xlsx-cdn-replacement.md
+++ b/.changeset/fix-xlsx-cdn-replacement.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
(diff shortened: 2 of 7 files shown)


packages/react-components/package.json
```diff
@@ -195,9 +195,10 @@
     "react-markdown": "^9.0.3",
     "remark-gfm": "^4.0.1",
     "utif": "^3.1.0",
-    "xlsx": "^0.18.5"
+    "xlsx-republish": "^0.20.3"
```
nfdiop: Could we explicitly pin to avoid any surprises in the future?

olliejenningsPT: The lockfile already pins it via integrity hash, so we wouldn't get a surprised unless someone did an `pnpm update`, but will change

nfdiop (approved):

PR #3445, opened 2026-06-04
Fix listogram bar border-radius and background token

Leolide:
## Summary
- Switch listogram bar fill from `scaleX()` to `width` percentage so border-radius renders uniformly on both sides. The `scaleX()` approach distorts the border-radius — on short bars the left corners look squished/pointy while the right side stays round. With `width`, both sides render identically regardless of fill level.
- Derive bar border-radius from the surface token at half size (`calc(var(--osdk-surface-border-radius) * 0.5)`) for a subtler look that scales with theming
- Replace non-existent `--osdk-palette-gray-100` background token with theme-aware `--osdk-background-tertiary`, fixing the bar track across light, dark, modern, and devcon themes

Before:
Screenshot 2026-06-04 at 12 07 37 PM

After
Screenshot 2026-06-04 at 12 05 59 PM

## Test plan
- [ ] Open Storybook FilterList stories, verify listogram bars have uniform rounded corners on both ends at all fill levels
- [ ] Switch between light, dark, modern, and devcon themes — bar track background should adapt correctly

### DIFF
--- a/.changeset/listogram-bar-styling.md
+++ b/.changeset/listogram-bar-styling.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
+Fix listogram bar border-radius distortion and use theme-aware background token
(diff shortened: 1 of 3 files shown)


nfdiop (approved):
looks great!

PR #3447, opened 2026-06-04
[Components] Alternative download

ziyunp:
Update getSnapshot() implementation to fetch data separately from table rendering
- if total row count exceeds row limit, return promise rejection.
- capping concurrent function-column requests during snapshot collection


packages/react-components-storybook/src/stories/ObjectTable/Recipes/WithLoadedDataDownload.stories.tsx
nfdiop, line 1: why delete the story?

packages/react-components/docs/ObjectTable.md
```diff
@@ -1290,6 +1291,63 @@ No additional configuration needed - these states are built-in!
 
 The ObjectTable automatically implements infinite scroll pagination, with page size of 50. As users scroll down, more data is loaded seamlessly. No configuration required!
 
+## Exporting Data
+
+Pass a `tableRef` to obtain an `ObjectTableHandle<Q, RDPs>`. Its `getSnapshot()` method loads **all** matching rows (up to `rowLimit`, default `10_000`) and returns a format-agnostic snapshot of the table's columns and rows, so you can export to CSV, Excel, JSON, the clipboard, or anywhere else.
```
nfdiop: Hm, I would prefer we default to only the visible rows. Else let the user pass the limit vs us passing the limit. If we set a default limit of 10,000 the browser can crash until the user realises they need to decrease the limit. Let me know what you think?

packages/react-components/docs/ObjectTable.md
```diff
@@ -1290,6 +1291,63 @@ No additional configuration needed - these states are built-in!
...
+
+The snapshot reflects the table's current column visibility, ordering, and pinning. Property, derived-property, and function-backed columns are all included. Custom-rendered columns have no underlying value and are omitted. Each row exposes a `getValue(columnId)` accessor; cells are the raw value, or the literal string `"Error"` if a function-backed cell failed to load (the promise still resolves with the rest of the snapshot).
```
nfdiop: Custom-rendered columns have no underlying value and are omitted": could we also fix in this PR? The idea is to ask users to pass the string version they want to use instead

packages/react-components/src/public/experimental/__tests__/object-table.test.tsx
ziyunp, line 1: Don't think we need tests in the public folder at all

packages/react-components/src/object-table/__tests__/ObjectTable.test.tsx
nfdiop, line 1: could we keep tests?

packages/react-components/src/object-table/hooks/__tests__/useFunctionColumnsData.test.ts
```diff
@@ -946,7 +919,6 @@ describe("useFunctionColumnsData", () => {
           objectOrInterfaceType: TestObjectType,
           objects,
           columnDefinitions,
-          primaryKeyApiName: PRIMARY_KEY_API_NAME,
```
nfdiop, line 949: curious why make changes to this file?

packages/react-components/src/object-table/hooks/__tests__/useObjectTableData.test.tsx
```diff
@@ -495,7 +495,6 @@ describe(useObjectTableData, () => {
       objectOrInterfaceType: TestObjectType,
       objects: mockBaseData,
       columnDefinitions,
-      primaryKeyApiName: undefined,
```
nfdiop, line 498: same question here

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
ziyunp, line 1: To do: Add a test

packages/react-components/src/object-table/hooks/useFunctionColumnsData.ts
```diff
@@ -87,50 +81,42 @@ export function useFunctionColumnsData<
...
-  // + filter with primary keys of each page's objects
+  // Construct an object set per page (base set filtered to each page's
+  // primary keys). `buildPagedObjectSets` returns `[]` for interfaces and
+  // when there's no work to do, which short-circuits the downstream queries.
   //
-  // When a new page loads, only that page's queries fire — old pages
-  // hit the dedupeIntervalMs cache since their params are unchanged.
+  // When a new page loads, only that page's queries fire — old pages hit the
+  // dedupeIntervalMs cache since their params are unchanged.
   const pagedObjectSets = useMemo(() => {
```
nfdiop, line 103: Curious why the change here?

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,240 @@
...
+  objectOrInterfaceType: Q;
+  table: Table<Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>>;
+  columnDefinitions?: Array<ColumnDefinition<Q, RDPs, FunctionColumns>>;
+  /**
+   * The object set the table resolved to (filter / ordering / unions applied,
+   * derived properties included). Rows are loaded by paginating this set.
+   */
+  objectSet: ObjectSet<Q, RDPs> | undefined;
+  /** Page size for chunking function-column queries across loaded rows. */
+  pageSize?: number;
```
nfdiop, line 66: nit: should we call it then functionsPageSize

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,240 @@
...
+  /**
+   * Total number of objects in the object set, as reported by the
+   * underlying list payload. Used to fail fast when `getSnapshot` is called
+   * with a `rowLimit` lower than the row count.
+   */
+  totalCount: string | undefined;
+
+  orderBy?: OrderBy<Q>;
```
nfdiop: nit: should we mark this and the other optional props here as "T | undefined" so the dev doesn't forget?

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,240 @@
...
+        Osdk.Instance<
+          Q,
+          "$allBaseProperties",
+          PropertyKeys<Q>,
+          RDPs
+        >
+      > = [];
+      if (objectSet != null && columnIds.length > 0 && rowLimit > 0) {
+        for await (
+          const object of objectSet.asyncIter({
```
nfdiop, line 162: should this break after we loaded enough rows?

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,241 @@
...
+      ) {
+        const pages = buildPagedObjectSets<Q, RDPs>(
+          client,
+          objectOrInterfaceType,
+          loadedObjects,
+          pageSize,
+        );
+
+        if (pages.length > 0) {
+          functionColumnValues = await fetchFunctionColumnValues<
```
nfdiop, line 203: Should we also respect max concurrent requests here?

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,241 @@
...
+      const rowLimit = options?.rowLimit ?? DEFAULT_SNAPSHOT_ROW_LIMIT;
+
+      // Fail fast when the caller's row cap is smaller than the matching set.
+      // `totalCount` is only available once the list payload has resolved; we
+      // skip the check otherwise rather than guess.
+      if (totalCount != null) {
+        const total = Number(totalCount);
+        if (Number.isFinite(total) && total > rowLimit) {
+          return Promise.reject(
+            `getSnapshot error: total row count exceeds row limit. `,
```
nfdiop: I don't think we should reject if we fetched more than what we want. worse case we can cap the results we return to the user

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,241 @@
...
+          .map((definition) => String(definition.locator.id)) ?? [],
+      ),
+    [columnDefinitions],
+  );
+
+  const getSnapshot = useCallback(
+    async (
+      options?: ObjectTableSnapshotOptions,
+    ): Promise<ObjectTableSnapshot<Q, RDPs>> => {
+      const rowLimit = options?.rowLimit ?? DEFAULT_SNAPSHOT_ROW_LIMIT;
```
nfdiop, line 121: Let's normalize the row limit like before:
```
function normalizeRowLimit(rowLimit: number | undefined): number | undefined {
  if (rowLimit == null || rowLimit <= 0 || !Number.isFinite(rowLimit)) {
    return undefined;
  }
  return rowLimit;
}
```

packages/react-components/src/object-table/utils/objectTableSnapshot.ts
nfdiop, line 1: should it be split into two different files instead?

nfdiop (commented):
Thanks for making the change

packages/react-components/docs/ObjectTable.md
ziyunp: No, I don't agree here. When you download a table of data, it should download the full dataset. Otherwise, for 2 users seeing the same table, one scrolled to the 3rd page and one stayinh on the first page, the data they download are different.

packages/react-components/docs/ObjectTable.md
ziyunp: Here the custom-rendered columns is re-worded to explicitly say it's for locator.type === "custom". Columns with custom renderCell will have value

packages/react-components/src/object-table/__tests__/ObjectTable.test.tsx
ziyunp: This test content is testing the getSnapshot function and the behaviour has completely changed. I added tests for useObejctTableSnapshot

packages/react-components/src/object-table/hooks/__tests__/useFunctionColumnsData.test.ts
ziyunp: Just removing the unneeded param after this PR https://github.com/palantir/osdk-ts/pull/3408

packages/react-components/src/object-table/hooks/useFunctionColumnsData.ts
ziyunp: To extract the common utils for use in fetching function columns

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
ziyunp: No, we reject early when rows exceed row limit

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
ziyunp: I don't think we should fetch data partially. In user's workflow, they would block user from executing download when row count is more than limit

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
ziyunp: I think rowLimit of 0 and infinity are acceptable values. If someone inputs less than 0, we should still reject than setting the default value

packages/react-components-storybook/src/stories/ObjectTable/Recipes/WithLoadedDataDownload.stories.tsx
ziyunp: Putting back

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
ziyunp: This is internal API. It's set via the top-level pageSize that applies to the full table

nfdiop: got it. I was confused because the comment mentioned functions here

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
ziyunp: Added another rejection guard when loadedObjects > rowLimit

packages/react-components/src/object-table/hooks/useObjectTableSnapshot.ts
```diff
@@ -0,0 +1,247 @@
...
+      const rowLimitExceededError =
+        `Error in getSnapshot: total row count exceeds row limit of ${rowLimit}.`;
```
nfdiop, line 123: nit: also add totalRowCount

nfdiop (approved):

PR #3448, opened 2026-06-04
[custom widgets] bump min versions in templates

amishshah:
Bump the custom widget versions to enforce using dev mode v2

### DIFF
--- a/packages/create-widget.template.react.v2/templates/package.json.hbs
+++ b/packages/create-widget.template.react.v2/templates/package.json.hbs
@@ -14,10 +14,10 @@
     "{{osdkPackage}}": "latest",
     "@osdk/client": "^2.0.0",
     "@osdk/react": "^0.13.0",
-    "@osdk/widget.client-react": "^3.3.0",
-    "@osdk/widget.client": "^3.3.0"
+    "@osdk/widget.client-react": "^3.26.0",
+    "@osdk/widget.client": "^3.26.0"
   },
   "devDependencies": {
-    "@osdk/widget.vite-plugin": "^3.3.0"
+    "@osdk/widget.vite-plugin": "^3.26.0"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/create-widget.template.react.v2/templates/package.json.hbs
```diff
@@ -14,10 +14,10 @@
...
     "@osdk/client": "^2.0.0",
     "@osdk/react": "^0.13.0",
-    "@osdk/widget.client-react": "^3.3.0",
-    "@osdk/widget.client": "^3.3.0"
+    "@osdk/widget.client-react": "^3.26.0",
+    "@osdk/widget.client": "^3.26.0"
   },
   "devDependencies": {
-    "@osdk/widget.vite-plugin": "^3.3.0"
+    "@osdk/widget.vite-plugin": "^3.26.0"
```
tzyl, line 21: Can you do this in both widget templates?

packages/create-widget.template.react.v2/templates/package.json.hbs
```diff
@@ -14,10 +14,10 @@
     "{{osdkPackage}}": "latest",
     "@osdk/client": "^2.0.0",
     "@osdk/react": "^0.13.0",
```
tzyl, line 16: Also noticed this has started releasing at 2.x so we may want to check in with them on if they're tracking updating the templates with any breaks
https://www.npmjs.com/package/@osdk/react?activeTab=versions

packages/create-widget.template.react.v2/templates/package.json.hbs
tzyl: Ah sorry you did it in both the templates but I saw the other diff as the examples package but that doesn't get regenerated because it gets swapped to workspace

tzyl (approved):

PR #3449, opened 2026-06-04
[Components] Expose focusedRow and onFocusedRowChange

ziyunp:
Expose focusedRow and onFocusedRowChange to allow user to persist that in their local state. Use case is to keep the last focused row after navigating away and back to the table

Screenshot 2026-06-04 at 15 56 01


packages/react-components/src/object-table/hooks/useFocusedRow.ts
```diff
@@ -0,0 +1,88 @@
...
+  onFocusedRowChanged?: (row: TData | null) => void;
+  /**
+   * Returns a stable id for a row
+   */
+  getRowId: (row: TData) => string;
+}
+
+export interface UseFocusedRowResult<TData> {
+  /** Effective focused row — controlled value or internal state. */
+  focusedRow: TData | null;
```
nfdiop: Curious if we could store some sort of id instead? I had many cases where I was storing the entire object then the underlying data changes and you are storing a stale value. We could have something like 
```
focusedRowId
focusedRow = function(data, rowId) -> row 
```

packages/react-components/src/object-table/hooks/useFocusedRow.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export function useFocusedRow<TData>({
+  focusedRow,
+  onFocusedRowChanged,
+  getRowId,
+}: UseFocusedRowProps<TData>): UseFocusedRowResult<TData> {
+  // Explicit check for undefined instead of != null
+  // because null is a valid value to clear a focused row
+  const isControlled = focusedRow !== undefined;
```
nfdiop: nice comment!

packages/react-components/src/object-table/TableEditContainer.tsx
```diff
@@ -25,14 +25,14 @@ interface TableEditContainerProps<
   TData extends RowData,
 > {
   editableConfig: EditableConfig<TData, unknown>;
-  focusedRowId?: string | null;
+  hasFocusedRow?: boolean;
```
nfdiop, line 28: I like this change!

packages/react-components/src/object-table/hooks/useFocusedRow.ts
ziyunp: I was using focusedRowId initially. Let me update it so that internally we store the id, but in the callback, we pass the full row data

packages/react-components/src/object-table/Table.tsx
```diff
@@ -219,7 +251,7 @@ function BaseTableInner<
     return () => {
       document.removeEventListener("pointerdown", handleClickOutside);
     };
-  }, [portalTracker]);
+  }, [portalTracker, isFocusControlled, setFocusedRowId]);
```
nfdiop: I think you also want to use isFocusControlled in the effect

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -594,6 +594,30 @@ export interface ObjectTableProps<
     object: Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>,
   ) => void;
 
+  /**
+   * The row to render as visually focused (the "last interacted" row).
+   * When provided, focus state is controlled by the caller.
+   *
+   * Pass `null` to render no row as focused.
+   */
+  focusedRow?:
```
nfdiop: should we also change to PrimaryKeyType<Q> like selectedRow?

packages/react-components/src/object-table/Table.tsx
```diff
@@ -132,12 +142,34 @@ function BaseTableInner<
...
   const [isLoadingMore, setIsLoadingMore] = useState(false);
-  const [focusedRowId, setFocusedRowId] = useState<string | null>(null);
+
+  const tableGetRowId = table.options.getRowId;
+  const getRowIdForFocus = useCallback(
+    (row: TData) => tableGetRowId?.(row, 0, undefined) ?? "",
+    [tableGetRowId],
+  );
+  const getRowById = useCallback(
+    (id: string) => table.getRow(id, true)?.original ?? null,
```
nfdiop: Could we try...catch. This will throw when the row is not found https://github.com/TanStack/table/blob/f4dc742b7b8bf01bb7dd10ee7d2f238400befcc0/packages/table-core/src/core/table.ts#L405

packages/react-components/src/object-table/ObjectTableApi.ts
ziyunp: Agree, will keep it consistent

packages/react-components/src/object-table/Table.tsx
ziyunp: Removed it from the dep array

nfdiop (approved):

PR #3451, opened 2026-06-04
@osdk/language-models: accept a Client in the helper functions

nihalbhatnagar:
The `@osdk/language-models` helpers (`createFetch`, `getFoundryToken`, `getOpenAiBaseUrl`, `getAnthropicBaseUrl`, `getGoogleBaseUrl`) now accept an OSDK `Client` in addition to a `PlatformClient`. The underlying client context is resolved internally, so callers no longer need to extract it by hand.

### DIFF
--- a/packages/language-models/src/utils.ts
+++ b/packages/language-models/src/utils.ts
@@ -14,7 +14,19 @@
  * limitations under the License.
  */
 
-import type { PlatformClient } from "@osdk/client";
+import type { Client, PlatformClient } from "@osdk/client";
+import { symbolClientContext as legacySymbolClientContext } from "@osdk/shared.client";
+import { symbolClientContext } from "@osdk/shared.client2";
+
+function resolveContext(client: Client | PlatformClient): PlatformClient {
+  if (symbolClientContext in client) {
+    return client[symbolClientContext] as PlatformClient;
+  }
+  if (legacySymbolClientContext in client) {
+    return client[legacySymbolClientContext] as PlatformClient;
+  }
+  return client as PlatformClient;
+}
 
 /**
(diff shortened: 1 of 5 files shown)


packages/language-models/src/utils.ts
```diff
@@ -39,9 +51,9 @@ import type { PlatformClient } from "@osdk/client";
  * ```
  */
 export function createFetch(
-  client: PlatformClient,
+  client: Client | PlatformClient,
 ): typeof globalThis.fetch {
-  return client.fetch;
+  return resolveContext(client).fetch;
```
ssanjay1, line 56: I think I'm missing something dumb, but im confused why we need to do this resolution here?

ssanjay1: Oh because you're taking client explicitly

ssanjay1: can't this just be SharedClientContext?

nihalbhatnagar: SharedClientContext is nested in `Client` behind the symbol

nihalbhatnagar: client.baseUrl would be undefined

ssanjay1: Yea sorry I was mixing things up in my head this makes sense, I guess only thing is that the resolution logic is pretty much duplicated from our fetch helper https://github.com/palantir/foundry-platform-typescript/blob/77068c0b65a8e5d3d26bec4fe8bc1d9400a111ba/packages/shared.net.platformapi/src/foundryPlatformFetch.ts#L78.

Nit refactor could be to export some sort of helper from shared.net to use in both places, but this is fine for now

ssanjay1 (approved):

PR #3453, opened 2026-06-05
Stop shipping node-only test infra in browser builds

olliejenningsPT:
## Why

`@osdk/api` and `@osdk/client` are universal packages (browser + node). Each emits a `build/browser/` artifact that should contain zero `node:*` imports. Today, two source files leak Node imports into the browser build:

- `packages/api/src/__quickinfo_snapshot__/probeUtils.ts` imports `node:child_process` and `node:path`. It's part of the quickinfo snapshot test harness; nothing public references it.
- `packages/client/src/tsserver.ts` imports `node:events`, `node:fs/promises`, and `node:path`. It's a TypeScript-language-server wrapper used by `intellisense.test.ts` in both `@osdk/client` and `@osdk/react`. It was reachable from outside the client package only through the `./internal-node` subpath export, which existed solely so the two test suites could share the harness.

Neither file is real runtime code. Both shipped to the browser tarball because the transpile tool walks `src/` recursively and emits everything that isn't a test file, and `isTestFile` only matches `.test.*` / `testUtils.*` segments.

Concretely, the cost was:
- `node:*` imports in `build/browser/*.js` for both packages. That's a sanity-check failure if anyone audits the browser bundle, and a real runtime failure if a consumer imported `@osdk/client/internal-node` from a browser context.
- ~4–5 kB per file (plus sourcemaps) of cosmetic bloat in the published tarballs.
- A published subpath (`./internal-node`) whose only purpose was internal test sharing, with no release tag and an "internal-" naming convention as its only signal.

### `@osdk/api`: use the existing `testUtils.` exclusion rule

The `__quickinfo_snapshot__/` directory is pure test infrastructure (a harness, typed probes, snapshot artifacts, and the test file itself). Renamed the harness files under the existing `testUtils.` prefix that `monorepo.tool.transpile` already treats as test-only:

| Before | After |
| --- | --- |
| `__quickinfo_snapshot__/probeUtils.ts` | `__quickinfo_snapshot__/testUtils.probe.ts` |
| `__quickinfo_snapshot__/probes/` | `__quickinfo_snapshot__/testUtils.probes/` |

No tool changes. The existing `s.startsWith("testUtils.")` segment match in `isTestFile` does the work. The harness now stays out of every build target (browser, esm, cjs, types). Snapshots regenerated to update their `// Source:` headers; README updated.

### `@osdk/client`: extract the harness to its own workspace package

Created a new private workspace package `@osdk/shared.test.intellisense` matching the existing `@osdk/shared.test` shape (private, ESM-only, node-target, internal-libraries archetype). Moved `tsserver.ts` into it and exposed `startTsServer` / `TsServer` / the event type-guards as its public surface.

Removed from `@osdk/client`:
- `src/tsserver.ts`
- `src/public/internal-node.ts` (the re-export)
- `internal-node.d.ts` (the legacy root-level stub for old-style module resolution)
- The `./internal-node` entry in the `exports` map (monorepolint auto-removes when `src/public/internal-node.ts` is gone)

Rewired both intellisense tests to import from the new package:
- `packages/client/src/intellisense.test.ts`
- `packages/react/src/intellisense.test.ts`

Both packages gained `"@osdk/shared.test.intellisense": "workspace:~"` as a `devDependency`.

## Why a new package instead of a filename/directory convention

A `.node.ts` suffix or a `node-src/` directory would have worked, but each requires inventing a new convention in `monorepo.tool.transpile` *and* in `.monorepolint.config.mjs`'s `ourExportsConvention`, and would leave `@osdk/client` publishing a node-only subpath whose only consumer was in-repo test code. That's a design smell the suffix/directory approaches hide.

Extracting answers the question "why is node-only test infrastructure in a published runtime SDK's exports?" by removing the premise. The relationship between the harness and the packages it tests becomes an explicit `devDependency` instead of a re-export. Zero new tooling, mirrors the existing `@osdk/shared.test` pattern.

## Build / CI impact

Roughly neutral, slightly faster on cold builds:

- `@osdk/client` loses ~8 transpile-passes worth of work (two files × four targets) and the api-extractor run on `internal-node` types.
- `@osdk/shared.test.intellisense` adds ~2 transpile-passes (one file × two targets: esm and types, no browser, no cjs).
- New package is a sibling of `@osdk/client` in the dependency graph, not a parent. Both depend on `@osdk/api`, neither blocks the other. Builds in parallel.
- One-time Turbo cache invalidation on landing for downstream packages. Subsequent runs cache normally.
- Intellisense tests themselves are unchanged. Same subprocess overhead, same wall-clock.
- Side benefit: future edits to `tsserver.ts` invalidate only the small harness package's cache, not `@osdk/client`'s four-target build plus api-extractor.

## Verification

- `pnpm turbo transpile`: 245 tasks, all green.
- `pnpm turbo typecheck test check-api lint --filter=@osdk/api --filter=@osdk/client --filter=@osdk/react --filter=@osdk/shared.test.intellisense`: all green; pre-existing lint warnings unchanged.
- `grep -rE 'node:' packages/{api,client}/build/browser --include='*.js'`: no matches.
- `npm pack --dry-run` on `@osdk/api` and `@osdk/client`: no `probeUtils`, no `probes/`, no `tsserver`, no `internal-node` entries.
- `pnpm mrl check`: clean.

## Versioning

Changeset at `.changeset/exclude-node-only-files-from-browser-build.md`:
- `@osdk/api`: **patch** (cosmetic; harness files were never public, never in the exports map).
- `@osdk/client`: **minor** (the `./internal-node` subpath export is removed; the `internal-` naming and lack of release tag suggest no external consumer was intended, but it was technically a published surface).
- `@osdk/react`: **patch** (one test file's import switched).

The new `@osdk/shared.test.intellisense` is `"private": true` so changesets skip it.


packages/api/src/__quickinfo_snapshot__/testUtils.probe.ts
```diff
@@ -161,7 +161,7 @@ function formatSnapshot(
   if (names.length === 0) return "";
   const header = [
     `// AUTO-GENERATED by quickInfoTypes.test.ts — do not edit by hand.`,
-    `// Source: probes/${probesFileName}`,
+    `// Source: testUtils.probes/${probesFileName}`,
```
ziyunp, line 164: Caught up in person - renamed as monorepolint will exclude testUtils

ziyunp (approved):

PR #3454, opened 2026-06-07
Ontology as Code | Allow "Exotic Strings" To Be Primary Keys on Object Creation

acaulkin:
### Summary

Simply allows the creation process to go through when a specific non-primitive type is set as a primary key, what I refer to here as the "exotic string", denoted by the `PropertyTypeTypeString` type in `PropertyTypeType.ts`. This behavior _is_ presently supported in OMA, so the idea here is to just get to parity with it. 

### Potential Improvements

We could make a generalized method to check if a given `PropertyTypeType` is a given `PropertyTypeType`'s type (E.g., combine the methods `isExoticString()` and `isStruct()` into a single method and make it more generalized), but `isStruct()` is used in a ton of spots and didn't want to change a bunch of logic for this FR - LMK if I am off base here and would be happy to implement the method! 😊

### Prior Behavior

1. User creates a new object in OAC with an exotic string as the primary key and pre-configured analyzers + render hints set on the exotic string
2. User attempts to build
3. User receives error `Primary key property <property-name-here> can only be primitive types` 
4. Sadness ensues 😭

### Intended Behavior after Change

1. User creates a new object in OAC with an exotic string as the primary key and pre-configured analyzers + render hints set on the exotic string
2. User attempts to build
3. User is able to create a new object with pre-configured analyzers and render hints
4. User is happy 😊

### DIFF
--- a/.changeset/shaky-lamps-retire.md
+++ b/.changeset/shaky-lamps-retire.md
@@ -1,0 +1,5 @@
+---
+"@osdk/maker": minor
+---
+
+Allow creation of object types with exotic strings to allow for analyzers and render hints to be set at time of creation from OAC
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/osdk-ts](https://github.com/palantir/osdk-ts), @acaulkin! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #3455, opened 2026-06-08
Mikhailprice/pdf viewer download filename

mikhailprice:
---
  PR 1 — mikhailprice/pdf-viewer-download-filename → main

  ## Fix PDF viewer toolbar download saving files as "[object Object]"

  ### Problem
  The PDF viewer's toolbar **Download** button wired its `onClick` directly to the
  viewer's `download(filename?)` handler:

  ```tsx
  <Button onClick={onDownload} … />   // onDownload === viewer.download

  React therefore passed the click event as the filename argument. The
  filename resolver returns its argument verbatim when non-null:

  function resolveDownloadFilename(src, filename) {
    if (filename != null) return filename;          // ← the event object
    if (typeof src === "string") return src.split("/").pop()?.split("?")[0] || "document.pdf";
    return "document.pdf";
  }
```
  So a.download was set to the event object and every download saved as
  [object Object] — for any consumer of the built-in toolbar.

  Fix

  Wrap the toolbar handler so the event is not forwarded; the filename is resolved
  from src as intended.
```
  const handleDownload = useCallback(() => {
    viewer.download();
  }, [viewer]);
  // onDownload={handleDownload}
```
  Testing

  - vitest run src/pdf-viewer → 159/159 pass
  - eslint + dprint clean
  - Changeset: @osdk/react-components patch

  ---

### DIFF
--- a/.changeset/pdf-viewer-download-event-fix.md
+++ b/.changeset/pdf-viewer-download-event-fix.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
+Fix PDF viewer toolbar download saving an invalid filename.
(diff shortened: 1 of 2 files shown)


ziyunp (approved):

ziyunp:
Not needed as we merged in PR [3456](https://github.com/palantir/osdk-ts/pull/3456)

PR #3456, opened 2026-06-08
Mikhailprice/blob download filename parameter

mikhailprice:
## Add `downloadFileName` prop to the PDF viewer

  > Stacked on #<PR1>. Review/merge that first; this branch contains its commit
  > and should be based on `mikhailprice/pdf-viewer-download-filename`.

  ### Motivation
  The toolbar download derives its filename from `src`. When `src` is a **Blob /
  ArrayBuffer** (or a `blob:` object URL), there is no name to parse, so the saved
  file falls back to a generic `document.pdf` — consumers have no way to set the
  intended name.

  ### Change
  Add an optional `downloadFileName` prop to `PdfViewerProps`; when provided, the
  toolbar download uses it. Behaviour is unchanged when omitted (still derived from
  `src`).

  ```ts
  // types.ts
  downloadFileName?: string;

  // PdfViewer.tsx
  const handleDownload = useCallback(() => {
    viewer.download(downloadFileName);
  }, [viewer, downloadFileName]);

  Testing

  - vitest run src/pdf-viewer → 159/159 pass
  - eslint + dprint clean
  - Changeset: @osdk/react-components minor

### DIFF
--- a/packages/react-components/src/pdf-viewer/types.ts
+++ b/packages/react-components/src/pdf-viewer/types.ts
@@ -222,6 +222,11 @@ export interface PdfViewerProps {
    */
   enableDownload?: boolean;
   /**
+   * Filename used when the user downloads the PDF via the toolbar button.
+   * If omitted, the name is derived from the `src` URL, falling back to "document.pdf".
+   */
+  downloadFileName?: string;
+  /**
    * Which sidebar panel to show: thumbnails or document outline.
    * @default "thumbnails"
    */
--- a/.changeset/pdf-viewer-download-event-fix.md
+++ b/.changeset/pdf-viewer-download-event-fix.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
+Fix PDF viewer toolbar download saving an invalid filename.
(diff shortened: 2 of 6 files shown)


packages/react-components/src/pdf-viewer/types.ts
```diff
@@ -221,6 +221,11 @@ export interface PdfViewerProps {
    * @default false
    */
   enableDownload?: boolean;
+  /**
+   * Filename used when the user downloads the PDF via the toolbar button.
+   * If omitted, the name is derived from the `src` URL, falling back to "document.pdf".
```
ziyunp, line 226: Maybe i'm missing something, where is the default downloadFileName set? I also don't see the fallback to document.pdf

mikhailprice: the function handles when filename == null and defaults to the pdf

ziyunp (dismissed):

packages/react-components/src/pdf-viewer/types.ts
ziyunp: Got it, thanks!

ziyunp (approved):

PR #3457, opened 2026-06-08
[Filter List] Fix filter list focus outline clipping

Leolide:
## Summary

- Override Blueprint's global `:focus { outline }` rule at the package level in `base.css`, so all OSDK components render without Blueprint's default focus ring
- Move Blueprint's CSS import into its own `bp` layer in Storybook, separate from the `storybook` layer, so CSS cascade ordering is explicit and correct

Before
Screenshot 2026-06-09 at 5 12 07 PM

After:
Screenshot 2026-06-09 at 5 11 33 PM

### 1. Suppress Blueprint focus ring for OSDK components

Blueprint applies a global `:focus { outline: rgba(33,93,176,0.752) solid 2px }` rule that bleeds into OSDK components (checkboxes, filter list rows, buttons). Added a top-level `:focus, :focus-visible { outline: none }` override in `base.css` that ships with the built `styles.css` inside `@layer osdk.tokens`. Since `osdk.tokens` is ordered after the `bp` layer, this cleanly wins in the CSS cascade for all consumers.

### 2. Separate Blueprint into its own CSS layer in Storybook

Previously Blueprint CSS was imported into the `storybook` layer alongside Storybook's own styles. Now it has a dedicated `bp` layer with explicit ordering: `bp, storybook, osdk.styles, themes`. This makes the cascade predictable and ensures OSDK token overrides reliably beat Blueprint defaults.

## Particular workflows to test

- [ ] Open FilterList story in Storybook — tab through checkboxes, buttons, and filter popover triggers; no blue focus outline should appear
- [ ] Verify the same in dark theme
- [ ] Check that keyboard navigation still works (focus should function, just without Blueprint's visible outline)
- [ ] Confirm other components (ObjectTable editable cells, action form inputs) also have no stray Blueprint focus ring


nfdiop (commented):
question: should we remove the outline instead?

ziyunp (approved):

ziyunp:
Action form still has the outlines around form fields

Screenshot 2026-06-10 at 10 39 20

PR #3462, opened 2026-06-08
Fix docs

ziyunp:
- Removed @beta from the install instructions
- Removed separate instructions for cbac-components as it's been moved into react-components
- Simplified css setup, user doesn't need to know about the layers under the hood
-


olliejenningsPT (approved):

Leolide:
Looks good to me

PR #3463, opened 2026-06-08
Add auto generation script for components props

ziyunp:
- Added a gen-props script to autogenerate a props table in the docs when the AUTOGEN marker is added.
- Added a check-gen-props script in the CI to run a check if the props table drifted from the current list of props. If the CI check fails, author has to run the gen-props script locally to update the docs.


packages/react-components/docs/DocumentViewer.md
```diff
@@ -46,21 +46,25 @@ import { DocumentViewer } from "@osdk/react-components/experimental/document-vie
...
+
+| Name                    | Type                                                                | Description                                                                                                                                                                   |
+| ----------------------- | ------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `media`                 | `Media`                                                             | **Required.** The Media object to render                                                                                                                                      |
+| `className`             | `string`                                                            | Additional CSS class name for the root element Defaults to `undefined`.                                                                                                       |
```
KuberSethi, line 51: hmm a lot of these default to undefined is that okay?

packages/react-components/scripts/gen-props.mjs
```diff
@@ -0,0 +1,652 @@
+#!/usr/bin/env node
```
KuberSethi, line 1: is it worth writing some tests for this, at least to verify it runs when it should / doesn't run when it shouldn't?

packages/react-components/scripts/gen-props.mjs
```diff
@@ -0,0 +1,652 @@
...
+      if (seen.has(key)) return []; // already resolving this type up the stack
+      seen.add(key);
+      const result = membersOfDeclaration(found.decl, found.sourceFile, seen);
+      seen.delete(key);
+      return result;
+    }
+    // A named type we were asked to pull members from but couldn't resolve
+    // (external/bare import, re-export, or aliased import). Surface it — a
+    // silently-incomplete table is the worst outcome for a drift checker.
+    console.warn(
```
KuberSethi: If a silent incomplete is bad should we console.error vs warn? Or does this work too?

packages/react-components/scripts/gen-props.mjs
```diff
@@ -0,0 +1,652 @@
...
+    }
+    return buildBlock(
+      buildRows(resolved.entries),
+      resolved.typeParams,
+      src,
+      interfaceName,
+    );
+  });
+
+  // Run the whole doc through dprint so the table alignment and surrounding
```
KuberSethi: Can we perhaps check before doing this as I think this might slow things down a lot potentially

KuberSethi (commented):
How long does the script take to run? My understanding of reading it is that we're recursively walking every markdown file under docs/ and regenerating against every autogen block it finds. I wonder if we could more intelligently figure out when to run this but if perf is fine rn then this is probably fine. Also is this gated to just components docs / can we make sure docs changes in other osdk-ts packages don't trigger the script? 

Mostly looks good just a few questions!

packages/react-components/docs/DocumentViewer.md
ziyunp: @olliejenningsPT Looking at this list of props, maybe it's better to consolidate all viewProps for different media type?

packages/react-components/scripts/gen-props.mjs
ziyunp: Added tests

ziyunp:
> How long does the script take to run? My understanding of reading it is that we're recursively walking every markdown file under docs/ and regenerating against every autogen block it finds. I wonder if we could more intelligently figure out when to run this but if perf is fine rn then this is probably fine. Also is this gated to just components docs / can we make sure docs changes in other osdk-ts packages don't trigger the script?
> 
> Mostly looks good just a few questions!

Yes, it walks docs/ in react-components only. The DOCS_DIR variable constructs the path from package root, and this script is only present in react-components. Turbo also handles this with the inputs for the check-gen-props job set to ["src/**/*.{ts,tsx}", "docs/**/*.md", "scripts/gen-props.mjs"].

packages/react-components/scripts/gen-props.mjs
ziyunp: Updated to only run dprint when there are actual changes, also added tests to ensure this

packages/react-components/scripts/gen-props.mjs
ziyunp: Changed to console.error

KuberSethi (approved):

Issue #3345, opened 2026-05-21
Expose option to ensure consistency with snapshots when loading objects with fetch page


nihalbhatnagar:
API Gateway supports snapshot consistency[ when loading an object set](https://www.palantir.com/docs/foundry/api/v2/ontologies-v2-resources/ontology-object-sets/load-object-set/), however this option is not exposed for `fetchPage` in the OSDK.

esinx:
Resolved via #3468

PR #3468, opened 2026-06-08
Add $snapshot option to fetchPage

esinx:
This PR addresses #3345.

> API Gateway supports snapshot consistency [when loading an object set](https://www.palantir.com/docs/foundry/api/v2/ontologies-v2-resources/ontology-object-sets/load-object-set/), however this option is not exposed for fetchPage in the OSDK.

`FetchPageArgs` now expects an optional `$snapshot` argument which corresponds to the [`snapshot` request parameter](https://www.palantir.com/docs/foundry/api/v2/ontologies-v2-resources/ontology-object-sets/load-object-set#request-body:~:text=yet%20generally%20available.-,snapshot,booleanoptional,-A%20flag%20to)

### DIFF
--- a/packages/api/src/object/FetchPageArgs.ts
+++ b/packages/api/src/object/FetchPageArgs.ts
@@ -129,7 +129,7 @@ export interface FetchPageArgs<
     & MODIFIERS
     & { [P in Exclude<keyof MODIFIERS, PropertyKeys<Q>>]: never };
   /**
-   * Ensures paging consistency by freezing the view at the time of query to prevent duplicate or missing items. New results cannot be added to page when $snapshot is set to true.
+   * Ensures paging consistency by freezing the view at the time of query to prevent duplicate or missing items. Setting $snapshot to false ensures that you will always get the latest results.
    * @default false
    */
   $snapshot?: boolean;
--- a/packages/api/src/object/FetchPageArgs.ts
+++ b/packages/api/src/object/FetchPageArgs.ts
@@ -68,10 +68,6 @@ export namespace ObjectSetArgs {
     $nextPageToken?: string;
     $pageSize?: number;
   }
-
-  export interface Snapshot {
-    $snapshot?: boolean;
-  }
 }
 
 export interface SelectArg<
@@ -101,8 +97,6 @@ export type SelectArgToKeys<
   : A["$select"] extends readonly string[] ? A["$select"][number]
   : PropertyKeys<Q>;
 
-export interface SnapshotArg extends ObjectSetArgs.Snapshot {}
-
 export interface FetchPageArgs<
   Q extends ObjectOrInterfaceDefinition,
   K extends string = PropertyKeys<Q>,
@@ -126,8 +120,7 @@ export interface FetchPageArgs<
     ORDER_BY_OPTIONS,
     PROPERTY_SECURITIES,
     MODIFIERS
-  >,
-  SnapshotArg
+  >
 {
   $nextPageToken?: string;
   $pageSize?: number;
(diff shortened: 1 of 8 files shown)


packages/client/src/object/fetchPage.test.ts
```diff
@@ -39,6 +40,58 @@ import {
...
+/**
+ * Builds a minimal client whose `fetch` is a vitest mock, so tests can inspect
+ * the request bodies the client sends. The mock resolves every request with an
+ * empty (`{ data: [] }`), 200 OK response, so callers only care about what was
+ * sent, not what comes back.
+ *
+ * @returns the `client` to drive under test and the `fetchFn` mock to assert against
+ *   (typically via {@link lastObjectSetRequest}).
+ */
+function makeCapturingClient() {
```
nihalbhatnagar: I like this abstraction! Can we move these helpers into a util file?

packages/client/src/object/fetchPage.test.ts
```diff
@@ -618,4 +671,25 @@ describe(fetchPage, () => {
       });
     });
   });
+
+  describe("snapshot", () => {
```
nihalbhatnagar: Can we move to ObjectSet.test.ts

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -68,6 +68,10 @@ export namespace ObjectSetArgs {
     $nextPageToken?: string;
     $pageSize?: number;
   }
+
+  export interface Snapshot {
```
nihalbhatnagar: I think we can just expose `$snapshot?: boolean;` directly on the FetchPage interface? Basically copy $nextPageToken and $pageSize. Or was there a specific reason you did it this way

nihalbhatnagar: Also could we add a short jsdoc

packages/client/src/util/mockCaptureClient.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Finds the most recent request the mock `fetchFn` received whose JSON body
+ * mentions an `objectSet` (i.e. the object-set load call fetchPage issues),
+ * and returns it parsed. Walks the recorded calls newest-first so the latest
+ * matching request wins.
+ *
+ * @param fetchFn the mock returned by {@link createMockCapturingClient}
+ * @returns the parsed request body, or `undefined` if no object-set request was sent
+ */
+export function getLastObjectSetRequest(
+  fetchFn: MockedFunction<typeof globalThis.fetch>,
```
nihalbhatnagar, line 61: I feel like we could FLUP and reuse this https://github.com/palantir/osdk-ts/blob/9c3e1c06f0c9e1214263806b473a13a25eebe394/packages/faux/src/mock/OntologiesV2/OntologyObjectSets.ts#L21

packages/client/src/object/fetchPage.ts
```diff
@@ -466,7 +466,14 @@ export async function fetchPageWithErrorsInternal<
...
 ): Promise<Result<FetchPageResult<Q, L, R, S, T>>> {
   try {
-    const result = await fetchPageInternal(client, objectType, objectSet, args);
+    const { $snapshot, ...restArgs } = args;
+    const result = await fetchPageInternal(
+      client,
+      objectType,
+      objectSet,
+      restArgs,
+      $snapshot ?? false,
```
nihalbhatnagar: hmm I think we can refactor `fetchPageInternal` to use the snapshot passed in args directly (since args is typed `FetchPageArgs` anyways) and then in `createObjectSet` for asyncIter we can force override false

packages/api/src/object/FetchPageArgs.ts
```diff
@@ -128,6 +128,11 @@ export interface FetchPageArgs<
     & ApplyModifiersArg<Q>
     & MODIFIERS
     & { [P in Exclude<keyof MODIFIERS, PropertyKeys<Q>>]: never };
+  /**
+   * Ensures paging consistency by freezing the view at the time of query to prevent duplicate or missing items. New results cannot be added to page when $snapshot is set to true.
```
nihalbhatnagar: This last sentence feels a bit confusing to me

nihalbhatnagar (approved):

PR #3471, opened 2026-06-09
ci: build once, run tests across Node version matrix

olliejenningsPT:
## What

Restructures `.github/workflows/ci.yml` so we **build once**, run the **Node-independent checks once each (in parallel)**, and run the **test suites across a Node version matrix** — instead of running the entire `pnpm check` + `pnpm build` pipeline 4× across `[18, 20, 22, 24]`.

The old `build` job ("Build and Test") fanned the full pipeline over a 4-version matrix. New shape (all the non-build jobs `needs: build` and reuse its turbo cache):

| Job | Host | Runs |
|-----|------|------|
| `build` | Node 24 | `turbo run transpile transpileTypes` — produces `build/{esm,cjs,browser,types}`, populates the turbo cache |
| `typecheck` | Node 24 | `typecheck` + `check-attw` + `check-api` + `check-bundle` |
| `lint` | Node 24 | `lint` + `//#check-mrl` |
| `cspell` | Node 24 | whole-repo `ci:cspell` |
| `test` | matrix `[18, 20, 22, 24]` | `turbo run test` against the built artifacts |
| `build-apps` | Node 24 | `turbo run build` — Storybook / examples / sandboxes (the packages with no `transpile`/`typecheck` task) |

All single-version jobs run on **Node 24**, matching `release.yml` and the deploy workflows (the host Node version does not affect emitted artifacts). The `test` matrix still covers `[18, 20, 22, 24]`.

`e2e`, `bundle-size`, `changesets`, `fork-guard` are unchanged. `ci-build` gates on `[build, build-apps, typecheck, lint, cspell, test, e2e, fork-guard]`.

## Why

- `tsc`/`babel`/`esbuild` emit identical output regardless of the host Node version (build target is controlled by `tsconfig target` / `tsup target`, not the runtime). Building 4× produced the same bytes 4×.
- Typecheck, lint, cspell, attw produce no per-Node output — no reason to matrix them.
- The load-bearing part of the matrix is running the code under each Node version **at test time** — preserved in full by the `test` matrix.

## CI performance: cold vs warm

**Cold / full rebuild (measured, apples-to-apples):** compared against a cold `main` run (a "Version Packages" commit that invalidated everything):

| | Old (cold) | New (cold) |
|---|---|---|
| Wall-clock | ~1337s | **~771s (~42% faster)** |
| Compute (Σ job-seconds) | ~5097s | **~3218s (~37% less)** |
| Builds | 4× | 1× |

The win comes from building once instead of four times, and from splitting the previously-serial check phase (typecheck + lint + cspell + attw all on one runner, with attw tailing behind typecheck) into parallel jobs — so the check phase ≈ the slowest single job (~8min) instead of their sum (~13min).

**Warm / incremental (hard to verify pre-merge):** most `main` commits are incremental, where the old workflow was ~380–480s (E2E-bound) because its per-node turbo cache (keyed per-SHA with `restore-keys`) carried check/test **results** across runs. To preserve that, each new job has **its own turbo cache** keyed per-job (and per-Node for `test`, since turbo's hash doesn't encode the runtime) with `restore-keys` that warm from a prior same-job cache, falling back to the build cache.

Caveat worth being explicit about: **this warm path can't be demonstrated on this PR.** GitHub scopes caches by ref, so the reliable warm-up is `main`→`main` (stable `refs/heads/main`) and PRs restoring `main`'s caches — exactly how the old workflow stayed fast. But (1) `main` still runs the old workflow, so it has no `turbo-…-typecheck-…`-style keys for a PR to restore yet, and (2) every run of *this* PR force-rebuilds (`--force`) because the PR changes `.github/workflows/ci.yml`, an infra-trigger file — so `build` and its dependents are cold here by design. Net: the **first** `main` run after merge is cold; the **second** should be warm (~E2E-bound, like the old workflow). The per-job caches use the same proven prefix-restore mechanism as the old per-node caches, so warm steady-state CI should match the old workflow while cold runs stay ~42% faster.

## Guarantees preserved

- **Node 18 support is genuinely tested** — the `test` matrix runs the suites under 18/20/22/24 against the built artifacts.
- **Node-18 syntax safety at build** — CJS (`tsup`) targets `es2022`; ESM/browser (Babel) only strips types; tsconfig `target` is `ES2022`. The build host's Node version never contributed to this, so dropping the build matrix loses nothing. Build tooling left untouched (see below).
- **Vite 7 / Storybook Node-18 exclusions preserved** on the Node-18 `test` leg (`@osdk/create-app.template.*`, `@osdk/create-widget.template.*`, `@osdk/widget.vite-plugin`, `@osdk/examples.*`, `@psdk/examples.*`, `@osdk/e2e.sandbox.todowidget`, `@osdk/ontology-explorer-app`, `@osdk/react-components-storybook`).
- **infra-changed `--force`** preserved on `build` (threaded to `build-apps` via a job output).
- **`ci-build` still fails loudly** if any required job fails.

## Supporting fixes (needed for build-once to be correct/green)

- `@osdk/react-components` tests run **isolated** in their own `turbo run test` invocation. Its large happy-dom suite was starving vitest's worker RPC ("Timeout calling onTaskUpdate") under the concurrent matrix on the slowest leg (Node 20) — all tests passed, only the reporter RPC timed out. Running it uncontended fixes it. No package change.
- `packages/ontology-explorer-app/turbo.json` (new): declare `codegen` outputs (`build/scss-types/**`) so the SCSS type declarations are cached/restored (else `tsc` fails `TS2307` when codegen is a cache hit).
- `packages/e2e.generated.1.1.x/turbo.json`: add `cts` to the `typecheck` outputs (its `typecheck.sh` renames CJS `*.d.ts` → `*.d.cts`) and make `typecheck` depend on its own `transpile` (they both write `build/cjs`) — fixes a `check-attw` "No types" failure.
- `changesets` job: pinned to Node 24 (was referencing `matrix.node` with no matrix — a latent bug).

## Build target & eslint-plugin-n

- **Build target** left at `es2022` (Node-18-safe). The "pin to node18" idea was investigated; `@osdk/monorepo.tool.transpile` already targets `es2022` for tsup and the ESM/browser path is Babel type-stripping with no target — no syntax-safety gain, and changing artifact-producing tooling is out of scope for a CI PR.
- **eslint-plugin-n** deferred to a follow-up (touches eslint config + package source).

## Verification

- `actionlint` clean (the one remaining `SC2016:info` is pre-existing in the untouched `bundle-size` `node -e` script).
- Full matrix green; `test` expands to `[18, 20, 22, 24]`; `ci-build` gates on all jobs.
- No published-package behavior changed, so no changeset required.

Draft for review.

### DIFF
--- a/.changeset/react-components-vitest-single-worker.md
+++ b/.changeset/react-components-vitest-single-worker.md
@@ -1,5 +1,0 @@
----
-"@osdk/react-components": patch
----
-
-Switch the test suite to the vitest forks pool to avoid a worker RPC timeout under concurrent CI load. No runtime changes.
--- a/.changeset/react-components-vitest-single-worker.md
+++ b/.changeset/react-components-vitest-single-worker.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
+Switch the test suite to the vitest forks pool to avoid a worker RPC timeout under concurrent CI load. No runtime changes.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,10 @@ jobs:
         run: u="$(git ls-files --others --exclude-standard)" && test -z "$u"
 
   cspell:
-    name: Spelling
+    # TODO: rename to "Spelling" once we can update the required status checks in
+    # branch protection. Kept as "Check spelling" for now so it matches the
+    # existing required check (branch protection can't be modified from here yet).
+    name: Check spelling
     timeout-minutes: 15
     runs-on: ubuntu-latest
     steps:
@@ -256,7 +259,11 @@ jobs:
   # the `build` job under every supported Node version. This is what actually
   # proves the published packages run on Node 18 (and up).
   test:
-    name: Test (Node ${{ matrix.node }})
+    # TODO: rename to "Test (Node ${{ matrix.node }})" once we can update the
+    # required status checks in branch protection. Kept as "Build and Test" for
+    # now so the matrix checks ("Build and Test (18)" … "(24)") match the existing
+    # required checks (branch protection can't be modified from here yet).
+    name: Build and Test
     needs: build
     timeout-minutes: 30
     runs-on: ubuntu-latest
@@ -498,7 +505,7 @@ jobs:
       - name: Setup Node.js environment
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: 22
+          node-version: 24
           cache: "pnpm"
 
       - name: Cache Turbo
@@ -709,7 +716,7 @@ jobs:
   ci-build:
     name: ci-build
     if: ${{ always() }}
-    needs: [build, build-apps, typecheck, lint, cspell, test, e2e, fork-guard]
+    needs: [build, build-apps, typecheck, lint, cspell, test, e2e, fork-guard, changes, storybook-tests]
     runs-on: ubuntu-latest
     steps:
       - name: Verify dependent jobs succeeded
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,12 +146,17 @@ jobs:
           node-version: 22
           cache: "pnpm"
 
-      - name: Restore Turbo cache from build
+      # Own per-job cache so typecheck/attw results persist across runs (warm).
+      # restore-keys fall back to a prior typecheck cache first, then this run's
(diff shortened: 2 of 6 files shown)


ssanjay1:
Why did we choose node 22?

olliejenningsPT:
> Why did we choose node 22?

So l was originally going to choose Node24, but thought might be too big of a jump for people and use the 2nd latest LTS, if people are happy l will change to use Node24?

tdeitch (approved):

PR #3472, opened 2026-06-09
address officeassignment sandbox review comments

KuberSethi:
follow-up to #3412 addressing review comments on the officeassignment e2e sandbox

• refactor the edit form onto ActionForm and surface action errors on the exclusion toggle
• rename the filter-state hook return, broaden filter reset, and give the drawer dialog an accessible name
• small cleanups: hoist static column defs, dedupe the latest-timestamp rdp, drop dead css, named app export

### DIFF
--- a/packages/e2e.sandbox.officeassignment/src/components/actions/UpdateAssignmentForm.tsx
+++ b/packages/e2e.sandbox.officeassignment/src/components/actions/UpdateAssignmentForm.tsx
@@ -15,103 +15,94 @@
  */
 
 import { useOsdkAction } from "@osdk/react";
+import {
+  ActionForm,
+  type FormFieldDefinition,
+  type FormState,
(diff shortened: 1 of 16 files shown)


packages/e2e.sandbox.officeassignment/src/components/actions/UpdateAssignmentForm.tsx
```diff
@@ -15,103 +15,94 @@
...
-        className={styles.primaryButton}
-        disabled={isPending}
-      >
-        Save changes
-      </button>
+    <>
+      <ActionForm
+        actionDefinition={updateAssignment}
+        formFieldDefinitions={fieldDefinitions}
+        onSubmit={handleSubmit}
```
nfdiop, line 100: nit: ideally we would let the form handle the submit to test that it works as expected

nfdiop (approved):

PR #3477, opened 2026-06-10
Sort decimal/long properties numerically in observable client

nfdiop:
## What

Sorts `decimal` and `long` properties numerically instead of lexicographically in the observable client. These property types arrive over the wire as strings, so `orderBy` previously sorted them as text (`"10" < "2"`), producing the wrong order. This also covers derived (RDP) properties and interface lists.

## How

- Adds `compareNumericStrings` — a BigInt-based comparator that orders numeric strings by value.
- Adds `resolvePropertyType`, which resolves a property's declared type at runtime via the object/interface definition or RDP metadata.
- Threads RDP type metadata (`RdpDefRef`) through object/interface construction and the strip/filter/merge rebuild paths so the type survives, and captures the property type for `min`/`max` aggregations.
- `SortingStrategy` uses these to sort numeric-typed properties by value.

## Before
before_fix

## After
after_fix

## Notes
This is the first half of a split from #3476. The numeric **comparison/filtering** side follows in a stacked PR (`nfdiop/comparison-numeric-properties`); the shared RDP-metadata infrastructure lands here since it ships first.

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects/InternalSymbols.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/InternalSymbols.ts
@@ -16,6 +16,7 @@
 
 import type { ObjectOrInterfaceDefinition, OsdkBase } from "@osdk/api";
 import type { PropertySecurities } from "@osdk/foundry.ontologies";
+import type { DerivedPropertyRuntimeMetadata } from "../../derivedProperties/derivedPropertyRuntimeMetadata.js";
 
 /** @internal */
 export const UnderlyingOsdkObject = Symbol(
@@ -25,6 +26,18 @@ export const UnderlyingOsdkObject = Symbol(
 /** @internal */
 export const ObjectDefRef = Symbol(
   process.env.MODE !== "production" ? "ObjectDefinition" : undefined,
+);
+
+/**
+ * Holds the runtime metadata for an object's derived (runtime-derived)
+ * properties. Unlike regular properties, derived properties aren't part of the
+ * object/interface metadata, so their types are captured here at construction
+ * time (e.g. so sorting can compare numeric-but-string types correctly).
+ *
+ * @internal
+ */
+export const RdpDefRef = Symbol(
+  process.env.MODE !== "production" ? "DerivedPropertyDefinitions" : undefined,
 );
 
 /** @internal */
@@ -46,5 +59,6 @@ export interface HolderBase<T extends ObjectOrInterfaceDefinition> {
   [UnderlyingOsdkObject]: OsdkBase<any>;
   [ObjectDefRef]?: T;
   [InterfaceDefRef]?: T;
+  [RdpDefRef]?: DerivedPropertyRuntimeMetadata;
   [PropertySecuritiesRef]?: PropertySecurities[];
 }
--- a/packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
@@ -20,7 +20,7 @@ import {
   InterfaceDefinitions,
 } from "../../ontology/OntologyProvider.js";
 import { createOsdkInterface } from "./createOsdkInterface.js";
-import { ObjectDefRef } from "./InternalSymbols.js";
+import { ObjectDefRef, RdpDefRef } from "./InternalSymbols.js";
 
 describe(createOsdkInterface, () => {
   it("works in the normal case", () => {
@@ -151,6 +151,125 @@ describe(createOsdkInterface, () => {
         "asdf": "hi mom"
       }"
     `);
+  });
+
+  it("carries runtime-derived properties and RDP metadata onto the interface", () => {
+    const rdpMetadata = {
+      "total": {
+        selectedOrCollectedPropertyType: { type: "decimal" },
+        definition: { type: "selection" },
+      },
+    };
+
+    const underlying = {
+      "foo": "hi mom",
+      // derived properties live alongside regular ones on the underlying object
+      "total": "10",
+
+      [RdpDefRef]: rdpMetadata,
+
+      [ObjectDefRef]: {
+        [InterfaceDefinitions]: {},
+        apiName: "Obj",
+        displayName: "",
+        interfaceMap: {
+          "IFoo": {
+            "asdf": "foo",
+          },
+        },
+        inverseInterfaceMap: {},
+        links: {},
+        pluralDisplayName: "",
+        primaryKeyApiName: "",
+        primaryKeyType: "string",
+        properties: {
+          "foo": {
+            type: "string",
+          },
+        },
+        type: "object",
+        titleProperty: "foo",
+        rid: "",
+        status: "ACTIVE",
+        icon: undefined,
+        visibility: undefined,
+        description: undefined,
+      } satisfies FetchedObjectTypeDefinition,
+    };
+
+    const iface = createOsdkInterface(underlying as any, {
+      "apiName": "IFoo",
+      displayName: "",
+      links: {},
+      properties: {
+        "asdf": {
+          type: "string",
+        },
+      },
+      rid: "",
+      type: "interface",
+      implements: [],
+      description: undefined,
+    });
+
+    // the derived value is exposed on the interface view ...
+    expect((iface as any).total).toBe("10");
+    expect(Object.keys(iface)).toContain("total");
+    // ... and the RDP metadata is carried so its type can be resolved.
+    expect((iface as any)[RdpDefRef]).toBe(rdpMetadata);
+  });
+
+  it("does not let a $-prefixed derived property clobber the $-metadata", () => {
+    // RDP names are developer-chosen aliases and never $-prefixed in practice,
+    // but a derived property literally named "$title" must not overwrite the
+    // interface's $title accessor.
+    const underlying = {
+      "foo": "hi mom",
+      "$title": "real title",
+      "$primaryKey": "pk-1",
+      "$apiName": "Obj",
+
+      [RdpDefRef]: {
+        "$title": {
+          selectedOrCollectedPropertyType: { type: "string" },
+          definition: { type: "selection" },
+        },
+      },
+
+      [ObjectDefRef]: {
+        [InterfaceDefinitions]: {},
+        apiName: "Obj",
+        displayName: "",
+        interfaceMap: { "IFoo": { "asdf": "foo" } },
+        inverseInterfaceMap: {},
+        links: {},
+        pluralDisplayName: "",
+        primaryKeyApiName: "$primaryKey",
+        primaryKeyType: "string",
+        properties: { "foo": { type: "string" } },
+        type: "object",
+        titleProperty: "foo",
+        rid: "",
+        status: "ACTIVE",
+        icon: undefined,
+        visibility: undefined,
+        description: undefined,
+      } satisfies FetchedObjectTypeDefinition,
+    };
+
+    const iface = createOsdkInterface(underlying as any, {
+      "apiName": "IFoo",
+      displayName: "",
+      links: {},
+      properties: { "asdf": { type: "string" } },
+      rid: "",
+      type: "interface",
+      implements: [],
+      description: undefined,
+    });
+
+    // $title remains the interface's title accessor, not the RDP value.
+    expect((iface as any).$title).toBe("real title");
   });
 
   it("works with mixed namespaces", () => {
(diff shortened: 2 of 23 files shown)


packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
```diff
@@ -153,6 +157,195 @@ describe(createOsdkInterface, () => {
...
+      type: "interface",
+      implements: [],
+      description: undefined,
+    });
+
+    // the derived value is exposed on the interface view ...
+    expect((iface as any).total).toBe("10");
+    expect(Object.keys(iface)).toContain("total");
+    // ... and the RDP metadata is carried so its type can be resolved.
+    expect((iface as any)[RdpDefRef]).toBe(rdpMetadata);
```
nfdiop, line 223: rest of the tests is using any so kept it

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
```diff
@@ -44,6 +45,20 @@ export function createOsdkInterface<
       // first to minimize hidden classes
       [UnderlyingOsdkObject]: { value: underlying },
 
+      // Expose each runtime-derived (RDP) property value. These aren't part of
+      // interfaceDef.properties, so copy them from the underlying object (the
+      // RDP type metadata is carried separately via [RdpDefRef] below, so
+      // consumers like orderBy sorting can resolve their numeric types).
+      // Defined before the "$"-metadata descriptors so a derived property can
+      // never clobber them -- a later same-named "$"-descriptor wins.
+      ...Object.fromEntries(
```
nfdiop, line 54: Required so we can sort by the derived properties properly. Would love expert eyes on it

packages/client/src/observable/internal/compareNumericStrings.ts
```diff
@@ -0,0 +1,64 @@
...
+ * exceed the precision of a double. Exotic forms (e.g. scientific notation)
+ * fall back to a best-effort Number() comparison.
+ *
+ * The fallback keeps a TOTAL order even for unparseable input: a value that
+ * isn't a finite number (e.g. malformed data on a numeric column) sorts after
+ * every finite value and is never reported equal to one. This matters because
+ * the function is used as an Array.sort comparator -- a NaN-vs-finite pair that
+ * returned 0 would make the comparator non-transitive and corrupt the sort.
+ */
+export function compareNumericStrings(a: string, b: string): number {
```
nfdiop, line 44: very simple function that won't handle for example scientific notations. But I don't expect those. We could use a more precise comparison/library if needed

packages/client/src/observable/internal/compareNumericStrings.ts
```diff
@@ -0,0 +1,74 @@
...
+ * and is never reported equal to a finite value, so the comparator stays a
+ * TOTAL order -- it's used as an Array.sort comparator, and a non-transitive
+ * NaN-vs-finite pair would corrupt the sort.
+ */
+export function compareNumericStrings(a: string, b: string): number {
+  if (!NUMERIC_STRING.test(a) || !NUMERIC_STRING.test(b)) {
+    // Empty/whitespace ("no value") is the smallest value; two such are equal.
+    const aEmpty = a.trim() === "";
+    const bEmpty = b.trim() === "";
+    if (aEmpty || bEmpty) {
```
nfdiop, line 49: matching workshop behaviour

packages/client/src/util/extractRdpDefinition.ts
```diff
@@ -110,6 +110,14 @@ async function extractRdpDefinitionInternal(
           case "collectList":
           case "collectSet":
           case "get":
+          // get/collect/min/max preserve the aggregated property's numeric type
```
KuberSethi, line 113: This comment is a little unclear to me, perhaps we can enumerate the "why" behind this block of code above on 109?

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
```diff
@@ -153,6 +157,195 @@ describe(createOsdkInterface, () => {
     `);
   });
 
+  it("carries runtime-derived properties and RDP metadata onto the interface", () => {
```
KuberSethi, line 160: Hmm I appreciate these tests being added but could they perhaps be added in a separate PR? Seems they're testing existing behavior vs the goal of this PR

packages/client/src/object/convertWireToOsdkObjects/InternalSymbols.ts
```diff
@@ -27,6 +28,18 @@ export const ObjectDefRef = Symbol(
...
+/**
+ * Holds the runtime metadata for an object's derived (runtime-derived)
+ * properties. Unlike regular properties, derived properties aren't part of the
+ * object/interface metadata, so their types are captured here at construction
+ * time (e.g. so sorting can compare numeric-but-string types correctly).
+ *
+ * @internal
+ */
+export const RdpDefRef = Symbol(
+  process.env.MODE !== "production" ? "DerivedPropertyDefinitions" : undefined,
```
KuberSethi, line 40: hmm maybe I'm not understanding but don't we need this defined even when in dev mode?

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
KuberSethi: Hmm I think this is technically safe but it does increase the complexity quite a bit. In `ObjectsHelper.ts` we actually deepEqual compare `ObjectHolder`s (we unwrap `InterfaceHolder`s into their underlying `ObjectHolder`) so while this is fine for us because we're just touching the interface, just calling out that if this leaks into the Object Holder this will trigger a really weird bug where we're now comparing RdpDefRefs too. I think this is fine but just calling out the additional complexity.

packages/client/src/util/extractRdpDefinition.ts
KuberSethi, line 113: nit: if this comment encompasses `get` should we move it above the `get`?

packages/client/src/observable/internal/compareNumericStrings.ts
```diff
@@ -0,0 +1,74 @@
+/*
```
KuberSethi, line 1: Out of curiosity is there a library which handles this sort of thing for us? Just wondering as surely we can't be the only ones to run into this, and i might prefer to delegate this complexity (regex, exotic form / scientific notation and other weirdness handling, etc) to that

packages/client/src/observable/internal/sorting/SortingStrategy.test.ts
```diff
@@ -0,0 +1,342 @@
...
+      fakeHolder({ agg }, { derivedUntyped: ["agg"] })
+    );
+    expect(sort({ agg: "asc" }, withNonNumeric).map((h) => h.agg)).toEqual([
+      "10",
+      "9",
+      "x",
+    ]);
+  });
+
+  it("sorts numbers numerically", () => {
```
KuberSethi, line 120: nit: is this really testing anything we've added vs just testing library behavior?

packages/client/src/observable/internal/list/ListQuery.test.ts
```diff
@@ -572,6 +579,86 @@ describe("ListQuery sort stability across pages", () => {
     const ids = payload!.resolvedList!.map((t) => t.id);
     expect(ids).toEqual([10, 30, 31, 20]);
   });
+
+  it("clientOrdered sorts decimal/long properties numerically, not lexicographically", async () => {
```
KuberSethi, line 583: nit: drop the `not lexicographically`, i think after this ships and things are sorted numerically (as one might expect) then this is a bit of a weird thing to put here

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
```diff
@@ -153,6 +157,195 @@ describe(createOsdkInterface, () => {
...
+
+    // the derived value is exposed on the interface view ...
+    expect((iface as any).total).toBe("10");
+    expect(Object.keys(iface)).toContain("total");
+    // ... and the captured type survives, so consumers resolve it as numeric.
+    expect(iface[RdpDefRef]?.total?.selectedOrCollectedPropertyType?.type)
+      .toBe("decimal");
+  });
+
+  it("does not let a $-prefixed derived property clobber the $-metadata", () => {
```
KuberSethi, line 294: I don't think we need this test or behavior - for context maybe this is a bad thing but we hold that users shouldn't collide with our protected symbols/names e.g. "$" or the PKs, so maybe this is fine to be safe but for SA generally I think unneeded as I think other things would break if we hit this case anyways

KuberSethi (commented):
Thanks for taking this on! I think this is a step in the right direction but I think this is a bit overengineered / overcomplicated and could be simplified greatly to reduce the potential blast radius as well as complexity.

Most of the PR from what I can tell is actually wiring through RDP typedef metadata into every object such that the sorting comparator can later read it, but in reality I think we actually only care that `resolvePropertyType` reads the `RdpDefRef` (and then its only caller in turn is just the orderBy comparator). This basically introduces a ton of new wiring and complexity all the way through the observable client just to carry a type that we actually already have I believe at the query level. 

`OrderBySortingStrategy` already knows the derived types from `extractRdpDefinition` so I wonder if we could just pass in an `orderByKey` type or enumerate the numeric types in the constructor in the sorting strategy and then write a `createOrderBySort` function to read + sort there. I think this way we might be able to drop the `RdpDefRef` wiring complexity! Let me know though if I'm misunderstanding something, very possible with this PR as it touches many layers / concerns. 

Overall I just think we can probably get away without having to make changes to `createOsdkObject/Interface` and related wiring.

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.ts
nihalbhatnagar: Isnt an RDP any key thats not prefixed by $ and not on the metadata? Do we need this?

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
nfdiop: Got it!

packages/client/src/observable/internal/list/ListQuery.test.ts
nfdiop: makes sense!

packages/client/src/observable/internal/sorting/SortingStrategy.test.ts
nfdiop: testing that i didn't introduce any regression in the existing code

packages/client/src/observable/internal/compareNumericStrings.ts
nfdiop: See this comment https://github.com/palantir/osdk-ts/pull/3477/changes/BASE..80f622046f20c4cc980e708ca5dfa0d288f50d14#r3395917713
I hesitated to use `big.js` but since this library is shipped everywhere I didn't want to risk increasing the bundle size. Happy to use it if we prefer

packages/client/src/object/convertWireToOsdkObjects/InternalSymbols.ts
nfdiop: Good question. Not sure. Copied the rest of the symbols

packages/client/src/object/convertWireToOsdkObjects/createOsdkInterface.test.ts
nfdiop: Hm, it is testing the change we made to the interfaces

packages/client/src/util/extractRdpDefinition.ts
nfdiop: sounds good!

nfdiop:
> Thanks for taking this on! I think this is a step in the right direction but I think this is a bit overengineered / overcomplicated and could be simplified greatly to reduce the potential blast radius as well as complexity.
> 
> Most of the PR from what I can tell is actually wiring through RDP typedef metadata into every object such that the sorting comparator can later read it, but in reality I think we actually only care that `resolvePropertyType` reads the `RdpDefRef` (and then its only caller in turn is just the orderBy comparator). This basically introduces a ton of new wiring and complexity all the way through the observable client just to carry a type that we actually already have I believe at the query level.
> 
> `OrderBySortingStrategy` already knows the derived types from `extractRdpDefinition` so I wonder if we could just pass in an `orderByKey` type or enumerate the numeric types in the constructor in the sorting strategy and then write a `createOrderBySort` function to read + sort there. I think this way we might be able to drop the `RdpDefRef` wiring complexity! Let me know though if I'm misunderstanding something, very possible with this PR as it touches many layers / concerns.
> 
> Overall I just think we can probably get away without having to make changes to `createOsdkObject/Interface` and related wiring.

Thanks! That was my initial approach to pass the required data to `OrderBySortingStrategy` but this one looked more consistent. Will switch to that approach here https://github.com/palantir/osdk-ts/pull/3493

PR #3479, opened 2026-06-10
Run Storybook interaction tests in CI

ziyunp:
Wire @storybook/addon-vitest into the storybook package so play functions run as Vitest browser tests (headless Chromium), and add a dedicated CI job that fails the build on any interaction-test failure.

- vitest.config.ts + .storybook/vitest.setup.ts run every story through the Storybook test plugin; silent: "passed-only" keeps the huge data-fetching story logs out of passing runs
- Add a MultipleSelection play function to ObjectTable as an example - https://palantir.github.io/osdk-ts/storybook/pr-3479/158fd4b560693064a783af31bcd97ee246a26824/?path=/story/components-objecttable-features--multiple-selection
- Give ActionForm SubmitSuccess's success waitFor headroom so the real MSW/OSDK round-trip doesn't flake under CI contention
- New storybook-tests CI job is gated by a changes job and only runs when react-components or react-components-storybook changed

Screenshot 2026-06-10 at 11 53 58

https://github.com/user-attachments/assets/9f55ba19-37f7-4f6d-bf5f-6643a00d30df

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,6 +216,8 @@ jobs:
     name: Detect package changes
     timeout-minutes: 5
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     outputs:
       # "true" when this push/PR touches the component packages whose stories
       # the interaction tests exercise.
@@ -255,6 +257,8 @@ jobs:
     if: needs.changes.outputs.storybook == 'true'
     timeout-minutes: 20
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
 
     steps:
       - name: Check out code
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -212,6 +212,87 @@ jobs:
       - name: Faux Publish
         run: pnpm publish -r --no-git-checks
 
+  changes:
+    name: Detect package changes
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    outputs:
+      # "true" when this push/PR touches the component packages whose stories
+      # the interaction tests exercise.
+      storybook: ${{ steps.filter.outputs.storybook }}
+    steps:
+      - name: Check out code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 2
+
+      - name: Detect changes to component packages
+        id: filter
+        run: |
+          BASE_SHA="${{ github.event.pull_request.base.sha || github.event.before }}"
+
+          # Fail open: if we can't resolve the base commit (e.g. a deeper push
+          # than the shallow clone, or the all-zero SHA of a branch's first
+          # push), run the tests rather than risk skipping a relevant change.
+          if [ -z "$BASE_SHA" ] || ! git cat-file -e "$BASE_SHA^{commit}" 2>/dev/null; then
+            echo "Base commit '$BASE_SHA' unavailable; running storybook tests."
+            echo "storybook=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          if git diff --name-only "$BASE_SHA" HEAD -- \
+            packages/react-components \
+            packages/react-components-storybook \
+            | grep -q .; then
+            echo "storybook=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "storybook=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  storybook-tests:
+    name: Storybook Interaction Tests
+    needs: changes
+    if: needs.changes.outputs.storybook == 'true'
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 2
+
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4.3.0
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 22
+          cache: "pnpm"
+
+      - name: Cache Turbo
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: .turbo/cache
+          key: turbo-${{ runner.os }}-${{ github.sha }}
+          restore-keys: |
+            turbo-${{ runner.os }}-
+
+      - name: Install dependencies
+        run: pnpm install
+
+      # The stories import the browser build of the workspace packages they
+      # render, so those must be transpiled before the tests run.
+      - name: Transpile story dependencies
+        run: pnpm exec turbo transpileBrowser --filter='@osdk/react-components-storybook^...'
+
+      # Vitest browser mode runs the play functions in headless Chromium.
+      - name: Install Playwright Chromium
+        run: pnpm --filter @osdk/react-components-storybook exec playwright install --with-deps chromium
+
+      - name: Run interaction tests
+        run: pnpm --filter @osdk/react-components-storybook test-storybook
+
   bundle-size:
     name: Bundle Size
     if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -212,6 +212,87 @@ jobs:
...
+          # Fail open: if we can't resolve the base commit (e.g. a deeper push
+          # than the shallow clone, or the all-zero SHA of a branch's first
+          # push), run the tests rather than risk skipping a relevant change.
+          if [ -z "$BASE_SHA" ] || ! git cat-file -e "$BASE_SHA^{commit}" 2>/dev/null; then
+            echo "Base commit '$BASE_SHA' unavailable; running storybook tests."
+            echo "storybook=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          if git diff --name-only "$BASE_SHA" HEAD -- \
```
nfdiop, line 245: Hum, what if we have breaking changes in the client?

.github/workflows/ci.yml
```diff
@@ -212,6 +212,87 @@ jobs:
...
+
+      - name: Install dependencies
+        run: pnpm install
+
+      # The stories import the browser build of the workspace packages they
+      # render, so those must be transpiled before the tests run.
+      - name: Transpile story dependencies
+        run: pnpm exec turbo transpileBrowser --filter='@osdk/react-components-storybook^...'
+
+      # Vitest browser mode runs the play functions in headless Chromium.
```
nfdiop, line 293: why not have it in our package.json?

packages/react-components-storybook/src/stories/ActionForm/ActionForm.stories.tsx
```diff
@@ -517,7 +523,9 @@ export const SubmitSuccess: Story = {
     );
     await userEvent.click(submitButton);
 
-    await waitFor(() => expect(successSpy).toHaveBeenCalled());
+    await waitFor(() => expect(successSpy).toHaveBeenCalled(), {
```
nfdiop: we will probably need to fix these afterwards and remove the timeout

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
nfdiop, line 1: nice tests!

packages/react-components-storybook/vitest.config.ts
```diff
@@ -0,0 +1,67 @@
...
+    testTimeout: 30_000,
+    // Several data-fetching stories log full Response payloads (including
+    // multi-MB PDFs as raw text) to the console. Intercept that noise on
+    // passing tests so the run shows clean progress instead of looking hung,
+    // while still surfacing logs for any test that fails.
+    silent: "passed-only",
+    // Retry flaky play tests in CI only. Browser interaction tests can flake
+    // under CI load (slow round-trips, timing). Locally we keep 0 retries so
+    // flakes surface immediately instead of being silently masked.
+    retry: process.env.CI ? 2 : 0,
```
nfdiop: retrying on CI just hides the flakes though - a test that only passes on attempt 3 still shows green, so the flake never gets fixed. can we have a FLUP and is `2` masking a real timing bug vs genuine runner contention?

packages/react-components-storybook/vitest.config.ts
```diff
@@ -0,0 +1,67 @@
...
+    setupFiles: [".storybook/vitest.setup.ts"],
+    // Interaction tests drive real (MSW-mocked) async round-trips; give them
+    // room above Vitest's 5s default so a loaded CI runner doesn't trip the
+    // per-test timeout before a play function's own waitFor resolves.
+    testTimeout: 30_000,
+    // Several data-fetching stories log full Response payloads (including
+    // multi-MB PDFs as raw text) to the console. Intercept that noise on
+    // passing tests so the run shows clean progress instead of looking hung,
+    // while still surfacing logs for any test that fails.
+    silent: "passed-only",
```
nfdiop: can we stop those stories logging multi-MB payloads to the console. ok to suppress for now

packages/react-components-storybook/package.json
```diff
@@ -31,23 +32,27 @@
...
     "@storybook/addon-mcp": "^0.2.3",
+    "@storybook/addon-vitest": "~10.2.10",
     "@storybook/icons": "^2.0.1",
     "@storybook/react-vite": "^10.2.10",
     "@tanstack/react-table": "^8.21.3",
     "@types/node": "^20.19.13",
     "@types/react": "^18.3.24",
     "@types/react-dom": "^18.3.7",
     "@vitejs/plugin-react": "^4.3.4",
+    "@vitest/browser": "3.2.4",
```
nfdiop: Apparently @vitest/browser@3.2.4  is in the affected range for GHSA-g8mr-85jm-7xhm, a Critical Vitest Browser Mode advisor https://github.com/vitest-dev/vitest/security/advisories/GHSA-g8mr-85jm-7xhm. So 3.2.5 instead?

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -700,6 +700,62 @@ export const MultipleSelection: Story = {
...
+      await expect(rowCheckbox).toBeChecked();
+    }
+
+    // Everything is selected, so the header label is "Deselect all rows" again.
+    // Clicking it clears the entire selection.
+    const deselectAllCheckbox = await canvas.findByRole("checkbox", {
+      name: /deselect all rows/i,
+    });
+    await userEvent.click(deselectAllCheckbox);
+    for (const rowCheckbox of rowCheckboxes) {
```
nfdiop: nit: these refs were captured before the header toggle re-rendered. probably fine since row keys are stable, but a fresh findAllByRole here would be safer against future DOM churn.

.github/workflows/ci.yml
```diff
@@ -212,6 +212,87 @@ jobs:
       - name: Faux Publish
         run: pnpm publish -r --no-git-checks
 
+  changes:
+    name: Detect package changes
```
nfdiop, line 216: can we have 
```
 permissions:
    contents: read
```
to all jobs?

.github/workflows/ci.yml
ziyunp: The storybook test should be contained for just the components interactions. @osdk/client is used to make real network call to fetch data and that should be tested in the e2e tests with sandbox apps

packages/react-components-storybook/vitest.config.ts
ziyunp: I was thinking the flakes is something like the mock data not loaded via the msw - I see this sometimes in storybook and have to reload the page to get data in.

ziyunp: Happy to remove this now and set it when needed

.github/workflows/ci.yml
ziyunp: We don't need to do this because we already added playwright as a dependency. Chromium is installed via `playwright install` which knows exactly which chromium version it needs

ziyunp: Here the command is telling playwright to only install the chromium browser, otherwise it will install other browsers like firefox

.github/workflows/ci.yml
ziyunp: added to the jobs I created

packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
ziyunp: done

packages/react-components-storybook/package.json
ziyunp: upgraded

packages/react-components-storybook/vitest.config.ts
ziyunp: set quiet: true

packages/react-components-storybook/src/stories/ActionForm/ActionForm.stories.tsx
ziyunp: Moved the timeout configuration to packages/react-components-storybook/.storybook/vitest.setup.ts for now so it's not buried in this story

.github/workflows/ci.yml
nfdiop: got it! I guess we can also expand later since we are also affected by changes inside osdk/react

nfdiop (approved):

olliejenningsPT (approved):

PR #3480, opened 2026-06-10
Lazily import optional @osdk/foundry.admin peer in @osdk/react platform-api hooks

ziyunp:
## Problem

Consumers running `vite build` against `@osdk/react` fail when they do **not** install the optional `@osdk/foundry.admin` peer dependency:

```
✗ Build failed in 3.19s
error during build:
node_modules/@osdk/react/build/browser/new/platform-apis/admin/useCbacBanner.js (17:9): "CbacBanners" is not exported by "__vite-optional-peer-dep:@osdk/foundry.admin:@osdk/react:false", imported by "node_modules/@osdk/react/build/browser/new/platform-apis/admin/useCbacBanner.js".
file: /app/.../repo/node_modules/@osdk/react/build/browser/new/platform-apis/admin/useCbacBanner.js:17:9

15:  */
16:
17: import { CbacBanners } from "@osdk/foundry.admin";
             ^
18: import React from "react";
19: import { usePlatformQuery } from "../../../utils/usePlatformQuery.js";
```

## Root cause

`@osdk/foundry.admin` is declared as an **optional** peer dependency (`peerDependenciesMeta.optional: true`) because only the admin platform-api hooks need it — the core object/ObjectSet hooks do not. When a consumer omits the peer, Vite replaces the module with an empty optional-peer-dep stub (`__vite-optional-peer-dep:@osdk/foundry.admin:@osdk/react:false`). The admin hooks used a **static** top-level `import { CbacBanners } from "@osdk/foundry.admin"`, so Rollup tries to resolve the named export `CbacBanners` against the empty stub and fails the build — even for consumers who never call the hook.

`useCbacBanner` is just the first one Rollup chokes on; **all 8 admin hooks** shared the same static value import (`CbacBanners`, `CbacMarkingRestrictionsObjects`, `Markings`, `Users`, `MarkingCategories`).

## Fix

Move the value imports into lazy `await import("@osdk/foundry.admin")` calls inside each hook's query callback. This is compatible with `usePlatformQuery`'s `() => Promise<T>` signature and mirrors the existing pattern in `foundry-sdk-generator` (`await import("@osdk/foundry.ontologies")`). Type-only imports (`type User`, `Marking`, `MarkingCategory`, `ListUsersResponse`, `UserStatus`) stay at the top since they're erased at build time.

This keeps the peer **truly optional**:
- Consumers who don't use admin hooks no longer pull `@osdk/foundry.admin` into their module graph → build passes, zero bundle cost (separate chunk, never loaded).
- Consumers who do use admin hooks install the peer and it resolves at runtime as before.

Keeping it as a *required* peer would have "fixed" the build only by forcing every `@osdk/react` consumer to install the full admin SDK — strictly worse and contrary to the optional design.

### DIFF
--- a/packages/react/src/new/platform-apis/admin/importFoundryAdmin.ts
+++ b/packages/react/src/new/platform-apis/admin/importFoundryAdmin.ts
@@ -17,6 +17,11 @@
 // eslint-disable-next-line @typescript-eslint/consistent-type-imports
 type FoundryAdminModule = typeof import("@osdk/foundry.admin");
 
+const MISSING_PEER_MESSAGE =
+  "@osdk/react: the optional peer dependency \"@osdk/foundry.admin\" is "
+  + "required to use the admin platform-api hooks (e.g. useCbacBanner, "
+  + "useMarkings, useFoundryUser). Install it to use these hooks.";
+
 /**
  * Lazily import the optional `@osdk/foundry.admin` peer dependency used by the
  * admin platform-api hooks.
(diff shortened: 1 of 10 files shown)


packages/react/src/new/platform-apis/admin/importFoundryAdmin.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Kept as a dynamic import (rather than a static top-level import) so consumers
+ * who don't use the admin hooks aren't forced to install the peer and don't
+ * fail their bundler build when it's absent. If the peer is missing, the import
+ * rejects with a clear, actionable error instead of an opaque module-resolution
+ * failure; that error surfaces through the hook's `error` state.
+ */
+export async function importFoundryAdmin(): Promise<FoundryAdminModule> {
+  try {
+    return await import("@osdk/foundry.admin");
+  } catch (cause) {
```
olliejenningsPT, line 44: One edge case worth calling out: this `catch` only fires when the import genuinely can't be resolved, like the peer not being installed in Node or an unbundled environment. Some bundlers instead replace a missing optional peer with an empty stub module, so the import resolves fine and the named export the caller destructures comes back as undefined. In that case they hit a TypeError at the call site rather than this nice message. It's an unusual setup (using an admin hook without installing the peer) so I don't think we need to handle it, but the actionable error isn't guaranteed in every environment.

PR #3031, opened 2026-04-16
Replace ESLint with oxlint for non-generator packages

olliejenningsPT:
## Summary

Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter.html) for linting across the monorepo, reducing lint time from ~30-60s (ESLint via Turbo across 77 packages) to **~3-4s** (single oxlint pass from root).

### Why

- **Performance**: oxlint lints 2871 files in ~3-4s on 16 threads. ESLint ran per-package via Turbo, each spawning a Node process with type-aware parsing — cold runs took 30-60s.
- **Simplicity**: One `oxlint .` from root replaces 77 per-package `eslint .` invocations orchestrated by Turbo.
- **Consistency with oxfmt**: We already migrated from dprint to oxfmt. oxlint completes the move to the [Oxc](https://oxc.rs) toolchain.

### What changed

**New files:**
- `.oxlintrc.json` — oxlint config with all rules mapped from `eslint.config.mjs`
- `plugins/oxlint-plugin-header.js` — JS plugin replicating `eslint-plugin-header` for Apache 2.0 license enforcement (with auto-fix)

**Removed:**
- 7 package-level `eslint.config.mjs` files (non-generator packages)
- `lint` / `fix-lint` scripts from 68 non-generator `package.json` files
- `eslint --fix` from lint-staged (oxlint runs in pre-commit instead)
- `dprint`, `eslint-formatter-autolinkable-stylish`, `eslint-import-resolver-typescript` devDependencies

**Updated:**
- `turbo.json` — added `//#oxlint` root task, included in `check` pipeline
- `.husky/pre-commit` — runs `oxlint` check (< 1s)
- Inline disable comments converted to `oxlint-disable` with updated rule prefixes (`@typescript-eslint/` → `typescript/`, `react-hooks/` → `react/`)

### Generator packages keep ESLint

9 codegen packages retain ESLint since they ship as standalone tools:
`generator`, `generator-utils`, `generator-converters`, `generator-converters.ontologyir`, `generator-converters.preview`, `foundry-sdk-generator`, `example-generator`, `typescript-docs-example-generator`, `osdk-docs-context-generator`

### Type-aware linting

[...]

PR #3481, opened 2026-06-10
chore: migrate shared.* to oxc toolchain via Ultracite (oxlint + oxfmt), sub-PR of #3031

olliejenningsPT:
## Summary

First incremental step of the repo-wide ESLint/dprint -> oxlint/oxfmt migration (#3031),
scoped to 4 published `shared.*` packages. **Stacked on #3471** (`ci-node-matrix-restructure`),
which runs `lint` + `//#check-mrl` once on Node 24 and matrixes only `test` — so oxlint
(Node 20+) needs no special-casing.

The oxlint/oxfmt config is **driven by [Ultracite](https://docs.ultracite.ai)**: root
`oxlint.config.ts` / `oxfmt.config.ts` that extend `ultracite/oxlint/core` + `ultracite/oxfmt`.

### Packages migrated (4)

`@osdk/shared.net.errors`, `@osdk/shared.client.impl`, `@osdk/shared.net.fetch`, `@osdk/shared.net`
(all `patch`). `@osdk/shared.test` (+ its companion `@osdk/shared.test.intellisense`) stay on
ESLint/dprint for now — deferred to a later increment.

### What changed

- **Config**: `oxlint.config.ts` + `oxfmt.config.ts` (Ultracite presets) replace `.oxlintrc.json` /
  `.oxfmtrc.json`. Keeps the repo's Apache-2.0 header plugin and ignore patterns;
  `vitest.config.mts` / `package.json` / `tsconfig.json` are left to monorepolint + dprint.
- **Strict preset, triaged**: adopt Ultracite's strict ruleset, with a documented override block in
  `oxlint.config.ts` for rules that would be pure churn (`func-style`, `no-use-before-define`,
  jsdoc, ...) or breaking for these published packages (`no-explicit-any` kept at `warn`,
  `custom-error-definition`, `default-param-last`, `require-await`, `eqeqeq {null: never}`). Two
  trivial safe fixes applied (`prefer-template`, `no-inferrable-types`); the rest auto-fixed.
- **Per-package (incremental)**: the 4 packages' `lint`/`fix-lint` call `oxlint`/`oxfmt` with the
  root `.ts` configs; everything else stays on ESLint/dprint. monorepolint `oxc` archetype,
  lint-staged carve-out, and the `//#global-oxc-config` turbo inputs all point at the new configs.

### Verification

- `pnpm turbo check` (whole repo) green locally: monorepolint, lint (all packages), typecheck,
  test, attw, api, bundle, cspell — 636/636.
- The 4 packages lint with 0 errors (only `no-explicit-any` warnings remain, by design).

### Deferred to later increments

Global root oxlint/oxfmt pass, migrating the remaining packages + removing ESLint/dprint,
`shared.test`/`shared.test.intellisense`, and Ultracite's whole-repo `ultracite check`/`fix` model.


.monorepolint.config.mjs
```diff
@@ -232,13 +228,32 @@ const archetypeRules = archetypes(
...
     ],
     {
       ...INTERNAL_LIBRARY_RULES,
     },
   )
+  // Packages migrated to the oxc toolchain (oxlint + oxfmt). As more packages
+  // are migrated, add them here (taking care not to capture generated
+  // namespace-prefix children via monorepolint's archetype matching). See #3031.
+  .addArchetype(
+    "oxc migrated libraries",
```
ssanjay1, line 245: I'm assuming you'll migrate the non standard ones all at once for each group

plugins/oxlint-plugin-header.js
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
```
ssanjay1, line 2: Are there no oxlint equivalents for the eslint header plugin?

.changeset/oxlint-shared-packages.md
```diff
@@ -0,0 +1,8 @@
+---
+"@osdk/shared.net.errors": patch
+"@osdk/shared.client.impl": patch
+"@osdk/shared.net.fetch": patch
+"@osdk/shared.net": patch
+---
+
+Lint with oxlint and format with oxfmt (configured via Ultracite) instead of ESLint and dprint (first increment of the repo-wide oxc migration)
```
ssanjay1: nit: can you add which packages you did here please

ssanjay1 (dismissed):
Approving to unblock, feel free to address comments as you see fit in FLUPs

.monorepolint.config.mjs
olliejenningsPT: yes

plugins/oxlint-plugin-header.js
olliejenningsPT: No oxlint-native option exists: oxlint has no built-in header/license rule, and there's no published oxlint header plugin. The ESLint plugins (eslint-plugin-header / eslint-plugin-headers) are configured entirely via rule options (comment style + header text/path), but oxlint's JS plugin system doesn't support passing options to plugin rules — I tested it and oxlint errors with "Rule does not accept options". So they can't be driven from oxlint, which is why this is a small self-contained plugin with the header baked in. Documented that rationale in the plugin header in 1714139c0d.

.changeset/oxlint-shared-packages.md
olliejenningsPT: Done in 1714139c0d — the changeset now lists the 4 packages (@osdk/shared.net, shared.net.errors, shared.net.fetch, shared.client.impl).

.monorepolint.config.mjs
olliejenningsPT: Yes — migrating per-group in small follow-up PRs. Next batch is queued (language-models + the sdk-docs libs), then the React packages once the Ultracite React preset is validated.

ssanjay1 (approved):

PR #3483, opened 2026-06-10
Lazy-load the CBAC banner in ObjectTable marking cells

ziyunp:
## Problem

Importing `ObjectTable` from `@osdk/react-components` dragged the CBAC picker — and the optional `@osdk/foundry.admin` peer it transitively reaches — into every consumer's **eager** module graph, even tables with no marking columns:

```
ObjectTable → useColumnDefs → DefaultCellRenderer
  → CbacMarkingCell / MandatoryMarkingCell
    → CbacBanner (cbac-picker)
      → @osdk/react/platform-apis → useCbacBanner → @osdk/foundry.admin
```

## Fix

New `LazyCbacBanner` wrapper (`React.lazy` + `Suspense`, following the existing `LazyDateCalendar` precedent), rendered by both `CbacMarkingCell` and `MandatoryMarkingCell`. The `CbacBanner` chunk now loads only when a `CBAC`/`MANDATORY` marking column actually renders. A `.catch` in the lazy factory degrades to an empty cell (logging once) if the chunk fails to load, so a load failure can't crash the table.

Both marking cells import `CbacBanner`, so the lazy boundary is placed at `CbacBanner` (one wrapper covers both cells) rather than at a single cell.

Fallback display when CbacBanner is not imported

Screenshot 2026-06-15 at 20 45 48

## Relationship to #3480

This is **code-splitting + graceful degradation**, not the build fix. The actual `vite build` failure (`"CbacBanners" is not exported by …foundry.admin…`) is fixed in #3480, which removes the static `@osdk/foundry.admin` named import inside the `@osdk/react` hooks. Lazy-loading alone can't fix the build because Rollup still validates named imports of statically-imported modules inside a dynamically-imported chunk. The two land together.

## Verification

- object-table + new `LazyCbacBanner` tests green, including a lazy-boundary test (banner is not rendered synchronously) and a guard test (degrades + warns once on chunk-load failure)
- transpiled browser output: **no static `cbac-picker` import** remains in the eager object-table graph; `LazyCbacBanner.js` carries the dynamic `import("../../cbac-picker/CbacBanner.js")`
- typecheck + lint clean for changed files

### DIFF
--- a/packages/react-components/src/object-table/components/LazyCbacBanner.tsx
+++ b/packages/react-components/src/object-table/components/LazyCbacBanner.tsx
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React, { Suspense } from "react";
+import type { CbacBannerProps } from "../../cbac-picker/CbacBanner.js";
+
+// Rendered when the CbacBanner chunk can't be loaded. CbacBanner pulls in
+// `@osdk/react/platform-apis`, whose admin hooks depend on the optional
+// `@osdk/foundry.admin` peer; if that chunk fails to load (peer not installed,
+// or a transient network error), degrade to an empty cell rather than letting
+// the failure propagate up and take down the whole table.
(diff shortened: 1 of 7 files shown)


packages/react-components/src/object-table/components/LazyCbacBanner.tsx
```diff
@@ -0,0 +1,61 @@
...
+);
+
+/**
+ * Suspense-wrapped, lazily-loaded {@link CbacBanner}. Same props as
+ * `CbacBanner`; renders nothing while the chunk loads, and falls back to the
+ * raw marking ids if the chunk fails to load.
+ */
+export function LazyCbacBanner(props: CbacBannerProps): React.ReactElement {
+  return (
+    <Suspense fallback={null}>
```
olliejenningsPT: `fallback={null}` means every marking cell paints empty and then pops the banner in once the chunk resolves, so the first render of a table with marking columns flashes and shifts. The chunk only loads once and subsequent renders are sync, but that first paint is a regression vs the old synchronous render.

`LazyDateCalendar` handles this by reserving space with a sized fallback (`<div className={styles.osdkDatePickerFallback} />`) instead of `null`. Could we do the same here with a placeholder matching the banner height to avoid the layout shift?

PR #3488, opened 2026-06-11
feat(create-app): add hidden TypeScript Library template

olliejenningsPT:
## Summary

Adds a **TypeScript Library** template to `@osdk/create-app` for scaffolding a publishable TypeScript library.

Unlike the app templates, a library has no use for the app-oriented prompts (`foundryUrl`, `scopes`, `clientId`, `ontology`, CORS, SDK version). So the template is registered with `hidden: true`: it stays out of the interactive `npm create @osdk/app` menu, but is still resolvable programmatically by id (`template-typescript-library`) and is present in the exported `TEMPLATES` array.

## What's included

- **New package `@osdk/create-app.template.typescript-library`** — the generic library skeleton under `templates/`:
  - `src/index.ts` (a starter `greet` export) and a colocated `src/__tests__/index.test.ts`
  - ESM `tsconfig.json` emitting declarations to `dist/`
  - [Ultracite](https://www.npmjs.com/package/ultracite) lint/format setup (`oxlint.config.ts`, `oxfmt.config.ts`) wired into `check`/`fix`/`lint`/`format` scripts, with `vitest` for tests
  - `README`, `AGENTS.md`, and `.claude/` config
- **Template registration** in `packages/create-app/codegen.mjs` — `id: typescript-library`, `label: "TypeScript Library"`, `hidden: true`, `buildDirectory: "dist"`. A new `onlyV2` flag pins it to the single `2.x` slot and skips the legacy `1.x` package discovery.
- **Generated example** committed at `examples/example-typescript-library-sdk-2.x` (produced by the example generator, like every other template).

## Design notes

- **Single, non-OSDK variant.** The library ships only a `2.x` files key and carries **no `@osdk/client` runtime dependency**. `cli.test.ts` asserts the *absence* of `@osdk/client` for this template (rather than the version-pin it asserts for the OSDK app templates).
- **Lint style override.** Ultracite's preset forces function *expressions* (`const f = () => …`). This template overrides `func-style` to **`["warn", "declaration"]`** — it steers toward named `function` declarations but only warns, so arrow functions remain allowed (inline arrow callbacks are unaffected).
- **`typescript` version.** The generated `package.json` uses the monorepo-standard `typescript ~5.5.4`. The template package is governed by the repo's shared lint rules, which pin that version, and the template packager merges the package's dev-dependency into the generated output.

## Testing

- `@osdk/create-app.template.typescript-library` builds (codegen + transpile + typecheck).
- `@osdk/create-app` test suite passes, including `cli.test.ts`.
- Confirmed the template is **absent** from the interactive menu but resolvable via `TEMPLATES.find(t => t.id === "template-typescript-library")`.
- The example-generator content check passes; the generated example builds (`tsc`), tests pass (Vitest), and `oxlint` is clean.
- `monorepolint`, `dprint`, and a full `turbo transpile` pass.


packages/create-app/codegen.mjs
```diff
@@ -54,6 +54,17 @@ export const TEMPLATES = [
...
   },
+  {
+    id: "typescript-library",
+    label: "TypeScript Library",
+    envPrefix: "",
+    buildDirectory: "dist",
+    hidden: true,
+    // A library only ships a single, non-OSDK-versioned variant. Pin it to the
+    // 2.x slot and skip the 1.x package discovery (which would otherwise also
+    // match the base package and emit a duplicate 1.x entry).
```
nfdiop: nit: I may be wrong but this is not required and already supported by the code

packages/create-app.template.typescript-library.beta/templates/AGENTS.md
nfdiop, line 1: As discussed, let's have this (maybe) in another version (and for all apps) if it still makes sense with the latest models

packages/create-app.template.typescript-library.beta/templates/_gitignore
nfdiop, line 1: nice touch!

examples/example-typescript-library-sdk-2.x/README.md
```diff
@@ -0,0 +1,11 @@
+# example-typescript-library
+
+This project was generated with [`@osdk/create-app`](https://www.npmjs.com/package/@osdk/create-app) from the `typescript-library` template. It is built against a locally generated SDK and a non-existent Foundry stack, so it is intended for reference purposes only.
+
+To quickly create your own version of this template run the following command and answer the prompts based on your Developer Console application:
+
+```
+npm create @osdk/app@latest -- --template typescript-library --sdkVersion 2.x
```
nfdiop: Should we drop  --sdkVersion 2.x since that is the only version?

examples/example-typescript-library-sdk-2.x/README.md
```diff
@@ -0,0 +1,11 @@
+# example-typescript-library
+
+This project was generated with [`@osdk/create-app`](https://www.npmjs.com/package/@osdk/create-app) from the `typescript-library` template. It is built against a locally generated SDK and a non-existent Foundry stack, so it is intended for reference purposes only.
+
+To quickly create your own version of this template run the following command and answer the prompts based on your Developer Console application:
```
nfdiop, line 5: Do we want to mention Developer Console application?

examples/example-typescript-library-sdk-2.x/.env.development
```diff
@@ -0,0 +1,34 @@
...
+# This URL is the Foundry host that your OSDK will use. It typically does not
+# need to be changed.
+
+FOUNDRY_API_URL=https://fake.palantirfoundry.com
+
+
+# This client ID must match the client ID given on the "OAuth & scopes" page of
+# Developer Console. It typically does not need to be changed.
+
+FOUNDRY_CLIENT_ID=123
```
nfdiop, line 27: I think we don't want these in the library?

packages/create-app.template.typescript-library.beta/templates/package.json.hbs
```diff
@@ -0,0 +1,32 @@
...
+    "check": "ultracite check",
+    "fix": "ultracite fix",
+    "format": "oxfmt .",
+    "lint": "oxlint"
+  },
+  "devDependencies": {
+    "oxfmt": "^0.52.0",
+    "oxlint": "^1.67.0",
+    "ultracite": "^7.8.0",
+    "vitest": "^4.1.7"
```
nfdiop, line 30: I assume there are plans to update the rest of the repo to use the same vitest version

nfdiop (commented):
Thanks for doing this! Super excited! Could we add `.beta` to the template to communicate wip?

packages/create-app.template.typescript-library.beta/templates/package.json.hbs
olliejenningsPT: Yep exactly

examples/example-typescript-library-sdk-2.x/README.md
olliejenningsPT: Good point, it might be done via the CLI in the future so maybe we don't want to, but l think its also good to direct users who might look at the templates

examples/example-typescript-library-sdk-2.x/README.md
olliejenningsPT: yeah we can

packages/create-app.template.typescript-library.beta/templates/AGENTS.md
olliejenningsPT: sure l can remove the agents file, it generally is there just to tell the AI to use ultracite instead of say oxlint or oxfmt since ultracite runs these

packages/create-app/codegen.mjs
olliejenningsPT: let me re-check again

olliejenningsPT:
> Thanks for doing this! Super excited! Could we add `.beta` to the template to communicate wip?

yeah am happy to, for the template id?

examples/example-typescript-library-sdk-2.x/README.md
nfdiop: So the reason I'm asking is that this template is Dev Console agnostic: it is also relevant to custom widgets for example. So I don't think we should keep the current wording

nfdiop:
> > Thanks for doing this! Super excited! Could we add `.beta` to the template to communicate wip?
> 
> yeah am happy to, for the template id?

Something like this to communicate that this may change https://github.com/palantir/osdk-ts/tree/main/packages/create-app.template.react.beta: Will defer to you

packages/create-app.template.typescript-library.beta/templates/package.json.hbs
nfdiop: Sounds good! I still think we can make the repo upgrades in a separate PR but deferring to you

packages/create-app.template.typescript-library.beta/templates/AGENTS.md
olliejenningsPT: i actually think we should keep this, since new tooling being introduced, and we want to make sure the agents understand this difference etc.

examples/example-typescript-library-sdk-2.x/.env.development
olliejenningsPT: these are added by the generators, so we can't remove

nfdiop: can it be at least a FLUP to remove them since we don't need them in library?

olliejenningsPT: Yes

nfdiop (approved):

PR #3489, opened 2026-06-11
Allow require implementation for iatc

anhhoang14022001:
We don't allow setting an interface action type constraint to be required. This PR enables an IATC to be required, but we will throw if an object type attempts to implement an interface type that has a required IATC because we don't have support for IATC implementation yet.

### DIFF
--- a/packages/maker/src/api/defineObject.ts
+++ b/packages/maker/src/api/defineObject.ts
@@ -194,21 +194,6 @@ export function defineObject(
   );
 
   objectDef.implementsInterfaces?.forEach(interfaceImpl => {
-    const requiredActionTypeConstraints =
-      (interfaceImpl.implements.actionTypeConstraints
-        ?? [])
-        .filter((actionTypeConstraint: any) =>
-          actionTypeConstraint.requireImplementation
-        );
-    invariant(
-      requiredActionTypeConstraints.length === 0,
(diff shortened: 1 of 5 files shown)


packages/maker/src/api/defineObject.ts
ethanai22, line 1: we should do this in convertObject since the action type constraints can be defined after the interface and object

packages/maker/src/conversion/toMarketplace/convertObject.ts
```diff
@@ -83,6 +83,19 @@ export function convertObject(
...
+  for (const impl of implementations) {
+    const requiredConstraints = (impl.implements.actionTypeConstraints ?? [])
+      .filter(constraint => constraint.requireImplementation);
+    invariant(
+      requiredConstraints.length === 0,
+      `Object "${objectType.apiName}" implements interface "${impl.implements.apiName}" which has required action type constraints: ${
+        requiredConstraints.map(constraint => constraint.metadata.apiName).join(
+          ", ",
+        )
+      }. Action type constraint implementation mappings are not yet supported in OAC.`,
```
ethanai22: can we add something like "Change the constraint to optional and manually apply later"

ethanai22 (dismissed):

ethanai22 (approved):

PR #3490, opened 2026-06-11
Upgrade TypeScript to 6.0.3

olliejenningsPT:
## Upgrade TypeScript `~5.5.4` → `~6.0.3`

Bumps the workspace TypeScript compiler to the latest stable (6.0.3) and makes
the small set of config/code adjustments TS 6 requires. Validated locally end
to end: **build, typecheck, test, check-api, and lint are all green.** Rebased
onto current `main`.

> [!NOTE]
> Draft for review. TypeScript 6.0 is explicitly a *transition* release toward
> the Go-based 7.0 — most of its breaking changes are new defaults and
> deprecations. Because this repo already pins `strict`/`target`/`module`/
> `moduleResolution`/`rootDir` explicitly, it sidesteps the bulk of 6.0's
> default-flip breakage.

### Validation

| Gate | Result |
|------|--------|
| `turbo build` | 290 / 290 ✓ |
| `turbo typecheck` | 204 / 204 ✓ |
| `turbo test` | 182 / 182 ✓ |
| `turbo check-api` | 136 / 136 ✓ (no API-report drift) |
| `turbo lint` | 211 / 211 ✓ |

## What changed

**Version bump**
- `LATEST_TYPESCRIPT_DEP` → `~6.0.3` in `.monorepolint.config.mjs`; `mrl --fix`
  propagated it to every package's `typescript` devDependency.

**Shared tsconfig (`packages/monorepo.tsconfig/base-4.9.json`)**
- `ignoreDeprecations: "6.0"` — neutralizes the `baseUrl` deprecation that
  **tsup's dts build emits unconditionally** (the only thing standing between
  this repo and a clean TS 6 build). See "tooling" below.
- `types: ["*"]` — restores pre-6.0 `@types` auto-discovery (TS 6 defaults
  `types` to `[]`, which otherwise drops Node globals / `node:*` builtins in
  node-tool packages).
- `noUncheckedSideEffectImports: false` — TS 6 turns this on by default and
  would flag CSS side-effect imports.

**Other config**
- Removed now-unnecessary `baseUrl` from 3 example/template tsconfigs (their
  `paths` are already relative, so `baseUrl` isn't needed and is deprecated).
- `ignoreDeprecations: "6.0"` on the 2 intentional `node10` test fixtures.
- `e2e.generated.1.1.x` (historically pinned to TS 4.9): `isolatedDeclarations:
  false`, since its vendored `generatedNoCheck` code predates the option.

**Code fixes for TS 6 strictness (each small, isolated, and commented)**
- `@osdk/api` `ShapeBuilder.ts` — explicit constructor type args on the
  `deriveLink` return (stricter generic-instantiation checking).
- `@osdk/aip-core` `streamStep.ts` — cast on `TextDecoderStream` (updated
  lib.dom stream variance with `Uint8Array<ArrayBufferLike>`).
- `@osdk/maker-experimental` — typed `resolvePath` wrapper for yargs `coerce`
  (TS 6 inferred `coerce: path.resolve` results as `unknown`).
- `@osdk/react-devtools` — non-generic forwarding cast + explicit `| undefined`
  on an optional ctor param (isolatedDeclarations).
- `@osdk/client` test util — `NonNullable` cast.
- `@osdk/generator` test — its in-memory compile check now ignores TS 6's
  diagnostic code `2307` ("cannot find module") in addition to the old `2792`.

**Tooling bumps (done in this PR)**
- `typescript-eslint` (+ parser + eslint-plugin) `8.43` → **`8.61.0`** — the
  first line with official TS 6 support is 8.58 (range `<6.1.0`); the lockfile
  resolves the current latest. Its newer recommended rules surfaced ~6
  pre-existing findings, all fixed: `no-useless-default-assignment`
  (`= undefined` → `?`, auto-fixed) and one `await-thenable` (lifted a
  synchronous `getRid()` out of a `Promise.all`).
- `@microsoft/api-extractor` `7.52` → **`7.58.8`** (bundled analysis engine
  5.8.2 → **5.9.3**) and `@microsoft/api-documenter` → `7.30.x`. Only effect on
  committed output: a one-line cosmetic change in `etc/api.report.api.md`
  (empty namespace export now rendered as `export {};`).

> The first push of this PR pinned the newest versions that satisfied the repo's
> 3-day `minimum-release-age` policy (8.60.1 / 7.58.7). That window has since
> cleared, so after rebasing the lockfile naturally resolves **8.61.0 / 7.58.8**.
> Functionally identical (8.61.0 supports TS 6; 7.58.8 still bundles the 5.9.3
> engine). See the "Reconcile TS 6 rebase with main" commit.

**Rebase reconciliation with `main`** — two packages new on `main` needed the
workspace TypeScript:
- `e2e.sandbox.officeassignment` → `~6.0.3` (was pinned `~5.5.4`).
- `shared.test.intellisense` → `~6.0.3` (its intellisense `tsserver` harness +
  typecheck now run on the workspace compiler; its tests still pass, no snapshot
  churn).

## What did NOT change

- **No runtime/behavior changes to published packages.** `check-api` is green;
  the only report delta is the cosmetic `export {};` rendering above.
- **Output is byte-identical under the new compiler.** As a spot check, the full
  `@osdk/api` `build/` tree (632 files) was built with both 5.5.4 and 6.0.3 and
  compared: every `.js`, `.d.ts`, and `.d.cts` is byte-identical. The only deltas
  were sourcemaps for `ShapeBuilder` — and only because this PR edits that file
  (type-only; line positions shifted). The repo's JS/ESM declarations come from
  esbuild + oxc (compiler-version-independent), and even the one tsc-produced
  artifact (tsup's `.d.cts`) came out identical.
- **No tooling upgrade was strictly *required* for TS 6** (verified): the prior
  `typescript-eslint@8.43` linted TS 6 fine, and `api-extractor@7.52` still
  produced stable reports. The bumps above are for official support and a newer
  analysis engine, not to unblock the build.

## What changes for template / create-app consumers

- **One shipped template file changed:** `create-app.template.react.beta`'s
  `templates/tsconfig.json` drops `baseUrl` (paths are relative; cleaner and
  TS-6-correct, harmless on older TS). No other shipped template's tsconfig had
  `baseUrl`.
- **Scaffolded apps' TypeScript version is unaffected** — the shipped
  `package.json.hbs` files don't pin a `typescript` version, so the `~6.0.3`
  bump stays internal to the monorepo (template packages' own dev deps + the
  committed example apps, which are regenerated to stay in sync — verified by
  the example-generator check).

## Follow-ups (not blocking, out of scope here)

- **api-extractor with a true TS 6 engine — gated on an upstream release.**
  No published api-extractor bundles TS 6.0 yet; the 7.58.x line (including
  7.58.8) tops out at the **5.9.3** engine. So 5.9.3 is the best available today.
  Revisit when a 6.0-bundling api-extractor ships — **not** on a fixed timer.
  Until then it harmlessly prints "newer than the bundled compiler" for 6.0.3.
- **Before TS 7.0:** TS 7 *removes* `baseUrl`/`node10` outright and
  `ignoreDeprecations` stops working. The `baseUrl` debt lives in tsup's dts
  build (`monorepo.tool.transpile`'s CJS path). Plan to migrate that path to
  [tsdown](https://tsdown.dev) (Rolldown/oxc, fits the repo's existing
  `isolatedDeclarations`) or wait for a tsup fix.


olliejenningsPT:
**Follow-up note — api-extractor TS 6 engine is gated on an upstream release, not a timer.**

This PR moves api-extractor to 7.58.7, whose bundled analysis engine is **TypeScript 5.9.3**. To be clear for whoever revisits this: **no released api-extractor bundles TS 6.0 yet** — even the (currently release-age-blocked) 7.58.8 still bundles 5.9.3. So bumping api-extractor again "in a few days" once 7.58.8 clears the 3-day `minimum-release-age` window will **not** yield a TS 6 engine.

The real follow-up is to bump api-extractor when a version that bundles TS 6.x is *published upstream* (and then clears the release-age window). Until then, 5.9.3 analyzes correctly (check-api green, reports stable) and only prints the cosmetic "newer than the bundled compiler" advisory for 6.0.3.

PR #3491, opened 2026-06-11
Export branchRid from generated SDK

zeyadkhaled:
Previously SDKs generated on a branch had the $branch present in the ontologyMetadata file but wasn't exported. This PR does export it to be consumed similar to $ontologyRid, renames to $branchRid to be consistent with $ontologyRid, and exports it unconditionally where the default value is set to undefined if the SDK was generated against a branch.

### DIFF
--- a/packages/generator/src/v2.0/generateMetadata.ts
+++ b/packages/generator/src/v2.0/generateMetadata.ts
@@ -40,7 +40,7 @@ export async function generateOntologyMetadataFile(
          * The RID of the Foundry branch this SDK was generated against, or
          * \`undefined\` if it was generated against the main branch.
          */
-        export const $branchRid: string | undefined = ${
+        export const $branch: string | undefined = ${
             ontology.raw.branch != null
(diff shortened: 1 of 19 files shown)


packages/generator/src/v2.0/generateMetadata.ts
```diff
@@ -36,13 +36,15 @@ export async function generateOntologyMetadataFile(
...
-      }
-       ${
-        ontology.raw.branch != null
-          ? `
-        export const $branch = "${ontology.raw.branch.rid}";
+        /**
+         * The RID of the Foundry branch this SDK was generated against, or
+         * \`undefined\` if it was generated against the main branch.
+         */
+        export const $branchRid: string | undefined = ${
```
ssanjay1: Can we keep it as `$branch` for now, in case someone was like pulling it internally. Also leaves it open for if this becomes liek a branchid or something

ssanjay1 (approved):

PR #3493, opened 2026-06-11
Sort decimal/long (incl. derived RDP) properties numerically in observable client orderBy

nfdiop:
## Summary

`long`/`decimal` properties (including derived/RDP ones) sorted lexicographically — `"10" < "2"` — when using `orderBy`, most visibly via `useOsdkObjects: true`. This makes the observable client's `orderBy` compare them numerically, resolving each property's declared type at sort time. This is the lower-risk alternative to #3477.

## Context

`long` and `decimal` properties arrive over the wire as strings, so the observable client's `orderBy` compared them lexicographically (`"10" < "2"`). Fixing this requires knowing a sorted property's declared type at comparison time. This applies to regular object properties, interface properties, and runtime-derived (RDP) properties.

## Decision

The comparator resolves the declared type at **sort time** from data that already exists:

- Regular object props: `holder[ObjectDefRef]` (already on every object holder).
- Interface props: `holder[InterfaceDefRef]` (already on every interface holder), with namespace re-qualification for namespaced interfaces.
- Derived props: a `DerivedPropertyRuntimeMetadata` map passed into `OrderBySortingStrategy` as a constructor argument, produced by `extractRdpDefinition` at the query layer. Both query paths that can carry derived properties thread it in: `ObjectSetQuery` (`observeObjectSet`) and `ListQuery` (`observeList` / `useOsdkObjects`). `SpecificLinkQuery` cannot carry derived properties (its `rdpConfig` is always `undefined`), so it passes none.

String-encoded numeric types (`long`, `decimal`) are compared by value via a BigInt-based `compareNumericStrings`. Among derived properties, only `min`/`max` aggregations need extra capture — they preserve the source `long`/`decimal` type. `count`/`exactDistinct`/`approximateDistinct` resolve to `integer` and `sum`/`avg`/`approximatePercentile` to `double`, all delivered as JS numbers, so they already sort numerically.

## Considered Options

We rejected the approach in #3477, which threaded an `RdpDefRef` symbol carrying the RDP type metadata through `createOsdkObject`/`createOsdkInterface` and the strip/filter/merge rebuild paths. That put new state on the hot object-construction path used by every wire-to-OSDK conversion, where a regression could corrupt objects far beyond sorting. Passing the metadata as a sort-time parameter keeps the construction path untouched.

## Consequences

- Adding `min`/`max` to `extractRdpDefinition`'s capture is construction-neutral: `modifyRdpProperties` only special-cases `count` and attachment/media/timeseries types, so a captured `long`/`decimal` type falls through and the raw string is returned unchanged.
- The type is resolved per comparison (`resolvePropertyType` is O(1)); no per-key caching is kept.
- Derived numeric **aggregations that are not `min`/`max`** rely on the SDK typing them as `integer`/`double`. If a future aggregation returns a string-encoded `long`/`decimal` without being captured, it will silently fall back to lexicographic order.

## Out of scope

Null placement is unchanged: the client-side sort keeps its pre-existing "nulls always last" behavior, which differs from Foundry's directional null ordering. That is orthogonal to numeric comparison, affects all types, and is left for a follow-up.


packages/client/src/observable/internal/sorting/SortingStrategy.ts
```diff
@@ -65,6 +68,8 @@ export class OrderBySortingStrategy implements SortingStrategy {
     private readonly orderBy: Canonical<
       Record<string, "asc" | "desc" | undefined>
     >,
+    public readonly derivedPropertyMetadata: DerivedPropertyRuntimeMetadata =
```
nfdiop, line 75: the main change with this approach is that we pass the derivedPropertyMetadata

packages/client/src/observable/internal/sorting/SortingStrategy.test.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    expect(sort({ value: "asc" }, holders).map((h) => h.value)).toEqual([
+      2,
+      9,
+      10,
+      100,
+    ]);
+  });
+
+  it("sorts nulls last regardless of direction", () => {
```
nfdiop, line 148: I realised the backend does the opposite: tracking to fix in a FLUP PR (if that makes sense)

packages/client/src/observable/internal/compareNumericStrings.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it("handles signs, including +, -, and -0", () => {
+    expect(compareNumericStrings("-1.5", "1.5")).toBe(-1);
+    expect(compareNumericStrings("-1.5", "-1.25")).toBe(-1);
+    expect(compareNumericStrings("+10", "9")).toBe(1);
+    expect(compareNumericStrings("-0.0", "0.0")).toBe(0);
+  });
+
+  it("falls back to a Number comparison for exotic forms (e.g. scientific)", () => {
```
nfdiop: I am not expecting these to come from the backend. Happy to support if they do

packages/client/src/observable/internal/compareNumericStrings.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Empty/whitespace values represent "no value" and sort as the smallest value:
+ * first when ascending, last when descending, matching Workshop's blank-cell
+ * ordering. (Note Number("") === 0, so they must NOT be compared as zero.)
+ *
+ * Any other non-numeric input (malformed data on a numeric column) sorts last
+ * and is never reported equal to a finite value, so the comparator stays a
+ * TOTAL order -- it's used as an Array.sort comparator, and a non-transitive
+ * NaN-vs-finite pair would corrupt the sort.
+ */
+export function compareNumericStrings(a: string, b: string): number {
```
nfdiop, line 36: As mentioned here, I debated this solution vs a library like `big.js`. Went with this approach because I assume it will handle the potential cases and we don't want to increase the bundle size of all apps (PS: I didn't test the bundle size impact in practice) https://github.com/palantir/osdk-ts/pull/3477#discussion_r3402128747

packages/client/src/observable/CONTEXT.md
nfdiop, line 1: I am using this skill https://github.com/mattpocock/skills/tree/main/skills/engineering/grill-with-docs and it adds Context that I find very helpful to understand the code. Happy to keep local

packages/e2e.sandbox.peopleapp/src/app/employees/EmployeesTable.tsx
nfdiop, line 1: Only added tests here. No example in Storybook since it relies on mocking that makes this useless

packages/client/src/observable/CONTEXT.md
KuberSethi: awesome, def think we should keep local but im a fan of that skill

packages/client/src/observable/internal/resolvePropertyType.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+  const selectedType = derivedPropertyMetadata[key]
+    ?.selectedOrCollectedPropertyType?.type;
+  return typeof selectedType === "string" ? selectedType : undefined;
+}
+
+/**
+ * Looks up an interface property's type by the key the interface holder exposes.
+ * The holder strips the namespace from a property whose namespace matches the
+ * interface's (createOsdkInterface), but `interfaceDef.properties` is keyed by
```
KuberSethi, line 87: are these comments still accurate now that we're not touching `createOsdkInterface`?

packages/client/src/observable/internal/compareNumericStrings.ts
```diff
@@ -0,0 +1,74 @@
+/*
```
KuberSethi, line 1: apologies if i missed this on the prior review but is there a library we can use for this since i doubt this is a novel problem only we're facing?

packages/client/src/observable/internal/sorting/SortingStrategy.test.ts
```diff
@@ -0,0 +1,296 @@
+/*
```
KuberSethi, line 1: Does this test need to be updated / scoped down now that we're taking an overall more scoped down approach?

packages/client/src/observable/internal/list/ListQuery.test.ts
```diff
@@ -56,6 +59,35 @@ function setupOntology(fauxFoundry: FauxFoundry) {
   fauxFoundry.getDefaultOntology().registerObjectType(
     stubData.todoWithLinkTypes,
   );
+  // Register objectTypeWithAllPropertyTypes with a self-referential ONE/MANY
+  // link pair (modeled on Employee's lead/peeps) so a derived property can
```
KuberSethi: are there other employee lead / peep mentions in this file

KuberSethi (commented):
Some high level comments (taking a deeper pass) but I think this looks largely good, thanks for making the changes!!

packages/client/src/observable/internal/resolvePropertyType.ts
nfdiop: yes because this is about resolving the interface type

packages/client/src/observable/internal/compareNumericStrings.ts
nfdiop: Replied here https://github.com/palantir/osdk-ts/pull/3493#discussion_r3403262862 but happy to try big.js

packages/client/src/observable/internal/sorting/SortingStrategy.test.ts
nfdiop: So most of it is around the long/decimal and I kept all tests to validate new behaviour/prevent regressions. Happy to scope down but I think it is good to have these

packages/client/src/observable/internal/list/ListQuery.test.ts
nfdiop: let me simplify the comment

packages/client/src/observable/internal/compareNumericStrings.ts
nfdiop: So I tried it and you can see that it does increase the bundle size of osdk/react for example so not sure it is worth it

**Now**
Screenshot 2026-06-12 at 21 11 04

**Before**
Screenshot 2026-06-12 at 21 11 37

packages/client/package.json
```diff
@@ -94,6 +94,7 @@
     "@osdk/shared.net.fetch": "workspace:~",
     "@types/geojson": "^7946.0.16",
     "@wry/trie": "^0.5.0",
+    "big.js": "^5.1.2",
```
KuberSethi, line 97: out of curiosity why this version? seeing lots of newer majors
https://www.npmjs.com/package/big.js?activeTab=versions

packages/client/package.json
```diff
@@ -118,6 +119,7 @@
     "@osdk/monorepo.api-extractor": "workspace:~",
     "@osdk/monorepo.tsconfig": "workspace:~",
     "@osdk/shared.test.intellisense": "workspace:~",
+    "@types/big.js": "^4.0.5",
```
KuberSethi, line 122: should this version align with the main `big.js` pkg version?

KuberSethi (approved):

PR #3494, opened 2026-06-11
Fix missing input borders under Storybook theme presets

nfdiop:
## What

The Storybook brand-theme-extractor's `"shadow"` role was wired to overwrite `--osdk-input-shadow`. Since "Workshop Light" is the active default preset for all stories, its flat elevation shadow replaced the composite box-shadow that draws the input's border ring — so inputs rendered with no border.

This removes `--osdk-input-shadow` from the elevation `"shadow"` role. The input border is themed correctly through the existing `"border"` (color) and `"border-width"` roles, which feed the composite token. The elevation-shadow control now only affects real surfaces (`--osdk-surface-shadow-2`).

## Why

`--osdk-input-shadow` isn't an elevation shadow — it's a composite that encodes the input's 1px border ring plus a faint inner shadow. Treating it as elevation erased the border under every preset.

Storybook-only change (`.storybook/` addon); the published `@osdk/react-components` package is unaffected.

## Before / after

Before: inputs show only a faint shadow, no border.
Screenshot 2026-06-11 at 21 06 57

After: inputs render their crisp 1px border under all presets.
Screenshot 2026-06-11 at 21 07 12


.changeset/green-things-hang.md
```diff
@@ -0,0 +1,2 @@
+---
```
KuberSethi: nit: we don't need an empty changeset

KuberSethi (dismissed):

PR #3495, opened 2026-06-11
add batched derived links store for simple links

KuberSethi:
list mode link store that batches simple pivot links into one observeLinks call

• single hop pivots with no filters resolve for the whole list in one request
• per source primary key snapshot and aggregated link status

### DIFF
--- a/packages/react/src/new/shapes/batchedDerivedLinksStore.ts
+++ b/packages/react/src/new/shapes/batchedDerivedLinksStore.ts
@@ -1,0 +1,405 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ObjectOrInterfaceDefinition, Osdk } from "@osdk/api";
+import type { ShapeDerivedLinkDef } from "@osdk/api/shapes-internal";
+import type {
+  LinkLoadConfig,
+  LinkStatus,
+  ShapeDefinition,
+} from "@osdk/api/unstable";
+import type { Client } from "@osdk/client";
+import { applyShapeTransformationsToArray } from "@osdk/client/unstable-do-not-use";
+import type {
+  ObservableClient,
+  Observer,
+  Unsubscribable,
+} from "@osdk/client/unstable-do-not-use";
+import {
+  createCachingNotifier,
+  createStoreSubscribe,
+  createVersionedCache,
+  wrapError,
+} from "../shared/storeUtils.js";
+import type { AnyShapeInstance } from "./types.js";
+import {
+  isBatchableLink,
+  NOOP_FETCH_MORE,
+  violationsToError,
+} from "./types.js";
+
+export interface BatchedDerivedLinksPayload<
+  S extends ShapeDefinition<ObjectOrInterfaceDefinition>,
+> {
+  linksBySourcePk: Map<string | number, Record<string, AnyShapeInstance[]>>;
+  linkStatusBySourcePk: Map<
+    string | number,
+    Partial<Record<string, LinkStatus>>
+  >;
+  aggregatedLinkStatus: Partial<Record<string, LinkStatus>>;
+  anyLoading: boolean;
+  anyError: boolean;
+}
+
+export interface BatchedDerivedLinksStore<
+  S extends ShapeDefinition<ObjectOrInterfaceDefinition>,
+> {
+  subscribe: (notify: () => void) => () => void;
+  getSnapshot: () => BatchedDerivedLinksPayload<S>;
+  updateSourceObjects: (
+    sourceObjects: Osdk.Instance<ObjectOrInterfaceDefinition>[],
+    transformedData: AnyShapeInstance[],
+  ) => void;
+  loadDeferred: (sourcePk: string | number, linkName: string) => void;
+  retry: (sourcePk?: string | number, linkName?: string) => void;
+  destroy: () => void;
+}
+
+interface TransformedSourceData {
+  data: AnyShapeInstance[];
+  error?: Error;
+}
+
+interface BatchableLinkState {
+  linkDef: ShapeDerivedLinkDef;
+  status: "init" | "loading" | "loaded" | "error" | "deferred";
+  error?: Error;
+  subscription?: Unsubscribable;
+  transformedBySourcePk: Map<string | number, TransformedSourceData>;
+  hasMore: boolean;
+  fetchMore: () => Promise<void>;
+}
+
+export function createEmptyBatchedDerivedLinksStore<
+  S extends ShapeDefinition<ObjectOrInterfaceDefinition>,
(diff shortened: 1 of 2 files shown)


KuberSethi:

- #3495 ◀
    - #3496
        - #3497
            - #3498
                - #3499
                    - #3500
                        - #3501
                            - #3502

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,405 @@
...
+          notifySubscribers();
+        },
+        complete: () => {},
+      },
+    );
+  }
+
+  function updateSourceObjects(
+    sourceObjects: Osdk.Instance<ObjectOrInterfaceDefinition>[],
+    transformedData: AnyShapeInstance[],
```
olliejenningsPT, line 308: `transformedData` is accepted here (and declared on the interface up at line 64) but it's never read anywhere in the store. Was it meant to feed into the snapshot, or is it a leftover from an earlier shape of this code? As written, a caller computes and passes the transformed source objects and we just drop them on the floor. I'd either wire it in or remove it from the signature so the contract isn't misleading.

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,405 @@
...
+  ): void {
+    if (destroyed) {
+      return;
+    }
+    currentSourceObjects = sourceObjects;
+
+    for (const state of linkStates.values()) {
+      if (state.status !== "deferred") {
```
olliejenningsPT: This restarts every link's `observeLinks` subscription on every call: `startObserveLinks` always tears down the existing subscription and re-subscribes, flipping status back to `loading`. There's no check on whether `sourceObjects` actually differs from `currentSourceObjects`. If the consumer calls this per render or per upstream emission, we'll churn subscriptions and flash a loading state even when the source set is identical. The per-object `createDerivedLinksStore` starts observation once and reuses the live subscription. Could we diff the pk set before restarting, or at least document that callers must only call this on a genuine source change? (The local `dedupeInterval` option is declared but never passed, so there's no dedupe safety net either.)

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,405 @@
...
+        linksBySourcePk,
+        linkStatusBySourcePk,
+        aggregatedLinkStatus,
+        anyLoading,
+        anyError,
+      };
+    });
+  }
+
+  const subscribe = createStoreSubscribe(subscribers, () => {}, cleanupAll);
```
olliejenningsPT: `onInit` is a no-op and `onCleanup` is `cleanupAll`, which only tears down subscriptions; it leaves `destroyed` false and keeps the link states and cached data around. So after the last unsubscribe the store goes dormant but not dead, and a later re-subscribe won't restart anything because `onInit` does nothing. The per-object store makes last-unsubscribe a one-way `destroy()`, which forces a fresh store on remount. The batched version is only safe if the consumer guarantees a new store per mount and always calls `updateSourceObjects` after subscribing. Can we make that contract explicit, or have `onInit` re-arm from `currentSourceObjects`?

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,405 @@
...
+
+  function loadDeferred(
```
olliejenningsPT, line 424: `loadDeferred` (and `retry` just below) take a `sourcePk` that's never used; both operate across all source objects. That's reasonable for batched semantics since one `observeLinks` covers every source, but the param is misleading: passing a specific pk still loads/retries for everyone. I'd drop the param or add a note that batched load/retry is necessarily all-or-nothing per link.

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
KuberSethi: good call, added a pk-set check up front so we skip the teardown + restart when the source pks haven't actually changed. I _think_ that's what kills the loading flash so we shouldn't end up needing dedupeInterval for it

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
KuberSethi: Good call out, fixed! `onInit` now re-arms from the retained source objects on re-subscribe, so we should properly restart observation. cleanupAll still tears down on the last unsubscribe so we should be good on that front

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
KuberSethi: This is used in the next PR in the stack!

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
KuberSethi: This is used in the next PR!

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,472 @@
...
+      return false;
+    }
+  }
+  return true;
+}
+
+export function createBatchedDerivedLinksStore<
+  S extends ShapeDefinition<ObjectOrInterfaceDefinition>,
+>(
+  shape: S,
```
olliejenningsPT, line 169: Heads up: `shape` and `client` are both accepted by the constructor but neither is referenced in the body. The per-object store needs them (client for buildObjectSetFromLinkDefByType, shape for __baseType / __derivedLinks), but the batched path goes through observeLinks and takes the links via allLinks, so these two are unused here. Are they wired up later in the stack like transformedData and sourcePk, or are they just mirroring createDerivedLinksStore's signature? If it's for symmetry a one-line comment would save the next reader a grep, and if neither they could come out. noUnusedParameters is not on in the base tsconfig so the compiler will not flag it.

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,472 @@
...
+        anyError,
+      };
+    });
+  }
+
+  // The first subscribe arrives after updateSourceObjects has already started
+  // observation. A later re-subscribe (remount / StrictMode) finds the
+  // subscriptions torn down by cleanupAll on the previous last-unsubscribe, so
+  // re-arm from the retained source objects instead of going dormant forever.
+  function onResubscribe(): void {
```
olliejenningsPT, line 411: The `next` and `error` callbacks both bail when `destroyed` is set, but the start path (onResubscribe -> startAllNonDeferred -> startObserveLinks) does not. onInit runs on the first subscriber, so if subscribe ever fires after destroy() we would create a live observeLinks subscription whose callbacks all no-op but which never gets torn down, i.e. a leak. In practice the owning hook should not subscribe after destroy, so this is defensive rather than a live bug, but a single `if (destroyed) return;` at the top of onResubscribe would close it cheaply and match the robustness of the observer callbacks.

packages/react/src/new/shapes/batchedDerivedLinksStore.ts
```diff
@@ -0,0 +1,472 @@
...
+      return;
+    }
+
+    state.status = "loading";
+    state.error = undefined;
+
+    state.subscription = observeLinksUntyped(
+      sourceObjects,
+      linkApiName,
+      {
```
olliejenningsPT, line 261: We only forward `pageSize` from the link config here. createDerivedLinksStore also honors autoFetchMore and streamUpdates. I am assuming observeLinks just does not support those for batched links, which lines up with isBatchableLink rejecting where/orderBy/limit, in which case no behavior change is needed. A one-line comment would help though, so nobody later wonders why a configured streamUpdates or autoFetchMore silently does nothing on a link that happens to be classified batchable. Related: the dedupeInterval field on the local alias is never passed, which matches your note above that the pk-set check makes it unnecessary, so it is just dead in the type now.

PR #3496, opened 2026-06-11
add per item fallback for filtered derived links

KuberSethi:
links with where, orderBy, limit, distinct, or set ops fall back to per item stores

• reconciles per item stores by primary key as the source list changes
• merges per item link data and status into the batched snapshot


KuberSethi:

- #3495
    - #3496 ◀
        - #3497
            - #3498
                - #3499
                    - #3500
                        - #3501
                            - #3502

PR #3497, opened 2026-06-11
add shape projection hooks

KuberSethi:
internal hooks powering the shape option for single objects and lists

• useShapeSingle minimizes fields and merges derived link data for one object
• useShapeList routes derived links through the batched store with per item status


KuberSethi:

- #3495
    - #3496
        - #3497 ◀
            - #3498
                - #3499
                    - #3500
                        - #3501
                            - #3502

PR #3498, opened 2026-06-11
add mock observable client test utility

KuberSethi:
a configurable mock observable client for testing hooks and stores

• drives object, list, and link observers with scripted payloads
• exported from the testing entrypoint

incorporates resolved review feedback from #2837.

PR #3500, opened 2026-06-11
wire shape option into useOsdkObjects

KuberSethi:
adds the shape option to the list hook and makes inline configs reactive

• dispatches to a dedicated useOsdkObjectsWithShape function guarded by a mode lock
• inline configs now re-derive on structural changes instead of freezing at mount

same __shapeId caveat as the single object hook: a config differing only in a transforms or links.via callback hashes identically and will not re-derive. structural changes are covered, callback-identity-only changes are not.

incorporates resolved review feedback from #2837.

### DIFF
--- a/.changeset/shapes-wire-objects.md
+++ b/.changeset/shapes-wire-objects.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react": patch
+---
+
+add shape option to useOsdkObjects and react to inline config changes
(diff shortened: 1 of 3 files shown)


KuberSethi:

- #3495
    - #3496
        - #3497
            - #3498
                - #3499
                    - #3500 ◀
                        - #3501
                            - #3502

PR #3505, opened 2026-06-12
Respect required-ness of individual struct fields in action parameters

### LINKED ISSUE #2358: `applyAction` with a struct parameter types does not respect required-ness of individual struct properties
### Repro

Given some action, `myAction`, that takes in a struct parameter where one or more fields are not required:
```
{
  firstName: string,
  lastName: string | undefined,
  email: string
}
```

The following snippet should be valid:
```typescript
  client(myAction).applyAction({
    "parameter" :{
      firstName:"",
      email :"",
    }
  })
```

instead, it throws:

`Property 'lastName' is missing in type '{ email: string; firstName: string; }' but required in type 'StructType<{ lastName: "string"; email: "string"; firstName: "string"; }>`

### Problem

This is because nullability is not considered for types within a struct:
https://github.com/palantir/osdk-ts/blob/9651f6738d2082b1219e952db908f7bcc29b2df0/packages/generator-converters/src/wireActionTypeV2ToSdkActionMetadata.ts#L98-L116

and is not stored in the `ActionDefinition` parameter struct type:
https://github.com/palantir/osdk-ts/blob/9651f6738d2082b1219e952db908f7bcc29b2df0/packages/api/src/ontology/ActionDefinition.ts#L93-L98

joshbouganim:
Fixes #2358

## Problem

`applyAction` with a struct parameter does not respect the required-ness of individual struct fields. The wire metadata (`OntologyStructField`) carries a per-field `required: boolean`, but `wireActionTypeV2ToSdkActionMetadata` drops it and `ActionMetadata.DataType.Struct` / `ActionParam.StructType` have no slot to store it, so every struct field is generated as required and non-nullable. Callers cannot omit optional struct fields at the type level, even though the `/apply` endpoint accepts absent optional struct keys (and the OSDK serializer already drops `undefined` values on the wire).

## Change

- `ActionMetadata.DataType.Struct` accepts a new `StructField` object form (`{ type, nullable }`) alongside the existing bare-string form.
- `ActionParam.StructType` maps nullable fields to optional properties (`field?: T | null`), mirroring how top-level optional parameters are generated (`key?` plus `| null`).
- `generator-converters` emits the object form **only** for `required: false` fields, so SDKs generated from all-required structs are byte-for-byte unchanged, and previously generated SDKs continue to typecheck against the widened constraints.
- `FauxFoundry` `validateAction` no longer rejects absent optional struct fields. It previously ran the type-match against `undefined`, so an optional-and-absent field was always `INVALID`; this matched the old (since-fixed) server behavior, not the current one.

Generated output for a struct with an optional field now looks like:

```ts
export interface Params {
  readonly address?: ActionParam.StructType<{
    city: 'string';
    state: 'string';
    streetName: { type: 'string'; nullable: true };
  }> | null;
}
```

and the resolved type of that struct is `{ city: string; state: string; streetName?: string | null }`.

## Backwards compatibility

- Required fields keep the bare-string encoding, so regenerating an ontology with no optional struct fields produces no diff.
- The new constraints on `DataType.Struct` and `StructType` are supersets of the old ones, so existing generated code compiles unchanged against the updated `@osdk/api`.
- The quickinfo snapshot confirms the hover display for all-required structs is unchanged (`{ name: string; age: number; joined: string }`).

The union encoding (bare string for required, `{ type, nullable }` for optional) is what preserves that compatibility. If a breaking change were acceptable, every field could use the uniform object form instead, which would drop the bare-string/object branch in three spots: the converter's per-field ternary, the `StructFieldType` union, and the `StructFieldValue` resolver. I went with the union form deliberately to keep this a non-breaking change. Note that it only tidies the *encoding* — the `StructType` mapped type still has to partition fields into required vs optional output properties (hence the two-half intersection + `FlattenIntersection`), which is inherent to the feature and unaffected by the encoding choice. Happy to switch to the uniform form if maintainers would rather take the break.

## Testing

- New converter unit test asserting required fields stay bare strings and optional fields gain `{ type, nullable: true }`.
- `createStructPerson` test fixture gained an optional `streetName` field; client tests assert the inferred param type accepts the field present, `null`, or omitted, rejects omission of required fields, and round-trips through the faux server.
- Quickinfo snapshot probe for the new encoding (`nickname?: string | null | undefined`).
- `turbo test` green across `@osdk/api`, `@osdk/generator-converters`, `@osdk/generator`, `@osdk/faux`, `@osdk/client`, `@osdk/shared.test`, `@osdk/foundry-sdk-generator`; full-repo `transpileTypes` green.

## Context

We hit this downstream after the Foundry server-side fix for absent optional struct keys rolled out (previously the `/apply` validator 400'd on absent optional struct fields, tracked separately with Palantir support). With the server accepting omitted fields, this SDK type gap is the last thing forcing consumers to cast around `StructType` at every struct-param call site.


palantirtech:
Thanks for your interest in [palantir/osdk-ts](https://github.com/palantir/osdk-ts), @joshbouganim! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #3506, opened 2026-06-12
Fix RDP cell value retains old value when cleared

ziyunp:
## Summary

Fixes the same class of bug as #3433 (function/derived ObjectTable cells not clearing when their value becomes null), but for **RDP (runtime-derived-property) columns**. The root cause is in the `@osdk/client` observable store.

## The problem
In mergeObjectFields, the old value is assigned to the RDP field when the field is missing (cleared) in the new data. 

The original code does not differentiate these when RDP field is missing from the new data:
1. When the RDP is missing because it was set to null.
2. When the RDP is missing from the data because the query does not include any RDP fields.

Repro-ed in https://github.com/palantir/osdk-ts/pull/3520

## Is there a workaround?
No, I tried doing observableClient.invalidateAll() and the old value still stay. Invalidation is already done on action completion - when setting an RDP field from null to some value, the cell reloads and shows new data.

## The fix
Get the query intent from sourceCacheKey to identify whether the query asked for the RDPs to decide if we should clear the RDP or keep the existing value. If the query contains the RDP field key, but the new data does not have the RDP field, then it means the RDP field is cleared.

**Implementation**
Pass the set of RDP field keys from sourceCacheKey (contains the intent) to mergeObjectFields as the sourceRdpFields and targetRdpFields. This means that the source data should contain the same set of RDP fields as the target cache we are writing to.

FLUP: To refactor to make this code more readable. Added comments for now to help understanding through the logic

### DIFF
--- a/packages/client/src/observable/internal/utils/rdpFieldOperations.ts
+++ b/packages/client/src/observable/internal/utils/rdpFieldOperations.ts
@@ -170,12 +170,12 @@ export function mergeObjectFields(
   targetRdpFields: ReadonlySet<string>,
   targetCurrentValue: ObjectHolder | undefined,
 ): ObjectHolder {
-  // Case: The destination (target) query does not need RDP
+  // When the destination (target) query does not need RDP
   if (targetRdpFields.size === 0) {
     return stripRdpFields(sourceValue, sourceRdpFields);
   }
 
-  // Case: Source contains more than the required RDP fields
+  // When source contains more than the required RDP fields
   if (isSuperset(sourceRdpFields, targetRdpFields)) {
     if (sourceRdpFields.size === targetRdpFields.size) {
       return sourceValue;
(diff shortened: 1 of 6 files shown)


packages/react-components/src/object-table/hooks/__tests__/useObjectTableData.test.tsx
ziyunp, line 1: Verified that the bug does not come from the Object Table, when a field is null, the data for that field is undefined.

packages/client/src/observable/internal/utils/rdpFieldOperations.ts
```diff
@@ -205,18 +217,13 @@ export function mergeObjectFields(
...
     const targetUnderlying =
       targetCurrentValue[UnderlyingOsdkObject] as SimpleOsdkProperties;
     for (const field of targetRdpFields) {
       if (field in targetUnderlying) {
-        // Preserve target's value when:
-        // 1. Source doesn't have this RDP field at all, OR
-        // 2. Source hasn't provided the value (undefined)
-        if (
-          !sourceRdpFields.has(field)
-          || newProps[field] === undefined
```
ziyunp, line 218: Removed the additional condition, this is a bug as it will override the field if it is a queried RDP field, but the field is cleared.

PR #3521, opened 2026-06-15
Implement client-side concurrency limit

esinx:
Adds a client-side concurrency limit so the OSDK client caps the number of HTTP requests it keeps in flight at once. Excess requests queue and run as slots free up, preventing the client from overwhelming the gateway under bursty load. **Does not expose configurable option to user end**

Defaults to 10 concurrent requests. Pass `Infinity` to disable throttling.

# Changes

- `@osdk/shared.net.fetch`: new `createLimitedFetch(fetchFn, maxConcurrent)` helper (built on p-limit) that wraps a fetch function with a concurrency gate.
  - Exports `createLimitedFetch` and `DEFAULT_MAX_CONCURRENT_REQUESTS = 10`.
  - Returns the original fn unchanged when `maxConcurrent === Infinity`. (For future-- if we ever want to make this configurable)
  - Throws on non-positive/non-integer values.

# Notes
- Tests added: `createLimitedFetch.test.ts` and concurrency coverage in `createClient.test.ts`.

### DIFF
--- a/etc/client.report.api.md
+++ b/etc/client.report.api.md
@@ -120,13 +120,20 @@ export const createClient: (baseUrl: string, ontologyRid: string | Promise<strin
     	logger?: Logger
     	UNSTABLE_DO_NOT_USE_BRANCH?: string
     	headers?: Record<string, string>
+    	concurrency?: {
+        		maxConcurrentRequests?: number
+        	}
 } | undefined, fetchFn?: typeof fetch | undefined) => Client;
 
 // @public
 export function createObjectSpecifierFromPrimaryKey<Q extends ObjectTypeDefinition>(objectDef: Q, primaryKey: PrimaryKeyType<Q>): ObjectSpecifier<Q>;
 
 // @public
-export function createPlatformClient(baseUrl: string, tokenProvider: () => Promise<string>, options?: undefined, fetchFn?: typeof globalThis.fetch): PlatformClient;
+export function createPlatformClient(baseUrl: string, tokenProvider: () => Promise<string>, options?: {
+    	concurrency?: {
+        		maxConcurrentRequests?: number
+        	}
+}, fetchFn?: typeof globalThis.fetch): PlatformClient;
 
 export { DerivedProperty }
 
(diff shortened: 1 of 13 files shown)


etc/client.report.api.md
```diff
@@ -120,13 +120,20 @@ export const createClient: (baseUrl: string, ontologyRid: string | Promise<strin
     	logger?: Logger
     	UNSTABLE_DO_NOT_USE_BRANCH?: string
     	headers?: Record<string, string>
+    	concurrency?: {
```
bryantpark04: We generally wouldn't expose this in the API surface unless there's a need for it

packages/shared.net.fetch/package.json
```diff
@@ -40,7 +40,8 @@
   },
   "dependencies": {
     "@osdk/shared.net.errors": "workspace:~",
-    "fetch-retry": "^6.0.0"
+    "fetch-retry": "^6.0.0",
+    "p-limit": "^3.1.0"
```
bryantpark04, line 44: new dep but it's already pulled in by something else judging from the diff in pnpm-lock, at least within this workspace. We should check if it's already there when a user installs `@osdk/client` `@osdk/api`

etc/client.report.api.md
esinx: Resolved: reverted option to be not configurable, just defaults to 10

packages/shared.net.fetch/package.json
esinx: From running `pnpm why p-limit` it seems like `p-limit` wasn't being installed because of those two packages, it's mainly due to `find-up` which is our dev dependency.

etc/client.report.api.md
nihalbhatnagar: I feel like if we ship this we should give users a way to quickly opt out tbh

packages/shared.net.fetch/src/createLimitedFetch.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import pLimit from "p-limit";
+
+const DEFAULT_MAX_CONCURRENT_REQUESTS = 10;
```
nihalbhatnagar, line 19: How'd we come up with this number

PR #3526, opened 2026-06-16
make the dev-mode action delay configurable

KuberSethi:
the observable client forced a 1s dev-mode delay on every action to make optimistic updates visible, which hurt local devx, especially for function-backed actions where optimistic updates don't apply

• skip the dev-mode delay when an action has no optimistic update, so function-backed actions stay fast
• add `devMode.actionDelayMs` to tune or disable it via the OsdkProvider prop and the createObservableClient option (0 disables, default 1000)
• log a one-time message explaining the delay and how to turn it off

### DIFF
--- a/.changeset/dev-mode-action-delay.md
+++ b/.changeset/dev-mode-action-delay.md
@@ -1,0 +1,6 @@
+---
+"@osdk/client": minor
+"@osdk/react": minor
+---
+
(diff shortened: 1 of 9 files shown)


KuberSethi:

- #3526 ◀
    - #3532
        - #3533

olliejenningsPT (approved):

PR #3532, opened 2026-06-16
add dev-mode observable client debug and loglevel config knobs

KuberSethi:
adds three dev-only configuration knobs to the observable client for debugging local behavior

• devMode.logLevel raises the client log level (default error) so its debug tracing is visible, cascading to every child logger
• devMode.debug.refCounts and devMode.debug.cacheKeys turn the hardcoded debug-flag module constants into per-client toggles for cache-internals logging
• all stay stripped from production builds and are exposed through createObservableClient and the OsdkProvider devMode prop

PR #3533, opened 2026-06-16
add client-wide default dedupe interval and page size config knobs

KuberSethi:
adds client-wide default options so teams don't have to thread the same value through every hook

• defaultDedupeInterval sets the revalidation dedupe window for subscriptions that don't specify their own
• defaultPageSize sets the list page size for subscriptions that don't specify their own
• exposed through createObservableClient options and OsdkProvider props, with per-call options still taking precedence


KuberSethi:

- #3532
    - #3533 ◀

PR #3534, opened 2026-06-17
Fix object type for pivot

Deuteu:
The fecthOne on link is failing with `PropertiesNotFound` error as it is at some point expecting the base object type instead of the target object type

```
const searchResult = await group.$link.members.fetchOneWithErrors(user.userId);
```

`{"errorName":"PropertiesNotFound", "errorInstanceId":"09ec0b4b-b141-4357-9882-4db15e74ec65", "parameters":{"objectType":"User","properties":["groupId"]}}`


packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -1336,4 +1338,31 @@ describe("ObjectSet", () => {
...
+    beforeAll(() => {
+      fauxFoundry.getDefaultDataStore().registerLink(
+        stubData.employee1,
+        "officeLink",
+        stubData.nycOffice,
+        "occupants",
+      );
+    });
+
+    // Office's primary key is `officeId` (string); Employee's is `employeeId` (integer).
```
nihalbhatnagar: Remove comment

packages/client/src/objectSet/ObjectSet.test.ts
```diff
@@ -359,31 +361,31 @@ describe("ObjectSet", () => {
...
     const state = player.address?.state;
-    expectTypeOf<typeof state>().toMatchTypeOf<
+    expectTypeOf<typeof state>().toExtend<
       string | undefined
     >;
     expect(state).toEqual("TN");
 
     const zipCode = player.address?.zipCode;
-    expectTypeOf<typeof zipCode>().toMatchTypeOf<
+    expectTypeOf<typeof zipCode>().toExtend<
```
nihalbhatnagar, line 388: Dont think these are needed

packages/client/src/objectSet/createObjectSet.ts
```diff
@@ -414,8 +415,14 @@ async function createWithPk(
   objectSet: WireObjectSet,
   primaryKey: PrimaryKeyType<ObjectTypeDefinition>,
 ) {
+  // `objectType` is the source carried by `createObjectSet`'s closure
```
nihalbhatnagar: Can we remove this comment this is already a pattern in the repo

packages/client/src/objectSet/ObjectSet.test.ts
Deuteu: The switch from `toMatchTypeOf` to `toExtend` or the test itself ?
The linter was complaining about `toMatchTypeOf` and requesting it to be updated to `toExtend`.
And the tests, I would prefer not to remove them in this PR as this is not related.

nihalbhatnagar (approved):

PR #3540, opened 2026-06-18
Make empty $and/$or/$not operands a compile-time error

tompersona:
https://github.com/palantir/osdk-ts/pull/3348 added a runtime check that throws on an empty `.where({})` but it caught more than intended: an empty `{}` tucked inside an `$and`/`$or`/`$not` array (think `cond ? {...} : {}`) now blows up at runtime, even though it compiled fine and used to work. Turns out an empty operand lowers to an empty AND on the wire, which some storage backends reject outright, so it was never really valid anyway, just silently tolerated on some paths.

This change makes that a compile-time error instead. Empty operands of `$and`/`$or`/`$not` are now rejected by the type system (via a new `NonEmptyWhereClause`), so the bug shows up while you're writing it rather than a surprise at runtime. Top-level `.where({})` still works as a no-op, and the runtime check stays as a backstop with a clearer message for JS callers.

### DIFF
--- a/.changeset/spicy-rivers-cheer.md
+++ b/.changeset/spicy-rivers-cheer.md
@@ -1,0 +1,6 @@
+---
+"@osdk/api": minor
+"@osdk/client": minor
+---
+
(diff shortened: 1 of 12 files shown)


tompersona:
Doing [Revert 3348](https://github.com/palantir/osdk-ts/pull/3544#top) instead.

PR #3541, opened 2026-06-18
fix rdp derived field clearing in the observable cache merge

KuberSethi:
runtime-derived-property columns kept showing a stale value after the derived value became null, because the cache merge resolved derived fields by object-schema membership and wire presence instead of the query's rdp set

• resolve derived fields by the query rdp set so the source owns what it computed and an omitted field clears
• drop the same-key wire-presence heuristic so a full refetch that omits a now-null derived value clears it
• stop dropping shared derived fields and select+derived values when propagating to sibling cache keys

this is the focused root-cause fix split out below the broader merge rearchitecture in #3523

### DIFF
--- a/packages/client/src/observable/internal/object/ObjectsHelper.ts
+++ b/packages/client/src/observable/internal/object/ObjectsHelper.ts
@@ -152,52 +152,28 @@ export class ObjectsHelper extends AbstractHelper<
 
     let valueToWrite = !dataChanged && existing ? existing.value : value;
 
-    // When a $select-filtered fetch returns partial objects, merge with
-    // existing cached data to preserve fields not in the select set.
+    const sourceRdpFields = this.store.objectCacheKeyRegistry.getRdpFieldSet(
+      sourceCacheKey,
+    );
+
+    // A $select fetch carries only the base props it selected, so merge it with
+    // the cached value. A full write owns every field and is stored as-is, so an
+    // omitted derived value clears instead of being refilled from stale cache.
     const existingHolder = existing?.value;
-    const canMergeSelectFields = dataChanged
+    if (
+      dataChanged
+      && valueToWrite !== tombstone
       && selectFields
       && selectFields.size > 0
       && existingHolder
-      && this.isObjectHolder(existingHolder);
-
-    if (canMergeSelectFields && valueToWrite !== tombstone) {
+      && this.isObjectHolder(existingHolder)
+    ) {
       valueToWrite = mergeSelectFields(
         valueToWrite,
         selectFields,
         existingHolder,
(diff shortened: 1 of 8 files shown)


KuberSethi:

- #3541 ◀
    - #3542

packages/client/src/observable/internal/object/ObjectsHelper.ts
```diff
@@ -152,54 +152,30 @@ export class ObjectsHelper extends AbstractHelper<
...
     let valueToWrite = !dataChanged && existing ? existing.value : value;
 
-    // When a $select-filtered fetch returns partial objects, merge with
-    // existing cached data to preserve fields not in the select set.
+    const sourceRdpFields = this.store.objectCacheKeyRegistry.getRdpFieldSet(
+      sourceCacheKey,
+    );
+
+    // A $select fetch carries only the base props it selected, so merge it with
+    // the cached value. A full write owns every field and is stored as-is, so an
```
olliejenningsPT: Flagging the load-bearing assumption behind removing the old same-key RDP preservation block here.

Real-world, the main writer that hits this path with `selectFields` undefined is the subscription handler: `ListQuery.onOswChange` (and `BaseListQuery.onOswChange`) calls `storeOsdkInstances([object], batch, this.rdpConfig, undefined, ...)`, which writes the streamed object straight to the RDP cache key as a "full write". With the merge gone, `valueToWrite` is the raw streamed object, so any derived field the payload doesn't carry gets cleared from cache.

That's correct only if the OSW stream actually delivers the `withProperties`-derived values (and omits a derived key only when it's genuinely null). If the stream drops derived fields wholesale, every subscription tick on an RDP-backed list will wipe the derived columns until the next refetch, which is exactly what the deleted block was guarding against (its comment cited subscription updates specifically).

There's also an asymmetry that points the same way: in the sibling path, `mergeForTarget` -> `mergeObjectFields` still preserves a target's derived value when the source didn't compute it, so a sibling RDP key keeps the value while the directly-written list key clears it. Two cache keys for the same pk would then disagree.

Could you confirm OSW updates include the derived values? If yes, worth a short comment here since it's load-bearing. If no, the same-key preservation needs to stay for the subscription path, and the "clears a derived value when a same-key refetch omits it" test would be encoding the wrong behavior.

olliejenningsPT (approved):

packages/client/src/observable/internal/object/ObjectsHelper.ts
KuberSethi: This isn't really a problem since we disable streamUpdates with withProperties usage (we don't support this behavior in the BE) however yeah its probably good to be defensive.

PR #3542, opened 2026-06-18
rearchitect rdp cache merge

KuberSethi:
builds on #3541. folds base-property and derived-field reconciliation into one rule in reconcileObject, so the cache merge has a single code path instead of separate select and rdp merges

• replace mergeRdpFields with reconcileObject: the source owns the base props it loaded and the derived fields it computed, an omitted owned field clears, otherwise the target's cached value is kept
• track loadedBaseFields so a full load clears the base props it omits while a partial load keeps the props it did not load
• resolve base props by schema and derived fields by the rdp sets in one pass

### DIFF
--- a/.changeset/rdp-merge-reconcile.md
+++ b/.changeset/rdp-merge-reconcile.md
@@ -1,0 +1,5 @@
+---
+"@osdk/client": patch
+---
+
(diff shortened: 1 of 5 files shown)


KuberSethi:

- #3542 ◀

PR #3543, opened 2026-06-22
fix: style number input stepper as buttons

Leolide:
## Summary
- Restyle number input stepper (increment/decrement) to look like proper buttons instead of an input field extension
- Use `--osdk-button-secondary-bg` background and a real CSS `border` on the stepper container so the border isn't covered by child hover backgrounds
- Extract new `--osdk-button-border-color` and `--osdk-button-drop-shadow-color` tokens from the raw `color-mix()` values previously inlined in `--osdk-button-shadow`
- Use `--osdk-iconography-color-muted` for stepper chevron icons instead of `--osdk-input-color`
- Match the button divider color to the container border color for consistency

## Before

https://github.com/user-attachments/assets/f05399ea-58e8-4c49-b67a-99b0bae2e8f9

## After

https://github.com/user-attachments/assets/ff0d6955-81d0-4a32-826b-871eadc941a1

## Test plan
- [ ] Open Storybook BaseForm story and verify the number input stepper buttons have a visible button-like border and background
- [ ] Hover each button and confirm the background darkens without the border disappearing or doubling
- [ ] Verify the 1px divider between increment/decrement buttons matches the outer border color
- [ ] Check dark mode appearance

### DIFF
--- a/packages/react-components/src/action-form/fields/NumberInputField.module.css
+++ b/packages/react-components/src/action-form/fields/NumberInputField.module.css
@@ -104,7 +104,7 @@
   padding: 0 calc(var(--osdk-surface-spacing) * 1.5);
   border: none;
   background: transparent;
-  color: var(--osdk-input-color);
+  color: var(--osdk-iconography-color-muted);
   cursor: pointer;
   line-height: 0;
   transition: background-color var(--osdk-input-transition-duration)
--- a/packages/react-components/src/action-form/fields/NumberInputField.module.css
+++ b/packages/react-components/src/action-form/fields/NumberInputField.module.css
@@ -29,7 +29,21 @@
 }
 
 .osdkNumberInputWrapper[aria-invalid] .osdkNumberInputStepper {
-  box-shadow: var(--osdk-input-shadow-error);
+  border-color: var(--osdk-button-border-color);
+  box-shadow: 0 1px 1px var(--osdk-button-drop-shadow-color);
+  background-color: var(--osdk-button-danger-bg);
+}
+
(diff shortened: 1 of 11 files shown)


nfdiop (commented):
Thanks for fixing! 
Two questions
- should we keep the border on hover?
- do we want the icon color to match the color of the select icon?

Image

packages/react-components/src/action-form/fields/NumberInputField.module.css
```diff
@@ -89,10 +89,10 @@
 .osdkNumberInputStepper {
   display: flex;
   flex-direction: column;
-  border: none;
+  border: 1px solid var(--osdk-button-border-color);
```
nfdiop: nit: maybe don't harcode 1px

Leolide:
> Thanks for fixing! Two questions
> 
> * should we keep the border on hover?
> * do we want the icon color to match the color of the select icon?

I think we should keep the border on hover as that is blueprint behaviour 

seems blueprint change the background color and the icon color for error / warning 

Screenshot 2026-06-22 at 1 50 48 PM

nfdiop (commented):
Thanks! One last ask: we currently copy the tokens also inside `react-component-styles` until we are GA

turbo.json
```diff
@@ -3,6 +3,7 @@
   "ui": "stream",
   "tasks": {
     "codegen": {
+      "dependsOn": ["^transpileEsm"],
```
nfdiop: Could we revert this? Have you tried merging main?

Leolide: reverted

nfdiop (approved):

PR #3545, opened 2026-06-22
add signMediaReferences param to useOsdkObjects and fetchPage

lucy-k-wang:
pre-work: https://github.palantir.build/foundry/api-gateway/pull/15705

## Changes in this PR
Expose `signMediaReferences` on the list and search object paths so callers can request signed `MediaReference.token` values. These paths currently return `token` as `null`, so media operations on list results fail for callers without direct view permission on the media set. Requires platform SDK 2.64.0 for the `referenceSigningOptions` request field.

Single-object reads (getObject, i.e. useOsdkObject / observeObject / fetchOne) already return signed MediaReference tokens unconditionally on the server, so $signMediaReferences is intentionally not surfaced there.

### DIFF
--- a/.changeset/sign-media-references-osdk.md
+++ b/.changeset/sign-media-references-osdk.md
@@ -4,4 +4,4 @@
 "@osdk/react": patch
 ---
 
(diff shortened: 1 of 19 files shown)


lucy-k-wang:
https://github.palantir.build/foundry/api-gateway/pull/16181

PR #3557, opened 2026-06-23
add enabled option to useOsdkAggregation

KuberSethi:
brings useOsdkAggregation in line with the other data hooks by letting callers conditionally defer the aggregation

• new `enabled` option (defaults to `true`); when `false` the hook skips the subscription and does not fetch from the server
• disabled hooks report `data: undefined`, `isLoading: false`, `error: undefined`, matching useOsdkObjects and the other hooks
• flipping `enabled` from `false` to `true` re-subscribes and fetches

### DIFF
--- a/packages/react/src/new/__tests__/useOsdkAggregation.test.tsx
+++ b/packages/react/src/new/__tests__/useOsdkAggregation.test.tsx
@@ -29,18 +29,20 @@ const AGGREGATE = { $select: { $count: "unordered" } } as const;
 function createMockObservableClient(): {
   client: ObservableClient;
   observeAggregation: ReturnType<typeof vi.fn>;
+  invalidateObjectType: ReturnType<typeof vi.fn>;
 } {
   const observeAggregation = vi.fn(
     (_args: unknown, _observer: Observer<unknown>) => ({
       unsubscribe: vi.fn(),
     }),
   );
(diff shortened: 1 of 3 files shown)


packages/react/src/new/__tests__/useOsdkAggregation.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+    it("calls observeAggregation by default when enabled is omitted", () => {
+      renderHook(
+        () => useOsdkAggregation(Employee, { aggregate: AGGREGATE }),
+        { wrapper: createWrapper(observableClient) },
+      );
+
+      expect(observeAggregation).toHaveBeenCalledTimes(1);
+    });
+
+    it("starts observing when enabled flips from false to true", () => {
```
nfdiop, line 110: could we also test from true to false?

nfdiop (commented):
Thanks for handling the issue promptly!

nfdiop (dismissed):

packages/react/src/new/__tests__/useOsdkAggregation.test.tsx
KuberSethi: done

nfdiop (dismissed):

nfdiop (approved):

PR #3560, opened 2026-06-24
CipherText codegen / decrypt

esinx:
## Summary

Adds end-to-end support for `cipherText` ontology properties: they now flow through codegen and are exposed on OSDK objects as a `CipherText` value whose plaintext can be retrieved on demand via `.decrypt()`.

Previously the generator-converters explicitly dropped `cipherText` properties (logging `"... is not a supported dataType"` and returning `undefined`), so they never appeared on generated objects.

## Changes

**API surface (`@osdk/api`)**
- New `CipherText` interface (`packages/api/src/object/CipherText.ts`): an opaque handle exposing `decrypt(): Promise<string>`.
- Registered `cipherText` in `BaseWirePropertyTypes` and mapped it to `CipherText` across `PropertyValueWireToClient` / `ClientToWire` / `WireToCreate`.

**Codegen (`@osdk/generator-converters`)**
- `wirePropertyV2ToSdkPropertyDefinition` now emits a real property definition for `cipherText` instead of skipping it (in both the dataType and objectPropertyType switches).

**Client (`@osdk/client`)**
- `CipherTextPropertyImpl` (`createCipherTextProperty.ts`) holds the `(objectApiName, primaryKey, propertyName)` triplet and calls the `CipherTextProperty.decrypt` ontologies endpoint, resolving to the plaintext.
- `createOsdkObject` treats `cipherText` as a special property type and instantiates the impl lazily.

**Faux / test infra (`@osdk/faux`, `@osdk/shared.test`)**
- New `CipherTextProperties.decrypt` mock handler returning a deterministic `decrypted:<objectType>:<pk>:<propertyName>` string.
- Added a `cipherText` property to the all-property-types stub object/objectTypeV2.

## Testing

- New `packages/client/src/object/ciphertext.test.ts` verifies the generated property is typed `CipherText | undefined` and that `.decrypt()` round-trips through the ontology endpoint.
- e2e sandbox runner (`runCipherTextTest.ts`) and generated `CipherTextTest` object added to the catchall fixtures.

## Usage

```typescript
const result = await client(CipherTextObject).fetchOne("pk-001");
const plaintext = await result.encryptedField?.decrypt();
```

Note that if the cipher text property key does not contain a value it will be omitted from the raw object itself; check for `Object.prototype.hasOwnProperty.call(result, "encryptedField")` for empty value checks

### DIFF
--- a/packages/api/src/object/CipherText.ts
+++ b/packages/api/src/object/CipherText.ts
@@ -15,7 +15,7 @@
  */
 
 /**
- * The client-side representation of a `cipherText` property. The encrypted
+ * Representation of a `cipherText` property. The encrypted
  * value is not stored; call {@link CipherText.decrypt} to retrieve the
  * decrypted plaintext on backend.
  *
--- a/packages/api/src/object/CipherText.ts
+++ b/packages/api/src/object/CipherText.ts
@@ -31,5 +31,5 @@ export interface CipherText {
    * Decrypts the ciphertext and resolves to the plaintext value.
    * @returns the decrypted plaintext
    */
-  decrypt(): Promise<string>;
+  decrypt(): Promise<string | undefined>;
 }
--- a/packages/api/src/object/CipherText.ts
+++ b/packages/api/src/object/CipherText.ts
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The client-side representation of a `cipherText` property. The encrypted
+ * value is held opaquely; call {@link CipherText.decrypt} to retrieve the
+ * decrypted plaintext.
+ *
+ * @example
+ * ```ts
+ * const employee = await client(Employee).fetchOne(12345);
+ * const ssn = await employee.encryptedSsn?.decrypt();
+ * ```
+ * @public
+ */
+export interface CipherText {
+  /**
+   * Decrypts the ciphertext and resolves to the plaintext value.
+   * @returns the decrypted plaintext
+   */
+  decrypt(): Promise<string>;
+}
--- a/packages/api/src/object/CipherText.ts
+++ b/packages/api/src/object/CipherText.ts
@@ -16,8 +16,8 @@
 
 /**
  * The client-side representation of a `cipherText` property. The encrypted
- * value is held opaquely; call {@link CipherText.decrypt} to retrieve the
- * decrypted plaintext.
+ * value is not stored; call {@link CipherText.decrypt} to retrieve the
+ * decrypted plaintext on backend.
  *
  * @example
  * ```ts
--- a/packages/client/src/createCipherTextProperty.ts
+++ b/packages/client/src/createCipherTextProperty.ts
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { CipherText } from "@osdk/api";
+import { decrypt as cipherTextDecrypt } from "@osdk/foundry.ontologies/CipherTextProperty";
+import invariant from "tiny-invariant";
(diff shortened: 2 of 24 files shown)


packages/client/src/object/ciphertext.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  it("decrypts the property value through the ontology endpoint", async () => {
+    const { data: [object1] } = await client(objectTypeWithAllPropertyTypes)
+      .where({ id: stubData.objectWithAllPropertyTypes1.id }).fetchPage();
+
+    const plaintext = await object1.cipherText?.decrypt();
+
+    // See packages/faux/src/handlers/createCipherTextHandlers.ts
+    expect(plaintext).toBe(
+      "decrypted:objectTypeWithAllPropertyTypes:1:cipherText",
```
esinx, line 48: Since we don't actually want to encrypt / decrypt in faux, the mock implementation just returns a deterministic map from the object type, primary key and property name.

packages/client/src/createCipherTextProperty.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.#triplet = [objectApiName, primaryKey, propertyName];
+  }
+
+  async decrypt(): Promise<string> {
+    const [objectApiName, primaryKey, propertyName] = this.#triplet;
+    const ontologyRid = await this.#client.ontologyRid;
+    const result = await cipherTextDecrypt(
+      this.#client,
+      ontologyRid,
+      objectApiName,
```
ssanjay1: nit: if you make the PK type `any` above should be able to just spread the triplet

packages/client/src/createCipherTextProperty.ts
```diff
@@ -0,0 +1,53 @@
...
+      primaryKey as string,
+      propertyName,
+    );
+    invariant(
+      typeof result.plaintext !== "undefined",
```
ssanjay1: What if the cipherText value is undefined/null, what does the backend return?

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -744,8 +744,6 @@ export class OntologyMetadataResolver {
       case "null":
       case "void":
         return Result.ok({});
-      case "mediaReference":
```
ssanjay1, line 747: Why this change?

packages/client/src/createCipherTextProperty.ts
esinx: The object doesn't contain the key at all. Might be more appropriate to make the check to be:
```typescript
result.hasOwnProperty("plaintext")
```

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
esinx: Was a quick patch for accidental duplicate case from a previous commit; there are duplicate cases to mediaReference and typeReference in lines 739 and 743 that gets resolved first (returns `Result.ok({})`). Removed for clarity.

packages/api/src/object/CipherText.ts
```diff
@@ -14,20 +14,22 @@
...
 
-import { type ConjureContext, conjureFetch } from "conjure-lite";
-import type {
-  BlockId as _api_BlockId,
-  OwnedBlockMetadata as _api_OwnedBlockMetadata,
-} from "../__components.js";
-
 /**
- * Get the metadata for a block.
+ * The client-side representation of a `cipherText` property. The encrypted
```
nihalbhatnagar: Comment doesnt make sense, shouldnt talk about "client side". Can we make this more clear?

packages/client/src/object/ciphertext.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      testSetup.apiServer.close();
+    };
+  });
+
+  it("generates a CipherText-typed property", async () => {
+    const { data: [object1] } = await client(objectTypeWithAllPropertyTypes)
+      .where({ id: stubData.objectWithAllPropertyTypes1.id }).fetchPage();
+
+    expectTypeOf(object1.cipherText).toEqualTypeOf<CipherText | undefined>();
+    expect(object1.cipherText).toBeDefined();
```
nihalbhatnagar: This check is useless since were checking it right below

packages/client/src/createCipherTextProperty.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor(args: {
+    client: MinimalClient;
+    objectApiName: string;
+    primaryKey: any;
+    propertyName: string;
+  }) {
+    const { client, objectApiName, primaryKey, propertyName } = args;
+    this.#client = client;
+    this.#triplet = [objectApiName, primaryKey, propertyName];
```
nihalbhatnagar: I know this is prior pattern but I dont like that we called this triplet in those other places can we name it something better here

packages/client/src/object/ciphertext.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    expect(typeof object1.cipherText?.decrypt).toBe("function");
+  });
+
+  it("decrypts the property value through the ontology endpoint", async () => {
```
nihalbhatnagar: We can consolidate into the test above

packages/e2e.sandbox.catchall/src/client.ts
```diff
@@ -73,6 +73,14 @@ export const mjClient: Client = createClient(
   loggingFetch,
 );
 
+export const eaOntologyClient: Client = createClient(
```
nihalbhatnagar: lets not call ea ontology since its public, maybe like "cipherTextOntology" for rn

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
nihalbhatnagar: Just for future reference always helpful to comment weird diffs (or drive-by changes) like this before sending out for review

nihalbhatnagar (approved):
Approving with flup to remove the | undefined

PR #3562, opened 2026-06-25
Added datasets and mediaSets to ScopeResources in @osdk/functions

oleh308:
datasets and mediasets are new fields that would allow functions to declare these types resources as part of the scope

```tsx

    scope: {
      resources: { datasets: ["myDataset"], mediasets: ["myMediaSet"], streams: ["myStreamingDataset"]},
      authorization: { read: ["marking1"], write: ["marking2"] },
    },
  };
  export function myFunction(): void {}

```

### DIFF
--- a/packages/functions/src/scope/types.ts
+++ b/packages/functions/src/scope/types.ts
@@ -36,7 +36,7 @@ export interface ScopeResources {
   links?: string[];
   models?: string[];
   datasets?: string[];
-  mediaSets?: string[];
+  mediasets?: string[];
   streams?: string[];
 }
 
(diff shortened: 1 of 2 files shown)


packages/functions/src/scope/types.ts
```diff
@@ -27,14 +27,17 @@ import type {
  * - A string alias from resources.json (e.g. `"myObject"`)
  * - An OSDK type reference imported from your generated ontology SDK (e.g. `Employee`)
  *
- * For `links`, and `models`, only string aliases are supported.
+ * For `links`, `datasets`, `mediaSets` and `models`, only string aliases are supported.
```
nfdiop: nit: should we also add `streams` ?

nfdiop (dismissed):

nfdiop (approved):

ssanjay1 (approved):

PR #3567, opened 2026-06-25
Add non-null assertion to CipherText

esinx:
Added the following line to `decrypt` on CipherText to always return `Promise<string>` instead of `Promise<string | undefined>`.

- Backend does not return null-plaintext result when the request is successful.
  - This is currently inconsistent with generated types but this PR is made to prevent a breaking change in the future.

```typescript
invariant(
      Object.hasOwn(result, "plaintext"),
      "Expected decryption result to have plaintext value",
    );
```

### DIFF
--- a/packages/client/src/createCipherTextProperty.ts
+++ b/packages/client/src/createCipherTextProperty.ts
@@ -16,6 +16,7 @@
 
 import type { CipherText } from "@osdk/api";
 import { decrypt as cipherTextDecrypt } from "@osdk/foundry.ontologies/CipherTextProperty";
+import invariant from "tiny-invariant";
 import type { MinimalClient } from "./MinimalClientContext.js";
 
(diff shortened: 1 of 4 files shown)


packages/client/src/createCipherTextProperty.ts
```diff
@@ -33,13 +34,17 @@ export class CipherTextPropertyImpl implements CipherText {
...
       this.#client,
       ontologyRid,
       ...this.#locator,
     );
-    return result.plaintext;
+    invariant(
+      Object.hasOwn(result, "plaintext"),
+      "Expected decryption result to have plaintext value",
+    );
+    return result.plaintext!;
```
esinx, line 48: Non-null assertion guaranteed by `invariant` call above

nihalbhatnagar (approved):

PR #3568, opened 2026-06-26
chore: migrate @osdk/react + react-devtools + react-components-storybook to oxc toolchain (oxc increment 6; React family)

olliejenningsPT:
Batch 6 of the #3031 oxc migration — React family (@osdk/react, @osdk/react-devtools, @osdk/react-components-storybook), the rung before @osdk/react-components (batch 7).

## What this does
Moves the three React-family packages (minus the @osdk/react-components giant, batch 7) from ESLint + dprint onto oxlint + oxfmt (via Ultracite), following the established per-package archetype + OXC_PACKAGES + dprint-exclusion mechanics:

- **@osdk/react** → new `oxc migrated react libraries with docs` archetype (react, oxc, extraPublishFiles, customTsconfigExcludes).
- **@osdk/react-devtools** → new `oxc migrated esm react libraries with css` archetype (react, oxc, ESM-only output, cssExport). styles.css build confirmed.
- **@osdk/react-components-storybook** → new `oxc migrated minimal packages` archetype (private, minimalChangesOnly, oxc).

The bulk of the diff is behavior-neutral oxfmt reformatting.

## Storybook specifics
- Instead of bloating the root config with storybook-only carve-outs, storybook carries a **nested `oxlint.config.ts` that `extends` the root** and holds its private-package rule relaxations. The root header jsPlugin path was absolutized so it survives `extends`.
- **eslint-plugin-storybook is kept, not dropped.** oxlint loads it through its ESLint-compatible JS plugin API (`jsPlugins`, the same mechanism this repo already uses for the license-header plugin), so the entire `flat/recommended` storybook ruleset keeps running under oxlint — verified including the filesystem-reading `no-uninstalled-addons` and the scope-analyzing `await-interactions`. The 11 rules are enabled in the nested config with the plugin's own severities and file-scoping (`*.stories.*` → story rules; `.storybook/main.*` → `no-uninstalled-addons`). Only the package's `eslint.config.mjs` is deleted; its rules move into the oxlint config.
- Its prior eslint config set `header/header` off, so its files keep the **palantir-header rule disabled** rather than mass-adding Apache headers.

## Override triage
New Ultracite-preset findings were moved to off/warn matching prior ESLint severity, not hand-rewritten. Notably, several **autofixes were disabled because they change behavior or break the build** (the repo targets ES2022 and ships isomorphic code): `no-array-sort`/`no-array-reverse` (→ ES2023 immutable copies), `prefer-dom-node-dataset` (dataset is HTMLElement-only), `prefer-code-point` (returns number|undefined), `no-useless-promise-resolve-reject` (unwrapped a load-bearing Promise.resolve around a thenable), `no-typeof-undefined` (`typeof process === 'undefined'` → `process === undefined` throws in the browser; broke OsdkProvider's __DEV__ guard in the Storybook interaction tests), and `prefer-at` (`arr[arr.length-1]` → `arr.at(-1)` returns `T | undefined`, breaking typechecking at non-optional use sites). oxfmt also ignores .json/.jsonc/.mdx (dprint's domain; .mdx would corrupt Storybook JSX).

## Verification
- `monorepolint check` → 0
- lint / typecheck / transpile / test / build for all three → 0 errors (react-devtools styles.css builds; Storybook interaction tests pass in chromium)
- @osdk/cbac-components re-lint → no regression
- whole-repo `turbo check` + `turbo transpile` → green, with one **out-of-scope, deferred** item below

## Known deferred item (not addressed here)
`@osdk/examples.typescript-library-sdk-2.x#check` is red on main independently of this PR (it does not gate CI; batch 5 merged with it present). Root cause: that scaffolded example runs its own `ultracite check`, and Ultracite's oxfmt formats `package.json`, which conflicts with monorepolint's package.json management. The example is generated from the published `create-app.template.typescript-library` template, so the real fix is a **shipped-template config change** (exclude package.json/tsconfig.json from the template's oxfmt) plus regenerating the example — that belongs to the template-config work in **batch 13**, not the React-family migration. Tracked under #3031.


packages/react-components-storybook/.storybook/crypto-polyfill.ts
```diff
@@ -19,7 +19,7 @@ export const randomUUID = (): string => {
     return crypto.randomUUID();
   }
   // Fallback for older browsers
-  return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, (c) => {
+  return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replaceAll(/[xy]/g, (c) => {
```
nfdiop, line 22: is this expected?

olliejenningsPT: yes, intend as new modern standards push for switching from `replace` -> `replaceAll`.

The industry standard is also to switch `sorted` -> `toSorted` (splice and others as well) but these are ES2023 and since we are targeting Node18 we disable these.

nfdiop: Yes aware of `sorted` to `toSorted` since this allows to create a new array vs modifying an existing one. 

But here there is a difference in behaviour between `replace` and `replaceAll`. Is it ok to replace all here?

olliejenningsPT: 100% fine here, no actual behaviour change if you see the replace has `/g` which is identical

nfdiop: I recommend we keep `replace` then or use `replaceAll` but drop the `g`.

nfdiop (approved):

PR #3573, opened 2026-06-26
Add null parameter type for struct action params

ssanjay1:
Previously, you could not pass in null for struct action params. This disallows people from clearing values in struct fields, we have now enabled that in the types.

### DIFF
--- a/etc/api.report.api.md
+++ b/etc/api.report.api.md
@@ -155,7 +155,7 @@ export namespace ActionParam {
             	} ? R extends true ? K : never : never]? : T[K] extends {
             		type: infer U
             		nullable: infer R
-            	} ? U extends keyof DataValueClientToWire ? R extends true ? DataValueClientToWire[U] | undefined : never : never : never };
+            	} ? U extends keyof DataValueClientToWire ? R extends true ? DataValueClientToWire[U] | null : never : never : never };
 }
 
 // @public (undocumented)
--- a/packages/client/src/actions/actions.test.ts
+++ b/packages/client/src/actions/actions.test.ts
@@ -211,7 +211,7 @@ describe.each([
     type InferredParamType = Parameters<typeof clientBoundActionTakesStruct>[0];
     expectTypeOf<{
       name: string;
-      address: { city: string; state: string; zipcode?: number | undefined };
+      address: { city: string; state: string; zipcode?: number | null };
     }>().toMatchTypeOf<InferredParamType>();
 
     const result = void (await client(createStructPerson).applyAction({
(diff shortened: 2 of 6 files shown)


packages/client/src/actions/actions.test.ts
```diff
@@ -206,19 +206,32 @@ describe.each([
...
         name: string;
-        address: { city: string; state: string; zipcode?: number | undefined };
+        address: { city: string; state: string; zipcode?: number | null };
       }
     >()
       .toMatchTypeOf<
         InferredParamType
       >();
 
-    const result = void await client(createStructPerson).applyAction({
```
ssanjay1: Had accidentally made this void in my last PR, reversed that change

packages/client/src/actions/actions.test.ts
```diff
@@ -206,19 +206,32 @@ describe.each([
...
+    const result = await client(createStructPerson).applyAction({
       name: "testMan",
       address: { city: "NYC", state: "NY", zipcode: 12345 },
     });
     expectTypeOf<typeof result>().toEqualTypeOf<undefined>();
     expect(result).toBeUndefined();
+
+    // Nullable struct fields accept null (to clear the value), undefined, or omission.
```
KuberSethi, line 224: nit: we say we're testing undefined here but there's no test for that unless I'm misunderstanding this code?

etc/api.report.api.md
```diff
@@ -155,7 +155,7 @@ export namespace ActionParam {
             	} ? R extends true ? K : never : never]? : T[K] extends {
             		type: infer U
             		nullable: infer R
-            	} ? U extends keyof DataValueClientToWire ? R extends true ? DataValueClientToWire[U] | undefined : never : never : never };
+            	} ? U extends keyof DataValueClientToWire ? R extends true ? DataValueClientToWire[U] | null : never : never : never };
```
KuberSethi, line 158: Is this technically a break / should this make the PR a minor at the least? Or am I missing context and this isn't used anywhere?

joshbouganim:
added context: follow up for #2364 and this comment: https://github.com/palantir/osdk-ts/pull/2364#issuecomment-4720974398

etc/api.report.api.md
ssanjay1: So shouldn't `really` manifest as a user facing break since we already generated these properties as optional as well e.g. `myStructField?: type | undefined` so users will still be able to leave it out or pass in undefined. If we want to be extra cautious can expand to `type | undefined | null` but seemed unnecessary

KuberSethi (approved):

PR #3574, opened 2026-06-26
remove action metadata calls when applying

ssanjay1:
We no longer need to pull the latest action metadata when parsing parameters, so removed it from our code paths

### DIFF
--- a/.changeset/cold-jars-repair.md
+++ b/.changeset/cold-jars-repair.md
@@ -1,0 +1,5 @@
+---
+"@osdk/client": patch
+---
+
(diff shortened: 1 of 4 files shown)


KuberSethi (dismissed):
looks good, only callout is that errors might manifest slightly differently i think? like if `await client.ontologyProvider.getActionDefinition` threw before users would see error handling from that vs now they don't but I think that's fine since we should be handling things fine in the cases of invalid / nonexistent actions regardless.

KuberSethi (approved):

PR #3579, opened 2026-06-29
chore: migrate @osdk/client.unstable + client.unstable.tpsa + client to oxc toolchain (oxc increments 8 & 9)

olliejenningsPT:
Batches 8 and 9 of the #3031 oxc migration (ESLint + dprint → oxlint + oxfmt via Ultracite), combined into one PR. Three packages: @osdk/client.unstable, @osdk/client.unstable.tpsa, @osdk/client.

## Generated code is linted + formatted (parity with the old toolchain)

@osdk/client.unstable and @osdk/client.unstable.tpsa are ~99% conjure-generated code under `src/generated/`. Ultracite's preset ignores `**/generated` by default, but this repo deliberately treats `src/generated/` as **first-class checked-in source** — only `src/generatedNoCheck{,2}/` are excluded, and ESLint linted + dprint formatted `src/generated/` before the migration. To preserve that coverage:

- The root `oxfmt.config.ts` no longer ignores `**/generated` (the other generated-globs stay), so oxfmt formats `src/generated/` wherever it runs.
- Each package carries a nested oxlint config that re-includes the generated tree and turns off the only two error-level rules it surfaces (`oxc/no-barrel-file`, `import/newline-after-import`). `oxlint --fix` is verified a no-op, so the generated diff is **pure oxfmt** (net −2.2k lines — oxfmt is more compact than dprint on the generated type definitions).
- Both packages live in one `oxc migrated force-bundle libraries with carve-outs` archetype (the former `forceBundle` archetype is removed).

## @osdk/client (the flagship SDK, ~300 hand-written files)

Reformatted with oxfmt. The ~50 error-level Ultracite rules its existing source first surfaces are turned off in `packages/client/oxlint.config.ts` (each a rule prior ESLint did not enforce at error level), so `oxlint --fix` makes no source rewrites and the diff is pure oxfmt. Two deliberate, behavior-preserving source touch-ups:
- `createObjectSet.ts`: hoist the `aggregate<Q, any>` instantiation expression to a local, because `(aggregate<Q, any>).bind(...)` (valid TS) trips oxfmt's parser.
- `intellisense.test.ts`: bump the hard-coded fixture line numbers by 1 to track the blank line oxfmt inserts between import groups.

The `etc/client.report.api.md` update is formatting-only (oxfmt drops redundant parens in two type constraints; the API is unchanged).

## Verification

- `monorepolint check` clean; per-package `lint typecheck transpile build test` (and `check-api` for client) green across all three packages; the reformatted generated trees typecheck and build.
- `oxlint --fix` is a confirmed no-op on every package → source diffs are pure oxfmt.
- Whole-repo `turbo transpile` → 250/250; whole-repo `turbo check --continue` → 644/645, the sole failure being the known pre-existing `examples.typescript-library-sdk-2.x#check` (reproduced with this branch stashed).
- Re-linted @osdk/cbac-components → no regression.

Rebased onto current `main` (batches 6 & 7 merged).


.changeset/oxc-client-increment9.md
nfdiop, line 1: can we have a single changeset? there is also .changeset/oxc-client-unstable.md

nfdiop (approved):
Approving. Assuming `oxlint --fix` is a no-op

PR #3581, opened 2026-06-29
chore: migrate @osdk/maker family to oxc toolchain (oxc increment 10)

olliejenningsPT:
Tenth increment of the #3031 oxc migration: the maker family (@osdk/maker, @osdk/maker-experimental, @osdk/maker-import).

Moves the three non-React maker libraries off ESLint + dprint and onto oxlint + oxfmt (configured via Ultracite), continuing the incremental migration.

## What changed
- Moved the three packages out of the `standardLibraries` monorepolint archetype into a new `oxc migrated libraries with carve-outs` archetype (`{ ...LIBRARY_RULES, oxc: true, oxcConfig: "./oxlint.config.ts" }`). `OUTPUT_NORMAL` transpile wiring is unchanged; only `lint`/`fix-lint` flip to oxlint/oxfmt.
- Added each package to `OXC_NESTED_CONFIG_PACKAGES` in `.lintstagedrc.mjs` and to the dprint exclude list in `dprint.json`.
- Added a nested `oxlint.config.ts` per package that `extends` the root config and turns off the error-level Ultracite rules each package's existing source first surfaces. Every disabled rule was not enforced at error level by the prior ESLint config, so the migration stays behavior-preserving.
- Reformatted each package with oxfmt.

## Second commit: format markdown with oxfmt
Before the migration, dprint formatted markdown (via its markdown plugin). Migrated packages are excluded from the dprint path, so their `.md` files were no longer formatted by anything (the shared oxfmt config ignored `**/*.md`). This commit removes that ignore so oxfmt formats markdown for migrated packages too, restoring parity. `.mdx` stays ignored (oxfmt would reflow the embedded JSX and break Storybook's indexer). Because the oxfmt config is shared, this reformats the `.md` files it surfaces across the already-migrated packages as well (client, react, react-components, cbac-components) — pure formatting (structure + embedded code fences), no prose changes.

## Behavior-preserving
`oxlint --fix` applies zero source rewrites once the carve-out rules are off, so the source diff is pure oxfmt. Verified `lint`, `typecheck`, `transpile`, `build`, and `test` are green for all three packages, whole-repo `lint` and `transpile` are green, and a prior migrated package (`@osdk/cbac-components`) still lints clean. `monorepolint check` reports 0. Whole-repo `turbo check` is green except the pre-existing `@osdk/examples.typescript-library-sdk-2.x#check` base failure (unrelated, reconfirmed across prior increments).

## Coverage parity
No lint coverage is lost: the maker packages have no generated directories, so every `.ts` source ESLint linted is still linted by oxlint, and no ESLint error-level rule was disabled (every disabled rule was a new Ultracite rule ESLint did not enforce, or enforced only as a warning). `package.json`/`tsconfig.json`/`vitest.config.mts` remain monorepolint-enforced. The linter is not type-aware (as with every migrated package), but `tsc` still runs in `typecheck`.


.changeset/oxc-maker-family.md
```diff
@@ -0,0 +1,7 @@
+---
+"@osdk/maker": patch
+"@osdk/maker-experimental": patch
+"@osdk/maker-import": patch
+---
+
+Migrate @osdk/maker, @osdk/maker-experimental, and @osdk/maker-import to lint with oxlint and format with oxfmt (configured via Ultracite) instead of ESLint and dprint (tenth increment of the repo-wide oxc migration; the maker family). Each package is reformatted with oxfmt, with its newly-surfaced error-level lint rules turned off in a nested oxlint config to keep the change behavior-preserving.
```
nfdiop: nit: I wouldn't add "tenth increment of the repo-wide oxc migration; the maker family" since this is user-facing

.monorepolint.config.mjs
```diff
@@ -247,6 +244,24 @@ const archetypeRules = archetypes(
       oxc: true,
     },
   )
+  // Same as "oxc migrated libraries" but with a nested oxlint config (oxcConfig)
+  // holding behavior-preserving carve-outs. The maker family is hand-written and
+  // first surfaces a set of error-level Ultracite rules the prior ESLint config
+  // did not enforce; each package's oxlint.config.ts `extends` the root and turns
+  // them off so the migration is a reformat, not a rewrite.
+  .addArchetype(
```
nfdiop, line 257: can we add a `TODO` to remove these archetypes later since we don't need to store the concept of migration?

nfdiop (commented):
The PR is touching more than the maker family. Could we address?

olliejenningsPT:
> The PR is touching more than the maker family. Could we address?

Its almost just makeing sure the markdown files of other packages are formatted aswell

.changeset/oxc-maker-family.md
olliejenningsPT: Done in 640cd89 — dropped the "tenth increment of the repo-wide oxc migration; the maker family" phrasing so the changeset reads as a plain user-facing note.

.monorepolint.config.mjs
olliejenningsPT: Added a TODO in 640cd89 above the "oxc migrated ..." archetypes noting they're transitional and should be collapsed back into the base library archetypes (dropping the oxc:true/oxcConfig plumbing) once oxc becomes the default in the final #3031 flip.

nfdiop (approved):

PR #3584, opened 2026-06-30
add devtools plugin contract and registry

KuberSethi:
define a shared plugin contract and subscribable registry so base tabs and advanced plugins use one shape

• define the DevToolsPlugin interface (id, label, icon, panel component, optional init/dispose) and DevToolsPanelProps
• implement a subscribable registry with idempotent registerDevToolsPlugin plus a useDevToolsTabs hook
• wire the ./advanced subpath for advanced plugin exports

### DIFF
--- a/.changeset/devtools-plugin-registry.md
+++ b/.changeset/devtools-plugin-registry.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-devtools": patch
+---
+
+Add the devtools plugin contract (DevToolsPlugin/DevToolsPanelProps), a subscribable global plugin registry, the useDevToolsTabs hook, and the ./advanced entrypoint so base tabs and advanced plugins share one shape.
--- a/packages/react-devtools/package.json
+++ b/packages/react-devtools/package.json
@@ -15,6 +15,13 @@
       },
       "default": "./build/esm/index.js"
     },
+    "./advanced": {
+      "import": {
+        "types": "./build/types/public/advanced.d.ts",
+        "default": "./build/esm/public/advanced.js"
+      },
+      "default": "./build/esm/public/advanced.js"
+    },
     "./vite": {
       "import": {
         "types": "./build/types/public/vite.d.ts",
--- a/packages/react-devtools/src/plugins/baseTabs.ts
+++ b/packages/react-devtools/src/plugins/baseTabs.ts
@@ -1,0 +1,20 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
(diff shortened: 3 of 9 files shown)


nfdiop (commented):
Sorry if I missed it but I don't see why we need a plugin system, even after looking at this [PR](https://github.com/palantir/osdk-ts/pull/3591). Why not have a single tab type `DevToolPluginTab` similar to your `DevToolPlugin`?

KuberSethi:
> Sorry if I missed it but I don't see why we need a plugin system, even after looking at this [PR](https://github.com/palantir/osdk-ts/pull/3591). Why not have a single tab type `DevToolPluginTab` similar to your `DevToolPlugin`?

We want some tabs to be optionally installable, right? And since a lot of the wiring for how the DevTools gets rendered is automatic / magic for users, we can't have them simply insert the JSX for each tab, so something more advanced is required.

nfdiop:
> We want some tabs to be optionally installable, right? And since a lot of the wiring for how the DevTools gets rendered is automatic / magic for users, we can't have them simply insert the JSX for each tab, so something more advanced is required.

What do you mean with "installable"? I want some tabs hidden by default (if that makes sense) and the user can enable them like in Chrome Devtools

KuberSethi:
> > We want some tabs to be optionally installable, right? And since a lot of the wiring for how the DevTools gets rendered is automatic / magic for users, we can't have them simply insert the JSX for each tab, so something more advanced is required.
> 
> What do you mean with "installable"? I want some tabs hidden by default (if that makes sense) and the user can enable them like in Chrome Devtools

Like when setting up Devtools `plugins={ComputeTab, ...}` to configure additional ones. Open to other approaches but is the concern semantic or what's the concern?

nfdiop:
> > > We want some tabs to be optionally installable, right? And since a lot of the wiring for how the DevTools gets rendered is automatic / magic for users, we can't have them simply insert the JSX for each tab, so something more advanced is required.
> > 
> > 
> > What do you mean with "installable"? I want some tabs hidden by default (if that makes sense) and the user can enable them like in Chrome Devtools
> 
> Like when setting up Devtools `plugins={ComputeTab, ...}` to configure additional ones. Open to other approaches but is the concern semantic or what's the concern?

We don't need to allow users to customise the plugins yet and this change is unfortunately making the code more complex. So I would rather we allow users to hide some tabs and we store the state in local storage

KuberSethi:
> We don't need to allow users to customise the plugins yet and this change is unfortunately making the code more complex. So I would rather we allow users to hide some tabs and we store the state in local storage

I'd argue we do, and storing state in local storage would be tech debt + not a path viable for some of our users.

PR #3601, opened 2026-07-01
[Components] Add contextItems to AipAgentChat

ziyunp:
- Add contextItems prop to AipAgentChat to allow users to ask questions about the ontology objects / interface.
- On selecting a context item, the objects are fetched via asyncIter() callback.
- The loaded objects are parsed into XML and send in the system prompt to the LLM endpoint.
- Fixed the storybook layout and chat composer styling 

https://github.com/user-attachments/assets/3b91df61-fa60-4d45-ab60-12f412f56874


packages/react-components/src/aip-agent-chat/AipAgentChatApi.ts
```diff
@@ -102,9 +105,57 @@ export interface AipAgentChatProps {
...
+   *
+   * When non-empty, a multi-select picker is rendered in the composer
+   * footer. Selecting a type fetches its objects via `useOsdkObjects`
+   * (lazily — nothing is fetched until selected) and appends a serialized
+   * snapshot to the system prompt. Requires the app to be wrapped in
+   * `OsdkProvider`, since `useOsdkObjects` reads the OSDK client from
+   * React context. If omitted, no picker is rendered and the chat behaves
+   * exactly as before.
+   */
+  objectTypes?: ReadonlyArray<ObjectOrInterfaceDefinition>;
```
nfdiop: We need a way to do this work without adding `objectTypes` to the API. Otherwise when we want to remove this, we will need some API break. Maybe we should take a list of `contextItems`. If `contextItems.length > 0` we render the "+" button.

We can also export a contextItem: (label, description, onClick) for adding objects.

ziyunp: I renamed it to contextItems for now. I think the props you suggest there - (label, description, onClick), can be applied to custom context item type in the future. I would keep ObjectOrInterfaceDefinition as one of the context item type now and handle the display and data fetching internally. onClick is exposed via onSelectedContextItemsChanged callback

packages/react-components/src/aip-agent-chat/buildObjectContext.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Serializes the loaded objects into a single block of text suitable for
+ * appending to a system prompt. Types with no objects are skipped, and an
+ * empty string is returned when nothing is loaded so callers can leave the
+ * prompt untouched.
+ */
+export function buildObjectContext(
+  loaded: ReadonlyArray<LoadedObjectContext>
+): string {
+  const sections = loaded
```
ziyunp, line 37: Looking to improve this

packages/react-components/src/aip-agent-chat/AipAgentChatApi.ts
```diff
@@ -15,15 +15,27 @@
...
-// names (mirrored from `quiver-language-model-service-utils`'
-// PREFERRED_MODEL_IDENTIFIERS) so consumers get IDE autocomplete on the
-// `model` / `defaultModel` / `availableModels` props. The union should
-// widen via `T | (string & {})` so registered-model API names still
-// type-check.
+/**
+ * A single piece of context the user can attach to an {@link AipAgentChat}
+ * conversation.
+ */
+export type AipAgentChatContextItem = AipAgentChatObjectTypeContextItem;
```
nfdiop, line 26: If I want to add a new file as a context item, what would it look like?

packages/react-components/src/aip-agent-chat/__tests__/BaseAipAgentChat.test.tsx
```diff
@@ -223,7 +223,43 @@ describe("BaseAipAgentChat", () => {
...
+        status="ready"
+        error={undefined}
+        onSendMessage={vi.fn()}
+        onStop={vi.fn()}
+        onClearError={vi.fn()}
+      />
+    );
+
+    const slot = screen.getByTestId("above-slot");
+    expect(slot.textContent).toBe("context-picker-stub");
```
nfdiop, line 246: I'm not sure about these tests: they are not really testing that we have the position.

packages/react-components/src/aip-agent-chat/__tests__/buildObjectContext.test.ts
nfdiop, line 1: nice series of tests!

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
```diff
@@ -0,0 +1,76 @@
...
+import type {
+  ObjectOrInterfaceDefinition,
+  ObjectTypeDefinition,
+} from "@osdk/api";
+import { useOsdkClient } from "@osdk/react";
+import * as React from "react";
+
+export interface AipAgentChatContextLoaderProps {
+  /** The object type to load. */
+  objectType: ObjectOrInterfaceDefinition;
```
nfdiop, line 26: Do we support interfaces?

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: AipAgentChatContextLoaderProps): null {
+  const client = useOsdkClient();
+
+  React.useEffect(() => {
+    let cancelled = false;
+
+    // Cast to the object-type overload so TS picks a single overload;
+    // both object and interface object sets expose `asyncIter()` via
+    // `MinimalObjectSet`, so this is safe at runtime.
+    const iterator = client(objectType as ObjectTypeDefinition).asyncIter();
```
nfdiop, line 54: I think we should take `ObjectTypeDefinition` directly if we don't support interfaces

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
```diff
@@ -0,0 +1,76 @@
...
+   * @param objects The fetched objects, or `undefined` while loading or
+   *   on unmount.
+   */
+  onLoaded: (
+    apiName: string,
+    objects: ReadonlyArray<unknown> | undefined
+  ) => void;
+}
+
+export function AipAgentChatContextLoader({
```
nfdiop, line 42: Instead of this, could we have a function that load objects then use it here?

packages/react-components/src/aip-agent-chat/AipAgentChat.tsx
```diff
@@ -109,30 +218,57 @@ export function AipAgentChat({
...
+      activeModel={activeModel}
+      models={availableModels}
+      onModelChange={handleModelChange}
+      disabled={isInFlight}
     />
+  ) : undefined;
+
+  return (
+    <>
+      {selectedContextItemIds.map((id) => {
```
nfdiop, line 244: Hm, I'm not sure we should have this because it is gonna overflow and could be hard to read. Maybe let's align on the design first?

nfdiop (commented):
Let's first set up the infra to add context items + align on the design we want

packages/react-components/src/aip-agent-chat/buildObjectContext.ts
ziyunp: I plan to use XMLBuilder from "fast-xml-parser", will do this as a FLUP as it will introduce a new dependency.

packages/react-components/src/aip-agent-chat/AipAgentChatApi.ts
ziyunp:
```
export interface AipAgentChatFileTypeContextItem {
    type: "file";
    mediaType: string;
    filename?: string;
}

export type AipAgentChatContextItem = AipAgentChatObjectTypeContextItem | AipAgentChatFileTypeContextItem;
```

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
ziyunp: I have not tested this, but it should work. We can construct an interface objectSet that has asyncIter() method, so we should be able to fetch all interface objects

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
ziyunp: See the previous comment, I think interface is supported (need to test). This cast is there because of client's type definition. If we wrote the below, TS does not complain:

```
const iterator = objectType === "object" ? client(objectType).asyncIter() : client(objectType).asyncIter();
```

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
ziyunp: Will refactor, this is not needed anymore. It was used previously when the implementation was using useOsdkObjects hook.

packages/react-components/src/aip-agent-chat/AipAgentChat.tsx
ziyunp: Removing it now, the pattern is no longer needed now that I don't use the useOsdkObjects hook

packages/react-components/src/aip-agent-chat/components/AipAgentChatContextLoader.tsx
```diff
@@ -0,0 +1,76 @@
...
+      for await (const object of iterator) {
+        if (cancelled) {
+          return;
+        }
+        data.push(object);
+      }
+      if (!cancelled) {
+        onLoaded(objectType.apiName, data);
+      }
+    })();
```
olliejenningsPT, line 67: Non-blocking: the async IIFE that drives `asyncIter()` has no `catch`, so if the iterator rejects partway through (network drop, a permission error on the object set) the rejection is just swallowed. `onLoaded` never fires, the context silently never shows up, and nothing is logged. Since the component already exposes `onError` for chat failures, it feels inconsistent for context-load failures to be invisible. Might be worth surfacing it through `onError` or at least logging so it's debuggable.

ziyunp:
Closing this PR, after discussing with @nfdiop, we will have a design (UI + API) to have a chat component that works in the long term.

PR #3603, opened 2026-07-01
chore: migrate the low-risk tail packages to the oxc toolchain (oxc increment 12)

olliejenningsPT:
## Summary

Twelfth increment of the repo-wide oxc migration (#3031): moves the remaining **non-generator** packages (the low-risk tail) from ESLint + dprint to oxlint + oxfmt (configured via Ultracite), **and closes a lint-coverage gap the migration had opened by re-linting `.js/.mjs/.cjs` files**.

Migrated in this PR:

- **CLIs:** `@osdk/cli`, `cli.common`, `cli.cmd.typescript`
- **Scaffolders + templates:** `create-app`, `create-widget`, and every `create-app.template.*` / `create-widget.template.*` package (their own `src` only — the shipped `templates/` scaffolding migrates separately)
- **Tools:** `tool.release`, `tool.generate-with-mock-ontology`, `version-updater`, `create-app.template-packager`
- **Vite plugins:** `vite-plugin-oac`, `vite-plugin-superrepo`, `vite-plugin-status-reporter`, `widget.vite-plugin`
- **e2e:** all `e2e.sandbox.*`, `e2e.generated.*`, `e2e.test.foundry-sdk-generator`, `client.test.ontology`
- **Apps / test libs:** `ontology-explorer-app`, `ontology-explorer-server`, `shared.test`, `shared.test.intellisense`

Each package is reformatted with oxfmt; packages that first surface error-level Ultracite rules carry a nested oxlint config (extends the root, re-applies its ignorePatterns) turning those rules off, so the change is a reformat, not a rewrite (`oxlint --fix` is a verified no-op). The `e2e.sandbox.*` nested configs also turn off `palantir-header/header`, matching the prior ESLint carve-out for those demo packages.

## ⭐ Newly-covered files

**Linting — `.js` / `.mjs` / `.cjs` are now linted (this was a gap).** The oxc migration had ignored `**/*.js|mjs|cjs` in the root `oxlint.config.ts` since increment 1, so build scripts, `bin/*` shims, and config files (`build-css.mjs`, `codegen.mjs`, `tsup.config.js`, `webpack.config.cjs`, …) were **not** linted at all under oxc — a regression from the prior ESLint setup, which linted them (base rules + the Apache header). This PR drops that ignore and adds a root JS override that:
- keeps the Apache `palantir-header/header` rule on (ESLint parity), and
- turns off the error-level rules existing JS first surfaces, so it stays behavior-preserving (`oxlint --fix` is a no-op on JS; **no JS source changes**).

A `TODO(#3031 follow-up)` in that override tracks re-enabling each carved-out rule (and fixing the offending JS) so JS eventually meets the full ruleset like `.ts`. All **74** oxc packages now lint their JS. The msw-generated `mockServiceWorker.js` is ignored.

**Formatting — no net-new coverage.** JS was already oxfmt-formatted (oxfmt formatting was never gated by the oxlint ignore), so the JS-linting change adds linting only. Formatting stays at parity with the prior dprint behavior for every migrated package (`.ts/.tsx/.mts/.cts/.js/.mjs/.cjs/.json/.md`).

## Deferred: `@osdk/typescript-sdk-docs-examples` (stays on ESLint + dprint)

Its `codegen` step reformats generated documentation examples via `pnpm run format`, and that step runs on **every** CI test-matrix leg (Node 18–24) because a generator, `@osdk/osdk-docs-context-generator`, depends on its codegen output (so a per-leg exclusion can't prevent it). oxfmt can't load the repo's TypeScript `oxfmt.config.ts` on Node < 22.18, so its format step must stay on the Node-version-independent dprint. It moves to oxc once codegen is restored from the Node-24 build cache on the matrix rather than re-executed per leg (#3031 follow-up). After this increment, the nine generators plus `typescript-sdk-docs-examples` remain on ESLint.

## Root config changes

- **oxlint:** drop `**/*.js|mjs|cjs` ignore (see above) + JS override; ignore `mockServiceWorker.js`.
- **oxfmt:** drop the `**/*.generated.*` ignore (it matched the `@osdk/e2e.generated.*` package *directory* names, making oxfmt skip those packages and exit non-zero; dprint formatted their hand-written barrels before, so this keeps parity — the repo has no real `*.generated.*` files relying on it). Ignore `**/*.html` (dprint had no html plugin) and `**/templates/` (shipped `.hbs` scaffolding, migrated separately; oxlint already ignores it).
- **oxfmt single version:** the only workspace package pinning a different oxfmt was `examples/example-typescript-library-sdk-2.x` (`^0.52.0`, generated from the ts-library template), so the lockfile carried both 0.52.0 and 0.53.0 and pnpm could hand a package a different oxfmt across platforms (a file formatted locally with 0.53.0 then failed `oxfmt --check` in CI). Bumped that example and its source template's scaffold to match root (`oxfmt ^0.53.0`, `oxlint ^1.68.0`, `ultracite ^7.8.2`) so oxfmt resolves to a single 0.53.0.

## Archetype splits

Generators stay on ESLint: `foundry-sdk-generator` in `consumerCliPackages`; `example-generator` / `typescript-docs-example-generator` / `osdk-docs-context-generator` in `internal clis`; the `generator-*` packages in `standardLibraries`.

## Verification

- `pnpm turbo lint` across all 74 oxc packages → green (JS included).
- `pnpm turbo typecheck / transpile / build / test` for the migrated packages → green; whole-repo `turbo transpile` → green.
- `monorepolint check` → 0.
- Whole-repo `turbo check --continue` → green modulo the known pre-existing base failure `@osdk/examples.typescript-library-sdk-2.x#check` (reconfirmed pre-existing by stashing).


.monorepolint.config.mjs
```diff
@@ -141,9 +142,9 @@ const archetypeRules = archetypes(
   .addArchetype(
     "consumerCliPackages",
     [
-      "@osdk/cli",
-      "@osdk/create-app",
-      "@osdk/create-widget",
+      // @osdk/cli, @osdk/create-app, and @osdk/create-widget migrated to the oxc
```
nfdiop, line 145: nice comment!

nfdiop (approved):

PR #3604, opened 2026-07-01
chore: migrate create-app/create-widget scaffold templates to oxlint/oxfmt (oxc increment 13; shipped templates)

olliejenningsPT:
Batch 13 of the #3031 oxc migration. Swap the shipped eslint config for oxlint/ultracite in the 8 create-app/create-widget templates (mirroring the ts-library template).

USER-FACING: changes what scaffolded projects lint with.

## What changed
Per template (react, react.beta, tutorial-todo-app, tutorial-todo-app.beta, tutorial-todo-aip-app, tutorial-todo-aip-app.beta, create-widget.react.v2, create-widget.minimal-react.v2):
- Delete `templates/eslint.config.mjs`; add `templates/oxlint.config.ts` (ultracite core + react presets) and `templates/oxfmt.config.ts`.
- `package.json.hbs` (and react.beta's .osdk/.psdk variants): replace the eslint `lint` script and eslint/prettier devDeps with `{ check: ultracite check, fix: ultracite fix, format: oxfmt ., lint: oxlint }` and `oxlint`/`oxfmt`/`ultracite` devDeps. Eslint plugin devDeps removed from each template package's own package.json (they were only merged into the scaffold).
- react.beta: drop `prettier.config.js`, point `.vscode` at the oxc extension.
- Scaffold sources updated to pass the ultracite oxlint preset cleanly (behavior-preserving): strict null comparisons, `re.test()` + unicode-flagged regex, async/await instead of promise chains, alphabetized object keys, aria-labels, etc.
- Regenerated the example apps; `example-generator --check` passes and each generated project is `oxlint`-clean, `oxfmt`-clean, and builds under `tsc` (the templates target ES2020).

## ESLint+dprint -> oxc coverage parity (audited behaviorally)
The shipped `oxlint.config.ts` disables only rules that are genuinely inappropriate for these scaffolds:
- `unicorn/filename-case` (PascalCase React components)
- `typescript/no-non-null-assertion` (the `createRoot(document.querySelector("#root")!)` entrypoint idiom)
- `unicorn/prefer-string-replace-all` (the templates target **ES2020**, where `String.prototype.replaceAll` (ES2021) doesn't exist; we keep `String#replace(/…/gu, …)`)

It keeps `func-style` at declaration and re-enables the React rules the ultracite preset does not turn on by default, so they match the old ESLint config:
- `react-hooks/rules-of-hooks`: error
- `react-hooks/exhaustive-deps`: warn
- `react/only-export-components`: [warn, { allowConstantExport: true }]  (the oxlint equivalent of eslint-plugin-react-refresh's only-export-components; verified it honors allowConstantExport)

Covered by the ultracite core+react presets (verified): eslint:recommended + typescript-eslint:recommended (core is a strict superset — `curly`, `no-unused-vars`, etc.), react recommended (`jsx-key`, …), `jsx-a11y`, `import/{default,namespace,no-duplicates,…}`, and prettier->oxfmt formatting.

Dropped: `no-console` (was warn on react.beta only; scaffolds log intentionally), `import/named` (nursery-only in oxlint / redundant with tsc), and `import/no-extraneous-dependencies`.

### On `import/no-extraneous-dependencies` (phantom dependencies)
This is the one meaningful coverage gap, and it is intentional. The rule is **not** native to oxlint and there is no committed plan to make it native (the enabling infrastructure — parsing `package.json` and exposing deps to rules — is only an open, undecided proposal: oxc#22364). It *can* be run via oxlint's JS-plugin bridge (`eslint-plugin-import`, fixed for js-plugins in oxc#19192 / oxlint 1.44+, verified working on 1.69), but wiring it in would ship `eslint-plugin-import` + ~19 transitive deps and an unmet `eslint` peer back into every scaffold — re-adding the ESLint dependency tree this migration exists to remove.

What actually goes uncaught: a *phantom* dependency — a package that is imported and resolves (because it was hoisted / installed transitively) but is not listed in the project's own `package.json`. `tsc` does not catch this (it validates resolution, not declaration — verified it exits 0 on a resolvable-but-undeclared import); a fully-missing import is still caught as `TS2307`. Under `pnpm`'s strict layout undeclared imports don't resolve at all, so this only affects `npm`/`yarn`-hoisted projects.

**Decision:** leave `no-extraneous-dependencies` dropped and close this gap repo-wide with the in-flight `knip` integration (see `origin/ea.add-knip` / the "Add knip to repo" work) rather than re-introducing `eslint-plugin-import` per-scaffold. knip's purpose is exactly unlisted/unused-dependency detection.

## Notes
- `oxfmt.config.ts` additionally ignores `package.json`/`tsconfig.json` (mirroring the repo root oxfmt config), since ultracite's `sortPackageJson` fights monorepolint's field order in the generated example packages.
- Stacked on / rebased over batch 12 (#3603), which migrated the template *packages'* own toolchain and already handles `dprint.json` (excludes the template dirs) and the root `oxfmt.config.ts` `**/templates/` ignore — so this PR no longer touches `dprint.json`.
- Whole-repo `turbo check` is green except the pre-existing `examples.typescript-library-sdk-2.x#check` (untouched here).

## Remaining #3031 work after this PR

**Batch 14 — the final global flip (config-only, ~zero source churn).** Invert the root config so oxlint/oxfmt are the default and the ~9 generator packages opt *out*; drop the now-transitional "oxc migrated …" monorepolint archetypes and the `OXC_PACKAGES` / `OXC_NESTED_CONFIG_PACKAGES` lists in `.lintstagedrc.mjs`; flip the dprint exclusion list into a generator-inclusion. Functionally a no-op since every non-generator package is already on oxc. (Marked by the transitional-archetype TODO in `.monorepolint.config.mjs`.)

**Deferred rule re-enablements carried as code TODOs from earlier batches (tracked under #3031):**
- Re-enable the behavior/API-risk rules the migrations parked to stay behavior-preserving — root `oxlint.config.ts` currently sets `typescript/no-explicit-any: warn`, `no-eq-null: off`, `require-await: off`, `require-unicode-regexp: off`, `unicorn/custom-error-definition: off`, `default-param-last: off`. Re-enable in a dedicated PR that fixes the underlying findings.
- Re-enable the cosmetic/high-churn rules disabled repo-wide (`sort-keys`, `curly`, `prefer-destructuring`, `complexity`, `unicorn/no-useless-switch-case`) as a group in a formatting-only PR.
- The JS-files override in the root config (`**/*.{js,mjs,cjs}`) disables ~14 rules *only* because existing build scripts/configs violate them; re-enable one-by-one (fixing the offending files), then delete the override.
- Per-package nested-config findings (`@osdk/api`, `client`, `maker`, `react-components`, …) each turned off a set of error-level rules to keep those migrations reformats-not-rewrites; address the underlying findings and remove the carve-outs.
- `@osdk/typescript-sdk-docs-examples` stays on ESLint + dprint (its codegen reformats on every CI matrix leg and oxfmt can't load the TS config on Node < 22.18); migrate once codegen is restored from the Node-24 cache per leg.
- The 9 generator packages remain on ESLint because the repo runs **type-aware** rules on `packages/*/src` (`tseslint strictTypeCheckedOnly`: `no-floating-promises`, `no-misused-promises`, `await-thenable`, and notably `@typescript-eslint/restrict-template-expressions`). Generators emit code via template literals, and `restrict-template-expressions` is the type-informed guard that keeps a non-string (object/Promise) from being interpolated into generated output (`[object Object]` in a generated SDK). oxlint offers these only via its beta `--type-aware` (tsgolint) mode — **not** `jsPlugins`, which can't provide type info — so migrating the generators is gated on the type-aware adoption below (incl. the `baseUrl` tsconfig migration), not a separate effort.

**Follow-ups discovered while reviewing this PR:**
- **`react-hooks/rules-of-hooks` is not enabled repo-wide.** The root config (and the ultracite react preset) only enable `react-hooks/exhaustive-deps` (warn), not `rules-of-hooks` — so the repo's own React packages (`@osdk/react`, `react-components`, `cbac-components`, `widget.client-react`, …) don't enforce it. (The scaffolds in this PR *do*.) Add it to the root with a repo-wide sweep, or fold into batch 14.
- **`expo.v2` (and `vue`/`vue.v2`) scaffolds ship no lint config.** `expo.v2` has React Native `.tsx` but no eslint (before) or oxc (now) config, so it's unlinted; a follow-up could give it the same `[core, react]` oxc config (Vue would need a Vue preset).

**Scheduled follow-up — type-aware linting (after the TS 6 uplift):**
- Enable `oxlint --type-aware` repo-internally (via `oxlint-tsgolint`, already installed) once the **TS 6 uplift** removes `baseUrl`, which is the current blocker: `--type-aware` errors on our tsconfigs' `baseUrl` (tsgolint/typescript-go removed it). Verified that removing `baseUrl` lets `--type-aware` run and `no-floating-promises` fire; the `baseUrl` is already vestigial (paths use relative `"@/*": ["./src/*"]`, which resolve without it on TS 5.0+), so it could even be dropped independently.
- This adds the type-informed typescript-eslint rules oxlint can't do today (`no-floating-promises`, `no-misused-promises`, `await-thenable`, `restrict-template-expressions`, `no-unsafe-*`), and **is the same unblock that lets the 9 generator packages move off ESLint** (they only stay for the type-aware `restrict-template-expressions` rule — see above).
- Scope it repo-internal (with `oxlint-tsgolint`), **not** in the shipped scaffold (would add the dep and slow users' `lint`). Type-aware is still beta (59/61 typescript-eslint rules) and adds build cost/memory, so gauge CI time when enabling.

**Under consideration (not scheduled):**
- **Replacing `tsc --noEmit` with `oxlint --type-check` or `tsgo`:** premature. `tsc` stays the type-correctness source of truth; `--type-check` and `typescript-go` are experimental/preview (the `baseUrl` rejection above is a concrete non-drop-in gap). Revisit when TS 7.0 / tsgo is stable.


olliejenningsPT:
Superseded by a newer PR which in-compasses some improvements over this one.

PR #3611, opened 2026-07-02
chore: re-enable oxlint require-unicode-regexp

olliejenningsPT:
## What

Re-enables `require-unicode-regexp` (`off -> error`) in the root `oxlint.config.ts` and adds the `u` (unicode) flag to the flagged regular expressions across all migrated packages. Second PR in the effort to re-enable the six behavior/API-risk rules the oxc migration parked at `off`.

434 findings total; this PR resolves all in migrated packages (the ~39 in not-yet-migrated generator packages still lint via eslint and are out of scope).

## Why require the `u` flag?

The `u` flag opts a regex into Unicode mode, which is stricter and more correct than the default (non-unicode) mode. Enforcing it repo-wide buys:

- **Correct handling of non-BMP characters.** Without `u`, `.`, character classes, and quantifiers operate on UTF-16 code units, so a single astral character (emoji, less-common CJK, etc.) counts as *two* "characters". A pattern like `/^.$/` does not match a single emoji, and `.` / `[^x]` can split a surrogate pair mid-character. With `u`, the engine matches whole code points, which is almost always what the author intended when a regex touches user-provided text.
- **Malformed patterns fail loudly instead of silently misbehaving.** In Unicode mode, invalid or unnecessary escapes (e.g. a stray `\-` or `\ `) and reserved sequences are syntax errors; in the default mode they are silently reinterpreted. That turns a class of latent regex bugs into build-time failures.
- **Access to modern Unicode features.** `u` is the prerequisite for `\u{...}` code-point escapes and `\p{...}` / `\P{...}` Unicode property escapes, so newly written regexes can use them without a follow-up migration.
- **One consistent regex dialect.** Every regex in the codebase behaves the same way instead of some relying on the looser default behavior, which removes a footgun for future edits.

The flip is behavior-preserving *for the existing code* (see verification), while making all of the above the enforced default going forward.

## Approach

oxlint's `require-unicode-regexp` fix is gated behind `--fix-dangerously` (it can change match semantics), so this was not a blind autofix. The `u` flag was inserted into exactly the reported regex literals, then **every** change was verified:

- **Syntactic validity under `/u`:** full `typecheck` + `transpile` + `build` pass (376 tasks), so no regex contains an escape that is illegal in unicode mode.
- **Behavior preservation:** all changed regexes are ASCII-only patterns (verified: zero non-ASCII characters in the diff) matched against non-astral inputs (version strings, RIDs, file paths, API names, test assertions). The affected packages' test suites pass.
- The one behavioral surface that changed is a `maker` inline snapshot that quotes a validation regex's `.source` in an error message (now shows the `u` flag); snapshot updated.

## Dynamic `new RegExp(...)` cases

9 findings are `new RegExp(...)` built from runtime or user-provided values (faux object-search from `where.value`, osdk-docs-context example search, git tag prefixes, superrepo route prefixes, version-updater scripts). Forcing the `u` flag there could change matching or throw on patterns that are valid without it, so each keeps its current behavior via a scoped inline `// oxlint-disable-next-line require-unicode-regexp` with justification.

## Verification

- `oxlint` (root + all nested configs): 0 `require-unicode-regexp` findings in migrated packages
- `pnpm turbo typecheck transpile build`: green (376 tasks)
- `pnpm turbo test`: migrated packages green (maker snapshot updated). The only failing suite is `@osdk/example-generator` (not migrated, unchanged here; a pre-existing `.expo` on-disk-diff base failure)
- `pnpm turbo lint`: green (217 tasks), including `oxfmt --check`
- `pnpm turbo check-api` (api, client): no API report changes
- `monorepolint check`: 0; `dprint check`: clean

## Changeset

One `patch` changeset covering the 22 changed public non-ignored packages.

### DIFF
--- a/packages/faux/src/FauxFoundry/filterObjects.ts
+++ b/packages/faux/src/FauxFoundry/filterObjects.ts
@@ -164,7 +164,7 @@ export function filterObjects(
         throw new Error("propertyIdentifier not supported");
       }
       invariant(field);
-      const searchTerms = where.value.toLowerCase().split(/\s+/);
+      const searchTerms = where.value.toLowerCase().split(/\s+/u);
       return objects.filter((obj) => {
         const fieldValue = obj[field];
         if (typeof fieldValue === "string") {
@@ -182,7 +182,7 @@ export function filterObjects(
         throw new Error("propertyIdentifier not supported");
       }
       invariant(field);
-      const searchTerms = where.value.toLowerCase().split(/\s+/);
+      const searchTerms = where.value.toLowerCase().split(/\s+/u);
       return objects.filter((obj) => {
         const fieldValue = obj[field];
         if (typeof fieldValue === "string") {
@@ -200,7 +200,7 @@ export function filterObjects(
         throw new Error("propertyIdentifier not supported");
       }
       invariant(field);
-      const searchTerms = where.value.toLowerCase().split(/\s+/);
+      const searchTerms = where.value.toLowerCase().split(/\s+/u);
       return objects.filter((obj) => {
         const fieldValue = obj[field];
         if (typeof fieldValue === "string") {
@@ -227,7 +227,7 @@ export function filterObjects(
         throw new Error("propertyIdentifier not supported");
       }
       invariant(field);
-      const searchTerms = where.value.toLowerCase().split(/\s+/);
+      const searchTerms = where.value.toLowerCase().split(/\s+/u);
       return objects.filter((obj) => {
         const fieldValue = obj[field];
         if (typeof fieldValue === "string") {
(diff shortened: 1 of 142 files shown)


packages/faux/src/FauxFoundry/filterObjects.ts
```diff
@@ -261,6 +261,7 @@ export function filterObjects(
         throw new Error("propertyIdentifier not supported");
       }
       invariant(field);
+      // oxlint-disable-next-line require-unicode-regexp -- dynamic pattern; adding the u flag could change matching or throw on patterns that are valid without it
```
nfdiop, line 264: thanks for adding context!

nfdiop (approved):

PR #3613, opened 2026-07-02
Restyle MonitoringPanel devtools tabs to Blueprint Tabs

nfdiop:
## What

Restyles the `@osdk/react-devtools` `MonitoringPanel` tab bar from a bespoke `<button role="tab">` strip to Blueprint `Tabs`/`Tab`, with each tab's content supplied via the `panel` prop. Visual/markup only — the same tabs (Performance / Compute / Intercept / Debugging), the same content, and the same default tab. Bespoke UPPERCASE + letter-spacing styling is dropped for Blueprint's default title-case look and active indicator.

## Notable decisions

- **Controlled `Tabs`** (`selectedTabId` + `onChange` backed by existing state) so the selected tab survives minimize/reopen — an uncontrolled `Tabs` would reset to the first tab on reopen.
- **All panels stay mounted** (`renderActiveTabPanelOnly` left at its default `false`) — preserves Intercept's CodeMirror editor state, Performance charts, and per-tab scroll positions across switches.
- `onChange`'s `TabId` is narrowed to the known tab-id union via a type guard — no unchecked assertions or double casts.
- `DegradationNotice` moved above the tabs; the `panel`-prop shape leaves no slot between the tab list and its panels.
- A divider is restored under the tab list to match the intended design.
- Focus outlines suppressed via Blueprint's `Classes.FOCUS_DISABLED`, scoped to the panel root (not the host page's `<html>`).
- Removed the now-dead custom tab CSS (tab mixins, tab-content classes, and the orphaned letter-spacing token).

## Testing

- Component tests migrated to role-based queries (`getByRole("tab" | "tabpanel")`), covering default selection, tab switching, all-panels-mounted, and selection surviving a minimize/reopen cycle.
- `typecheck`, `lint`, and the package tests pass.

## Before
Screenshot 2026-07-03 at 15 32 16

## After
Screenshot 2026-07-03 at 13 59 27

Screenshot 2026-07-03 at 10 39 07

## Also

- `e2e.sandbox.officenetwork`: the `dev` script now watches and transpiles deps, for local iteration on the panel.


packages/e2e.sandbox.officenetwork/package.json
```diff
@@ -27,11 +27,12 @@
     "build": "vite build",
     "check-spelling": "cspell --quiet .",
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
-    "dev": "vite",
+    "dev": "trap 'kill 0' EXIT; pnpm turbo watch transpileBrowser --filter='@osdk/e2e.sandbox.officenetwork^...' & vite",
```
nfdiop: Added for HMR

packages/react-devtools/src/components/_mixins.scss
```diff
@@ -87,44 +87,6 @@
   margin: 0;
 }
 
-// Tab bar container
```
nfdiop, line 90: using the blueprint tabs

packages/react-devtools/src/components/MonitoringPanel.module.scss
```diff
@@ -195,33 +195,41 @@
 
 .content {
   flex: 1;
-  overflow: auto;
+  min-height: 0;
```
nfdiop, line 198: Because now moving overflow to the tabs. this meant that the callout for the DegradationNotice could overflow but added a fix below

packages/react-devtools/src/components/MonitoringPanel.module.scss
```diff
@@ -195,33 +195,41 @@
...
-  flex-shrink: 0;
-  pointer-events: auto;
-}
-
-.tabButton {
-  @include m.dt-tab-button;
+.notice {
+  // Blueprint's Callout is width: 100%; combined with its inline margin that
+  // overflows the panel on the right. Let it size to the padded width instead.
+  :global(.bp6-callout) {
```
nfdiop: note to self: we have some mechanism internally to use `:global(.#{bp.$ns}-` so you don't have to change for every new BP version. If I have time, will look into it

packages/react-devtools/src/components/MonitoringPanel.tsx
```diff
@@ -413,6 +426,8 @@ export const MonitoringPanel: React.FC<MonitoringPanelProps> = ({
 
   const panelClassName = classNames(
     styles.panel,
+    // Class needed to remove the useless outline added on tabs/buttons
+    Classes.FOCUS_DISABLED,
```
nfdiop, line 449: This is what we use internally. Otherwise, we will have useless outlines

packages/react-devtools/src/components/MonitoringPanel.tsx
```diff
@@ -527,71 +542,45 @@ export const MonitoringPanel: React.FC<MonitoringPanelProps> = ({
...
-
       <div className={styles.content}>
         {(!fiberCapabilities.hookInstalled ||
           !fiberCapabilities.fiberAccessWorking) && (
           <DegradationNotice onRetry={() => validateFiberAccess()} />
         )}
 
-        <div
-          className={
-            activeTab === "performance"
```
nfdiop: Blueprint tabs by default keep the tabs mounted

packages/react-devtools/src/components/MonitoringPanel.test.tsx
```diff
@@ -65,4 +69,56 @@ describe("MonitoringPanel", () => {
...
+    expect(
+      screen.queryByRole("tab", { name: "Performance", selected: true })
+    ).toBeNull();
+    // The single visible tabpanel is the one owned by the Compute tab.
+    expect(
+      screen.getByRole("tabpanel").getAttribute("aria-labelledby")
+    ).toContain("compute");
+  });
+
+  it("keeps all four tab panels mounted across a tab switch", () => {
```
nfdiop, line 132: These tests will change with the tabs enabled

packages/react-devtools/src/components/MonitoringPanel.tsx
```diff
@@ -402,6 +415,7 @@ export const MonitoringPanel: React.FC<MonitoringPanelProps> = ({
         <div
           className={styles.minimized}
           data-dt-theme={resolvedTheme}
+          aria-label="View OSDK Devtools"
```
nfdiop: For testing. But we should also use a proper button later

packages/e2e.sandbox.officenetwork/package.json
```diff
@@ -27,7 +27,7 @@
     "build": "vite build",
     "check-spelling": "cspell --quiet .",
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
-    "dev": "vite",
+    "dev": "trap 'kill 0' EXIT; pnpm turbo watch transpileBrowser transpileEsm --filter='@osdk/e2e.sandbox.officenetwork^...' & vite",
```
nfdiop, line 30: to avoid rebuilding packages we depend on including the devtools

packages/react-devtools/src/components/MonitoringPanel.module.scss
```diff
@@ -195,33 +195,41 @@
...
-.tabs {
-  @include m.dt-tabs;
-  padding: 0 12px;
-  flex-shrink: 0;
-  pointer-events: auto;
-}
-
-.tabButton {
-  @include m.dt-tab-button;
+.notice {
```
olliejenningsPT, line 204: This `.notice` rule is never applied. `styles.notice` is not referenced anywhere (grepped the whole package), and `<DegradationNotice>` is rendered bare inside `.content`. Because CSS-module classes are hashed, the `:global(.bp6-callout)` override only takes effect under an element carrying the hashed `.notice` class, so this width fix is currently inert and the Callout overflow it describes will still happen. Looks like the intent was to wrap the notice in `<div className={styles.notice}>` and it got dropped. Either wire it up or remove the rule.

olliejenningsPT (commented):
A few review notes below. The Blueprint Tabs swap itself looks correct: inactive panels stay mounted by default, so component state survives tab switches, and the new tests lock that in.

packages/e2e.sandbox.officenetwork/package.json
olliejenningsPT, line 30: This `dev` script change is unrelated to the devtools tab restyle and is not mentioned in the changeset. Presumably a local convenience for running the sandbox, but it does not belong in this PR. Consider splitting it out, or confirm it is intentional. (No changeset needed since this package is not published.)

packages/react-devtools/src/components/MonitoringPanel.tsx
olliejenningsPT: Minor a11y: this reopen affordance is a clickable `<div>` with `onClick` but no `role="button"`, `tabIndex`, or keyboard handler, so it is mouse-only and not exposed as a control to assistive tech. The added `aria-label` lets the test find it but does not make it operable. Pre-existing (the div was already click-only), so not a regression, but since the PR touches it, consider making it a real button.

packages/react-devtools/src/hooks/usePersistedState.ts
```diff
@@ -118,3 +118,11 @@ export function usePersistedState<T>(
 
   return [value, setPersistedValue];
 }
+
+/**
+ * Function exported for testing.
+ * We don't want tests to use localStorage directly since the storage layer could change
+ */
+export function clearPersistedState(): void {
+  localStorage.clear();
```
olliejenningsPT: `localStorage.clear()` wipes every key, not just the `osdk-monitor-*` keys this hook owns. Fine for the test `afterEach` and it is not re-exported from `index.ts`, so there is no public-API impact. But a prefix-scoped clear would match the `clearPersistedState` name better and avoid surprises. Non-blocking.

packages/react-devtools/src/components/MonitoringPanel.tsx
nfdiop: Yes. mentioned here that this should be fixed https://github.com/palantir/osdk-ts/pull/3613#discussion_r3519679423 but later

packages/e2e.sandbox.officenetwork/package.json
nfdiop: I kept it because it was small enough + made dev easier. see https://github.com/palantir/osdk-ts/pull/3613#discussion_r3520158858

packages/react-devtools/src/components/MonitoringPanel.module.scss
nfdiop: good catch. I accidentally removed. adding back

packages/react-devtools/src/hooks/usePersistedState.ts
nfdiop: let me make the change

packages/react-devtools/src/hooks/usePersistedState.ts
```diff
@@ -16,6 +16,8 @@
 
 import React from "react";
 
+const OSDK_DEVTOOLS_KEY_PREFIX = "osdk-";
```
nfdiop, line 19: this should be changed later to something more specific but I didn't want to break the keys

packages/react-devtools/src/fiber/DegradationNotice.tsx
```diff
@@ -202,7 +209,7 @@ export const DegradationNotice: React.FC<DegradationNoticeProps> = ({
           <Tag
             key={feature}
             intent={Intent.WARNING}
-            minimal
+            minimal={true}
```
nfdiop, line 212: change after formatting

KuberSethi (dismissed):

KuberSethi (approved):

PR #3614, opened 2026-07-02
chore: re-enable oxlint require-await

olliejenningsPT:
## What

Re-enables `require-await` and `typescript/require-await` (`off -> error`) in the root `oxlint.config.ts`, and resolves all findings. Fourth PR in the effort to re-enable the six behavior/API-risk rules the oxc migration parked at `off`.

## How the findings were resolved

`require-await` flags `async` functions with no `await`. The findings fell into three groups, each handled by its safest fix:

**1. Test callbacks (~127) — drop the redundant `async`.**
`it(...)` / `describe(...)` / `beforeEach(...)` and mock functions that never `await`. Removing `async` is behavior-identical and has no API surface. (15 of the originally-flagged test callbacks turned out to be typed mocks that *require* `async` to match a `() => Promise<T>` slot — those were kept async + disabled instead; caught by typecheck.)

**2. Shipped-source functions (76) — keep `async`, add a scoped disable.**
These are async to satisfy a `Promise`-returning contract: cache getters (`AsyncClientCache`/`WeakAsyncCache`), token providers, `ObservableClient` methods, `async next()` iterators, mock `ObjectSet` methods, `validateEdit` callbacks. Removing `async` would break their types (verified by typecheck) or change throw-timing / the published `Promise` signature, so per the migration's behavior-preserving rule each gets a justified `// oxlint-disable-next-line require-await`.

**3. Generated conjure trees (571) — disable in the nested configs.**
`client.unstable` (527) and `client.unstable.tpsa` (44) lint their checked-in `src/generated/**` trees, where the conjure generator emits every service method as `async` (many with no `await`). This is regenerated code, not hand-editable, so `require-await` is turned off in those packages' nested configs — matching their existing generated-tree rule opt-outs (`no-barrel-file`, `import/newline-after-import`).

Also dropped a redundant `async` from `e2e.generated.1.1.x`'s tsup `onSuccess` hook.

> Note: these generated-tree findings are only surfaced by the packages' *own* nested configs (which re-include `**/generated`), not by a root-level lint pass — so `pnpm turbo lint` (what CI runs) is the ground truth here.

## Verification

- `pnpm turbo lint` (root + all nested configs): **217 tasks green** — 0 `require-await` findings
- `pnpm turbo typecheck transpile build`: green (376 tasks) — confirms no `async` was removed where its `Promise<T>` type is load-bearing
- Affected test suites: green (api, client, react, react-components, aip-core, oauth, language-models, unit-testing, functions, version-updater)
- `pnpm turbo check-api` (api, client, functions): no API report changes
- `monorepolint check`: 0; `dprint check`: clean

## Changeset

One `patch` changeset covering the 18 changed public non-ignored packages.

### DIFF
--- a/packages/client/src/object/object.test.ts
+++ b/packages/client/src/object/object.test.ts
@@ -57,7 +57,7 @@ describe.each([
   let client: Client;
   let apiServer: SetupServer;
 
-  beforeAll(async () => {
+  beforeAll(() => {
     ({ client, apiServer } = startNodeApiServer(
       new LegacyFauxFoundry(baseUrl),
       createClient
@@ -205,7 +205,7 @@ describe.each([
         "ri.phonograph2-objects.main.object.88a6fccb-f333-46d6-a07e-7725c5f18b61"
       );
     });
-    it("objects are enumerable in an sdk", async () => {
+    it("objects are enumerable in an sdk", () => {
       const objects = Object.keys($Objects);
       expect(objects.sort()).toStrictEqual(
         [
@@ -251,7 +251,7 @@ describe.each([
       });
     });
 
-    it("clones and updates an object with another osdk object", async () => {
+    it("clones and updates an object with another osdk object", () => {
       const updatedEmployee = createOsdkObject(
         client[additionalContext],
         {
@@ -304,7 +304,7 @@ describe.each([
       });
     });
 
-    it("correctly scopes up with another OSDK object", async () => {
+    it("correctly scopes up with another OSDK object", () => {
       const firstEmployee = { $clone: () => {} } as unknown as Osdk.Instance<
         Employee,
         never,
@@ -315,7 +315,7 @@ describe.each([
       >();
     });
 
-    it("Correctly preserves keys from original and new with distinct property key sets", async () => {
+    it("Correctly preserves keys from original and new with distinct property key sets", () => {
       const firstEmployee = { $clone: () => {} } as unknown as Osdk.Instance<
         Employee,
         never,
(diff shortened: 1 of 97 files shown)


olliejenningsPT:
Follow-up (commit 02e5877): classified the 76 shipped-source disables against the **type-aware** `typescript/require-await` rule (`oxlint --type-aware`) to see which would survive type-aware linting:

- **36** are on functions that return a Promise — type-aware does **not** flag these, so their disable becomes redundant once type-aware linting is enabled. Each now carries a `// TODO(oxc type-aware): ... remove this disable once type-aware linting is enabled.` above it.
- **40** are on functions that return a sync value to satisfy a `Promise<T>` contract — type-aware **still** flags these (dropping `async` would break the type), so they keep a plain disable with no TODO.

Note: a `// oxlint-disable-next-line require-await` also suppresses the type-aware `typescript/require-await` variant (verified), so these disables won't error even if/when type-aware linting is turned on — the TODOs are purely a cleanup signal.

packages/client/src/object/object.test.ts
```diff
@@ -886,61 +886,59 @@ describe.each([
   });
 });
 
-export async function shouldError(client: Client): Promise<Osdk<Employee>> {
+export function shouldError(client: Client): Promise<Osdk<Employee>> {
```
ssanjay1, line 889: These should technically still be async right because they're functions returning promises?

olliejenningsPT: Not exactly, so a function can be `sync` but still return a promise, e.g.

```ts
function example(): Promise<X> {
  return new Promise();
}
```

The `async` keyword lets you use `await` inside the body of it, or auto-wraps the return value in a promise. So if the function has no `await` inside of the body, the `async` keyword is redundant.

The only aspect is when you throw in a sync function before the promise, it doesn't handle as a rejection (standard error), but since this is a test file it is fine, but we should be explicit in prod code.

ssanjay1 (approved):

PR #3621, opened 2026-07-03
feat: import brand theme from uploaded CSS or a site's CSS

Leolide:
## Summary

Adds an **Import** button to the Storybook Brand Theme addon that builds a complete OSDK theme from real design tokens in an uploaded file — a **CSS stylesheet** or a **DESIGN.md**. No network calls and no screenshots: if a file has no usable tokens it shows an error instead of guessing.

This is a **Storybook-only** change (private `@osdk/react-components-storybook`) — it touches no shipped package, so there's **no changeset** and nothing reaches consumers.

## What it does

- **Upload CSS** (e.g. your app's `tokens.css`) → recovers custom properties, button/anchor colors, radii, and fonts.
- **Upload DESIGN.md** → supports both the structured `design.md` YAML frontmatter and freeform prose brand-analysis files (hex codes listed under headings).
- Extracted values anchor a synthesized, contrast-checked, clamped palette, so every theme role ends up populated and coherent.

## Suggested review order

| File | What to look at |
|------|-----------------|
| `Panel.tsx` (+20) | Where the Import dropdown is wired into the panel |
| `ImportDropdown.tsx` (+345) | The UI — file picker, loading/error states, applying the result |
| `design-md-extractor.ts` (+589) | DESIGN.md → tokens: YAML frontmatter parser + prose fallback |
| `css-extractor.ts` (+771) | CSS → tokens: reads `--custom-property` / `:root` declarations, resolves `var()` aliases, then synthesizes a contrast-checked, clamped palette |
| `palette-extractor.ts` (+336) | Turns the recovered anchors into a full, harmonized palette |
| `token-map.ts` / `presets.ts` / `types.ts` / `decorator.tsx` (small) | New token roles + defaults |

## Testing

- `oxlint`, `oxfmt --check`, and `typecheck` all pass for the package.
- Verified extraction against real inputs — CSS token sheets and the `awesome-design-md` DESIGN.md collection (stripe, notion, supabase, spotify, tesla, …): correct primary / background / text and a coherent theme; token-free input correctly reports "no usable tokens".

## Not in this PR

The themeable button token (`--osdk-button-border-radius`) is split into #3622.

### DIFF
--- a/.changeset/brand-theme-font-availability.md
+++ b/.changeset/brand-theme-font-availability.md
@@ -1,0 +1,4 @@
+---
+---
+
+Brand-theme Storybook addon: ignore imported fonts that aren't installed locally (fall back consistently) and warn on import when a named brand font is missing. Dev-only addon change — no published package is affected.
--- a/packages/react-components-storybook/.storybook/addons/brand-theme-extractor/ImportDropdown.tsx
+++ b/packages/react-components-storybook/.storybook/addons/brand-theme-extractor/ImportDropdown.tsx
@@ -21,7 +21,6 @@ import { styled } from "storybook/theming";
 import {
   type CssExtractionResult,
   extractTokensFromCssText,
-  extractTokensFromWebsite,
 } from "./css-extractor.js";
 import { extractTokensFromDesignMarkdown } from "./design-md-extractor.js";
 import type { ThemeColorMode, TokenAssignment } from "./types.js";
@@ -29,8 +28,6 @@ import type { ThemeColorMode, TokenAssignment } from "./types.js";
 interface ImportDropdownProps {
   onApply: (assignments: TokenAssignment[], colorMode: ThemeColorMode) => void;
 }
-
-type Mode = "upload" | "website";
 
 // The resulting theme roles worth previewing, in reading order.
 const PREVIEW_ROLES: Array<{ role: string; label: string }> = [
@@ -91,35 +88,6 @@ const Body = styled.div({
   gap: 8,
 });
 
-const Segmented = styled.div(({ theme }) => ({
-  display: "flex",
-  gap: 2,
-  padding: 3,
-  background: theme.background.app,
-  border: `1px solid ${theme.appBorderColor}`,
-  borderRadius: 8,
-}));
-
-const Segment = styled.button<{ active: boolean }>(({ active, theme }) => ({
-  flex: 1,
-  background: active ? theme.background.content : "transparent",
-  border: "none",
-  borderRadius: 6,
-  boxShadow: active ? "0 1px 2px rgba(0,0,0,0.12)" : "none",
-  color: active ? theme.color.defaultText : theme.color.mediumdark,
-  cursor: "pointer",
-  fontSize: 12,
-  fontWeight: 600,
-  padding: "6px 10px",
-  transition: "background 120ms ease, color 120ms ease",
-}));
-
-const Row = styled.div({
-  display: "flex",
-  gap: 6,
-  alignItems: "center",
-});
-
 const ChooseButton = styled.button<{ disabled?: boolean }>(
   ({ disabled, theme }) => ({
     width: "100%",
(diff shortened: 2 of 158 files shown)


packages/react-components-storybook/.storybook/addons/brand-theme-extractor/css-extractor.ts
nfdiop, line 1: Could we simplify this logic? Could you look into https://github.com/jnsahaj/tweakcn/blob/main/utils/parse-css-input.ts#L9 and how they import css?

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/decorator.tsx
```diff
@@ -20,8 +20,57 @@ import { useEffect, useMemo } from "react";
...
+import type { TokenAssignment } from "./types.js";
 
 const STYLE_ID = "brand-theme-overrides";
+const FONTS_LINK_ID = "brand-theme-fonts";
+
+/**
+ * First families that are system/preinstalled or proprietary — never on Google
+ * Fonts, so we don't try to load them (the font stack's fallbacks handle them).
+ */
+const NON_GOOGLE_FONT =
```
nfdiop: This list could be hard to keep in sync. 
Instead, can we ignore the font for now? It would simplify the PR. It's also safer not to download fonts on behalf of users

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/design-md-extractor.ts
nfdiop, line 1: is there a css library we can use to do this from here? https://github.com/google-labs-code/design.md

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/ImportDropdown.tsx
```diff
@@ -0,0 +1,345 @@
...
+} from "./css-extractor.js";
+import { extractTokensFromDesignMarkdown } from "./design-md-extractor.js";
+import type { ThemeColorMode, TokenAssignment } from "./types.js";
+
+interface ImportDropdownProps {
+  onApply: (assignments: TokenAssignment[], colorMode: ThemeColorMode) => void;
+}
+
+// The resulting theme roles worth previewing, in reading order.
+const PREVIEW_ROLES: Array<{ role: string; label: string }> = [
```
nfdiop, line 55: could we try extracting only some of these from the css file so it is easier to parse?

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/design-md-extractor.ts
Leolide: Dug into `@google/design.md` — one blocker for us:
**Won't bundle for the browser** (the addon runs in the Storybook manager). Its `/linter` entry hard-imports `node:fs`/`path`/`process`/`url`/`module`, and there's no browser-safe subpath.

So I kept a small, dependency-free reader (browser-safe). Options: **(a)** keep it, or **(b)** add a Vite `node:*` stub so we can use the lib anyway. Preference?

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/decorator.tsx
Leolide: Done — dropped font loading entirely. No more downloading fonts on users' behalf, and that sync-prone list is gone.

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/css-extractor.ts
Leolide: Simplified - swapped the heuristic engine for a small alias reader that just reads the declared custom props and resolves `var()`s (~1583→771 lines). Looked at tweakcn, but it does a full CSS parse; felt like overkill for reading vars that are already declared, so I kept this dependency-free. Happy to switch if you'd rather lean on it.

packages/react-components-storybook/.storybook/addons/brand-theme-extractor/ImportDropdown.tsx
Leolide: Yeah — it now only pulls a small curated set (primary/secondary/background/surface/text, plus font + radius/spacing) via an alias list, instead of trying to parse everything out.

ziyunp:
Did we remove the capability to update a screenshot or visit a website to extract theme?

Leolide:
> Did we remove the capability to update a screenshot or visit a website to extract theme?

Yes because it does not work very well

nfdiop (commented):
one blocking comment: the font used in the filter list looks off 

Image

Leolide:
> one blocking comment: the font used in the filter list looks off
> 
> Image

I did some exploration seems that the token in filter list should be using the same token as the rest of the components, can you tell me how to repro?

.changeset/brand-theme-font-availability.md
nfdiop, line 1: nit: I don't think this changeset is correct but it should be fine since this is Storybook

nfdiop (approved):

nfdiop (approved):
Approving and let's track to refactor the code after testing

PR #3624, opened 2026-07-04
chore: re-enable oxlint no-useless-switch-case

olliejenningsPT:
Re-enables the `unicorn/no-useless-switch-case` oxlint rule, which we parked repo-wide during the oxc migration, and drops the handful of redundant case labels it flags.

The rule only flags an **empty** `case` that falls straight through to `default`, e.g.:

```ts
case "manual":   // empty, falls through
default:
  return generateManualChanges(employees);   // "manual" lands here anyway
```

Removing `case "manual":` is behavior-neutral (it still hits `default`). That's all the changes here are: 8 files where an empty case fell through to a `default` that already handled it. It does not touch exhaustive `switch`es that use a `const _: never` check — those have an unreachable `default`, so nothing falls through to flag.

### Scope / what's intentionally left off

This started as a batch of formatting rules; based on review feedback the rest are staying off:

- **`prefer-destructuring`** — off. Array destructuring (`const [, x] = arr`) read worse than indexed access, and object destructuring wasn't worth the churn either, so the whole rule stays disabled.
- **`curly`** — off. Forcing braces on guard clauses like `if (x) continue;` reads worse.
- **`complexity`** — its own PR (#3687), since it can't be autofixed and its extractions add shipped code.
- **`sort-keys`** — off (alphabetizing object keys reads worse than intentional grouping).

Scope is limited to packages already on the oxc toolchain.

### Config change

Removed the `unicorn/no-useless-switch-case` off-override from the root `oxlint.config.ts`. Everything else stays off.

### Verification

Green across the repo: `lint` + `typecheck` (305/305) and the full test suite (maker 154, maker-experimental 67, react-devtools 558). Behavior-preserving.

### DIFF
--- a/.changeset/reenable-no-useless-switch-case.md
+++ b/.changeset/reenable-no-useless-switch-case.md
@@ -1,0 +1,9 @@
+---
+"@osdk/maker": patch
+"@osdk/maker-experimental": patch
+"@osdk/react-devtools": patch
+"@osdk/react-sdk-docs": patch
+"@osdk/typescript-sdk-docs": patch
+---
+
+Re-enable the `unicorn/no-useless-switch-case` oxlint rule and drop the redundant empty `case` labels that fell straight through to `default`. Behavior-preserving with no runtime or API changes.
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -53,13 +53,9 @@ export default defineConfig({
     "unicorn/no-lonely-if": "off",
     "jsdoc/require-param-description": "off",
     "jsdoc/require-returns-description": "off",
-    // Cyclomatic-complexity ceiling; the large discriminated-union switch helpers
-    // in the docs packages exceed it. Splitting them up is pure churn here.
+    // Cyclomatic-complexity ceiling. Not a pure autofix (requires extracting
+    // helpers, which adds bundle size), so it's handled in its own PR.
     "complexity": "off",
-    // Several switches list known discriminant values (e.g. "string", "unknown")
-    // that share the `default` body. Keeping the explicit cases documents intent;
-    // removing them would be churn with no behavioral change.
-    "unicorn/no-useless-switch-case": "off",
 
     // Rules whose autofix would change behavior or the public API of these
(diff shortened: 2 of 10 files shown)


packages/cli.cmd.typescript/src/generate/handleGenerate.mts
```diff
@@ -338,7 +338,9 @@ export async function getDependencyVersions(): Promise<{
 
 let cachedOurPackageJsonPath: string;
 async function getOurPackageJsonPath() {
-  if (cachedOurPackageJsonPath) return cachedOurPackageJsonPath;
+  if (cachedOurPackageJsonPath) {
```
bryantpark04: I'm not sure these kinds of changes make our code more readable; I would prefer they stay as is

packages/client/src/observable/internal/Store.test.ts
```diff
@@ -1795,7 +1795,7 @@ describe(Store, () => {
           );
           await waitForCall(ifaceSub, 2);
 
-          const ifacePayload = ifaceSub.next.mock.calls[1][0];
+          const [, [ifacePayload]] = ifaceSub.next.mock.calls;
```
bryantpark04: Same with this array destructuring. I understand for objects it saves some duplication but this seems a lot less clear to me.

packages/cli.cmd.typescript/src/generate/handleGenerate.mts
bryantpark04: especially for `if (x) continue;`

olliejenningsPT: 100% fine with me, i'll keep this rule off

packages/client/src/observable/internal/Store.test.ts
olliejenningsPT: sounds good to me, let me keep the object destructuring but turn off the array. These are the recommended rules, but again we should tweak to what people feel, and l agree array destructuring is one of those, not so clear to understand.

bryantpark04 (approved):

PR #3626, opened 2026-07-06
add copy-prompt builder and button for devtools recommendations

KuberSethi:
adds a one-click copy-prompt action that turns a performance recommendation into an ai-ready fix prompt

• new copyPrompt builder that formats a recommendation (and a copy-all variant) into a prompt
• CopyPromptButton with clipboard write and a copied confirmation
• extends the performance recommendation engine types the button consumes

### DIFF
--- a/packages/react-devtools/src/components/CopyPromptButton.tsx
+++ b/packages/react-devtools/src/components/CopyPromptButton.tsx
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Button } from "@blueprintjs/core";
+import React, { useCallback, useRef, useState } from "react";
+
+import {
+  buildCopyAllPrompt,
+  buildCopyPrompt,
+} from "../recommendations/copyPrompt.js";
+import type { Recommendation } from "../utils/PerformanceRecommendationEngine.js";
+
+export type CopyPromptButtonProps =
+  | { recommendation: Recommendation; label?: string }
+  | { recommendations: Recommendation[]; label?: string };
+
+export const CopyPromptButton: React.FC<CopyPromptButtonProps> = (props) => {
+  const [copied, setCopied] = useState(false);
+  const copyTimeoutRef = useRef<ReturnType<typeof setTimeout>>(undefined);
+
+  const isSingle = "recommendation" in props;
+  const prompt = isSingle
+    ? buildCopyPrompt(props.recommendation)
+    : buildCopyAllPrompt(props.recommendations);
+  const label = props.label ?? (isSingle ? "Copy prompt" : "Copy all");
+
+  const handleCopy = useCallback(async () => {
+    try {
+      await navigator.clipboard.writeText(prompt);
+      setCopied(true);
+      clearTimeout(copyTimeoutRef.current);
+      copyTimeoutRef.current = setTimeout(() => setCopied(false), 1500);
+    } catch {
+      // Clipboard writes can reject (e.g. denied permission); ignore silently.
+    }
+  }, [prompt]);
+
+  return (
+    <Button
+      variant="minimal"
+      size="small"
+      icon={copied ? "tick" : "clipboard"}
+      onClick={() => void handleCopy()}
+    >
+      {label}
+    </Button>
+  );
+};
--- a/packages/react-devtools/src/components/CopyPromptButton.tsx
+++ b/packages/react-devtools/src/components/CopyPromptButton.tsx
@@ -15,8 +15,9 @@
  */
 
 import { Button } from "@blueprintjs/core";
-import React, { useCallback, useRef, useState } from "react";
+import React from "react";
 
+import { useCopyToClipboard } from "../hooks/useCopyToClipboard.js";
 import {
   buildCopyAllPrompt,
   buildCopyPrompt,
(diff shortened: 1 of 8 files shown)


packages/react-devtools/src/components/CopyPromptButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  buildCopyAllPrompt,
+  buildCopyPrompt,
+} from "../recommendations/copyPrompt.js";
+import type { Recommendation } from "../utils/PerformanceRecommendationEngine.js";
+
+export type CopyPromptButtonProps =
+  | { recommendation: Recommendation; label?: string }
+  | { recommendations: Recommendation[]; label?: string };
+
+export const CopyPromptButton: React.FC<CopyPromptButtonProps> = (props) => {
```
nfdiop, line 31: nit: I find it nice to use something like https://usehooks-ts.com/react-hook/use-copy-to-clipboard. If we have this pattern in another part of the codebase, happy to use it

packages/react-devtools/src/recommendations/copyPrompt.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    effort: "Low",
+    suggestion: "Use $select to only fetch used properties",
+    dismissible: true,
+    ...overrides,
+  };
+}
+
+describe("buildCopyPrompt", () => {
+  it("always renders intro, issue, suggestion, guidance and task sections", () => {
+    const prompt = buildCopyPrompt(makeRec());
```
nfdiop: curious why not test the full prompt once?

packages/react-devtools/src/recommendations/copyPrompt.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    );
+    expect(prompt).toContain("OSDK GUIDANCE:");
+    expect(prompt).toContain("TASK: Apply the suggested fix");
+  });
+
+  it("omits the LOCATION line when filePath is absent", () => {
+    expect(buildCopyPrompt(makeRec())).not.toContain("LOCATION:");
+  });
+
+  it("does not claim a location in the task when filePath is absent", () => {
```
nfdiop: extra nit: this could be combined with the above test

packages/react-devtools/src/recommendations/copyPrompt.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    const prompt = buildCopyPrompt(makeRec());
+
+    expect(prompt).not.toContain("the location above");
+    expect(prompt).toContain(
+      "Use the issue and suggested fix above to find the component or query"
+    );
+  });
+
+  it("points the task at the location when filePath is present", () => {
```
nfdiop, line 67: can be a FLUP: but I would expect the filePath to show up in the prompt

packages/react-devtools/src/recommendations/copyPrompt.test.ts
nfdiop, line 1: nice adding tests

packages/react-devtools/src/recommendations/copyPrompt.ts
```diff
@@ -0,0 +1,79 @@
...
+  return sections.join("\n\n");
+}
+
+/**
+ * Builds a single combined prompt covering several recommendations, ordered as
+ * given. Each recommendation is numbered and the entries are separated by a
+ * horizontal rule.
+ */
+export function buildCopyAllPrompt(recs: Recommendation[]): string {
+  const preamble = `Here are ${recs.length} OSDK optimizations, ordered by priority:`;
```
nfdiop, line 78: maybe I missed it. But where do we order?

packages/react-devtools/src/utils/PerformanceRecommendationEngine.ts
```diff
@@ -47,8 +47,26 @@ export interface Recommendation {
   suggestion: string;
   code?: string;
   dismissible: boolean;
+  filePath?: string;
+  lineNumber?: number;
+  currentCode?: string;
+  osdkGuidance?: string;
 }
 
+export const OSDK_GUIDANCE_BY_CATEGORY: Record<RecommendationCategory, string> =
```
nfdiop, line 56: nice!

packages/react-devtools/src/utils/PerformanceRecommendationEngine.ts
```diff
@@ -132,6 +156,9 @@ export class PerformanceRecommendationEngine {
         suggestion: "Use $select to only fetch used properties",
         code: offender.suggestion,
         dismissible: true,
+        filePath: locatedBinding?.filePath,
```
nfdiop, line 159: nit: can be a FLUP but we could pass a location maybe? location: locatedBinding

nfdiop (commented):
This is exciting

packages/react-devtools/src/recommendations/copyPrompt.test.ts
KuberSethi: not tractable unfortunately; sourcemaps are hard lol

packages/react-devtools/src/recommendations/copyPrompt.test.ts
KuberSethi: will address later in a FLUP

packages/react-devtools/src/recommendations/copyPrompt.test.ts
KuberSethi: likely to change and it doesn't test too much, that would be a bad test imo

packages/react-devtools/src/components/CopyPromptButton.tsx
KuberSethi: nice find! did something similar

packages/react-devtools/src/utils/PerformanceRecommendationEngine.ts
KuberSethi: if not required doing in a flup

nfdiop: yes can be a FLUP when we get the chance

packages/react-devtools/src/recommendations/copyPrompt.test.ts
nfdiop: wait then why do we have the test if we cannot guarantee the file is present in it? It would be weird if the message said "look at the location above" and there is no location above

KuberSethi: ah misunderstood the original question, didn't test originally as we filter out every react-devtools stack frame as internal, so inside the package's own tests parseStackTrace sees only internal frames and returns nothing, there would be no real file/line to assert.

KuberSethi: thought you were asking about something else

ziyunp (approved):

packages/react-devtools/src/recommendations/copyPrompt.ts
KuberSethi: added a comment!

PR #3627, opened 2026-07-06
add client metrics for the devtools performance tab

KuberSethi:
add the client metrics the redesigned performance tab reads

• derive nine headline metrics (cache hit rate, requests saved, latencies, optimistic coverage, rollback rate) from the metrics snapshot
• each metric exposes a value that stays undefined until it has enough samples, shown as N/A
• the base-tab plugin contract that used to live here now ships with the shell PR (#3628)


.changeset/dt2-devtools-plumbing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/react-devtools": patch
+---
+
+add internal devtools plumbing: the base-tab plugin contract and canonical request-based metrics that the redesigned tabs build on
```
nfdiop: could you update the changeset + PR description?

packages/react-devtools/src/metrics/clientMetrics.test.ts
```diff
@@ -0,0 +1,280 @@
...
+  "rollbackRate",
+];
+
+describe("getCanonicalMetrics", () => {
+  let store: MetricsStore;
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+
+    if (globalThis.requestIdleCallback === undefined) {
```
nfdiop, line 112: could we add a comment why we need this?

packages/react-devtools/src/metrics/canonicalMetrics.test.ts
```diff
@@ -0,0 +1,280 @@
...
+    store.dispose();
+    vi.useRealTimers();
+  });
+
+  it("exposes exactly the nine canonical metric keys", () => {
+    const metrics = getCanonicalMetrics(makeSnapshot({}));
+    expect(Object.keys(metrics).sort()).toEqual([...CANONICAL_KEYS]);
+  });
+
+  it("locks the MIN_SAMPLES thresholds", () => {
```
nfdiop: Curious why test a constant?

packages/react-devtools/src/plugins/types.ts
nfdiop, line 1: can we remove as discussed?

packages/react-devtools/src/plugins/baseTabs.ts
nfdiop, line 1: to remove?

packages/react-devtools/src/metrics/canonicalMetrics.ts
```diff
@@ -0,0 +1,165 @@
...
+    optimisticCoverage,
+    avgPerceivedSpeedupMs,
+    rollbackRate,
+  };
+}
+
+/** Format a metric as a display string, or a placeholder until it is ready. */
+export function formatMetric(m: Metric): string {
+  if (!m.ready) {
+    return "collecting data…";
```
nfdiop: I'm not sure about this because without enough space, this will overflow or look off. Instead can we use something like "N/A" instead?

packages/react-devtools/src/metrics/canonicalMetrics.ts
```diff
@@ -0,0 +1,165 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { MetricsSnapshot } from "../types/index.js";
+
+export interface Metric {
+  value: number;
+  sampleCount: number;
+  ready: boolean;
```
nfdiop: rather than having "ready" can we have  `value: number | undefined`?

packages/react-devtools/src/hooks/useCanonicalMetrics.ts
```diff
@@ -0,0 +1,34 @@
...
+  monitorStore: MonitorStore
+): CanonicalMetrics {
+  const store = monitorStore.getMetricsStore();
+  const subscribe = useCallback(
+    (callback: () => void) => store.subscribe(callback),
+    [store]
+  );
+  const getSnapshot = useCallback(() => store.getSnapshot(), [store]);
+  const snapshot = useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
+  return useMemo(() => getCanonicalMetrics(snapshot), [snapshot]);
```
nfdiop: Question: do we expect to use this hook everywhere? My worry is that when we do even when an uncorrelated metric changes, a panel can re-render again. For example if I use this metric in the network panel and the cache data changes, I wouldn't expect the panel to re-render.

packages/react-devtools/src/metrics/clientMetrics.ts
```diff
@@ -0,0 +1,165 @@
...
+    avgCachedMs,
+    avgNetworkMs,
+    optimisticCoverage,
+    avgPerceivedSpeedupMs,
+    rollbackRate,
+  };
+}
+
+/** Format a metric as a display string, or a placeholder until it is ready. */
+export function formatMetric(m: Metric): string {
```
nfdiop, line 158: This is nice!!!

packages/react-devtools/src/hooks/useCanonicalMetrics.ts
KuberSethi: Yup used in PerformancePanel!

packages/react-devtools/src/metrics/clientMetrics.ts
```diff
@@ -0,0 +1,169 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { MetricsSnapshot } from "../types/index.js";
+
+export interface Metric {
+  /** The computed value, or undefined until enough samples have accrued. */
```
nfdiop, line 20: nice comment!

nfdiop (dismissed):

.changeset/dt2-devtools-plumbing.md
KuberSethi: done, renamed the changeset and updated the pr description

packages/react-devtools/src/metrics/clientMetrics.test.ts
KuberSethi: added a comment explaining why we polyfill requestIdleCallback

packages/react-devtools/src/metrics/canonicalMetrics.test.ts
KuberSethi: good call, dropped that one

packages/react-devtools/src/plugins/types.ts
KuberSethi: removed

packages/react-devtools/src/plugins/baseTabs.ts
KuberSethi: removed

packages/react-devtools/src/metrics/canonicalMetrics.ts
KuberSethi: switched to N/A

packages/react-devtools/src/metrics/canonicalMetrics.ts
KuberSethi: done, it's value: number | undefined now

packages/react-devtools/src/hooks/useCanonicalMetrics.ts
KuberSethi: added a doc note on the hook. it's meant for one surface that renders all the metrics together, so a shared re-render is expected there. easy to add a selector if we ever read a single metric on its own

nfdiop (approved):

PR #3628, opened 2026-07-06
rework devtools shell to render base tabs with a close control

KuberSethi:
reworks the monitoring panel shell around the base-tab list and a real close control

• renders BASE_TABS and persists the active tab, with a close (X) that collapses to the launcher
• panel always boots closed on page load, independent of persisted geometry
• removes the old compute, intercept, debugging and performance tabs and their dead helpers
• lifts in-panel tooltips above the panel so they stay readable

the large deletion (~5.7k lines) is just those old tabs and their support code: mock editors, the mock data generator, the hook inspector, issue and component cards, their tests, and scss. nothing shared is dropped, and the tabs are replaced by the hardcoded base-tab list plus the close control.


KuberSethi:
closing: the devtools redesign dropped the reworked shell and the plugin system. the tab panels now land as standalone PRs targeting main (#3630 components, #3631 performance, #3632 console, #3633 cache).

PR #3630, opened 2026-07-06
rebuild devtools components tab as an ontology tree

KuberSethi:
rebuilds the components tab to show each mounted component as an ontology tree

• object types, actions, properties and react props a component uses, filterable by name
• derives the ontology from hook usage and the property access tracker
• health badge summarizing wasted renders and over-fetch

## Screenshot

components tab

### DIFF
--- a/packages/react-devtools/src/components/components/ComponentCard.tsx
+++ b/packages/react-devtools/src/components/components/ComponentCard.tsx
@@ -33,6 +33,8 @@ function moreRow(depth: number, hidden: number): React.ReactNode {
   if (hidden <= 0) {
     return null;
   }
+  // TODO: make this row expandable so users can reveal the hidden items instead
+  // of only seeing a static "+ N more" count.
   return <TreeRow depth={depth} leaf label={`+ ${hidden} more`} />;
 }
 
@@ -124,6 +126,8 @@ export const ComponentCard: React.FC<ComponentCardProps> = ({
                   label={group.objectType}
                   count={group.names.length}
                 >
+                  {/* TODO: show each property's read value on hover once the
+                      access tracker captures values, not just names. */}
                   {visible.map((propertyName) => (
                     <TreeRow
                       key={propertyName}
--- a/packages/react-devtools/src/components/components/ComponentCard.tsx
+++ b/packages/react-devtools/src/components/components/ComponentCard.tsx
@@ -27,10 +27,8 @@ interface ComponentCardProps {
   ontology: ComponentOntology;
 }
 
-/** How many leaf rows to show before collapsing the rest into "+N more". */
 const VISIBLE_LEAVES = 5;
 
-/** A muted, non-interactive "+N more" leaf row. */
 function moreRow(depth: number, hidden: number): React.ReactNode {
   if (hidden <= 0) {
     return null;
@@ -38,12 +36,6 @@ function moreRow(depth: number, hidden: number): React.ReactNode {
   return <TreeRow depth={depth} leaf label={`+ ${hidden} more`} />;
 }
 
-/**
- * One live component rendered as an ontology tree: the object types and actions
- * it uses (inferred from hook usage) plus the instances and properties it read
- * and its React props. Health is summarized in the header badge; the per-object
- * detail is available on expand.
- */
 export const ComponentCard: React.FC<ComponentCardProps> = ({
   name,
   ontology,
--- a/packages/react-devtools/src/components/components/ComponentCard.tsx
+++ b/packages/react-devtools/src/components/components/ComponentCard.tsx
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2025 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import classNames from "classnames";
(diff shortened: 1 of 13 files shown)


packages/react-devtools/src/components/components/ComponentCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { TreeRow } from "./TreeRow.js";
+
+import styles from "./ComponentsPanel.module.scss";
+
+interface ComponentCardProps {
+  name: string;
+  ontology: ComponentOntology;
+}
+
+const VISIBLE_LEAVES = 5;
```
KuberSethi, line 30: can change this if we want, looked fine / matched mocks well enough

packages/react-devtools/src/components/components/ComponentsPanel.tsx
```diff
@@ -0,0 +1,167 @@
...
+    <div className={styles.panel}>
+      <div className={styles.toolbar}>
+        <InputGroup
+          leftIcon="search"
+          placeholder="Search ontology"
+          value={search}
+          onChange={(event) => setSearch(event.currentTarget.value)}
+        />
+      </div>
+      <ComponentsHeader
```
nfdiop, line 137: nice decomp of the component

packages/react-devtools/src/components/components/deriveComponentOntology.test.ts
```diff
@@ -0,0 +1,192 @@
...
+      })
+    ).toBe("Parcel");
+    expect(
+      objectTypeOf({ type: "action", actionName: "createParcel" })
+    ).toBeNull();
+  });
+});
+
+describe("deriveComponentOntology", () => {
+  it("surfaces object types and actions from hook usage, not hook names", () => {
```
nfdiop: can be a FLUP but this test name should be changed or we need to use the hooks to show that we are not using the hook names

packages/react-devtools/src/components/components/deriveComponentOntology.test.ts
```diff
@@ -0,0 +1,192 @@
...
+            componentName: "C",
+            propertyName: "b",
+            totalRenders: 10,
+            accessCount: 0,
+          },
+        ],
+      }
+    );
+    expect(result.healthy).toBe(false);
+    expect(result.warning).toContain("3 wasted");
```
nfdiop, line 177: not blocking: it is not clear to me what the "3" means herre

packages/react-devtools/src/components/components/deriveComponentOntology.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+}
+
+export interface OntologyObjectType {
+  name: string;
+  instances: string[];
+}
+
+export interface OntologyProperties {
```
nfdiop: not blocking but I'm not sure what this type represents. Given the name I would expect more than objectTypes

packages/react-devtools/src/fiber/HookStateInspector.ts
```diff
@@ -412,6 +416,63 @@ export interface DiscoveredComponent {
...
   hooks: OsdkHookMetadata[];
   sourceLocation: { fileName?: string; lineNumber?: number } | null;
+  props?: Record<string, string>;
+}
+
+const MAX_PROP_KEYS = 12;
+// Longest a prop-value preview may be before it's truncated, in characters.
+const MAX_PROP_VALUE_LEN = 80;
+
+function summarizePropValue(value: unknown): string {
```
nfdiop, line 428: Let's FLUP and allow a tooltip to show the full value

nfdiop (commented):
Thanks for extracting the components tab :). This is exciting!
 
Notes from testing. These can be FLUPs since the tab is not wired yet.
- we need to remove the caret next to the Object since it is not showing any item under it
- it would be good to show the prop value on hover

Side note: Blueprint has a Tree component we can use instead of our own. Let's FLUP when we get the chance

Image
Image

nfdiop (dismissed):

packages/react-devtools/src/components/components/ComponentCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+      <TreeRow depth={0} icon="widget" label={name} badge={badge}>
+        {ontology.objectTypes.length > 0 ? (
+          <TreeRow
+            depth={1}
+            label="Related objects"
+            count={ontology.objectTypes.length}
+            defaultOpen
+          >
+            {ontology.objectTypes.map((objectType) => {
+              const visible = objectType.instances.slice(0, VISIBLE_LEAVES);
```
nfdiop, line 67: Notes:
- In the office sandbox app, the instances is always empty. Is that expected? 
- FLUP: when I click on the more tabs, I need to be able to see more items

packages/react-devtools/src/components/components/deriveComponentOntology.test.ts
KuberSethi: renamed it to describe what it actually checks

packages/react-devtools/src/components/components/deriveComponentOntology.test.ts
KuberSethi: added a comment explaining the wasted-render count

packages/react-devtools/src/components/components/deriveComponentOntology.ts
KuberSethi: renamed it to ObjectTypeProperties with a doc line

packages/react-devtools/src/fiber/HookStateInspector.ts
KuberSethi: added a todo, we'd need to keep the untruncated value to show it on hover

packages/react-devtools/src/components/components/ComponentCard.tsx
KuberSethi: the empty instances are expected, they only fill in from property reads the tracker sees. added a todo to make the '+ N more' row expandable

KuberSethi:
fixed the empty-node caret. left todos for showing prop values on hover and swapping in blueprint's tree when we get to it

nfdiop (approved):

PR #3631, opened 2026-07-06
fold recommendations into overview

KuberSethi:
surface the performance recommendations as a new section on the overview tab

• add a recommendations section to the overview tab, reusing the shared overview section
• add a client metrics module and hook that compute cache, latency, and optimistic-update effectiveness from the metrics store
• give copy-prompt buttons intent and variant props and stop the copy prompt from emitting missing fields as "undefined"


packages/react-devtools/src/components/performance/PerformancePanel.module.scss
```diff
@@ -0,0 +1,146 @@
...
+  opacity: 0.55;
+  margin-top: 4px;
+}
+
+.section {
+  display: flex;
+  flex-direction: column;
+  gap: 6px;
+
+  // Blueprint minimal buttons default to dark text; keep the copy-prompt
```
KuberSethi: Unsure what we typically do for blueprint button text on dark backgrounds / callouts, thoughts maybe @Leolide ?

packages/react-devtools/src/components/performance/MetricStat.tsx
```diff
@@ -0,0 +1,59 @@
...
+import styles from "./PerformancePanel.module.scss";
+
+interface MetricStatProps {
+  label: string;
+  metric: Metric;
+  /** Plain-language explanation shown on the info tooltip (progressive disclosure). */
+  explain: string;
+}
+
+export const MetricStat: React.FC<MetricStatProps> = ({
```
nfdiop: blocking: Let's re-use the metrics component for the UI to be consistent and match the mocks: https://github.com/palantir/osdk-ts/blob/main/packages/react-devtools/src/components/Metrics.tsx

packages/react-devtools/src/components/performance/PerformancePanel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    "Share of actions that showed an optimistic update before the server responded.",
+  rollbackRate:
+    "Share of actions whose optimistic update had to roll back after the server rejected it. Lower is better.",
+};
+
+interface PerformancePanelProps {
+  monitorStore: MonitorStore;
+}
+
+export const PerformancePanel: React.FC<PerformancePanelProps> = ({
```
nfdiop: I'm confused. Now how is this different from the Overview tab? Which metrics do we expect to see here vs in the Overview tab?

nfdiop (commented):
I appreciate the effort to improve this tab but I don't think we should keep this tab as agreed in the design mocks. Instead, let's keep the metrics inside the `Overview` tab and let's move the Recommendations inside the `Overview` tab as another section

Image

Also the UI looks off:

Screenshot 2026-07-07 at 11 58 10

packages/react-devtools/src/components/performance/MetricStat.tsx
KuberSethi: done, dropped MetricStat. recommendations live on the overview tab now which uses the shared metric/section components

packages/react-devtools/src/components/performance/PerformancePanel.tsx
KuberSethi: removed the duplication, overview is the single home for metrics now and the separate tab is gone

KuberSethi:
done, dropped the separate performance tab. metrics stay on the overview tab and recommendations moved in as their own section

.changeset/dt2-client-metrics.md
nfdiop, line 1: after merging main, we can remove one of these changesets

nfdiop (approved):

PR #3632, opened 2026-07-06
add console base tab with errors and logs views

KuberSethi:
adds the console tab: a structured error feed plus a raw console output view

• aggregates action failures, window errors and console.error into one feed
• dedupes a console.error mirroring a window error, matching on the first arg
• separate logs view for raw console output

## Screenshot

console tab


packages/react-devtools/src/components/console/collectErrors.ts
```diff
@@ -0,0 +1,115 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { MonitorStore } from "../../store/MonitorStore.js";
+import type { Issue } from "../issueTypes.js";
+
+// A console.error within ±100ms of a window error with the same message is
```
nfdiop: (not-blocking) 100ms feels arbitrary. How did we came up with that value?

packages/react-devtools/src/components/console/ConsoleErrorsView.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { collectErrors } from "./collectErrors.js";
+import { ErrorCard } from "./ErrorCard.js";
+
+import styles from "./ConsolePanel.module.scss";
+
+interface ConsoleErrorsViewProps {
+  monitorStore: MonitorStore;
+}
+
+export const ConsoleErrorsView: React.FC<ConsoleErrorsViewProps> = ({
```
nfdiop: Hm, I don't think we need this view. If my understanding is correct, we already allow the user to filter by log type. Also the UI with big red cards is overwhelming

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
```diff
@@ -0,0 +1,95 @@
...
+import React, { useMemo, useState } from "react";
+
+import { useConsoleLogs } from "../../hooks/useConsoleLogs.js";
+import type { ConsoleLogLevel } from "../../store/ConsoleLogStore.js";
+import type { MonitorStore } from "../../store/MonitorStore.js";
+import { LogEntryCard } from "../LogEntryCard.js";
+
+import styles from "./ConsolePanel.module.scss";
+
+const MAX_RENDERED = 200;
```
nfdiop, line 36: Can we have a TODO to virtualise the list and render all as a FLUP?

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <div className={styles.view}>
+      <div className={styles.toolbar}>
+        <InputGroup
+          leftIcon="search"
+          placeholder="Filter logs…"
+          value={search}
+          onChange={(event) => setSearch(event.currentTarget.value)}
+        />
+        <ButtonGroup>
```
nfdiop: Blocking: 
- The text input should fill the full width
- The "Clear" button should be on the right

Question:
- Should we colour code the filter button to match the colours used for the logs

Image

packages/react-devtools/src/components/console/ConsolePanel.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { afterEach, describe, expect, it } from "vitest";
+
+import { createMockMonitorStore } from "../testHelpers.js";
+import { ConsolePanel } from "./ConsolePanel.js";
+
+afterEach(() => {
+  cleanup();
+});
+
+describe("ConsolePanel", () => {
```
nfdiop, line 28: nit: Sorry if I may miss it but I'm not seeing the value of having these tests

packages/react-devtools/src/components/console/ConsolePanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+  monitorStore: MonitorStore;
+}
+
+export const ConsolePanel: React.FC<ConsolePanelProps> = ({ monitorStore }) => {
+  const [view, setView] = useState<ConsoleView>("errors");
+
+  return (
+    <div className={styles.panel}>
+      <div className={styles.segmented}>
+        <ButtonGroup>
```
nfdiop: As mentioned above, let's drop this. The user can filter below 

Image

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <Button size="small" variant="minimal" icon="trash" onClick={clear}>
+          Clear
+        </Button>
+      </div>
+      {filtered.length === 0 ? (
+        <div className={styles.empty}>No console output captured.</div>
+      ) : (
+        <div className={styles.list}>
+          {filtered.slice(-MAX_RENDERED).map((entry) => (
+            <LogEntryCard key={entry.id} entry={entry} />
```
nfdiop: Can be a FLUP (when we wire this component)
- Can we make the card non-interactive? They look clickable but nothing happens when I click on it?

nfdiop (commented):
This is gonna be exciting for people who don't have access to the console logs. Thanks for adding!

Notes from testing:

When I log inside my app, looks like the same log is duplicated and that duplicate comes from something called `installHook.js` (see first image)

Also the UI is not working once I start scrolling (see second image)

Image

Image

packages/react-devtools/src/components/console/collectErrors.ts
KuberSethi: moot now, dropped the errors view and that file with it

packages/react-devtools/src/components/console/ConsoleErrorsView.tsx
KuberSethi: dropped it, the logs view plus the filter covers this

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
KuberSethi: added the todo

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
KuberSethi: done, full-width input, and the filter + clear moved into the funnel with color-coded severity pills

packages/react-devtools/src/components/console/ConsolePanel.test.tsx
KuberSethi: reworked it to check the actual logs view instead

packages/react-devtools/src/components/console/ConsolePanel.tsx
KuberSethi: dropped

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
KuberSethi: done, only the expandable rows are interactive now

KuberSethi:
fixed both. the dupes were react devtools re-emitting through our wrapper, so we skip installHook-sourced calls now. and reworked the layout so the list scrolls in its own container

packages/react-devtools/src/components/console/ConsoleLogsView.tsx
```diff
@@ -0,0 +1,152 @@
...
+          onChange={(event) => setSearch(event.currentTarget.value)}
+        />
+        <div className={styles.issuesBar}>
+          <span className={styles.issuesCount}>{entries.length} issues</span>
+          <Tooltip content="Errors" compact={true}>
+            <span className={classNames(styles.pill, styles.pillError)}>
+              {counts.error}
+            </span>
+          </Tooltip>
+          <Tooltip content="Logs" compact={true}>
```
nfdiop, line 115: not-blocking: when we have the chance, let's also show the info and use the same code color as in the cards

packages/react-devtools/src/store/ConsoleLogStore.ts
```diff
@@ -273,10 +288,12 @@ export class ConsoleLogStore extends SubscribableStore {
         const skipCapture =
           store.suppressed || store.capturing || isBrowserLoggerCall(args);
         const source = skipCapture ? undefined : getCallerLocation();
+        const skip =
```
nfdiop, line 291: nice fixing!

nfdiop (commented):
Thanks for improving the UI. 

Note from testing: 
- the tooltip shows up behind (maybe this will fixed by merging main?)
- the filter should be on the right?

Image

nfdiop:
Tooltip appears now but I cannot click any of the options. Not sure if I am missing something. The `Clear logs` button is also not working

Screenshot 2026-07-07 at 23 41 16

KuberSethi:
> Tooltip appears now but I cannot click any of the options. Not sure if I am missing something. The Clear logs button is also not working

Fixed! Good catch

nfdiop (approved):

PR #3633, opened 2026-07-06
add cache base tab with split inspector and timeline

KuberSethi:
adds the cache tab: the cache entry inspector paired with a live activity timeline

• split view of the cache entry inspector and a cache hit/miss timeline
• draggable divider with a persisted position
• timeline limited to cache hits and misses

## Screenshot

cache tab


packages/react-devtools/src/components/cache/CacheTimeline.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+import { createPollingStore } from "../../hooks/createPollingStore.js";
+import { useMetrics } from "../../hooks/useMetrics.js";
+import type { MonitorStore } from "../../store/MonitorStore.js";
+import type { Operation } from "../../types/index.js";
+import { formatTime } from "../../utils/format.js";
+
+import styles from "./CacheTimeline.module.scss";
+
+const MAX_RECENT_OPERATIONS = 50;
```
KuberSethi, line 30: can lower if desired, think its fine to start but if there are rendering perf issues we can lower

packages/react-devtools/src/components/CacheInspectorTab.tsx
```diff
@@ -316,7 +316,10 @@ export const CacheInspectorTab: React.FC<CacheInspectorTabProps> = ({
                   </Tag>
                   <span className={styles.objectType}>{entry.objectType}</span>
                   {entry.metadata.isOptimistic && (
-                    <Tooltip content="Has optimistic updates">
+                    <Tooltip
```
nfdiop: FLUP to remove after this PR https://github.com/palantir/osdk-ts/pull/3647/changes

packages/react-devtools/src/components/cache/CachePanel.tsx
```diff
@@ -0,0 +1,28 @@
...
+import type { FC } from "react";
+
+import type { MonitorStore } from "../../store/MonitorStore.js";
+import { CacheSplitView } from "./CacheSplitView.js";
+
+interface CachePanelProps {
+  monitorStore: MonitorStore;
+}
+
+export const CachePanel: FC<CachePanelProps> = ({ monitorStore }) => {
```
nfdiop, line 31: nit: is there a reason we don't inline `CacheSplitView` here?

packages/react-devtools/src/components/cache/CacheSplitView.tsx
```diff
@@ -0,0 +1,103 @@
...
+import styles from "./CacheSplitView.module.scss";
+
+const MIN_FRACTION = 0.2;
+const MAX_FRACTION = 0.8;
+
+export interface CacheSplitViewProps {
+  monitorStore: MonitorStore;
+}
+
+export const CacheSplitView: React.FC<CacheSplitViewProps> = ({
```
nfdiop: I like the idea of allowing users to drag these sections. However, I would defer to later and instead have an abstraction that we can use for all tabs. 
Instead, can we use sections here like in the `Overview` tab so that the UI is easier to read ( I can rename `OverviewSection` to `TabSection` so it is easier to use)
See the mocks below.

If short on time, let's start by just showing the CacheInspectorTab

Image

packages/react-devtools/src/components/cache/CacheTimeline.tsx
nfdiop: Let's track a FLUP later to virtualize

packages/react-devtools/src/components/cache/CacheTimeline.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+export interface CacheTimelineProps {
+  monitorStore: MonitorStore;
+}
+
+export const CacheTimeline: React.FC<CacheTimelineProps> = ({
```
nfdiop, line 36: Unfortunately, I am still confused by this UI. I don't have a good solution but we can FLUP with the designer. For now, we can add a Section with description

packages/react-devtools/src/components/cache/CacheSplitView.tsx
KuberSethi: done, replaced the drag split with stacked sections

packages/react-devtools/src/components/cache/CachePanel.tsx
KuberSethi: inlined it

packages/react-devtools/src/components/cache/CacheTimeline.tsx
KuberSethi: it's in its own 'cache history' section now so it reads clearer. happy to add a one-line description if you still want one

packages/react-devtools/src/components/cache/CacheTimeline.tsx
KuberSethi: added a todo to virtualize

packages/react-devtools/src/components/CacheInspectorTab.tsx
KuberSethi: we got #3647 in so i think we're good here

packages/react-devtools/src/components/cache/CachePanel.tsx
```diff
@@ -0,0 +1,42 @@
...
+import styles from "./CachePanel.module.scss";
+
+interface CachePanelProps {
+  monitorStore: MonitorStore;
+}
+
+export const CachePanel: FC<CachePanelProps> = ({ monitorStore }) => {
+  return (
+    <div className={styles.panel}>
+      <OverviewSection title="Cache inspector">
```
nfdiop, line 34: I noticed this adds some padding by default. Will fix it in a separate PR. Otherwise the UI looks off (see image)

Image

nfdiop (approved):

packages/react-devtools/src/components/cache/CachePanel.tsx
KuberSethi: thanks!

PR #3635, opened 2026-07-06
[1/4]: add Overview tab primitives

nfdiop:
## What

Adds an Overview tab to the react-devtools monitoring panel, plus the primitives it's built from:

- `OverviewTab` — an at-a-glance summary of the monitored client's ontology usage and health: object/action/link counts, cache hit rate, network requests, response time, duplicate requests, overfetching, and errors & warnings. Shows a "no ontology" empty state when nothing is registered.
- `Metric` / `Metrics` — a single metric cell and the bordered grid that lays them out, with hairline dividers.
- `MetricLegend` + `metricIntent` — a color key for a metric's help tooltip, and the shared `MetricIntent → Blueprint color` mapping so a value and its legend swatch can't disagree on what a color means.
- `OverviewSection` — a collapsible Blueprint `Section` wrapper for grouping metrics.
- `PanelContainerContext` + `BpTooltip` — portal tooltips/overlays into the panel instead of `document.body`, so they inherit the panel's theme and stacking context. `MonitoringPanel` now provides the context.

## Notes

- The tab runs on hardcoded dummy data (see the `TODO: Fetch the real data later`); the real data hooks are intentionally left out of this change.

## Next
This PR is the first one to split and clean this [PR](https://github.com/palantir/osdk-ts/pull/3634). Next is to wire real data into this PR

Screenshot 2026-07-06 at 15 15 05

### DIFF
--- a/packages/react-devtools/src/components/Metric.module.scss
+++ b/packages/react-devtools/src/components/Metric.module.scss
@@ -1,0 +1,63 @@
+@use 'tokens' as t;
+@use 'mixins' as m;
+
+// A single `Metric` cell — borderless; the enclosing `Metrics` grid draws the
+// dividers around it.
+.metricItem {
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
+  padding: 12px;
+  background: t.$bg-surface;
+}
+
+.metricItemHeader {
+  display: flex;
+  align-items: flex-start;
+  justify-content: space-between;
+  gap: 8px;
+}
+
+.metricHelp {
+  flex-shrink: 0;
+  color: t.$text-tertiary;
+  cursor: help;
+}
+
+// `Metric` help-tooltip body: the definition text plus, for colored metrics, a
+// color key supplied by the caller.
+.metricHelpContent {
+  display: flex;
+  flex-direction: column;
+  gap: 8px;
+  max-width: 220px;
+  line-height: 1.4;
+}
+
+.metricValue {
+  @include m.dt-metric-value;
+
+  &.success {
+    color: t.$green;
+  }
+
+  &.warning {
+    color: t.$orange;
+  }
+
+  &.danger {
+    color: t.$red;
+  }
+
+  // No data yet — muted, so it reads as absent rather than a real measurement.
+  &.na {
+    color: t.$text-tertiary;
+  }
+}
+
+// Push the footer to the bottom of the (equal-height) cell so the "View in …"
+// links line up across a row even when metric labels wrap to different heights.
+.metricFooter {
+  margin-top: auto;
+  padding-top: 6px;
+}
--- a/packages/react-devtools/src/components/Metric.module.scss
+++ b/packages/react-devtools/src/components/Metric.module.scss
@@ -37,17 +37,6 @@
 .metricValue {
   @include m.dt-metric-value;
 
-  &.success {
-    color: t.$green;
-  }
-
-  &.warning {
-    color: t.$orange;
-  }
-
-  &.danger {
-    color: t.$red;
-  }
 
   // No data yet — muted, so it reads as absent rather than a real measurement.
   &.na {
--- a/packages/react-devtools/src/components/Metric.tsx
+++ b/packages/react-devtools/src/components/Metric.tsx
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
-import { Classes, Icon } from "@blueprintjs/core";
-import type { Intent } from "@blueprintjs/core";
+import { AnchorButton, Classes } from "@blueprintjs/core";
 import classNames from "classnames";
 import React from "react";
 
 import { BpTooltip } from "./common/BpTooltip.js";
+import { METRIC_INTENT_COLOR } from "./metricIntent.js";
+import type { MetricIntent } from "./metricIntent.js";
 
 import styles from "./Metric.module.scss";
 
@@ -33,10 +34,10 @@ export interface MetricProps {
    */
   value: string | number | null | undefined;
   /**
-   * Colors the value via Blueprint's `Intent`; defaults to `"none"` (no color).
-   * Only `success`/`warning`/`danger` are colored. Ignored when `value` is empty.
+   * Colors the value using its `MetricIntent` color; defaults to `"none"`
+   * (inherits the surrounding text color). Ignored when `value` is empty.
    */
-  intent?: Intent;
+  intent?: MetricIntent;
   /**
    * When set, renders a top-right "?" icon whose tooltip shows this content — the
    * metric's definition and, where the value is colored, its color key.
@@ -67,7 +68,9 @@ export function Metric({
           <BpTooltip
             content={<div className={styles.metricHelpContent}>{help}</div>}
           >
-            <Icon
+            <AnchorButton
+              variant="minimal"
+              size="small"
               icon="help"
               className={styles.metricHelp}
               aria-label={`About ${title}`}
@@ -76,10 +79,8 @@ export function Metric({
         )}
       </div>
       <span
-        className={classNames(
-          styles.metricValue,
-          isEmpty ? styles.na : styles[intent]
-        )}
+        className={classNames(styles.metricValue, isEmpty && styles.na)}
+        style={isEmpty ? undefined : { color: METRIC_INTENT_COLOR[intent] }}
       >
         {isEmpty ? "N/A" : value}
       </span>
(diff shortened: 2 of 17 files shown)


packages/react-devtools/src/components/Metric.module.scss
```diff
@@ -0,0 +1,63 @@
+@use 'tokens' as t;
+@use 'mixins' as m;
+
+// A single `Metric` cell — borderless; the enclosing `Metrics` grid draws the
+// dividers around it.
+.metricItem {
+  display: flex;
+  flex-direction: column;
+  gap: 4px;
```
nfdiop, line 9: not happy with these values but I can later use bp grid size in the app consistently

packages/react-devtools/src/components/Metric.tsx
```diff
@@ -0,0 +1,91 @@
...
+   * Only `success`/`warning`/`danger` are colored. Ignored when `value` is empty.
+   */
+  intent?: Intent;
+  /**
+   * When set, renders a top-right "?" icon whose tooltip shows this content — the
+   * metric's definition and, where the value is colored, its color key.
+   */
+  help?: React.ReactNode;
+  /** Optional content rendered below the value, e.g. a "View in …" link. */
+  footer?: React.ReactNode;
```
nfdiop, line 47: Image

packages/react-devtools/src/components/common/BpTooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { PanelContainerContext } from "../PanelContainerContext.js";
+
+/**
+ * Blueprint `Tooltip` pre-wired to portal into the devtools panel (via
+ * `PanelContainerContext`) instead of `document.body`, so the tooltip inherits
+ * the panel's theme and stacking context. Takes all the usual `TooltipProps`;
+ * an explicitly passed `portalContainer` still wins.
+ */
+export function BpTooltip({
```
nfdiop, line 29: will FLUP to use in other tooltips. Without this, the tooltip appears behind

packages/react-devtools/src/components/MetricLegend.tsx
```diff
@@ -0,0 +1,57 @@
...
+export interface MetricLegendProps {
+  /** The color-key rows, top to bottom. */
+  entries: readonly MetricLegendEntry[];
+}
+
+/**
+ * A small color key for a `Metric`'s help tooltip, mapping each swatch color to
+ * the value range it denotes.
+ */
+export function MetricLegend({
```
nfdiop, line 40: Image

packages/react-devtools/src/components/common/BpTooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { Tooltip } from "@blueprintjs/core";
+import type { TooltipProps } from "@blueprintjs/core";
+import React from "react";
+
+import { PanelContainerContext } from "../PanelContainerContext.js";
+
+/**
+ * Blueprint `Tooltip` pre-wired to portal into the devtools panel (via
+ * `PanelContainerContext`) instead of `document.body`, so the tooltip inherits
```
KuberSethi, line 25: Why do we need to use a React Context here? We should use them sparingly

packages/react-devtools/src/components/Metric.tsx
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Colors the value using its `MetricIntent` color; defaults to `"none"`
+   * (inherits the surrounding text color). Ignored when `value` is empty.
+   */
+  intent?: MetricIntent;
+  /**
+   * When set, renders a top-right "?" icon whose tooltip shows this content — the
+   * metric's definition and, where the value is colored, its color key.
+   */
+  help?: React.ReactNode;
```
KuberSethi, line 45: nit: rename to tooltip to follow conventions?

packages/react-devtools/src/components/OverviewTab.tsx
```diff
@@ -0,0 +1,218 @@
+/*
```
KuberSethi, line 1: Should we add tests to match the other tabs?

KuberSethi (commented):
the description says the tab's wired into MonitoringPanel behind IS_OVERVIEW_TAB_ENABLED (currently true), but i don't see that flag or any change to the tabs list in the diff. Can you either wire it in or update the description to say its not hooked up yet? 

Also it's marked ready for review but the body says draft until real data lands, might want to flip it back to draft.

packages/react-devtools/src/components/Metrics.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from "react";
+
```
KuberSethi, line 18: nit: remove whitespace

nfdiop: ah it automatically get added by oxc

packages/react-devtools/src/components/OverviewTab.tsx
nfdiop: Didn't add because this contains mostly dummy data - I would need to rewrite everything when the data is wired. Not sure it is worth it

packages/react-devtools/src/components/common/BpTooltip.tsx
nfdiop: Because I need access to the panelRef in all tooltips. Otherwise they show up behind. Couldn't find a better solution. See https://github.com/palantir/osdk-ts/pull/3635#discussion_r3529251585
Screenshot 2026-07-06 at 20 00 09

packages/react-devtools/src/components/Metric.tsx
nfdiop: which conventions? I generally don't name them `tooltip` because then when the container changes and we use a popover instead, it looks off

nfdiop:
> the description says the tab's wired into MonitoringPanel behind IS_OVERVIEW_TAB_ENABLED (currently true), but i don't see that flag or any change to the tabs list in the diff. Can you either wire it in or update the description to say its not hooked up yet?
> 
> Also it's marked ready for review but the body says draft until real data lands, might want to flip it back to draft.

my bad, updated the PR description. Decided to keep IS_OVERVIEW_TAB_ENABLED in another PR

packages/react-devtools/src/components/common/BpTooltip.tsx
KuberSethi: I solved this for another tab in https://github.com/palantir/osdk-ts/pull/3628/changes#diff-a53fd5c8fc910c01c750cc26eb8cc9b4b20297afe303a4fed880811df95451adR402

I think you can just use `portalClassName` + `z-index` rules instead, would strongly bias towards this vs a context, w the context we're basically converting `panelRef` from `useRef` to a `useState` callback ref which'll force an extra render of the createPortal tree on mount i think. This would then mean `panelEl` is null on first render so I think we'd fall back to this bug on initial render anyways. Also not to mention the overhead of contexts / providers. 

I think we should try the more lightweight approach but worst case the context is fine but I'd be surprised if there was no other way out of this.

nfdiop: Fair on keeping contexts sparse. Reason I reached for one here: the tooltip has two problems, not just stacking.

It has to land above the panel AND inside the panel's theme scope. The `--dt-*` vars are scoped to `.panel` via `[data-dt-theme]`, and the tooltip CSS itself lives under the panel (`.bp5-tooltip .bp5-popover-content` in `MonitoringPanel.module.scss`). Portaling into `panelEl` covers both. `portalClassName` + z-index only covers stacking, since the tooltip ends up in `document.body` outside `.panel`, so it drops the vars and that rule, and I'd still have to re-scope the CSS to the portal class and thread the resolved light/dark theme down (prop or context).

On the null-on-first-render thing: I don't think it actually hits tooltips. They only portal when they open on hover, which is well after mount. The callback-ref `useState` fires one extra render on mount, and that's what pushes the real `panelEl` to the consumer, long before anyone hovers, so `portalContainer` isn't null by open time. The null is there for a single render and never reaches the user. You'd only get bitten if the overlay opened on mount, which tooltips don't.

KuberSethi: Fair enough on the perf. piece but I think the theming changes wouldn't be that bad and would let us avoid a lot of tech debt & complexity. Feels like this is an antipattern but to unblock will approve.

KuberSethi (approved):

PR #3637, opened 2026-07-06
Render AipAgentChat as a Storybook folder and align its docs

Leolide:
## Summary

Renders the **AipAgentChat** Storybook story as a proper sidebar folder (matching CbacPicker/ActionForm) and brings its docs in line with the other component docs.

### Storybook

- Story title is `Components/AipAgentChat`; a new `AipAgentChat.mdx` (`Components/AipAgentChat/Docs`) sits alongside it. Having a `Docs` sub-node under the same prefix promotes AipAgentChat to a **folder** in the sidebar — a bare `Components/AipAgentChat` stays a component even with multiple stories. This mirrors the `CbacPicker` pattern exactly.
- The `beta` tag badge is still auto-injected by the indexer in `.storybook/main.ts` for any title under `Components/`.
- Drops the earlier redundant `BaseChat` leaf.

### Docs (`packages/react-components/docs/AipAgentChat.md`)

- Added a **Table of Contents**.
- Aligned section language with sibling docs: `Usage` → `Basic Usage`, `Minimal` → `Minimal Example` (with a lead-in sentence), `Props` → `Props Reference`.
- Verified the doc is up to date against source — props (`AipAgentChatApi.ts`), exports, and the `useChat`/`foundryModel` import paths all match.

### Design tokens

- Cross-checked all 24 `--osdk-aip-agent-chat-*` tokens: names and default values in `CSSVariables.md` match `src/tokens/component-tokens/aip-agent-chat.css` exactly.
- Fixed a stale ToC in `CSSVariables.md` that was missing the **AIP Agent Chat** entry, and pointed the doc's Theming section at the exact token source/anchor.

No changeset for `@osdk/react-components` (docs only); the existing changeset covers `@osdk/react-components-storybook`.

### DIFF
--- a/packages/react-components-storybook/src/stories/AipAgentChat/AipAgentChat.stories.tsx
+++ b/packages/react-components-storybook/src/stories/AipAgentChat/AipAgentChat.stories.tsx
@@ -140,7 +140,7 @@ function InteractiveChat(
 }
 
 const meta: Meta<typeof InteractiveChat> = {
-  title: "Components/BaseChat/AipAgentChat",
+  title: "Components/AipAgentChat/BaseChat",
   component: InteractiveChat,
   render: (args) => (
     <div style={{ height: "600px", maxWidth: "700px" }}>
--- a/packages/react-components-storybook/src/stories/AipAgentChat/AipAgentChat.stories.tsx
+++ b/packages/react-components-storybook/src/stories/AipAgentChat/AipAgentChat.stories.tsx
@@ -140,7 +140,7 @@ function InteractiveChat(
 }
 
 const meta: Meta<typeof InteractiveChat> = {
-  title: "Components/AipAgentChat",
+  title: "Components/BaseChat/AipAgentChat",
   component: InteractiveChat,
   render: (args) => (
     <div style={{ height: "600px", maxWidth: "700px" }}>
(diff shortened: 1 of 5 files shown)


packages/react-components-storybook/src/stories/AipAgentChat/AipAgentChat.stories.tsx
```diff
@@ -140,7 +140,7 @@ function InteractiveChat(
 }
 
 const meta: Meta<typeof InteractiveChat> = {
-  title: "Beta/AipAgentChat",
+  title: "Components/BaseChat/AipAgentChat",
```
nfdiop: I think it should be "Components/AipAgentChat" instead

packages/react-components-storybook/src/stories/AipAgentChat/AipAgentChat.stories.tsx
```diff
@@ -140,7 +140,7 @@ function InteractiveChat(
 }
 
 const meta: Meta<typeof InteractiveChat> = {
-  title: "Beta/AipAgentChat",
+  title: "Components/AipAgentChat/BaseChat",
```
nfdiop: I don't think there is a baseChat. Can we do `"Components/AipAgentChat"

nfdiop (approved):

PR #3638, opened 2026-07-06
Implement CipherText edits

esinx:
## Summary

Reworks `cipherText` object edits so create and update take distinct, ergonomic inputs, and `getEdits()` (and the wire request sent by the writeable client) emits the backend shape.

Previously both `create` and `update` accepted the same `CipherText` type and the value was stored/emitted as-is. There was no way to hand back a fetched `CipherText` to reuse its encrypted envelope, and no way to control the cipher channel on update.

## What changed

**Input types (what callers pass):**
- `create` accepts `{ plaintext }` (`CreateCipherText`) or an existing `CipherText` to reuse.
- `update` accepts `{ plaintext, strategy? }` (`UpdateCipherText`) or an existing `CipherText`. `strategy` is anew `CipherChannelStrategy` enum (`PREFER_EXISTING` (default) / `PREFER_DEFAULT` / `EXISTING_ONLY` / `DEFAULT_ONLY`) controlling which cipher channel is used when re-encrypting. `strategy` is update-only — passing it to `create` is a type error.

**Wire types (what `getEdits()` returns / the writeable client sends):**
- A reused `CipherText` is reduced to its encrypted value `{ ciphertext }` (`CipherTextValue`) — no decryption round-trip.
- A plaintext input passes through unchanged (`{ plaintext }` / `{ plaintext, strategy? }`).

**Implementation:**
- `@osdk/api`: split the property-value maps into create/update × input/wire variants (`PropertyValueWireToCreate`/`WireToUpdate`/`CreateToWire`/`UpdateToWire`) sharing a common base, plus matching `OsdkObject{Create,Update}{,Wire}PropertyType` helpers. Added `CreateCipherText`, `UpdateCipherText`, `CipherTextValue`, and `CipherChannelStrategy`.
- `@osdk/client`: `CipherTextPropertyImpl` now retains the raw encrypted value and exposes an internal `getValue()`; `createOsdkObject` threads the raw value in. Re-exports the new API symbols.
- `@osdk/functions`: edit interfaces now store the *wire* property types while the `EditBatch` `create`/`update` signatures accept the *input* types (new `*InputProps` / `*WireProps` type families in `edits/types.ts`). New `toWireEditProperties` normalizer converts reused `CipherText` instances (and arrays thereof) to `{ ciphertext }` on both the in-memory batch path and the writeable-client transaction path.

## Public API

New exports from `@osdk/api` and `@osdk/client`: `CipherChannelStrategy` (value), `CreateCipherText`, `UpdateCipherText`, `CipherTextValue`, `OsdkObjectCreateWirePropertyType`, `OsdkObjectUpdatePropertyType`, `OsdkObjectUpdateWirePropertyType`. API reports updated.

## Testing

- Unit tests: `edits/toWireEditProperties.test.ts` (normalizer: reuse → `{ ciphertext }`, plaintext passthrough, primitives/nullish/struct untouched, array multiplicity) and additions to `createEditBatch.test.ts` (create/update wire-shape output + `@ts-expect-error` guards that reject the wire-only `{ ciphertext }` input and `strategy` on create).
- E2E sandbox: `runCipherTextTest.ts` exercises fetch/decrypt, `$isNull`/`$eq` filters, and create/update edits(plaintext, plaintext + strategy, reused ciphertext) against the `CipherTextTest` object.

## Changeset

`minor` for `@osdk/api`, `@osdk/client`, `@osdk/functions`.


packages/e2e.sandbox.catchall/src/runPropertyValueFormattingTest.ts
esinx, line 1: This got picked up as a part of the formatting commit, can be disregarded.

packages/api/src/mapping/PropertyValueMapping.ts
esinx, line 1: CipherText is the first instance of a type that has a different create / update interface, so additional interfaces were declared to distinguish them.

packages/client/src/createCipherTextProperty.ts
```diff
@@ -23,16 +23,19 @@ import type { MinimalClient } from "./MinimalClientContext.js";
 export class CipherTextPropertyImpl implements CipherText {
   #client: MinimalClient;
   #locator: [string, any, string];
+  #value: string;
```
esinx, line 26: Internal value stored for copy assignments

packages/api/src/object/CipherText.ts
```diff
@@ -32,4 +32,47 @@ export interface CipherText {
...
+   * to forward an existing ciphertext value into an object edit without
+   * decrypting it.
+   *
+   * @internal
+   * @returns the encrypted value
+   */
+  getValue(): string;
+}
+
+// TODO: re-export from @osdk/foundry.ontologies
```
nihalbhatnagar: Can we do this now?

esinx: Yup will do

packages/api/src/Definitions.ts
```diff
@@ -58,3 +63,48 @@ export type OsdkObjectCreatePropertyType<
...
+      MaybeArray<T, GetCreateWirePropertyValueFromWire<T["type"]>>
+    >;
+
+/**
+ * The wire type emitted for an update object property edit.
+ *
+ * @param T - ObjectMetadata.Property in literal form
+ * @param STRICTLY_ENFORCE_NULLABLE - S for strict. If false, always `|undefined`
+ */
+export type OsdkObjectUpdateWirePropertyType<
```
nihalbhatnagar, line 78: Should these types just live in osdk/functions? whats the upside of having them here

nihalbhatnagar: NVM I think this is fine

packages/cli/src/util/configLoader.ts
```diff
@@ -20,10 +20,10 @@ import { loadFoundryConfig } from "@osdk/foundry-config-json";
 
 let siteConfigPromise:
   | Promise<LoadedFoundryConfig<"site"> | undefined>
-  | undefined = undefined;
+  | undefined;
```
nihalbhatnagar, line 23: Are these like from the new lint rules? Why are they showing up now

packages/e2e.sandbox.officenetwork/src/components/LoadingIndicator.tsx
```diff
@@ -48,7 +48,7 @@ export function StatusBadge({
   if (isLoading) {
     return (
       <span className="officenetwork-badge officenetwork-badge-info flex items-center gap-1.5">
-        <LoadingIndicator isLoading size="sm" />
+        <LoadingIndicator isLoading={true} size="sm" />
```
nihalbhatnagar, line 51: Why this?

packages/functions/src/edits/toFunctionEditValue.ts
```diff
@@ -0,0 +1,61 @@
+/*
```
nihalbhatnagar, line 1: Can we rename this file to `toFunctionEditValue` or something to convey this is what the function runtime expects, I feel like the wire term is super overloaded

packages/api/src/mapping/PropertyValueMapping.ts
```diff
@@ -122,6 +121,34 @@ export interface PropertyValueWireToCreate {
...
+
+/**
+ * User facing update types for each properties
+ */
+export interface PropertyValueWireToUpdate extends PropertyValueWireToCreateBase {
+  cipherText: UpdateCipherText | CipherText;
+}
+
+/**
+ * Wire facing creation types for each properties
```
nihalbhatnagar: Yeah like can we a bit more explicit about what "wire means in the comment"

packages/api/src/mapping/PropertyValueMapping.ts
```diff
@@ -60,9 +62,6 @@ export type GetClientPropertyValueFromWire<
     ? { [K in keyof T]: PropertyValueWireToClient[T[K]] }
     : never;
 
-/**
- * Map from the PropertyDefinition type to the typescript type that we accept
```
nihalbhatnagar, line 64: Wait why delete these

packages/api/src/mapping/PropertyValueMapping.ts
nihalbhatnagar: Specifically maybe flag that this is what types are accepted by the backend for edits

packages/api/src/object/CipherText.ts
```diff
@@ -32,4 +38,39 @@ export interface CipherText {
...
+
+  /**
+   * Returns the raw encrypted value backing this ciphertext. Used internally
+   * to forward an existing ciphertext value into an object edit without
+   * decrypting it.
+   *
+   * @internal
+   * @returns the encrypted value
+   */
+  getValue(): string;
```
nihalbhatnagar: Discussed offline but lets leave this off the type

nihalbhatnagar (approved):

PR #3643, opened 2026-07-06
fix: stop listogram filter rows reordering when a checkbox is toggled

nfdiop:
## What
Clicking a checkbox in a listogram filter no longer moves the clicked row. Selection was used as a sort key, so the list was partitioned selected-first on every render and toggling snapped the row to the top.

## Fix
- Render values in their natural count/value order regardless of selection.
- Collapsed view: show the first `maxVisibleItems` in natural order, then append any *selected* values below that slice at the tail (Workshop-style) — no hoisting, no displacing the head.
- `hasMore` and "View all (N)" read the natural-order list (unchanged count).
- Selection still drives checked state, aria-pressed, and filtered-out styling — it just no longer affects ordering.

Confined to `ListogramInput`; no public props change; no other filter input touched.

## Tradeoff (intentional)
Unchecking a **below-the-fold** row in the collapsed view drops it from view immediately (it's no longer selected and sits past the fold). "View all" brings it back. The row briefly vanishes under the cursor rather than staying put for a re-check in place.

This is a deliberate choice to keep the component simple. An earlier revision preserved the row by tracking unchecked below-fold values in extra instance state, but that added real complexity — a monotonically growing state set, a head-vs-below-fold distinction in the toggle handler, and pins that persisted across data refreshes (aggregation counts change whenever a sibling filter changes). The behavior isn't worth that surface area for a below-fold row. **We can bring the pin behavior back if a user actually asks for it.**

Head rows are unaffected — they never vanish on toggle.

## Tests
- New `ListogramInput.test.tsx`: order stable across toggle on/off, natural order regardless of selection, below-fold selection visible at the tail under `maxVisibleItems`, and a below-fold row dropping from the collapsed view once unchecked. Sibling tests unaffected.
- Storybook interaction tests (`WithCheckbox`, `WithBelowFoldSelection`) covering the same paths at the FilterList integration layer.

### Before
bug_sorting

### After
sorting_stopped

below_fold_bug

### DIFF
--- a/packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
+++ b/packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
@@ -30,7 +30,7 @@ import { ObjectTable } from "@osdk/react-components/experimental/object-table";
 import type { Meta, StoryObj } from "@storybook/react-vite";
 import { useCallback, useMemo, useState } from "react";
 import { useArgs } from "storybook/preview-api";
-import { fn } from "storybook/test";
+import { expect, fn, userEvent, waitFor, within } from "storybook/test";
 
 import { fauxFoundry } from "../../mocks/fauxFoundry.js";
 import { Employee } from "../../types/Employee.js";
@@ -1278,6 +1278,52 @@ export const WithCheckbox: Story = {
     },
   },
   render: (args) => <WithCheckboxStory {...args} />,
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+    const expectedDepartmentOrder = [
+      "Engineering",
+      "Marketing",
+      "Design",
+      "Data",
+      "Finance",
+    ];
+
+    const visibleDepartmentOrder = () =>
+      canvas
+        .getAllByRole("button", {
+          name: /^(Engineering|Marketing|Design|Data|Finance)\s+\d+/u,
+        })
+        .map((row) => {
+          const label = expectedDepartmentOrder.find((name) =>
+            row.textContent?.includes(name)
+          );
+          if (label == null) {
+            throw new Error(
+              `Unable to identify department row from "${row.textContent}"`
+            );
+          }
+          return label;
+        });
+
+    await canvas.findByRole("button", { name: "Marketing 4" });
+    await expect(visibleDepartmentOrder()).toEqual(expectedDepartmentOrder);
+
+    await userEvent.click(canvas.getByRole("button", { name: "Marketing 4" }));
+    await waitFor(() =>
+      expect(
+        canvas.getByRole("button", { name: "Marketing 4" })
+      ).toHaveAttribute("aria-pressed", "true")
+    );
+    await expect(visibleDepartmentOrder()).toEqual(expectedDepartmentOrder);
+
+    await userEvent.click(canvas.getByRole("button", { name: "Marketing 4" }));
+    await waitFor(() =>
+      expect(
+        canvas.getByRole("button", { name: "Marketing 4" })
+      ).toHaveAttribute("aria-pressed", "false")
+    );
+    await expect(visibleDepartmentOrder()).toEqual(expectedDepartmentOrder);
+  },
 };
 
 function WithRemovableFiltersStory(args: Partial<EmployeeFilterListProps>) {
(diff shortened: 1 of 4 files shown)


packages/react-components/src/filter-list/base/inputs/ListogramInput.tsx
```diff
@@ -95,19 +91,66 @@ function ListogramInputInner({
...
   }, [stableValues, searchQuery, renderValue]);
 
-  const sortedValues = useMemo(() => {
-    const selected = filteredValues.filter((v) => selectedSet.has(v.value));
-    const unselected = filteredValues.filter((v) => !selectedSet.has(v.value));
-    return [...selected, ...unselected];
-  }, [filteredValues, selectedSet]);
+  const toggleValue = useCallback(
+    (value: string) => {
+      if (selectedSet.has(value)) {
```
nfdiop, line 79: to match Workshop behaviour. Otherwise the value can disappear below the fold

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1278,6 +1278,103 @@ export const WithCheckbox: Story = {
...
+        code: `<FilterList
+  objectType={Employee}
+  filterDefinitions={[
+    { type: "PROPERTY", key: "department", label: "Department", filterComponent: "LISTOGRAM", filterState: { type: "EXACT_MATCH", values: ["Sales"] } },
+  ]}
+/>`,
+      },
+    },
+  },
+  render: (args) => <WithBelowFoldSelectionStory {...args} />,
```
nfdiop, line 1377: I didn't add a "Play" because I want devs to be able to click the story and see the behaviour. If we had the ability to Collapse again the list, I would have added it

packages/react-components/src/filter-list/base/inputs/__tests__/ListogramInput.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+    expect(getRenderedOrder()).toEqual(["Engineering", "Sales", "Support"]);
+
+    fireEvent.click(supportRow!);
+
+    expect(queryRow("Support")).toBeNull();
+    expect(getRenderedOrder()).toEqual(["Engineering", "Sales"]);
+    expect(screen.getByRole("button", { name: "View all (4)" })).not.toBeNull();
+  });
+
+  it("shows the View all button in the collapsed view and expands to all values", () => {
```
nfdiop, line 111: this and below were missing tests

packages/react-components/src/filter-list/base/inputs/__tests__/ListogramInput.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+  { value: "Marketing", count: 12 },
+  { value: "Support", count: 6 },
+];
+
+describe("ListogramInput ordering", () => {
+  it("keeps row order unchanged when a checkbox is toggled", () => {
+    render(<ControlledListogram values={mockValues} maxCount={42} />);
+
+    const before = getRenderedOrder();
+    expect(before).toEqual(["Engineering", "Sales", "Marketing", "Support"]);
```
ziyunp, line 38: nit: Can we keep the expected order / natural order in a constant var? 

const EXPECTED_ORDER = ["Engineering", "Sales", "Marketing", "Support"]

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
ziyunp: You can write the play tests to end in the initial state. I think we are missing the View Less button. Let's do this in a FLUP

ziyunp (approved):

packages/react-components/src/filter-list/base/inputs/__tests__/ListogramInput.test.tsx
nfdiop: So I didn't save because it wasn't always the same variable

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
nfdiop: sounds good!

PR #3645, opened 2026-07-06
Extract cache tab

nfdiop:
## What

Promotes the cache inspector out of the collapsible section nested inside the **Debugging** tab and into its own top-level Devtools tab.

- `MonitoringPanel` now renders `CacheInspectorTab` as a dedicated tab alongside Performance, Compute, Intercept, and Debugging.
- `DebuggingTab` drops the collapsible **Cache** section and its supporting polling state (`cacheCount`), since the inspector is no longer nested there.

## Why

`CacheInspectorTab` is already a self-contained tab (own title, stats, search, and clear/refresh controls), so nesting it behind a collapsible section inside Debugging was redundant. Surfacing it as a first-class tab makes it easier to reach.

cache_Demo

### DIFF
--- a/packages/react-devtools/src/components/DebuggingTab.tsx
+++ b/packages/react-devtools/src/components/DebuggingTab.tsx
@@ -24,7 +24,6 @@ import { useConsoleLogs } from "../hooks/useConsoleLogs.js";
 import type { MonitorStore } from "../store/MonitorStore.js";
 import type { ComponentHookBinding } from "../utils/ComponentQueryRegistry.js";
 import { formatTime } from "../utils/format.js";
-import { CacheInspectorTab } from "./CacheInspectorTab.js";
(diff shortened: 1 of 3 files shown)


packages/react-devtools/src/components/DebuggingTab.tsx
```diff
@@ -186,7 +185,6 @@ function collectIssues(monitorStore: MonitorStore, now: number): Issue[] {
 
 export const DebuggingTab: React.FC<DebuggingTabProps> = ({ monitorStore }) => {
   const [searchQuery, setSearchQuery] = useState("");
-  const [cacheExpanded, setCacheExpanded] = useState(false);
```
nfdiop, line 189: Removed from here since it is now first-class. I wouldn't expect users to complain

nfdiop:
Deferring to @KuberSethi

PR #3646, opened 2026-07-06
Extract active components list into its own Components tab

nfdiop:
## What

Moves the active-components list out of the **Debugging** tab into its own top-level **Components** tab in the devtools `MonitoringPanel`.

- New `ComponentsTab` — self-contained: its own search box, `useActiveComponents`, and issue polling to drive the per-component health dots in `ComponentCard`. Takes only `monitorStore`.
- Extracted the pure `collectIssues` helper into `collectIssues.ts` so both `DebuggingTab` and `ComponentsTab` derive issues from one source instead of duplicating the logic.
- `DebuggingTab` now keeps Errors / Console / Improvements / Cache; the Components section and its now-dead supporting code are removed.
- `MonitoringPanel` gains the `components` tab button and content pane.

## Note on a dropped interaction

The old Components section wired `handleComponentClick` into the **error** `IssueCard`s so clicking a component tag scrolled to it. Error issues never carry a `componentId` (only wasted-render / unused-field issues do, and those weren't rendered as cards in the Errors section), so that handler never fired — it's dropped rather than preserved. Component-linking from issues into the new tab would be a separate, deliberate feature.

## Verification

Typecheck, lint (oxlint + oxfmt), and the package test suite all pass.

### DIFF
--- a/.changeset/witty-melons-cheer.md
+++ b/.changeset/witty-melons-cheer.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-devtools": patch
+---
+
+Move the active components list out of the Debugging tab into its own Components tab in the devtools panel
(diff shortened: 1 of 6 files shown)


nfdiop:
Deferring to @KuberSethi

PR #3647, opened 2026-07-06
isolate react-devtools styling in a shadow root

KuberSethi:
enabling the devtools overrode the host app's brand tokens because it injected an unlayered copy of blueprint css globally into the page

• render the panel and click-to-inspect overlay into a shadow root, injecting all devtools + blueprint css there instead of document.head
• rewrite blueprint's :root/html/body rules to :host and reset inheritance at the boundary so app and devtools styles stay independent both ways
• route blueprint overlays into the shadow root via PortalProvider, drop the vite plugin's global stylesheet, and fix the panel's dead bp5 overrides to bp6


packages/react-devtools/src/styles.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-// Placeholder. scripts/build-css.mjs replaces the transpiled
-// build/esm/styles.js with a runtime-injection module that adds devtools
-// styles (including Blueprint) to document.head when the package is
-// imported.
+// Placeholder. scripts/build-css.mjs overwrites the transpiled
```
ziyunp: Do we need a placeholder file? The build-css script should be able to create a new file?

packages/react-devtools/src/vite/index.ts
```diff
@@ -39,18 +33,15 @@ export interface OsdkDevToolsOptions {
   verbose?: boolean;
 
   /**
-   * Automatically inject CSS.
-   * Default: true
+   * @deprecated No longer used. Devtools styles load into a shadow root at
```
ziyunp: No need to deprecate, can just remove

packages/react-devtools/src/shadow/ShadowHost.ts
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { devtoolsCss } from "../styles.js";
+
+export const HOST_ID = "__osdk_react_devtools_host__";
+const MOUNT_ATTR = "data-osdk-devtools-mount";
+
+let cachedMount: HTMLElement | null = null;
```
ziyunp: Is there a way to write this without a global let var?

packages/react-devtools/src/styles.ts
ziyunp: Oh got it, ShadowHost is importing the devtoolsCss from here, hmmm

ziyunp (dismissed):

ziyunp:
Approved to unblock the fix

ziyunp (approved):

PR #3651, opened 2026-07-07
feat: make Components Overview tabs fill full available height

Leolide:
## Summary

Makes the **Components / Overview** Storybook page use the full available height so the `ObjectTable`, `FilterList`, and `DocumentViewer` fill the viewport instead of being capped at fixed pixel heights.

## Changes

- Root container: `100vh` flex column with a `flex: 1` tab-content area, so tabs resolve a real height.
- **Data** tab: inner container `height: 100%` (was fixed `500`) → FilterList stretches full height, ObjectTable fills the remaining width/height.
- **Viewers** tab: inner container `height: 100%` (was fixed `400`) → DocumentViewer fills.

The components themselves already set `height: 100%` on their roots; they just needed a sized parent chain to resolve against.

## Notes

- `@osdk/react-components-storybook` is `private`, so no changeset is required.

### DIFF
--- a/packages/react-components-storybook/src/stories/Overview/Overview.stories.tsx
+++ b/packages/react-components-storybook/src/stories/Overview/Overview.stories.tsx
@@ -306,7 +306,7 @@ function DataTab(): React.ReactElement {
   const [collapsed, setCollapsed] = useState(false);
 
   return (
-    <div style={{ display: "flex", height: 500 }}>
+    <div style={{ display: "flex", height: "100%", minHeight: 0 }}>
       <div
         style={{
           ...FILTER_WRAPPER_STYLE,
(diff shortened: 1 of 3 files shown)


packages/react-components-storybook/src/stories/Overview/Overview.stories.tsx
```diff
@@ -427,11 +427,19 @@ function ComponentOverview(): React.ReactElement {
   );
 
   return (
-    <div>
+    <div
+      style={{
+        display: "flex",
+        flexDirection: "column",
+        height: "100vh",
```
nfdiop: nit: not best practice to use "vh" because it doesn't always work well in every display

nfdiop (dismissed):

nfdiop (approved):
Tested every story since we are making a global CSS change and looks to be working like before. Approving

PR #3655, opened 2026-07-07
Add interaction tests for ObjectTable stories

ziyunp:
Add play functions covering all ObjectTable stories.

### DIFF
--- a/packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
@@ -31,7 +31,15 @@ import type {
 } from "@osdk/react-components/experimental/object-table";
 import type { Meta, StoryObj } from "@storybook/react-vite";
 import { useCallback, useState } from "react";
-import { expect, fn, userEvent, waitFor, within } from "storybook/test";
+import {
+  expect,
+  fireEvent,
+  fn,
+  screen,
+  userEvent,
+  waitFor,
+  within,
+} from "storybook/test";
 
 import { fauxFoundry } from "../../mocks/fauxFoundry.js";
 import { Employee } from "../../types/Employee.js";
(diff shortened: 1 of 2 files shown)


packages/react-components-storybook/src/stories/ObjectTable/ObjectTable.stories.tsx
```diff
@@ -430,6 +438,122 @@ const columnDefinitions: ColumnDefinition<
...
+ * toggling selection re-renders the rows, and any checkbox refs captured
+ * earlier could go stale.
+ */
+const findRowCheckboxes =
+  (root: Canvas): (() => Promise<HTMLElement[]>) =>
+  () =>
+    root.findAllByRole("checkbox", { name: /select row/iu });
+
+const findSelectAllCheckbox = (root: Canvas): Promise<HTMLElement> =>
+  root.findByRole("checkbox", { name: /select all rows/iu });
```
olliejenningsPT: `/select all rows/iu` is a substring match, so it also matches "**De**select all rows". This reintroduces the ambiguity the previous exact-string version specifically guarded against (the comment removed in this PR read: *"exact-string match so it doesn't also match 'Deselect all rows'"*).

It won't fail today, since there is only ever one header checkbox and nothing is selected at the call site, so `findByRole` returns the single match. But `MultipleSelection` no longer verifies that the label actually flips back to "Select all rows" after deselecting; if the component regressed and left it labeled "Deselect all rows", this helper would still match and the test would stay green.

Suggest an exact string (`name: "Select all rows"`) or an anchored regex (`/^select all rows$/iu`). `findDeselectAllCheckbox` is fine as-is, since "Select all rows" does not contain "deselect all rows".

ziyunp: Ah good catch!

olliejenningsPT (dismissed):

olliejenningsPT (approved):

PR #3656, opened 2026-07-07
render only components, console, and cache tabs in devtools

KuberSethi:
trims the osdk react devtools panel down to three core tabs

• MonitoringPanel now renders only Components, Console, and Cache, replacing the Performance, Compute, Intercept, and Debugging tabs
• Console renders the ConsolePanel added in #3632
• Components is the default active tab; the Overview component stays in the tree but is intentionally not wired up yet
• OverviewSection interior padding is now opt-in via a padded prop, defaulting off so the cache sections render flush

### DIFF
--- a/packages/react-devtools/src/components/MonitoringPanel.tsx
+++ b/packages/react-devtools/src/components/MonitoringPanel.tsx
@@ -36,11 +36,11 @@ import { usePersistedState } from "../hooks/usePersistedState.js";
 import { getDevtoolsShadowMount } from "../shadow/ShadowHost.js";
 import type { MonitorStore } from "../store/MonitorStore.js";
 import type { PanelPosition } from "../types/index.js";
-import { ComputeTab } from "./ComputeTab.js";
-import { DebuggingTab } from "./DebuggingTab.js";
-import { InterceptTab } from "./InterceptTab.js";
+import { CachePanel } from "./cache/CachePanel.js";
+import { ComponentsPanel } from "./components/ComponentsPanel.js";
+import { ConsolePanel } from "./console/ConsolePanel.js";
 import { MonitorErrorBoundary } from "./MonitorErrorBoundary.js";
-import { PerformanceTab } from "./PerformanceTab.js";
+import { OverviewTab } from "./OverviewTab.js";
 
 import styles from "./MonitoringPanel.module.scss";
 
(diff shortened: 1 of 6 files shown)


packages/react-devtools/src/components/MonitoringPanel.tsx
```diff
@@ -99,10 +99,10 @@ const UI_CONSTANTS = {
 };
 
 const DEVTOOLS_TAB_IDS = [
-  "performance",
-  "compute",
-  "intercept",
-  "debugging",
+  "overview",
```
nfdiop: let's not add the overview yet. wiring the data here https://github.com/palantir/osdk-ts/pull/3644/changes

nfdiop (commented):
Could you also add `padded` to the Overview tab and disable by default?

packages/react-devtools/src/components/MonitoringPanel.tsx
```diff
@@ -36,11 +36,10 @@ import { usePersistedState } from "../hooks/usePersistedState.js";
 import { getDevtoolsShadowMount } from "../shadow/ShadowHost.js";
 import type { MonitorStore } from "../store/MonitorStore.js";
 import type { PanelPosition } from "../types/index.js";
-import { ComputeTab } from "./ComputeTab.js";
-import { DebuggingTab } from "./DebuggingTab.js";
-import { InterceptTab } from "./InterceptTab.js";
+import { CachePanel } from "./cache/CachePanel.js";
```
nfdiop, line 39: Could you remove the extra padding around the cache sections. Don't want to merge and someone drafts a release. I'm thinking making the padding optional or you can remove and I can add back https://github.com/palantir/osdk-ts/blob/7a16c41a8b3156b54466ae553a6c492faa38ab15/packages/react-devtools/src/components/OverviewSection.module.scss#L14

KuberSethi: Good call, done!

packages/react-devtools/src/components/OverviewSection.module.scss
```diff
@@ -11,7 +11,12 @@
   }
 
   .overviewSectionCard {
-    padding: 12px;
     background: t.$bg-surface;
+
+    // Interior padding is opt-in via the `padded` prop; sections whose content
+    // owns its own spacing (e.g. the cache lists) render flush without it.
+    &.padded {
```
nfdiop, line 18: thanks!

nfdiop (approved):

PR #3662, opened 2026-07-09
Add hidden React (Experimental) create-app template

nfdiop:
## What

Adds a new **hidden, 2.x-only** `create-app` template — `@osdk/create-app.template.react-experimental` — forked from `react.beta`. It's a batteries-included starting point for the newer OSDK React ecosystem: the scaffolded app ships with `@osdk/react-components` and the `@osdk/react-devtools` Vite plugin pre-wired.

Reachable via:

```
npm create @osdk/app@latest -- --template react-experimental
```

It's `hidden`, so it doesn't appear in the interactive picker — you have to ask for it by name (like the `tutorial-*` and `typescript-library` templates).

## What's wired

Same Home/starter UX as `react.beta`, plus:

- `@osdk/react-components` (with `@osdk/aip-core` + `classnames` peers) — styles imported in `index.css` via `@import "@osdk/react-components/styles.css"`, following the package's documented CSS setup
- `#root { isolation: isolate }` — required for the Base UI portals `@osdk/react-components` relies on
- `@osdk/react-devtools` — `osdkDevTools()` added to the Vite plugins (dev-only, no-op in production builds)

## Notes

- **2.x only.** `codegen.mjs` gets an `only2x` flag that suppresses the 1.x loader, so the un-suffixed package name doesn't resolve as both SDK versions.
- **`react-experimental` (hyphen), not `react.experimental`.** `create-app`'s project-name validation rejects dots, and both the CLI tests and `example-generator` derive project names from the template id. The template package is private/unpublished, so the name is internal-only.
- Includes the generated `example-react-experimental-sdk-2.x` reference app.


packages/create-app.template.react-experimental/templates/package.json.osdk.hbs
```diff
@@ -0,0 +1,33 @@
...
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "{{osdkPackage}}": "latest",
+    "@osdk/aip-core": "^0.8.0",
+    "@osdk/client": "{{clientVersion}}",
+    "@osdk/foundry": "latest",
+    "@osdk/oauth": "^1.1.0",
+    "@osdk/react": "{{clientVersion}}",
+    "@osdk/react-components": "^0.37.0",
```
olliejenningsPT, line 23: These OSDK-adjacent pins (`@osdk/react-components`, `@osdk/aip-core`, `@osdk/react-devtools`, and `classnames`) live only in the `.hbs`, not in the template package's own `package.json`. That means they aren't governed by monorepolint and aren't `Object.assign`-overridden by the packager, so they'll silently drift stale when those packages are version-bumped in the monorepo. CI won't catch it, since the generated example resolves them to `workspace:*`.

Contrast with `@osdk/client`/`@osdk/react` (dynamic `{{clientVersion}}`, always fresh) and `@blueprintjs/*` (declared in the template package's `package.json`, so monorepolint keeps it current). Consider moving these into the template package's `package.json` so monorepolint governs them, assuming that doesn't trip knip's unused-dependency check on the template package.

packages/example-generator/src/run.ts
```diff
@@ -454,6 +454,21 @@ const UPDATE_PACKAGE_JSON: Mutator = {
         /"@osdk\/oauth": "\^.*?"/,
         `"@osdk/oauth": "workspace:*"`,
       )
+      .replace(
+        // Use the local package in the monorepo
+        /"@osdk\/react-components": "\^.*?"/,
```
olliejenningsPT, line 459: Nit: these three new replacements match `"\^.*?"` (caret only), whereas the sibling `@osdk/client`/`@osdk/react` replacements above use `"[\^~].*?"`. Harmless today since all three pins use `^`, but if any pin ever switches to `~` it would silently fail to be rewritten to `workspace:*`. Using `[\^~]` here too would be more robust and consistent.

olliejenningsPT (commented):
A few non-blocking notes from review:

**Missing `.vscode/`.** `react.beta`'s template ships `templates/.vscode/{settings.json,extensions.json}` (format-on-save + recommended extensions), and it propagates to its generated example. This fork drops `.vscode/` entirely, so `example-react-experimental-sdk-2.x` and user-generated apps won't get it. Every other scaffold file is a faithful copy of `react.beta`, so this looks like an accidental omission rather than a deliberate choice. If it's intentional, feel free to ignore; otherwise consider copying `.vscode/` over for parity.

(Two inline comments below on dependency pinning and a regex nit.)

PR #3663, opened 2026-07-10
Add $includeAllBaseObjectProperties option to useObjectSet and observeObjectSet

nfdiop:
## What

Adds an opt-in `$includeAllBaseObjectProperties` option to `useObjectSet` (`@osdk/react`) and `observeObjectSet` (`@osdk/client`).

For Interface Object Sets, returned rows remain interface views (`$apiName` is the interface). When the option is enabled, `row.$as(ConcreteType)` also exposes the implementing object's complete Base Object Properties. The default interface-only behavior is unchanged.

## Behavior

- Defaults off and resubscribes when the hook option changes.
- Forwards the option through `observeObjectSet` to `fetchPage` only when the Object Set may resolve to an interface.
- Handles filtered and type-preserving wrappers, `narrowToType`, pivots, and every operand of unions, intersections, and subtractions.
- Separates normalized cache entries for each interface-only result shape and for all-base-property results.
- Merges shared-property updates across those shapes without stripping shape-specific fields, while keeping ordinary object/list entries current.
- Has no effect on concrete Object Sets.

## Manual example

Adds `/person-interface-properties` to the peopleapp sandbox. The page compares the default and opt-in results side by side, limits each panel to five rows, and shows the existing RDP result in both variants without claiming additional RDP support.

Browser verification confirmed:

- both panels keep `$apiName === "Person"`;
- the default panel does not expose the Employee-only full name;
- the opt-in panel exposes the full name through `$as(Employee)`;
- the existing RDP has the same result in both panels;
- no console errors were emitted.

## Testing

- Client Object Set tests: 29 passed.
- React `useObjectSet` tests: 23 passed.
- Peopleapp component tests: 2 passed.
- Client and peopleapp typechecks passed.
- Client and React API Extractor checks passed.
- Global `pnpm turbo transpile`: 264 tasks passed.

## Out of scope

- Broader interface Runtime Derived Property support, including preservation across `$as(...)`.
- Local membership reconciliation for interface Object Sets.


packages/client/src/observable/internal/objectset/ObjectSetHelper.ts
```diff
@@ -179,6 +185,98 @@ export class ObjectSetHelper extends AbstractHelper<
...
 
+    // The flag is interface-only on the server. Keep it only when the base set
+    // resolves to an interface, so it never fragments the cache over a no-op flag
+    // for a Base Object Set. This ignores pivotTo, so a pivot onto objects can
+    // still send the flag; gating on the true post-pivot type needs async
+    // metadata resolution on this synchronous path, which isn't worth it. Results
+    // stay correct (the server ignores the flag); the cost is cache fragmentation.
+    if (
+      options.$includeAllBaseObjectProperties &&
+      baseSetMayResolveToInterface(
```
nfdiop, line 196: Honestly, I want to drop this check since it adds complexity (see function below + tests). How bad would it be to allow cache fragment and allowing potentially the same data in the cache? How often do users do this?

PR #3664, opened 2026-07-10
fix: escape comment terminators in generated JSDoc

tdeitch:
## Problem

`@osdk/generator` interpolates ontology descriptions and display names verbatim into generated JSDoc block comments. A block comment is terminated by the first `*/` sequence, so when a description contains `*/`, it prematurely closes the comment and leaves the remainder of the text as invalid TypeScript — breaking SDK generation.

This was hit in the wild by a query (pro-code function) parameter whose description contained a `*/` (e.g. text like ``a/*/b``). Generation failed while formatting the generated query file:

```
SyntaxError: Property or signature expected.
  23 |                 /**
> 24 |  *   description: ... (a/*/b) ...
  25 |  */
  26 | readonly "dryRun"?: QueryParam.PrimitiveType<"boolean">,
    at generateV2QueryFile (@osdk/generator/.../generatePerQueryDataFiles.js)
```

## Fix

Add a shared `escapeJsDocText` helper that neutralizes the comment terminator (`*/` → `*\/`, which is also TSDoc-correct, so renderers still show the original text). Apply it at every site that embeds ontology metadata into a generated comment:

- `getDescriptionIfPresent` — query & action descriptions, action parameter descriptions
- `queryParamJsDoc` — query parameter descriptions
- `propertyJsdoc` — object/interface property descriptions, display names, and deprecation message/deadline/replacedBy
- `generatePerActionDataFiles` `jsDocBlock` — action description and `@param` descriptions

Metadata embedded into `__DefinitionMetadata` is unaffected because it already goes through `JSON.stringify` (a `*/` inside a string literal is harmless).

## Tests

- New unit tests for `escapeJsDocText`.
- New regression test in `generatePerQueryDataFiles` that generates a query with a parameter description containing `*/` and asserts the output escapes it and parses with zero TypeScript syntax diagnostics.
- Full `@osdk/generator` suite passes with no snapshot changes (descriptions without `*/` are byte-identical — no behavior change).

### DIFF
--- a/.changeset/escape-jsdoc-comment-terminator.md
+++ b/.changeset/escape-jsdoc-comment-terminator.md
@@ -1,0 +1,5 @@
+---
+"@osdk/generator": patch
+---
+
+Escape `*/` in ontology descriptions and display names so they cannot prematurely close generated JSDoc comments and break SDK generation.
--- a/packages/generator/src/shared/propertyJsdoc.ts
+++ b/packages/generator/src/shared/propertyJsdoc.ts
@@ -16,6 +16,7 @@
 
 import type { ObjectMetadata } from "@osdk/api";
 import type { PropertyV2 } from "@osdk/foundry.ontologies";
+import { escapeJsDocText } from "../util/escapeJsDocText.js";
 
 export function propertyJsdoc(
   property: ObjectMetadata.Property,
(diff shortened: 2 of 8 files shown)


ssanjay1 (approved):
Seems good for now, a bigger refactor is probably in order for some of this jsdoc generation. Someone new coming into this and adding docs may not know when/when not to escape text

PR #3665, opened 2026-07-10
Added support for actions in function Scope

oleh308:
actions are new fields that would allow functions to declare these types resources as part of the scope

```tsx

    scope: {
      resources: { actions: ["actionTypeA"]},
    },
  };
  export function myFunction(): void {}

```

```tsx
  import { ActionTypeA } from "@ontology/sdk"
  
    scope: {
      resources: { actions: [ActionTypeA]},
    },
  };
  export function myFunction(): void {}

```

### DIFF
--- a/etc/functions.report.api.md
+++ b/etc/functions.report.api.md
@@ -4,6 +4,7 @@
 
 ```ts
 
+import type { ActionDefinition } from '@osdk/client';
 import { Attachment } from '@osdk/client';
 import type { Client } from '@osdk/client';
 import type { CompileTimeMetadata } from '@osdk/client';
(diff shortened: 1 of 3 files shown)


etc/functions.report.api.md
```diff
@@ -4,6 +4,7 @@
 
 ```ts
 
+import type { ActionDefinition } from '@osdk/client';
```
oleh308, line 7: the CI was failing because this was not here, is this expected? I run pnpm build but it was not generated

KuberSethi: Yeah generally we should be regenning these APIs, will fails checks otherwise

KuberSethi (approved):

PR #3668, opened 2026-07-13
feat(react-components): accept Uint8Array and Blob as PdfViewer src

olliejenningsPT:
## What

Widens the PDF source input on `BasePdfViewer` (and the `usePdfDocument` hook / `PdfViewerContent` building block) to accept in-memory bytes, not just a URL or `ArrayBuffer`. Adds a new exported `PdfSource` type:

```ts
type PdfSource = string | ArrayBuffer | Uint8Array | Blob;
```

URL / `ArrayBuffer` / `Uint8Array` resolve synchronously; a `Blob` is read via `.arrayBuffer()` before rendering, with cancellation if the hook unmounts mid-read. Purely additive and backward-compatible.

## Input matrix

| Component | Accepts | Notes |
|---|---|---|
| `DocumentViewer` | `media: Media` only | Routes by media type; forwards `pdfViewerProps` (with `src` / `className` omitted) |
| `PdfViewer` (OSDK wrapper) | `media: Media` only | Fetches media into an `ArrayBuffer`, forwards to `BasePdfViewer` |
| `BasePdfViewer` | `src: PdfSource` only | **Changed:** `PdfSource` now includes `Uint8Array` and `Blob` |

The new Blob / Uint8Array support reaches the `BasePdfViewer` layer. `PdfViewer` and `DocumentViewer` remain media-only.

## Tests

- `usePdfDocument`: added `Uint8Array` (sync), `Blob` (async read + byte-equality), Blob-read rejection into error state, and unmount-before-read cancellation.
- Full `@osdk/react-components` suite green (1246 tests).

## Storybook / docs

- New **Features -> With Blob** story alongside *With Pdf Url*, rendering through `BasePdfViewer`.
- `PdfViewer.md` updated to document the widened `src`.

## Changeset

`minor` for `@osdk/react-components`.

### DIFF
--- a/packages/react-components/src/pdf-viewer/hooks/usePdfDocument.ts
+++ b/packages/react-components/src/pdf-viewer/hooks/usePdfDocument.ts
@@ -29,21 +29,17 @@ type GetDocumentParams = Parameters<typeof getDocument>[0];
 /**
  * Resolve a {@link PdfSource} to pdfjs `getDocument` parameters.
  *
- * String URLs and in-memory bytes (ArrayBuffer/Uint8Array) resolve
- * synchronously; a Blob is read into an ArrayBuffer first, so it resolves to a
- * promise. Callers can branch on `instanceof Promise` to keep the synchronous
- * path synchronous.
+ * A Blob is read into an ArrayBuffer first; URLs and in-memory bytes
+ * (ArrayBuffer/Uint8Array) resolve immediately. Always returns a promise so
+ * callers have a single code path regardless of source type.
  */
-function toDocumentParams(
-  src: PdfSource
-): GetDocumentParams | Promise<GetDocumentParams> {
-  if (typeof src === "string") {
-    return { url: src };
-  }
+function toDocumentParams(src: PdfSource): Promise<GetDocumentParams> {
   if (src instanceof Blob) {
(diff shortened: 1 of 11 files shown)


packages/react-components/src/pdf-viewer/hooks/usePdfDocument.ts
```diff
@@ -22,6 +22,30 @@ const pdfWorkerUrl = new URL(
...
+ * Resolve a {@link PdfSource} to pdfjs `getDocument` parameters.
+ *
+ * String URLs and in-memory bytes (ArrayBuffer/Uint8Array) resolve
+ * synchronously; a Blob is read into an ArrayBuffer first, so it resolves to a
+ * promise. Callers can branch on `instanceof Promise` to keep the synchronous
+ * path synchronous.
+ */
+function toDocumentParams(
+  src: PdfSource
+): GetDocumentParams | Promise<GetDocumentParams> {
```
ziyunp: Should we standardise the return type to always return Promise<GetDocumentParams>?

olliejenningsPT: addressed

ziyunp (approved):

PR #3669, opened 2026-07-13
feat: export ObjectTable headless hooks from @osdk/react-components

ziyunp:
## What

Exports the 13 building-block hooks that `ObjectTable` composes internally from `@osdk/react-components/experimental/object-table`, so consumers can build a **fully headless table** while keeping OSDK data fetching, sorting, selection, editing, and snapshot behavior.

Hooks now exported: `useObjectTableData`, `useColumnDefs`, `useTableSorting`, `useRowSelection`, `useSelectionColumn`, `useColumnPinning`, `useColumnResize`, `useColumnVisibility`, `useFocusedRow`, `useEditableTable`, `useObjectTableSnapshot`, `useFunctionColumnsData`, `useCellContextMenu`.

## API review / cleanup

Beyond the mechanical export, this reviews and tidies the surface:

- **Normalized type names** to `Use<Hook>Props` / `Use<Hook>Result` across all hooks (was a mix of `Props`/`Results`/`Options`).
- **`useObjectTableData` refactored** from 9 positional arguments to a single options object (`UseObjectTableDataOptions`). Call site in `ObjectTable.tsx` and all tests updated.
- **Promoted supporting types** referenced by the hook signatures: `EditableConfig`, `EditModeState`, `OrderBy`, `ObjectSetOptions`, `RowSelectionChange`, `AsyncCellData`, `PopoverPosition`, `FunctionColumnData`.
- **Docs**: new "Headless Hooks" section in `docs/ObjectTable.md` with a worked example + TOC entry.


packages/react-components/package.json
```diff
@@ -237,6 +237,7 @@
     "@osdk/monorepo.api-extractor": "workspace:~",
     "@osdk/monorepo.tsconfig": "workspace:~",
     "@osdk/react": "workspace:*",
+    "@tanstack/react-table": "^8.21.3",
```
olliejenningsPT: `@tanstack/react-table` is listed here in `devDependencies` as well as in `dependencies` (line 203). The devDependency was needed while this was a peer dep, but now that it's a regular dependency it's already installed, so this line is redundant and can be dropped.

packages/e2e.sandbox.peopleapp/src/app/offices/CustomOfficeTable.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const table = useReactTable<OfficeRow>({
+    data: data ?? EMPTY_ROWS,
+    columns: allColumns,
+    state: { sorting, rowSelection },
+    onSortingChange,
+    enableRowSelection,
+    manualSorting: true,
+    getCoreRowModel: getCoreRowModel(),
+    getRowId: (row) => row.$primaryKey,
```
olliejenningsPT, line 80: `getRowId` needs to return a string. This works here only because Office's primary key happens to be a string. Someone copying this example for an object with a numeric or boolean primary key would hit a type error. The docs snippet uses `String(row.$primaryKey)`, so matching that here would keep the example portable and safe to copy-paste.

ziyunp: I think it's okay to leave it here as this is just an example app.

olliejenningsPT (approved):

PR #3670, opened 2026-07-13
fix devtools console log source attribution always resolving to ConsoleLogStore

KuberSethi:
console log source attribution in the devtools panel always resolved to "ConsoleLogStore" instead of the real caller, aside from being a bad practice this just isn't very useful. This PR replaces the old brittle regex matching / hardcoded approach to a new approach leveraging newer browser features when available and a worse, but still smarter approach otherwise. Also moves a lot of the nasty browser compatibility stuff into utils since that type of code will always be a bit hard to reason about next to regular code.


packages/react-devtools/src/utils/browser-compat/callerLocation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { afterEach, describe, expect, it } from "vitest";
+
+import type { CallerLocation } from "./callerLocation.js";
+import {
+  captureCallerLocation,
+  formatCallerLocation,
+} from "./callerLocation.js";
+
+describe("callerLocation", () => {
+  describe("captureCallerLocation", () => {
```
olliejenningsPT, line 28: This suite now only covers the V8 path and the throw path. The non-V8 fallback (`new Error().stack` -> `locateCallerLine` -> `parseLineLocation`) isn't exercised anywhere, since the test env is always V8. That's the branch that had the off-by-one fixed earlier on this branch (fab55d9594), and it's real production code for Firefox/Safari, so a regression there would ship silently.

The old test file drove it with a synthetic headerless stack (by deleting `Error.captureStackTrace`). Could we bring that fallback test back? Right now it's the one path most likely to break and the only one with no coverage.

packages/react-devtools/src/utils/browser-compat/callerLocation.ts
```diff
@@ -0,0 +1,129 @@
...
+  readonly column: number;
+}
+
+// Non-V8 engines (Firefox/Safari) format frames as `name@file:line:column`.
+// Anchored to end-of-line and excluding `@` from the captured group so a line
+// it can't parse fails closed (undefined) instead of matching a partial
+// location.
+const LINE_LOCATION_REGEX = /@([^@]*):(\d+):(\d+)\s*$/u;
+
+export function parseLineLocation(line: string): CallerLocation | undefined {
```
olliejenningsPT, line 34: `parseLineLocation` and `locateCallerLine` are both exported but only called inside this file now, and the direct unit tests for them were dropped in the same commit that removed the "exported for direct unit testing" note. So the exports don't buy us anything at the moment.

Either re-add the direct tests (my preference, ties into the fallback coverage comment) or make them module-private.

packages/react-devtools/src/utils/browser-compat/callerLocation.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  const site = callSites?.[0];
+  if (!site) {
+    return undefined;
+  }
+  const fileName = site.getFileName() ?? site.getEvalOrigin();
+  const line = site.getLineNumber();
+  const column = site.getColumnNumber();
+  if (fileName == null || line == null || column == null) {
```
olliejenningsPT: `column` never makes it into the output, `formatCallerLocation` only emits `name:line`, so bailing to `undefined` when just the column is missing throws away a perfectly good file:line. Some native/eval call sites report a null column.

The old regexes required a column too, so this isn't a regression, but since we don't render it, could we gate only on `fileName`/`line`?

packages/react-devtools/src/utils/browser-compat/callerLocation.ts
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Returns the caller's stack line, skipping V8's "Error" header if present.
+ * `framesBeforeCaller` = frames from the capture point to the caller.
+ */
+export function locateCallerLine(
+  stack: string,
+  framesBeforeCaller: number
+): string | undefined {
+  const lines = stack.split("\n");
+  const hasHeader = parseLineLocation(lines[0]) === undefined;
```
olliejenningsPT, line 55: Minor robustness point: this assumes `lines[0]` always parses on the non-V8 path. It normally does, since that frame is `captureCallerLocation` with a real location. But if a top frame ever fails to parse as `@file:line:col` (old Safari without column numbers, or a native/anonymous frame), `hasHeader` flips to true and we return `lines[3]`, one frame too deep, rather than failing closed like the rest of the module does.

Low likelihood, but this is the exact spot the earlier off-by-one lived, so a defensive check or a test would be reassuring.

packages/react-devtools/src/utils/browser-compat/callerLocation.test.ts
KuberSethi: good call, added some parseLineLocation and locateCallerLine tests

i deliberately left out the old integration test because it's really going to end up testing against our own (mocked) reimplementation of firefox's format, so it's really testing the mock, not the test behavior. don't think it makes sense to try and test that sort of thing but lmk if you disagree

packages/react-devtools/src/utils/browser-compat/callerLocation.ts
KuberSethi: re-added the direct unit tests for both, so the exports are used again.

packages/react-devtools/src/utils/browser-compat/callerLocation.ts
KuberSethi: fixed.

packages/react-devtools/src/utils/browser-compat/callerLocation.ts
KuberSethi: added a test to hopefully protect against this, lmk if you had something else in mind

olliejenningsPT (approved):

PR #3671, opened 2026-07-13
Add Vite plugin for preview error reporter 

bhsz:
Adds a Vite plugin that posts runtime errors to the parent from an iframe so that a broken preview shows a reason instead of a blank white screen.

When running in a given mode (default: code-workspaces), it injects a small inline script that forwards uncaught errors, failed resource loads, and unhandled rejections to the parent.

Dev-only and gated on the mode, so it never ships in production builds and adds nothing to a normal local `npm run dev` without the mode set.

### DIFF
--- a/packages/vite-plugin-code-workspace-preview/package.json
+++ b/packages/vite-plugin-code-workspace-preview/package.json
@@ -1,0 +1,69 @@
+{
+  "name": "@osdk/vite-plugin-code-workspace-preview",
+  "version": "0.0.0",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/palantir/osdk-ts.git"
+  },
+  "exports": {
+    ".": {
+      "browser": "./build/browser/index.js",
+      "import": {
+        "types": "./build/types/index.d.ts",
+        "default": "./build/esm/index.js"
+      },
+      "require": "./build/cjs/index.cjs",
+      "default": "./build/browser/index.js"
+    },
+    "./*": {
+      "browser": "./build/browser/public/*.js",
+      "import": {
+        "types": "./build/types/public/*.d.ts",
+        "default": "./build/esm/public/*.js"
+      },
+      "require": "./build/cjs/public/*.cjs",
+      "default": "./build/browser/public/*.js"
+    }
+  },
+  "scripts": {
+    "check-attw": "attw --pack .",
+    "check-spelling": "cspell --quiet .",
+    "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
+    "fix-lint": "oxlint -c ./oxlint.config.ts --fix . && oxfmt -c ../../oxfmt.config.ts .",
+    "lint": "oxlint -c ./oxlint.config.ts . && oxfmt -c ../../oxfmt.config.ts --check .",
+    "test": "vitest run",
+    "transpileBrowser": "monorepo.tool.transpile -f esm -m normal -t browser",
+    "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
+    "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
+    "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
+    "typecheck": "tsc --noEmit --emitDeclarationOnly false",
+    "watch": "tsc --watch"
+  },
+  "peerDependencies": {
+    "vite": "*"
+  },
+  "devDependencies": {
+    "@osdk/monorepo.api-extractor": "workspace:~",
+    "@osdk/monorepo.tsconfig": "workspace:~",
+    "@types/node": "^20.19.13",
+    "typescript": "~5.5.4"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "build/cjs",
+    "build/esm",
+    "build/browser",
+    "build/types",
+    "CHANGELOG.md",
+    "package.json",
+    "templates",
+    "*.d.ts"
+  ],
+  "main": "./build/cjs/index.cjs",
+  "module": "./build/esm/index.js",
+  "types": "./build/cjs/index.d.cts",
+  "type": "module"
+}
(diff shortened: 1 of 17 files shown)


packages/vite-plugin-code-workspace-preview/vitest.config.mts
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+import { configDefaults, defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    pool: "forks",
+    exclude: [...configDefaults.exclude, "**/build/**/*"],
+    fakeTimers: {
```
olliejenningsPT: Minor: this `fakeTimers` block is not required here. `test.fakeTimers.toFake` only takes effect once `vi.useFakeTimers()` is called, and these tests never call it.

packages/vite-plugin-code-workspace-preview/src/reporterScript.ts
```diff
@@ -0,0 +1,57 @@
...
+  window.addEventListener("error", function (e) {
+    reportIfBlank({ kind: "error", message: e.message, filename: e.filename, lineno: e.lineno, colno: e.colno, stack: e.error && e.error.stack });
+  });
+
+  window.addEventListener("unhandledrejection", function (e) {
+    var r = e.reason;
+    reportIfBlank({ kind: "unhandledrejection", message: String((r && r.message) || r), stack: r && r.stack });
+  });
+
+  // reset any previous errors
```
olliejenningsPT: Tiny wording nit: the `ready` message does not reset anything on its own, it signals the parent to clear any previously reported errors. Something like `// tell the parent to clear any previous errors` reads a little truer to what is happening.

packages/vite-plugin-code-workspace-preview/src/codeWorkspacePreviewPlugin.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export function codeWorkspacePreviewPlugin(
+  options: CodeWorkspacePreviewOptions = {}
+): Plugin {
+  const activeMode = options.mode ?? "code-workspaces";
+  return {
+    name: "osdk-code-workspace-preview",
+    apply: "serve",
+    transformIndexHtml(_html, ctx) {
+      // apply: "serve" guarantees ctx.server is present.
```
olliejenningsPT: Small inconsistency: the comment says `apply: "serve"` guarantees `ctx.server` is present, but the line below still uses optional chaining (`ctx.server?.config.mode`).

packages/vite-plugin-code-workspace-preview/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "access": "public"
+  },
+  "files": [
+    "build/cjs",
+    "build/esm",
+    "build/browser",
+    "build/types",
+    "CHANGELOG.md",
+    "package.json",
+    "templates",
```
olliejenningsPT: `templates` is listed here but there is no `templates/` directory in this package (looks copied over from the sibling plugin). No real effect since npm skips missing entries, but it could be dropped to avoid confusion.

olliejenningsPT (commented):
A few inline notes. The main one is a question about whether resource/module load failures are meant to be in scope for the error listener; the rest are minor nits. The package scaffolding all looks consistent with the sibling vite-plugin-status-reporter.

packages/vite-plugin-code-workspace-preview/src/reporterScript.ts
```diff
@@ -0,0 +1,57 @@
+/*
```
olliejenningsPT, line 1: I think for this file we can have the reporter itself being exported on top of the script variable stringified, this means we can fully test the reporter script:

```ts
export function installReporter( 
    win: Window,
    doc: Document,
    schedule: (task: () => void) => void,
  ): void {
    if (win.parent === win) return;

    function post(data: Record<string, unknown>): void {
      data.source = "osdk-preview";
      try {
        win.parent.postMessage(data, "*");
      } catch (_e) {
        /* parent gone or cross-origin refusal; ignore */
      }
    }

    function appHasContent(): boolean {
      const root = doc.getElementById("root");
      return root != null && root.childElementCount > 0;
    }

    function reportIfBlank(data: Record<string, unknown>): void {
      // So this is an optional thing, the reason why this might be helpful
      // is since this is explicitly is because error handlers throw instantly when error is thrown
      // You might want to wait until things like React have finished rendering, which is to wait for all micro-tasks to be finished (setTimeout is basically a macro-task).
      // again completely optional 
      schedule(function () {
        if (appHasContent()) return;
        post(data);
      });
    }

    // capture: true so we also see resource/module load failures (a failed
    // <script>/<link>/<img>, e.g. a 404 on a chunk). Those dispatch on the
    // element and do NOT bubble, so a normal window listener would miss them.
    win.addEventListener(
      "error",
      function (e) {
        const target = e.target; // put as a constant due to e.target potentially being undefined when used
        if (target != null && target !== win) {
          // Resource load failure: a plain Event, so there is no message/stack.
          // The useful detail is the failing element's URL.
          const el = target as Element & { src?: string; href?: string };
          const url = el.src ?? el.href ?? "";
          reportIfBlank({
            kind: "error",
            message: "Failed to load " + el.tagName.toLowerCase()
              + (url ? ": " + url : ""),
          });
          return;
        }
        // Thrown exception: dispatched at window, so this is a real ErrorEvent.
        reportIfBlank({
          kind: "error",
          message: e.message,
          filename: e.filename,
          lineno: e.lineno,
          colno: e.colno,
          stack: (e.error as Error | undefined)?.stack,
        });
      },
      true,
    );

    // unhandledrejection is always dispatched at window and never bubbles from an
    // element, no need for the capture option here
    win.addEventListener("unhandledrejection", function (e) {
      const r: { message?: unknown; stack?: unknown } | null | undefined = e.reason;
      reportIfBlank({
        kind: "unhandledrejection",
        message: String(r?.message ?? r),
        stack: r?.stack,
});
    });

    // Signal a fresh load so the parent can clear any previously reported error.
    // This happens instantly before any scheduled error
    post({ kind: "ready" });
  }

  /**
   * serialized as a self-contained IIFE for injection. `schedule` is wired to a real macro-task so tit runs after the current render tick.
   */
  export const REPORTER_SCRIPT: string = `(${installReporter.toString()})(`
    + `window, document, function (task) { setTimeout(task, 0); });`;
```

olliejenningsPT (approved):

PR #3672, opened 2026-07-13
Support interface link actions in maker-experimental

ethanai22:


packages/generator-converters.ontologyir/src/OntologyIrToFullMetadataConverter.ts
```diff
@@ -1069,6 +1069,9 @@ export class OntologyIrToFullMetadataConverter {
             );
           }
         }
+        case "addInterfaceLinkRuleV2":
+        case "deleteInterfaceLinkRule":
+          return [];
```
davidpaquin, line 1074: Any reason to not fill this in?

ethanai22: `LogicRule` doesn't have a type for these rules (operations on the full metadata type is only used by the generator to resolve object edits). We actually replace these actions types later in the preview path (`ActionLogicRuleConverter.ts`)

davidpaquin: ah got it

davidpaquin (approved):

PR #3675, opened 2026-07-13
Make latest-registered mock client stub win

nihalbhatnagar:
## Summary

Re-registering a mock client stub with the same pattern now **overrides** the previous stub (latest registered wins) instead of being silently shadowed by the first one.

Previously, `stubs` were stored in an array and resolution walked it front-to-back returning the first match. So calling `.when(...).thenReturnObjects(...)` twice with the same pattern appended a second stub that was never reached — the original value always won. This surprised users who expected re-stubbing to update the return value (and forced them to `clearStubs()`, which wipes *every* stub).

Now both resolution loops (`resolveStub` for object sets, `resolveQuery` for queries) iterate newest-first, so the most recently registered matching stub wins.

## Changes
- `resolveStub` iterates stubs newest-first
- `resolveQuery` iterates query stubs newest-first
- Regression test: re-registering the same pattern overrides the previous stub
- Changeset (`@osdk/unit-testing`: patch)

### DIFF
--- a/packages/unit-testing/src/mock/createMockClient.ts
+++ b/packages/unit-testing/src/mock/createMockClient.ts
@@ -75,7 +75,8 @@ export function createMockClient(): MockClient {
     );
 
   const resolveQuery = (queryApiName: string, params: unknown): unknown => {
-    for (const stub of queryStubs) {
+    for (let i = queryStubs.length - 1; i >= 0; i--) {
+      const stub = queryStubs[i];
       if (stub.queryApiName !== queryApiName) continue;
       if (deepEqual(stub.params, params)) {
         if (stub.error !== undefined) {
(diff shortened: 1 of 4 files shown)


esinx (approved):

bryantpark04 (approved):

packages/unit-testing/src/mock/createMockClient.ts
```diff
@@ -75,7 +75,8 @@ export function createMockClient(): MockClient {
     );
 
   const resolveQuery = (queryApiName: string, params: unknown): unknown => {
-    for (const stub of queryStubs) {
+    for (let i = queryStubs.length - 1; i >= 0; i--) {
```
bryantpark04, line 78: Would prefer [`toReversed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toReversed) if available

PR #3677, opened 2026-07-13
add experimental hydrate osdk object

ssanjay1:
Adds an experimental hydrate osdk object method. This does not work for interface or objects defined with RDPs as of now

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects.ts
@@ -69,7 +69,8 @@ export async function convertWireToOsdkObjects(
   interfaceToObjectTypeMappingsV2?: Record<
     InterfaceTypeApiName,
     InterfaceToObjectTypeMappingsV2
-  >
+  >,
+  objectDefsByApiName?: Record<string, ObjectMetadata>
 ): Promise<Array<InterfaceHolder>>;
 export async function convertWireToOsdkObjects(
   client: MinimalClient,
@@ -87,7 +88,8 @@ export async function convertWireToOsdkObjects(
   interfaceToObjectTypeMappingsV2?: Record<
     InterfaceTypeApiName,
     InterfaceToObjectTypeMappingsV2
-  >
+  >,
+  objectDefsByApiName?: Record<string, ObjectMetadata>
 ): Promise<Array<ObjectHolder>>;
 export async function convertWireToOsdkObjects(
   client: MinimalClient,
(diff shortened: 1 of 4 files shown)


packages/client/src/object/hydrateOsdkObject.ts
```diff
@@ -0,0 +1,64 @@
...
+export function hydrateOsdkObject(
+  client: Client,
+  objects: OntologyObjectV2[],
+  objectDefsByApiName?: Record<string, ObjectMetadata>,
+  propertySecurities?: PropertySecurities[]
+): Promise<Array<Osdk.Instance<any>>> {
+  return convertWireToOsdkObjects(
+    client[additionalContext],
+    objects,
+    /* interfaceApiName */ undefined,
```
ssanjay1, line 54: Purposefully left out interface stuff for now, easy to add/break later since this is all unstable

packages/client/src/object/convertWireToOsdkObjects.ts
```diff
@@ -141,9 +145,15 @@ export async function convertWireToOsdkObjects(
   const isInterfaceScoped = Object.keys(effectiveMappings).length > 0;
   const ret = [];
   for (const rawObj of objects) {
-    const objectDef = await client.ontologyProvider.getObjectDefinition(
-      rawObj.$apiName
-    );
+    // If caller supplied object definitions, use them exclusively instead of
```
bryantpark04, line 148: comment is unnecessary I think

packages/client/src/object/hydrateOsdkObject.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param client - the OSDK client
+ * @param objects - the raw wire objects to hydrate
+ * @param objectDefsByApiName - optional map of object type api name to its
+ * definition. When provided, every object's api name must be present.
+ * @param propertySecurities - optional property securities for the objects
+ */
+export async function hydrateOsdkObject(
+  client: Client,
+  objects: OntologyObjectV2[],
+  objectDefsByApiName?: Record<string, ObjectMetadata>,
```
bryantpark04, line 48: Do we have an existing ObjectTypeApiName type that aliases to string we could use instead of naming the parameter `objectDefsByApiName`?

ssanjay1: We do not unfortunately, and didn't really want to add one right now just for this experimental method. If we promote this later can add

bryantpark04 (approved):

packages/client/src/object/hydrateOsdkObject.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Unstable API. Do not depend on this outside of experimental usage.
+ *
+ * @param client - the OSDK client
+ * @param objects - the raw wire objects to hydrate
+ * @param objectDefsByApiName - optional map of object type api name to its
+ * definition. When provided, every object's api name must be present.
+ * @param propertySecurities - optional property securities for the objects
+ */
+// oxlint-disable-next-line require-await -- intentionally async
```
esinx, line 44: I might be missing the picture here but why is it intentionally async when it only needs to forward a call to another async function?

PR #3678, opened 2026-07-13
Fix pagination in asyncIterLinks

bryantpark04:
I had forgotten to pass the received `pageToken` back into the next `loadLinks` PSDK call. This means that for object sets > 1000 items in size, the method would never finish loading links. We unfortunately didn't catch this in testing because I had only tested with object sets larger than 5000 items (the limit at the time) using PSDK when I was writing the endpoint. E2E through the OSDK I used a different object type with more types of links but fewer instances.

### DIFF
--- a/.changeset/public-grapes-post.md
+++ b/.changeset/public-grapes-post.md
@@ -1,0 +1,5 @@
+---
+"@osdk/client": patch
+---
+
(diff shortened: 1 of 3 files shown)


nihalbhatnagar (approved):
wow links

PR #3679, opened 2026-07-14
Introduce branched SDKs CLI

zeyadkhaled:
# `unstable branch` subcommand

The `branch` subcommand works with branched Foundry SDK versions published on a global branch.

```
npx @osdk/cli unstable branch [subcommand]
```

Common options

| Option       | Description                                                                             |
| ------------ | --------------------------------------------------------------------------------------- |
| --branchName | Global branch to select branched SDKs for. Defaults to the current git branch. [string] |

The `branch` subcommand is not runnable without passing a specific subcommand (i.e. `install`).

### `install` subcommand

Discovers and installs the latest versions of branched Foundry SDKs published on a global branch, updating the dependencies in `package.json`.

```
npx @osdk/cli unstable branch install
```

Install options

| Option        | Description                                                                                   |
| ------------- | --------------------------------------------------------------------------------------------- |
| --packageName | Operate on this single SDK package instead of detecting branched SDKs automatically. [string] |
| --dryRun      | Print the intended version change(s) without modifying any files. [boolean] [default: false]  |

Only SDKs whose declared version differs from the published one are installed; those already in sync are left untouched. `install` makes no changes and exits `0` when:

- you are on `main`/`master`, a detached `HEAD`, or not in a git repo (pass `--branchName` to force a branch),
- there are no branched SDKs on a global branch yet, or
- every branched SDK is already in sync.

After a successful install, review and commit the updated `package.json` and lockfile.

## Help and debugging

To see all supported subcommands and their expected arguments, run a command with the `--help` flag

```
npx @osdk/cli unstable branch --help
```

For additional information for debugging, use the `--verbose` flag

```
npx @osdk/cli unstable branch install --verbose
```


packages/cli/src/commands/branch/install/decideAction.ts
```diff
@@ -0,0 +1,37 @@
...
+  resolved,
+}: {
+  current: string | undefined;
+  resolved: string;
+}): Action {
+  if (current === resolved) {
+    return { kind: "noop", reason: `already in sync at ${resolved}` };
+  }
+  return { kind: "install", version: resolved };
+}
```
amishshah: Would there ever be a 3rd variant to this enum? Otherwise I think it makes more sense to make the function self-describing with something like e.g. `doesInstalledSdkVersionMatchBranchLatest` that returns a boolean? Think that reduces the complexity a bit

packages/cli/src/commands/branch/utils/resolveBranchedPackages.mts
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Identify every branched SDK and its published version for the current branch.
+ *
+ * The branched SDKs are whichever `@<scope>/sdk` candidates (or the explicit
+ * `--packageName`) publish the branch's `${branch}-latest` npm dist tag. npm has
+ * no tag -> package reverse lookup, so each candidate is probed with `npm view`.
+ * All matching candidates are returned.
+ */
+export async function resolveBranchedPackages(
```
amishshah, line 69: IMO think this should become:

- `resolveFoundryBranch` which does lines 75-80
- `resolveBranchedPackages(branch)` which is the remainder

Do you think there's a strong case for discriminating `no-sdk-on-branch` and `match`? It feels like we should have  `resolveBranchedPackages` just return what is now the `match` variant, and the caller should do something like:

```ts
const branch = resolveFoundryBranch(...);
if (branch == null) {
  throw new Error("...");
}

const packages = resolveBranchedPackages(branch);
if (packages.length === 0) {
  // no-sdk-on-branch case, if needed
}
// match case
```

I think this is more idiomatic and simple

packages/cli/src/commands/branch/install/decideAction.ts
zeyadkhaled: For the time being, I can't third variant. So I just simplified this fully and inlined the equality check.

packages/cli/src/commands/branch/utils/resolveBranchedPackages.mts
```diff
@@ -0,0 +1,192 @@
...
+  };
+}
+
+async function readPackageJson(deps: ResolveDeps): Promise<PackageJsonLike> {
+  const packageJsonPath = await findUp("package.json", { cwd: deps.cwd });
+  if (packageJsonPath == null) {
+    throw new ExitProcessError(
+      1,
+      "Could not find a package.json in the current directory or any parent.",
+      "Run osdk unstable branch from within your project."
```
amishshah:
```suggestion
      "Run 'osdk unstable branch' from within your project."
```

amishshah (dismissed):
Looks good! I don't think this would work with [aliased packages](https://docs.npmjs.com/cli/v8/using-npm/package-spec#aliases), but I think that's not the priority right now

amishshah (approved):

PR #3686, opened 2026-07-14
Add ObservableClient.applyChanges for external cache updates

olliejenningsPT:
## What

Adds a new `applyChanges` method (and an `ObservableClient.Changes` type) to `@osdk/client`'s `ObservableClient`. It pushes externally-sourced object upserts and deletes straight into the observable cache that the `@osdk/react` / `@osdk/react-components` hooks read.

```ts
export namespace ObservableClient {
  export interface Changes {
    upserts?: ReadonlyArray<Osdk.Instance<ObjectOrInterfaceDefinition, "$allBaseProperties">>;
    deletes?: ReadonlyArray<{ objectType: string; primaryKey: string | number }>;
  }
}

interface ObservableClient {
  applyChanges(changes: ObservableClient.Changes): Promise<void>;
}
```

## How

It reuses the existing ingest paths rather than inventing a new one, all inside a single `Store.batch`:

- **upserts** go through `objects.storeOsdkInstances` -> `writeToStore` -> `propagateWrite`, the same path server page fetches and websocket stream updates use.
- **deletes** tombstone every registered cache variant of the object via `deleteFromStore` -> `propagateWrite` (the same path action-result deletes use). Objects that were never loaded have no registered variants and are ignored.

Because it runs in one batch, all observing hooks (objects, lists, object sets, links, dependent functions) update reactively without a network round trip.

## Why

This is the seam an embedded or offline ontology engine uses to stream locally-synced deltas into the hooks. Combined with the already-public `createClientWithSubscriptionConnection` (subscription transport), the `fetchFn` override on `createClient` (reads/writes), and `objectFactory` (wire to instance conversion), it lets such an engine drive every hook online and offline with live updates.

## Compatibility

Additive and non-breaking. Callers that never invoke `applyChanges` are unaffected. The `ObservableClient` interface lives on the `@osdk/client/observable` subpath, which is not part of the api-extractor main-entry report, so no API report changes.

## Testing

- New `Store.applyChanges.test.ts` (5 tests, FauxFoundry harness, `mode: "offline"`): upsert updates an observed object, upsert inserts a matching object into an observed list, delete removes an object from list + cache, combined upsert/delete in one batch, and empty no-op. All exercise cache-only updates with no network fetch.
- `pnpm turbo typecheck --filter=@osdk/client --filter=@osdk/react-devtools` passes.
- `pnpm turbo check-api --filter=@osdk/client`: report up to date.
- Full `@osdk/client` observable suite passes (356 passed, 1 skipped).

A changeset (`@osdk/client: minor`) is included.

### DIFF
--- a/packages/client/src/observable/ObservableClient.ts
+++ b/packages/client/src/observable/ObservableClient.ts
@@ -63,6 +63,44 @@ export namespace ObservableClient {
   export interface ApplyActionOptions {
     optimisticUpdate?: (ctx: OptimisticBuilder) => void;
   }
+
+  /**
+   * A batch of externally-sourced object changes to apply directly to the
+   * observable cache via {@link ObservableClient.applyChanges}.
+   *
+   * Changes are written through the same store ingest path that server fetches,
+   * action results, and websocket stream updates use, so every observing hook
+   * (objects, lists, object sets, links, and dependent functions) updates
+   * reactively without a network round trip. This is the seam an embedded or
+   * offline ontology engine uses to push locally-synced deltas into the hooks.
+   */
+  export interface Changes {
+    /**
+     * Full object or interface instances to insert or update.
+     *
+     * Instances must be produced by the client's object factory — i.e. wire
+     * data converted through the client's `objectFactory`, or an existing
+     * instance's `$clone` — so they carry the internal holder representation
+     * the cache stores. Each instance should carry all of its base properties
+     * (`"$allBaseProperties"`); an instance carrying only a subset of
+     * properties merges into any existing cached value rather than replacing
+     * it.
+     */
+    upserts?: ReadonlyArray<
+      Osdk.Instance<ObjectOrInterfaceDefinition, "$allBaseProperties">
+    >;
+
+    /**
+     * Objects to remove from the cache, identified by their concrete object
+     * type API name and primary key. Removing an object tombstones it along
+     * with every registered cache variant, so lists and object sets that
+     * contain it drop it reactively.
+     */
+    deletes?: ReadonlyArray<{
+      objectType: string;
+      primaryKey: string | number;
+    }>;
+  }
 }
 
 export interface CacheSnapshot {
@@ -566,6 +604,26 @@ export interface ObservableClient extends ObserveLinks {
     action: Q,
     args: Parameters<ActionSignatureFromDef<Q>["applyAction"]>[0]
   ) => Promise<ActionValidationResponse>;
+
+  /**
+   * Apply a batch of externally-sourced object changes directly to the cache,
+   * bypassing the network.
+   *
+   * Upserts and deletes flow through the same store ingest path used by server
+   * fetches, action results, and websocket stream updates, so all observing
+   * hooks update reactively. Intended for data that arrives outside the normal
+   * fetch/action flow, e.g. an embedded or offline ontology engine streaming
+   * locally-synced deltas, or an external system pushing updates.
+   *
+   * Applying an empty set of changes is a no-op. This method is additive and
+   * non-breaking for callers that never invoke it.
+   *
+   * @param changes - Object upserts and/or deletes to apply
+   * @returns Promise that resolves once the changes have been written to the
+   *   cache. Object observers are notified synchronously; list and object-set
+   *   propagation happens reactively.
+   */
+  applyChanges(changes: ObservableClient.Changes): Promise<void>;
 
   /**
    * Invalidates the entire cache, forcing all queries to refetch.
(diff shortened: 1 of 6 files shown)


packages/client/src/observable/ObservableClient.ts
```diff
@@ -63,6 +63,44 @@ export namespace ObservableClient {
   export interface ApplyActionOptions {
     optimisticUpdate?: (ctx: OptimisticBuilder) => void;
   }
+
+  /**
+   * A batch of externally-sourced object changes to apply directly to the
+   * observable cache via {@link ObservableClient.applyChanges}.
+   *
+   * Changes are written through the same store ingest path that server fetches,
+   * action results, and websocket stream updates use, so every observing hook
```
KuberSethi, line 72: Hm, why are we doing this here / is there a reason we don't want to reuse or extend the existing Changes class?

packages/client/src/observable/ObservableClient.ts
```diff
@@ -63,6 +63,44 @@ export namespace ObservableClient {
...
+   * (objects, lists, object sets, links, and dependent functions) updates
+   * reactively without a network round trip. This is the seam an embedded or
```
KuberSethi, line 74: We shouldn't mention embedded ontology / overfit, things should be pretty pure at this level imo so would suggest ripping out these mentions

KuberSethi, line 74: FYI this isn't fully accurate, this says links and dependent functions update reactively but they don't. Also applyChanges only hits objects, lists, and object sets and link queries/ dependsOn functions get skipped

packages/client/src/observable/ObservableClient.ts
```diff
@@ -63,6 +63,44 @@ export namespace ObservableClient {
...
+     * the cache stores. Each instance should carry all of its base properties
+     * (`"$allBaseProperties"`); an instance carrying only a subset of
+     * properties merges into any existing cached value rather than replacing
+     * it.
+     */
+    upserts?: ReadonlyArray<
+      Osdk.Instance<ObjectOrInterfaceDefinition, "$allBaseProperties">
+    >;
+
+    /**
```
KuberSethi, line 93: a delete for an objectType that was never loaded / doesn't exist would silently do nothing, thisi s kind of a footgun. Might be worth commenting or figuring out how to warn users. Might be being a bit pedantic tho

packages/client/src/observable/internal/Store.ts
```diff
@@ -322,6 +323,51 @@ export class Store {
...
+   * Deletes reuse the existing tombstone path used by action results: every
+   * registered cache variant of the object is tombstoned via
+   * `deleteFromStore` -> `propagateWrite`. Objects that were never loaded have
+   * no registered variants and are ignored (there is nothing observing them).
+   *
+   * Everything runs inside a single {@link batch} so the accumulated changes
+   * propagate to observing queries in one revalidation pass. Applying no
+   * changes is a no-op.
+   */
+  public applyChanges(changes: ObservableClient.Changes): Promise<void> {
```
KuberSethi, line 340: Hm, why is this here vs in ObservableClientImpl? Not super opposed but it is a bit weird, probably less code but in terms of leaky abstractions I don't ;love we're leaking these details into the store

packages/client/src/observable/ObservableClient.ts
```diff
@@ -63,6 +63,44 @@ export namespace ObservableClient {
...
+     * Objects to remove from the cache, identified by their concrete object
+     * type API name and primary key. Removing an object tombstones it along
+     * with every registered cache variant, so lists and object sets that
+     * contain it drop it reactively.
+     */
+    deletes?: ReadonlyArray<{
```
KuberSethi, line 99: FYI we should chat offline about this maybe, I don't think we're guaranteed to get back a PK from them

packages/client/src/observable/internal/Store.ts
```diff
@@ -322,6 +323,51 @@ export class Store {
...
+            objectType,
+            primaryKey
+          )) {
+            this.queries.peek(cacheKey)?.deleteFromStore("loaded", batch);
+          }
+        }
+      });
+    }
+
+    return Promise.resolve();
```
KuberSethi, line 368: the returned promise resolves before the list/object set revalidation actually runs do we have to change things / await?

KuberSethi (commented):
Good first stab, could be fine to dismiss for v0 but we might need to put in more thought on how to get this to work for links and functions

olliejenningsPT:
Not needed

PR #3687, opened 2026-07-15
chore: re-enable oxlint complexity rule (ceiling 40)

olliejenningsPT:
Re-enables the `complexity` oxlint rule, the last of the rules parked during the oxc migration.

Follow-up to #3624 (which enabled `no-useless-switch-case`). Rebuilt fresh on `main` now that #3624 has merged — the earlier stacked version was stale.

## The threshold: 40, not 20

The preset default is 20, but on the current codebase that flags **108 functions** across ~20 packages, most of them discriminated-union dispatchers (big `switch (type)` mappers) that read fine as one flat switch — splitting them is pure churn and, for shipped packages, adds bundle size. So the ceiling is set to **40**, which flags only the genuinely unwieldy outliers.

## What this does

Enables `["error", 40]` and refactors the **12 functions** that exceeded 40, by extracting behavior-preserving helper functions:

- `client` — `toDataValueQueries`
- `faux` — `filterObjects`, `validateActionParameterType`
- `maker` — a `defineAction` target-parameter builder (was 112), `renderHintFromBaseType`
- `maker-experimental` — `ActionTypeShapeExtractor.convert`, `getMarkingShapes`, `renderHintFromBaseType`
- `react-components` — `buildWhereClause`, `LinkedPropertyInput` content switch
- `e2e.sandbox.officenetwork` — `useNetworkConnections`, `EmployeePanel`

Each refactor is mechanical (pulling cohesive branches / lookup tables / render sections into helpers) with no signature, export, or public-API change. The ceiling can be ratcheted down later as functions are tidied.

## Code-quality improvements beyond the rule change

While refactoring the functions that exceeded the complexity ceiling, a few readability and type-safety improvements were made that stand on their own:

- **`ActionTypeShapeExtractor.convert`**: removed the `{ type: t, [t]: {} } as unknown as BaseParameterType` computed-key construction in favour of explicit per-type literals (via a `convertSimpleParameterType` helper and concrete returns in the decimal / object-reference / interface-reference converters). The only cast left is the single pre-existing `as BaseParameterType` on the struct path.
  - Benefit: the `as unknown as` cast had switched off type checking entirely for these returns; the explicit literals are checked by the compiler, so a wrong discriminant, misnamed field, or missing property is now a build error instead of a silent runtime bug.
- **`defineAction`**: factored out `isListParameter`, `objectReferenceParamType`, and `interfaceReferenceParamType`.
  - Benefit: the list-detection predicate (previously the same `typeof ... === "object" && "listLength" in ...` expression copy-pasted five times) and the object/interface reference `type` blocks (duplicated across the modify/delete builders) now have a single source of truth, so the behaviour can only drift in one place and each builder reads at a glance.
- **`renderHintFromBaseType`** (maker + maker-experimental): the fixed-hint types are now a small `simpleRenderHint` switch instead of a large inline conditional, with the validation-dependent cases (`string`, `marking`) handled explicitly.
  - Benefit: the type-to-hint mapping is now easy to scan and extend, and the cases that genuinely depend on allowed values are visibly separated from the ones that do not.
- **`useNetworkConnections`**: the connection-building logic is now a set of small pure functions that each return their connections and are composed by lens mode, replacing inline closures that mutated a shared array.
  - Benefit: each builder can be read (and tested) in isolation with no shared mutable state to track, and guard clauses (`if (!x) return []`) remove the nested ternaries that previously obscured the control flow.

All changes are behavior-preserving: no API changes, no runtime behavior changes, and existing tests pass.


nfdiop (commented):
Questions:
- how did we test that the refactoring didn't drop anything? (I may have missed it but is every change tested)
- There are not that many changes in this PR: making me think that the complexity issue doesn't happen a lot. And I am worried it could encourage agents to write less clear code. Is there evidence this is improving the agents output?

Side note: I'm not sure of the changes we made here: https://github.com/palantir/osdk-ts/pull/3624/changes#diff-bb74d47b77763f955ac4f1aa50c369328c445a697494dd8abbeaa650907870ae. Generally we recommend exhaustiveness checking so we can easily detect new values and make sure we support them (https://gibbok.github.io/typescript-book/book/exhaustiveness-checking/)

olliejenningsPT:
> * There are not that many changes in this PR: making me think that the complexity issue doesn't happen a lot. And I am worried it could encourage agents to write less clear code. Is there evidence this is improving the agents output?

1. Most of the code has existing tests, with the parts that don't have indirect tests, and only 1 part being part of the sandbox app.

2. you are right, l actually doubled the complexity rule value (the initial 20 resulting in 120 functions being flagged). On the agent complexity concern, i can't really say if it will worsen it or improve it, but having a high ceiling value, means that at least most code will never get it, except for some rare cases (like the main one in this review is a single function that became huge, which in itself is bad practice). So i see it as more of a preventative measure for massive functions / bad practices.

In terms of the exhaustiveness check, l actually went back through and checked the old ESLint configs, and this repo never had this rule enabled. I did find a couple of files that manually did the `never` stuff for the exhaustive check. I myself normally prefer the exhaustiveness check on switch statements, but since ultracite prefers the no-useless ones, and its more of an industry standard config, l went with its one this time. But l am happy to enable the exhaustiveness check (and disable the no-useless one) in the FLUP if you want?

packages/faux/src/FauxFoundry/validateAction.ts
```diff
@@ -222,89 +165,96 @@ function validateActionParameterType(
...
-      if (
-        typeof value !== "string" ||
-        !dataStore.ontology.getObjectTypeFullMetadata(value)
-      ) {
-        ret.result = "INVALID";
-        ret.parameters[paramKey] = {
-          ...baseParam,
-        };
-        return;
+    default: {
```
nfdiop, line 175: Can we keep the enum values in this function so we are intentional about what we want for each enum value?

packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ActionTypeShapeExtractor.ts
```diff
@@ -77,228 +146,113 @@ class BaseParameterTypeConverter {
...
+      case "struct":
+      case "structList":
+        return this.#convertStruct(t, pt);
+
+      default:
+        return convertSimpleParameterType(t);
+    }
+  }
+
+  #convertDecimal(
```
nfdiop: note: this doesn't seem to follow the convention in this class. Is there a reason it is a method and not a function?

nfdiop:
> > * There are not that many changes in this PR: making me think that the complexity issue doesn't happen a lot. And I am worried it could encourage agents to write less clear code. Is there evidence this is improving the agents output?
> 
> 1. Most of the code has existing tests, with the parts that don't have indirect tests, and only 1 part being part of the sandbox app.
> 2. you are right, l actually doubled the complexity rule value (the initial 20 resulting in 120 functions being flagged). On the agent complexity concern, i can't really say if it will worsen it or improve it, but having a high ceiling value, means that at least most code will never get it, except for some rare cases (like the main one in this review is a single function that became huge, which in itself is bad practice). So i see it as more of a preventative measure for massive functions / bad practices.
> 
> In terms of the exhaustiveness check, l actually went back through and checked the old ESLint configs, and this repo never had this rule enabled. I did find a couple of files that manually did the `never` stuff for the exhaustive check. I myself normally prefer the exhaustiveness check on switch statements, but since ultracite prefers the no-useless ones, and its more of an industry standard config, l went with its one this time. But l am happy to enable the exhaustiveness check (and disable the no-useless one) in the FLUP if you want?

1. Got it. Generally when we add new rules, we don't modify existing files since it can create unnecessary bugs for something that is nice to have but I will defer to you if you have tested it
2. Sounds good. We can see how it works in practice
Yes that would be nice to enable

oxlint.config.ts
```diff
@@ -53,9 +53,10 @@ export default defineConfig({
     "unicorn/no-lonely-if": "off",
     "jsdoc/require-param-description": "off",
     "jsdoc/require-returns-description": "off",
-    // Cyclomatic-complexity ceiling; the large discriminated-union switch helpers
-    // in the docs packages exceed it. Splitting them up is pure churn here.
-    "complexity": "off",
+    // Ceiling of 40 (not the preset's 20): the default flagged ~100
+    // discriminated-union dispatchers that read fine as one flat switch. 40
+    // targets the real outliers and can be ratcheted down later.
```
nfdiop, line 58: since this is subjective, let's make this "warn" level instead and we can change later

packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ActionTypeShapeExtractor.ts
olliejenningsPT: I was just using the modern private syntax that is runtime private rather than compile-time private, can make this adopt the rest of the class conventions if we want it to match the rest of the file

oxlint.config.ts
olliejenningsPT: Yep will do

packages/faux/src/FauxFoundry/validateAction.ts
olliejenningsPT: Yep, let me add it back in

packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ActionTypeShapeExtractor.ts
olliejenningsPT: Offline discussion, will change to `private` to keep with rest of the file

olliejenningsPT:
Closing, as l think the general plan is this will not offer much in terms of immediate value and changing current code to meeting this lint rule poses a risk we don't need

PR #3691, opened 2026-07-15
Restructure ObjectTable docs into a one-pager

ziyunp:
- ObjectTable.mdx: render usage, props, and styling tokens below the live demo via the one-pager renderer (title/description/TOC from frontmatter)
- ObjectTable.md: use the gen-props AUTOGEN props reference generated from ObjectTableApi.ts; keep the consolidated styling-tokens table
- ObjectTableExamples.md: new; holds the examples and extended guidance (data sources, columns, sorting, selection, editing, exporting, headless hooks, TypeScript tips, troubleshooting, theming), linked at the bottom


waynecai1213:
Love the component preview on the Docs page! 
Some comments:
- I asked claude and seems Storybook has a native `addon-docs` floating table of content panel. Like the screenshot: 
  - image
  - Also the table of content item section in this PR opens a new page without the left sidebar which I supposed is not intentional. 
- Could we fix the prop table overflowing? 
- Clicking on the "ObjectTable examples and guide" link at the bottom doesn't show the page at the moment. 
- Is there a way to better group the Features, for example, "Single selection", "Multiple selection", "Controlled Selection" should be placed closer together as they are all about selection?

PR #3696, opened 2026-07-16
[Storybook] Refactor: split ObjectTable stories into feature groups

ziyunp:
## Summary

Restructures the ObjectTable Storybook stories to mirror the file layout introduced in #3519, without changing any story content.

The monolithic `ObjectTable.stories.tsx` (~3.1k lines, 28 stories) is split so each feature area lives under its own sidebar group:

Screenshot 2026-07-21 at 12 09 58

## Structure

- `objectTableStoryHelpers.tsx` — shared `meta`, column/query definitions, and interaction-test helpers only. No story objects live here.
- Each `Features/*.stories.tsx` (Columns, Sorting, Selection, Editing, DataSources, Display, Interactions, Overlays) holds its own full story definitions, importing the shared meta/helpers.
- `ObjectTable.stories.tsx` is trimmed to the component-root `Default` landing demo.

## Notes
- Story bodies are unchanged — verified all 28 stories are re-exported exactly once, with no duplicates or omissions.


packages/react-components-storybook/.storybook/preview.tsx
```diff
@@ -102,13 +102,42 @@ const preview: Preview = {
...
+           * The ordering within each component folder:
+           * Docs
+           * Default
+           * Features
+           * Building Blocks
+           */
+          const sectionOrder = ["Docs", "__root__", "Features"];
+          const aSub = aParts.length > 2 ? aParts[2] : "__root__";
+          const bSub = bParts.length > 2 ? bParts[2] : "__root__";
+          const aRank =
```
olliejenningsPT: I feel like `aRank` and `bRank` code is hard to read, understand the need for it, just the nested ternarys make it difficult to read

olliejenningsPT: Might be a cleaner way:

```ts
if (
    aParts[0] === "Components" &&
    bParts[0] === "Components" &&
    aParts[1] === bParts[1]
  ) {
    // Use a lookup table, which makes added new sections a 1 line change
    const sectionRank = {
      Docs: 0,
      __root__: 1, // root stories ("Default")
      Features: 2,
      // unknown: 3 => this you can use if you want to have a fixed unknown
      "Building Blocks": 4,
    };
    function rankOf(parts): number {
      const sub = parts.length > 2 ? parts[2] : "__root__";
      const rank = sectionRank[sub];
      return rank === undefined ? 3 : rank; // using 3, akin to your 99 to make building blocks always bottom
    };

    const aRank = rankOf(aParts);
    const bRank = rankOf(bParts);
    if (aRank !== bRank) return aRank - bRank;

    // Here again we have your default fallback logic.
    const aIsDefault = a.name === "Default";
    const bIsDefault = b.name === "Default";
    if (aIsDefault !== bIsDefault) return aIsDefault ? -1 : 1;
  }
```

ziyunp: Made it cleaner using array's index now.

packages/react-components-storybook/src/stories/ObjectTable/Recipes/WithConfigureColumnsButton.stories.tsx
nfdiop, line 1: not blocking: we may want to move these out and move to a Recipes folder

nfdiop (approved):
nit: since we are also changing packages/react-components-storybook/.storybook/preview.tsx which affects other files, worth updating the PR description. Not blocking for the changeset since this is for storybook

PR #3697, opened 2026-07-16
Allow overriding labels in ObjectTable

ziyunp:
Add a `labels` prop to `ObjectTable`/`BaseTable` (and the standalone `ColumnConfigDialog`/`MultiColumnSortDialog`) so every user-facing string can be overridden for localization or wording


packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -614,6 +615,13 @@ export interface ObjectTableProps<
   tableRef?: React.Ref<ObjectTableHandle<Q, RDPs>>;
 
   className?: string;
+
+  /**
+   * Overrides for the table's user-facing strings. Provide any subset; unset keys fall
+   * back to the built-in English defaults. Use this to localize the table or
+   * adjust wording. See {@link ObjectTableLabels}.
+   */
+  labels?: Partial<ObjectTableLabels>;
```
olliejenningsPT, line 626: `EditableCell` is wrapped in `React.memo` (`EditableCell.tsx:365`), and the surrounding code goes out of its way to keep that memo effective: `filterCellEditsToRow` returns a stable `undefined` reference specifically so unedited rows can skip re-rendering when edits change elsewhere in the table (see the comment in `DefaultCellRenderer.tsx`).

`EditableCellInner` now reads the labels context via `useObjectTableLabels()`. A memoized component still re-renders whenever a context it consumes changes identity, regardless of props. When `labels` is passed as an inline object, the provider's `useMemo` produces a new merged object every render, so the context value changes every render and every `EditableCell` re-renders on every table render (e.g. on each keystroke while editing). That quietly defeats the row-level memoization.

The default path is fine: with no `labels` (or a memoized one) the provider returns the stable default constant. Could we add a line to this JSDoc noting that `labels` should be a stable/memoized reference? It matches the package rule about always memoizing non-primitive props and closes the trap.

packages/react-components/src/object-table/ColumnConfigDialog.tsx
```diff
@@ -54,14 +67,26 @@ interface ColumnItem {
 }
 
 export function ColumnConfigDialog({
+  labels,
+  ...props
+}: ColumnConfigDialogProps): React.ReactElement | null {
```
olliejenningsPT: Nit: this wrapper now always returns the `ObjectTableLabelsProvider` element, so the `| null` here is unreachable (the inner component keeps `ReactElement | null`, which is correct). `MultiColumnSortDialog`'s wrapper is typed as just `ReactElement`, so it's also a small inconsistency between the two.

packages/react-components/src/object-table/ObjectTableApi.ts
ziyunp: Handled this in ObjectTableLabels by wrapping labels in useRef

packages/react-components/src/object-table/ObjectTableLabels.tsx
```diff
@@ -0,0 +1,260 @@
...
+ */
+export function useObjectTableLabels(): ObjectTableLabels {
+  return useContext(ObjectTableLabelsContext);
+}
+
+/**
+ * Wraps `Inner` so it accepts an optional `labels` prop and supplies the merged
+ * {@link ObjectTableLabels} to its subtree.
+ *
+ * For generic components, e.g. `BaseTable use
```
olliejenningsPT: missing ` after BaseTable

olliejenningsPT (approved):

PR #3708, opened 2026-07-17
Cbac banner loading state

lukeMuratore12:
- Add `isLoading` prop to the `react-components` CbacBanner that allows users to render a skeleton placeholder in the case that their markings are loading

https://github.com/user-attachments/assets/c5cb507c-a20c-470e-a5f2-422d77661915

### DIFF
--- a/packages/react-components/src/cbac-picker/CbacBanner.tsx
+++ b/packages/react-components/src/cbac-picker/CbacBanner.tsx
@@ -25,6 +25,7 @@ export interface CbacBannerProps {
   onClick?: () => void;
   onDismiss?: () => void;
   className?: string;
+  isLoading?: boolean;
 }
 
 export function CbacBanner({
@@ -32,8 +33,11 @@ export function CbacBanner({
   onClick,
   onDismiss,
   className,
+  isLoading = false,
 }: CbacBannerProps): React.ReactElement {
-  const { banner } = useCbacBanner({ markingIds });
+  const { banner, isLoading: isFetchingCbacMarkings } = useCbacBanner({
+    markingIds,
+  });
   const resolved = resolveBannerDisplay(banner);
 
   return (
@@ -44,6 +48,7 @@ export function CbacBanner({
       onClick={onClick}
       onDismiss={onDismiss}
       className={className}
+      isLoading={isFetchingCbacMarkings || isLoading}
     />
   );
 }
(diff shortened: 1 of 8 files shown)


packages/react-components/src/cbac-picker/CbacBanner.tsx
```diff
@@ -44,6 +48,7 @@ export function CbacBanner({
       onClick={onClick}
       onDismiss={onDismiss}
       className={className}
+      isLoading={isFetchingCbacMarkings || isLoading}
```
lukeMuratore12, line 51: Also blocking rendering on loading from `useCbacBanner`

packages/react-components/src/cbac-picker/base/BaseCbacBanner.tsx
```diff
@@ -57,6 +60,10 @@ export function BaseCbacBanner({
     [onDismiss]
   );
 
+  if (isLoading) {
+    return <BannerSkeleton className={className} />;
```
ziyunp, line 64: Can we reuse LoadingCell / LoadingCellContent please?

lukeMuratore12: I believe [LoadingCell](https://github.com/palantir/osdk-ts/blob/main/packages/react-components/src/object-table/LoadingCell.tsx#L33-L39) renders a `<td>` so I'm not sure if it would fit the same needs of this banner. 

We already reuse `SkeletonBar` here from `base-components`, which `LoadingCellContent` seems to be a [wrapper around with table-based styling](https://github.com/palantir/osdk-ts/blob/main/packages/react-components/src/object-table/LoadingCell.module.css#L24-L25). I think using `LoadingCellContent` here might mess up the banner height (and potentially some other styling) because of the height override [here](https://github.com/palantir/osdk-ts/blob/main/packages/react-components/src/object-table/LoadingCell.module.css#L18)

ziyunp: Oh got it, sorry, I thought it's in the ObjectTable.

ziyunp (approved):

packages/react-components/src/cbac-picker/base/BaseCbacBanner.tsx
lukeMuratore12: No worries!

olliejenningsPT: Is this potentially introducing an issue where you have a `classificationString` already, but loading is set to true, so you then revert to showing a loading state?

E.g. the hook refetches data, suddenly all cbac going into loading state in say an object table.

I know the CBACPicker has a guard against this, so if loading and no data, it showing specific loading state

PR #3713, opened 2026-07-18
Add CBAC picker tests and refactor the max-classification callout

olliejenningsPT:
## Summary
The CBAC (classification-based access control) picker in `@osdk/react-components` shipped with only two test files (`selectionLogic`, `validationMessages`), leaving its components, error/formatting utilities, OSDK data-fetching hooks, and the top-level component unverified. This PR adds 11 test files (74 tests) across every layer of the feature, relocates the two existing tests so the whole feature follows the package's `__tests__/` convention, and refactors an internal anti-pattern (a hook that returned JSX) into a proper component. The full `cbac-picker/` suite now runs at 105 passing tests, with `pnpm turbo check` green.

## Why this matters
This is security-sensitive UI: it decides which classification markings a user applies to data, so a silent regression here has real consequences. The new tests pin down the user-facing behavior and the non-obvious internal logic (banner stickiness, controlled-selection reset, overflow collapsing, the max-classification cap) that a refactor could quietly break.

## Test coverage added

**Top-level component (`CbacPicker`)**
- `onChange` fires with the correct marking ids when a marking is toggled (the component's primary output, previously unverified end to end).
- `readOnly` prevents selection changes.
- Dismissing the banner clears the selection and fires `onChange([])`.
- The max-classification callout renders only when a constraint is provided.

**OSDK-wrapper hooks (platform APIs mocked)**
- `useCbacPickerState`: permissive defaults before restrictions load, single-error passthrough vs multi-error aggregation into an `AggregateError`, combined loading state, `retry` fanning out to every underlying query, and banner stickiness (the last banner is retained while the next one is being fetched, which prevents flicker).
- `useCbacSelection`: seeding from initial ids, resetting when `initialMarkingIds` changes, conjunctive add/remove on toggle, dismiss, and reset.

**Base components (OSDK-agnostic, Testing Library)**
- `MarkingButton`: toggling fires with the right id, selected and implied markings report `aria-pressed`, implied markings render a parenthesized label, and disallowed or read-only markings cannot be toggled.
- `BaseCbacPicker`: category rendering, the error state hiding the marking grid, initial-loading vs reload behavior (existing categories stay rendered while refetching so the UI does not flash), validation-warning gating, and banner render plus dismiss.
- `CategoryMarkingGroup`: the grid overflow logic that keeps the first markings visible and collapses the rest behind a "+N more" control.
- `CbacPickerDialogFooter`: cancel and confirm callbacks, plus the submit button being disabled (and confirm blocked) when a disabled reason is supplied.

**Pure utilities (fast, no mocking)**
- `formatCbacError`: the full error-to-message mapping (session expiry, permission denied, not found, server errors, network detection, and fallbacks), including `AggregateError` unwrapping to the first underlying error. This is exactly the text a user sees when a request fails.
- `cbacPickerUtils`: banner background composition (gradient vs single color vs transparent), UNMARKED defaults, marking-to-category grouping with the id-fallback path, and required-group name resolution.
- `selectionStateHelpers`: the display rules (implied markings shown in parentheses, per-state tooltip text).

## Refactor: `useConstraintCallout` to `ConstraintCallout`
The "Maximum allowed classification" callout was produced by `useConstraintCallout`, a hook that returned JSX. A hook returning rendered markup is an anti-pattern: it mixes data fetching with rendering behind a `use*` name and is awkward to test. It is now a `ConstraintCallout` component that fetches its banner data and renders `MaxClassificationField`, and the two call sites (`CbacPicker`, `CbacPickerDialog`) render it conditionally through a memoized `validationCallouts`.

Benefits:
- Data fetching lives in a component, rendering lives in JSX, which matches the package's stated architecture (outer OSDK-aware component fetches, base component stays presentational).
- The `enabled` flag and `?? EMPTY_ARRAY` fallback are gone. Conditional mounting replaces them, so the banner request only runs when a constraint exists.
- It is now testable as an ordinary component, which is why `ConstraintCallout` gets its own test file.

This is an internal-only change: no public API surface changes (`useConstraintCallout` was never exported).

## Housekeeping
- Relocated the two pre-existing tests (`selectionLogic`, `validationMessages`) into `__tests__/` with `git mv`, so history is preserved and the entire feature now matches the convention documented in `packages/react-components/CLAUDE.md`.
- Tests follow the established package style: vitest with `@testing-library/react` (`fireEvent`, plain attribute assertions, `afterEach(cleanup)`), and `vi.mock` for the OSDK platform-API hooks.
- `patch` changeset added (internal refactor, no API or behavior change).

## Verification
- 105 tests passing across 13 files.
- `pnpm turbo check --filter=@osdk/react-components` green; typecheck and lint/oxfmt clean.
- Security review: no findings (React composition plus tests, no unsafe rendering sinks). Code review: the two actionable items it raised (wrapper-level coverage of the callout wiring, and the CLAUDE.md rule to memoize non-primitive props) are addressed in this PR.

## Out of scope
The dialog and banner-popover wrappers (`CbacPickerDialog`, `BaseCbacPickerDialog`, `CbacBanner`, `CbacBannerPopover`) are not directly covered here and can follow up if we want fuller coverage.


packages/react-components/src/cbac-picker/__tests__/ConstraintCallout.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  it("requests banner data for the constraint's marking ids", () => {
+    const constraint: MaxClassificationConstraint = {
+      bannerClassificationString: "SECRET",
+      markingIds: ["m1", "m2"],
+    };
+    render(<ConstraintCallout constraint={constraint} />);
+
+    expect(useCbacBanner).toHaveBeenCalledWith({ markingIds: ["m1", "m2"] });
```
nfdiop, line 63: Hum, I'm not sure of the value of this test. This is testing that the hook is called by this component: which may not track valuable regression. Is there something else we want to test here or can we delete?

packages/react-components/src/cbac-picker/__tests__/useCbacPickerState.test.ts
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  it("passes a single hook error straight through", () => {
+    setHooks({ errors: { categories: new Error("boom") } });
+    const { result } = renderHook(() => useCbacPickerState([]));
+    expect(result.current.error?.message).toBe("boom");
+    expect(result.current.error).not.toBeInstanceOf(AggregateError);
+  });
+
+  it("combines multiple hook errors into an AggregateError", () => {
```
nfdiop, line 124: Nice!! This can easily be broken

packages/react-components/src/cbac-picker/__tests__/useCbacPickerState.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      textColor: "#000000",
+      backgroundColors: ["#123456"],
+      markingIds: ["m1"],
+    };
+    setHooks({ banner });
+    const { result, rerender } = renderHook(() => useCbacPickerState(["m1"]));
+    expect(result.current.banner).toBe(banner);
+
+    // The next fetch is in flight and returns no banner yet.
+    setHooks({ banner: undefined });
```
nfdiop, line 152: Sorry if I'm missing it but what guarantees that results are refetching here?

packages/react-components/src/cbac-picker/__tests__/ConstraintCallout.test.tsx
olliejenningsPT: So my thought here is, because am treating the CBAC stuff as highly sensitive, since its used for security purposes, this test would normally been seen pointless, but am seeing it as a test that if it fails, the person changing the Component `ConstraintCallout` is actively making the decision that they no longer want to call this hook, and there must be a valid reason why.

packages/react-components/src/cbac-picker/__tests__/useCbacPickerState.test.ts
olliejenningsPT: So nothing guarantees per-say since in this test everything is mocked out. But the idea is if we have a banner, and then the banner is being refetched. Whilst it is currently being refetched, we do not lose the current banner information we have. So its a retention test

nfdiop: By the time we call `expect(result.current.banner).toBe(banner);` after the re-render what guarantees that we are still refetching? If we don't have guarantee, hard for me to see the value of the test.

packages/react-components/src/cbac-picker/__tests__/ConstraintCallout.test.tsx
nfdiop: That is true for every one of our components. What is the expected behaviour here? Can we test the behaviour instead of the implementation details?

olliejenningsPT: I would argue that not displaying the right CBAC banner and having some miss-classified or accidentally leaked due to a classification bug is more serious than say the PDF viewer having a bug in its auto-size feature.

The expected behaviour is this component should always be calling the hook, and if that is being changed, the test should be introducing friction.

packages/react-components/src/cbac-picker/__tests__/useCbacPickerState.test.ts
olliejenningsPT: Network latency, which can be unstable means that it could re-render instantly or in X seconds. The `renrender` method on the line below means that we trigger a re-render of the hook. There is little value in adding a fake timer to act as network latency, setting the hook to be undefined covers this behaviour

nfdiop: To clarify: I didn't ask for fake timer or similar things. The question was how do we know that we are refetching and this reply `setting the hook to be undefined covers this behaviour` answers it. So all good for me

packages/react-components/src/cbac-picker/__tests__/ConstraintCallout.test.tsx
nfdiop: Non-blocking: I'd still test the rendered banner output instead of the hook call: it would better catch what would actually go wrong.

nfdiop (approved):

PR #3714, opened 2026-07-18
Add vector property support to osdk/maker

ekatz-pt:
Adds the vector property type to @osdk/maker, so you can author vecto with defineObject, defineSharedPropertyType, and on interfaces. Follows the existing exotic-variant pattern (struct/decimal/marking); the IR target already exists in @osdk/client.unstable.

Usage:

  embedding: {
    type: {
      type: "vector",
      dimension: 768,
      supportsSearchWith: "COSINE_SIMILARITY",
    },
  }

Changes:
- New vector authoring type (dimension + supportsSearchWith required; embeddingModel/quantization optional).
- Emits vector IR on both the object/SPT and interface conversion paths.
- Vectors skip render hints and text-search indexing (they use KNN search).
- Rejects vectors in struct fields and array: true on a vector.
- Tests for objects, SPTs, and interfaces.

Not included (v1): vectors as action parameters, nested in structs or arrays, and embedding-model auto-provisioning. Backend constraints (Object Storage V2, dimension bounds) stay with the installer.

### DIFF
--- a/packages/maker/src/api/defineObject.ts
+++ b/packages/maker/src/api/defineObject.ts
@@ -50,7 +50,11 @@ import type {
 import type { ObjectTypeDefinition } from "./object/ObjectTypeDefinition.js";
 import type { ObjectTypeStatus } from "./object/ObjectTypeStatus.js";
 import type { PropertyTypeType } from "./properties/PropertyTypeType.js";
-import { isExotic, isStruct } from "./properties/PropertyTypeType.js";
+import {
+  isExotic,
+  isStruct,
+  isValidVector,
+} from "./properties/PropertyTypeType.js";
 // From https://stackoverflow.com/a/79288714
 const ISO_8601_DURATION =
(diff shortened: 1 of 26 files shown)


packages/maker/src/api/defineObject.ts
```diff
@@ -337,6 +341,10 @@ function convertUserObjectPropertyType(
   apiName: string,
   property: ObjectPropertyTypeUserDefinition,
 ): ObjectPropertyType {
+  invariant(
+    isValidVector(property.type, property.array),
```
ethanai22: feels like all the invocations of `isValidVector` should be moved to the property conversion step.

packages/maker/src/api/properties/PropertyTypeType.ts
```diff
@@ -80,6 +90,16 @@ type PropertyTypeTypeDecimal = {
   scale?: number;
 };
 
+export type PropertyTypeTypeVector = {
+  type: "vector";
+  /** The dimensionality of the vector (must be at least 1). */
+  dimension: number;
+  /** The similarity function(s) supported for vector search. Must specify exactly one. */
+  supportsSearchWith: VectorSimilarityFunction[];
```
ethanai22: why make this a list if it must be exactly one?

packages/maker/src/conversion/toMarketplace/propertyTypeTypeToOntologyIrInterfaceType.ts
```diff
@@ -113,6 +113,17 @@ export function propertyTypeTypeToOntologyIrInterfaceType(
         },
       };
 
+    case typeof type === "object" && type.type === "vector":
```
ethanai22, line 116: can you add the same handling for the `maker-experimental` conversion path? SuperRepos depend on that

aep000 (approved):

PR #3717, opened 2026-07-19
Add local Vitest coverage tooling

olliejenningsPT:
### Why

We run Vitest across ~39 packages but have never had a way to measure test
coverage. This adds a simple local command to check coverage so we can see where
we stand.

Nothing is wired into CI and nothing is published. This is a local developer
tool only, until we decide where to surface coverage.

### What it adds

- `pnpm run coverage` runs every package's suite with coverage, prints a
  per-package line-coverage summary and the repo totals in the terminal, and
  writes a full HTML report to `coverage/index.html`.
- Coverage is collected per package with `@vitest/coverage-v8` (already a
  dependency) via turbo, then merged into one report with `istanbul-lib-*`.
- Coverage config (include `src/**`; exclude tests, generated code, and
  `index.ts` barrels) lives in each `vitest.config.mts`, generated centrally by
  monorepolint so all packages stay consistent.
- Documented in the README under "Test coverage".

Current baseline is ~61% lines overall, ranging from ~1% (some CLI/generator
tooling) to 100%.

### Notes

- Bumps `@osdk/react-devtools` from Vitest 2.x to 3.x so it works with the v3
  coverage provider (it was the only package still on Vitest 2). Changeset
  included.
- `@osdk/example-generator` is excluded from collection: its only test is a
  codegen-drift check that calls `process.exit`, unrelated to coverage.
- `@osdk/react-components` is collected in its own turbo pass so its large
  happy-dom suite does not starve the others.

### Follow-ups (not in this PR)

- Somewhere to publish coverage (CI job, PR comment, or badge) once we decide
  where.
- Per-package thresholds to guard against regressions.


scripts/coverage/merge-coverage.mjs
```diff
@@ -0,0 +1,127 @@
...
+// report under `coverage/` at the workspace root.
+//
+// Outputs:
+//   coverage/coverage-final.json    merged raw per-file coverage
+//   coverage/coverage-summary.json  totals + per-file summary (machine readable)
+//   coverage/index.html             HTML report (open in a browser)
+//   coverage/lcov.info              lcov (handy for editors / external tools)
+// ...and prints the aggregate totals to stdout.
+
+import { globSync, mkdirSync, readFileSync, rmSync } from "node:fs";
```
nfdiop, line 29: globSync is not supported in node 18 I think

packages/cli/vitest.config.mts
```diff
@@ -20,6 +20,18 @@ export default defineConfig({
...
+    coverage: {
+      include: ["src/**"],
+      // Exclude tests, generated code, and index.ts barrels (no logic).
+      exclude: [
+        "**/*.test.*",
+        "**/__tests__/**",
+        "**/__mocks__/**",
+        "**/generatedNoCheck/**",
+        "**/*.d.ts",
+        "**/index.ts",
```
nfdiop, line 32: I think some of the files like packages/cli/src/commands/site/index.ts have logic in it

package.json
```diff
@@ -7,6 +7,9 @@
     "check-mrl": "mrl check --verbose",
     "ci:publish": "./scripts/ci-publish.sh",
     "ci:publishSnapshot": "pnpm run prePublish && pnpm exec changeset version --snapshot && pnpm exec changeset publish --no-git-tag --snapshot --tag=next",
+    "coverage": "pnpm run coverage:collect; ec=$?; pnpm run coverage:merge; exit $ec",
```
nfdiop, line 10: maybe 
```
"coverage": "pnpm run coverage:collect; collect_rc=$?; pnpm run coverage:merge; merge_rc=$?; exit $((collect_rc != 0 || merge_rc != 0))"
```
so that when merge fail, we also fail (if that is wanted)?

nfdiop (commented):
Thanks for adding coverage! 

(not-blocking at all) Curious how do you account for the fact that we can have 100% test coverage but the tests are not correct?

packages/cli/vitest.config.mts
olliejenningsPT: They do, this was an intentional decision because around ~3 `index.ts` files have logic in them but all the rest don't, so l felt like the best decision for now was to include all, rather than having to build some special logic, which would involve modifying the `monorepolint` setup so that we could include the `index.ts` for specific folders. Which we would then need to update every time we wanted special `index.ts` includes.

scripts/coverage/merge-coverage.mjs
olliejenningsPT: This only runs on `node24` stage if we run it on the CI, we don't perform any merge coverage on the test matrix stages. This is because we only want to collect the test coverage from one stage, we don't care about the other X stages.

package.json
olliejenningsPT: Yeah thats not a bad shout, l felt for the initial part it would be non-blocking to not interfere with workflows, and allow for discussion around want we want from the whole test-coverage type setup

olliejenningsPT:
> Thanks for adding coverage!
> 
> (not-blocking at all) Curious how do you account for the fact that we can have 100% test coverage but the tests are not correct?

So l complete agree with you that they won't catch anything, and if we focus on that aspect / number alone its pointless. The way l think about this is:

1. Coverage tells us what the tests touch, not correctness (you can have tests for functions that assert nothing). Its more of a friction to introduce habits, with correctness coming from how we write the test assertions.
2. The real value, it not the day-1 bug catch, its what happens when we change tested code. Even if a test asserts a wrong assumption, the fact that we make that assumption explicit, so when changes the code and it breaks, they are forced to stop and think. That thinking aspect is what we care about.

If we really want to make our tests valuable, we can do mutation testing, which is semi-industry practice (done at places like Google, Meta, etc):

* [Stryker - Mutation Testing Framework for JS (integrates with Vitest)](https://stryker-mutator.io/docs/)
* [Meta - llms-are-the-key-to-mutation-testing-and-better-compliance](https://engineering.fb.com/2025/09/30/security/llms-are-the-key-to-mutation-testing-and-better-compliance/)

scripts/coverage/merge-coverage.mjs
nfdiop: Got it!

packages/cli/vitest.config.mts
nfdiop: Got it! Maybe worth a comment?

nfdiop (approved):

PR #3721, opened 2026-07-20
Write dependency file for maker-experimental

ethanai22:


packages/maker/src/api/defineOntology.ts
```diff
@@ -779,7 +779,7 @@ function getEntityTypeName(type: string): string {
   }[type]!;
 }
 
-function writeDependencyFile(dependencyFile: string): void {
+export function writeDependencyFile(dependencyFile: string): void {
```
davidpaquin, line 782: can we just port this over instead of making it public?

ethanai22: any specific reason you prefer that? The internal state/dependencies are already shared between maker and maker-experimental via `initializeOntologyState`

davidpaquin (approved):

packages/maker/src/api/defineOntology.ts
davidpaquin: i guess if the door has already been opened, go for it 🤷 
just seems unfortunate to expose the internal stuff

PR #3725, opened 2026-07-21
Add "View less" collapse toggle to ListogramInput

nfdiop:
## What

Makes the ListogramInput **"View all (N)"** button a two-way toggle. Today it expands the truncated list one-way and then unmounts — there's no way to collapse back. Now clicking it again (**"View less"**) collapses the list to its head.

## Details

- Render condition changed from `hasMore && !isExpanded` to `hasMore`, so the button stays mounted and relabels rather than unmounting.
- Label reads `View all (N)` when collapsed and `View less` when expanded; `onClick` flips `isExpanded`.
- Added `aria-expanded` on the toggle for screen-reader state. No `aria-controls`/wrapper — this is a truncated-list show-more/less affordance, not an APG disclosure.
- Collapse reuses the existing display memo (head + below-fold-selected tail) — no new layout logic, no public API change, no CSS change.

## Storybook

Added an interaction play to the `WithBelowFoldSelection` story exercising the full round-trip: a below-fold selected value sits at the tail → unselect it (drops below the fold) → expand via "View all" → re-select it → "View less" collapses back to the initial state.

## Testing

- Unit tests in `ListogramInput.test.tsx` cover the label branch, render condition, and `aria-expanded`, and preserve the existing ordering/tail invariants.
- The Storybook interaction test verifies the collapse round-trip end-to-end.

demo

### DIFF
--- a/packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
+++ b/packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
@@ -1409,7 +1409,8 @@ export const WithBelowFoldSelection: Story = {
         story:
           "A selected value that sorts below the collapsed fold stays visible, " +
           "appended at the tail of the collapsed view rather than hoisted to " +
-          'the top. The "View all" button remains so the rest can be revealed.',
+          'the top. The "View all" toggle reveals the rest, and "View less" ' +
+          "collapses back to this state.",
       },
       source: {
         code: `<FilterList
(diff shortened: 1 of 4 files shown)


packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1422,6 +1423,95 @@ export const WithBelowFoldSelection: Story = {
     },
   },
   render: (args) => <WithBelowFoldSelectionStory {...args} />,
+  play: async ({ canvasElement }) => {
```
nfdiop, line 1426: FLUP from https://github.com/palantir/osdk-ts/pull/3643/changes#r3613686336

packages/react-components/src/filter-list/base/inputs/__tests__/ListogramInput.test.tsx
```diff
@@ -132,7 +132,7 @@ describe("ListogramInput ordering", () => {
     expect(screen.queryByRole("button", { name: "View all (4)" })).toBeNull();
   });
 
-  it("omits the View all button when all values fit in the collapsed view", () => {
+  it("omits the view toggle entirely when all values fit in the collapsed view", () => {
```
nfdiop, line 135: changed to also assert we don't show "View less"

packages/react-components/src/filter-list/base/inputs/ListogramInput.tsx
```diff
@@ -202,14 +203,15 @@ function ListogramInputInner({
...
 
-          {hasMore && !isExpanded && (
+          {hasMore && (
             <Button
               type="button"
               className={styles.viewAllButton}
+              aria-expanded={isExpanded}
               // eslint-disable-next-line react/jsx-no-bind
-              onClick={() => setIsExpanded(true)}
+              onClick={() => setIsExpanded((v) => !v)}
```
ziyunp: Can we move this to a useCallback and remove the eslint-disable pls?

packages/react-components/src/filter-list/base/inputs/__tests__/ListogramInput.test.tsx
```diff
@@ -147,8 +147,98 @@ describe("ListogramInput ordering", () => {
...
+
+    fireEvent.click(screen.getByRole("button", { name: "View all (4)" }));
+    expect(getRenderedOrder()).toEqual([
+      "Engineering",
+      "Sales",
+      "Marketing",
+      "Support",
+    ]);
+
+    const collapseToggle = getToggle();
```
ziyunp: Prefer to explicitly get View less button here. getToggle function matches both view all and view less

nfdiop: So it's by design. I liked the idea to call it toggle since I see it as the important bit (toggling the button). But happy to name viewLess

ziyunp (dismissed):

packages/react-components/src/filter-list/base/inputs/__tests__/ListogramInput.test.tsx
ziyunp: I'm not referring to the naming of the function, but the function matches both View all and View less. I prefer to have an explicit match in the test where we look for "View less" when the list is expanded

  const toggles = screen.queryAllByRole("button", {
    name: /View (all|less)/u,
  });

ziyunp (approved):

PR #3727, opened 2026-07-21
feat: add leading/trailing options to useDebouncedCallback

nfdiop:
## What

Adds an additive, optional `{ leading, trailing }` options argument to `@osdk/react`'s `useDebouncedCallback`. The hook was trailing-only; callers can now opt into a leading-edge invocation as well.

```ts
useDebouncedCallback(fn, 500, { leading: true, trailing: true });
```

## Why

ActionForm's Phase-2 re-validation needs leading + trailing debouncing (500ms), matching Forge behavior.

## Non-breaking

- The existing `(callback, delay)` call shape is unchanged — no options means trailing-only, exactly as before.
- `.cancel()` and `.flush()` are unchanged.
- Defaults: `leading` false, `trailing` true.

## Behavior pinned by tests

- Leading-only fires synchronously and re-arms after a quiet window; trailing collapses to the last args.
- Leading + trailing fires the first args on the leading edge and the last args on the trailing edge; a lone call fires once (no double-fire).
- `{ leading: false, trailing: false }` never invokes.
- `cancel()` / `flush()` behave correctly under every configuration, including as no-ops when nothing is pending.


nfdiop:
Closing in favour of https://github.com/palantir/osdk-ts/pull/3739

PR #3728, opened 2026-07-21
feat(react-components): add creatable combobox support to DropdownField

nfdiop:
## What

Adds a creatable ("Create …") capability to `DropdownField`'s searchable Combobox.

Two optional props on `DropdownFieldProps`:

- `createNewItemFromQuery?: (query) => V | undefined` — presence enables creatable mode (implies searchable). When the trimmed query is non-empty and matches no existing option, a synthetic "Create …" item is offered; selecting it (by click, or by pressing Enter when the create row is the highlighted option) commits the value this callback returns via `onChange`. The **caller** does the coercion here (e.g. `coerceFieldValue`) and returns `undefined` to reject a query. Keeps `DropdownField` OSDK-agnostic.
- `createNewItemRenderer?: (query) => ReactNode` — overrides the **visible** content of the "Create …" item (defaults to `Create "<query>"`). The item's accessible name stays `Create "<query>"` regardless, so an override only changes visuals.

Works for both single-select and multi-select (multi behaves like a creatable tag input, appending each created value). The synthetic item is injected as a real, already-coerced `V` into base-ui's item list so base-ui handles click and keyboard selection natively. An already-selected value is never re-offered, and created values absent from `items` are surfaced back into the list so they stay visible — and, in multi-select, uncheckable — after selection.

## Scope

This is the reusable building block for the `oneOf` + `otherValuesAllowed: true` slice.

**Known follow-up:** Enter-to-create through `autoHighlight` (not added in this PR because it doesn't play nicely with the AsyncDropdown)

## Verification

- `pnpm turbo check --filter=@osdk/react-components`
- `pnpm --dir packages/react-components exec vitest run src/action-form` (creatable tests cover: synthetic item appears/suppressed, click + Enter commit with coercion, multi-select append, rejected coercion, custom renderer)

update

### DIFF
--- a/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
@@ -1338,6 +1338,13 @@ const creatableDropdownFormContent: ReadonlyArray<FormContentItem> = [
     fieldComponentProps: {
       items: DEPARTMENT_ITEMS,
       createNewItemFromQuery: coerceStringQuery,
+      // createNewItemRenderer overrides only the visible content of the
+      // "Create …" item; the accessible name stays the default.
+      createNewItemRenderer: (query) => (
+        <span>
+          <strong>Add</strong> new department "{query}"
+        </span>
+      ),
       placeholder: "Search or create a department...",
     },
   }),
@@ -1363,7 +1370,7 @@ export const WithCreatableDropdown: Story = {
     docs: {
       description: {
         story:
-          'Passing `createNewItemFromQuery` turns a dropdown into a creatable combobox: typing a value that matches no existing item offers a "Create …" option that commits the typed value (by click or by pressing Enter). Works for single-select and multi-select, where it behaves like a tag input.',
+          'Passing `createNewItemFromQuery` turns a dropdown into a creatable combobox: typing a value that matches no existing item offers a "Create …" option that commits the typed value (by click or by pressing Enter). Works for single-select and multi-select, where it behaves like a tag input. Pass `createNewItemRenderer` to customize the visible content of the "Create …" option (the accessible name stays the default).',
       },
       source: {
         code: `const coerceStringQuery = (query) => query.trim();
--- a/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
+++ b/packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
@@ -1330,6 +1330,79 @@ export const WithMultiSelectDropdown: Story = {
   },
 };
 
+const creatableDropdownFormContent: ReadonlyArray<FormContentItem> = [
+  field({
+    fieldKey: "department",
+    fieldComponent: "DROPDOWN",
+    label: "Department (creatable)",
+    fieldComponentProps: {
+      items: DEPARTMENT_ITEMS,
+      createNewItemFromQuery: coerceStringQuery,
+      placeholder: "Search or create a department...",
+    },
+  }),
+  field({
+    fieldKey: "tags",
+    fieldComponent: "DROPDOWN",
+    label: "Tags (creatable multi-select)",
+    fieldComponentProps: {
+      items: TAG_ITEMS,
+      isMultiple: true,
+      createNewItemFromQuery: coerceStringQuery,
+      placeholder: "Search or create tags...",
+    },
+  }),
+];
+
+export const WithCreatableDropdown: Story = {
+  args: {
+    formContent: creatableDropdownFormContent,
+    onSubmit: handleSubmit,
+  },
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'Passing `createNewItemFromQuery` turns a dropdown into a creatable combobox: typing a value that matches no existing item offers a "Create …" option that commits the typed value (by click or by pressing Enter). Works for single-select and multi-select, where it behaves like a tag input.',
+      },
+      source: {
+        code: `const coerceStringQuery = (query) => query.trim();
+
+const formContent = [
+  {
+    fieldKey: "department",
+    fieldComponent: "DROPDOWN",
+    label: "Department (creatable)",
+    fieldComponentProps: {
+      items: ["Engineering", "Marketing", "Sales", "Finance", "Operations", "Legal"],
+      createNewItemFromQuery: coerceStringQuery,
+      placeholder: "Search or create a department...",
+    },
+  },
+  {
+    fieldKey: "tags",
+    fieldComponent: "DROPDOWN",
+    label: "Tags (creatable multi-select)",
+    fieldComponentProps: {
+      items: ["Urgent", "Review", "Follow-up", "Archived", "Pinned"],
+      isMultiple: true,
+      createNewItemFromQuery: coerceStringQuery,
+      placeholder: "Search or create tags...",
+    },
+  },
+];
+
+// createNewItemFromQuery coerces the raw query into an item value.
+// Return undefined to reject a query (no create option shown).
+<BaseForm
+  formContent={formContent}
+  onSubmit={(formState) => console.log("Submitted:", formState)}
+/>`,
+      },
+    },
+  },
+};
+
 const richDropdownLabelFormContent: ReadonlyArray<FormContentItem> = [
   field({
     fieldKey: "assigneeUserId",
@@ -2310,6 +2383,10 @@ export const WithGridSection: Story = {
     onSubmit: handleSubmit,
   },
 };
+function coerceStringQuery(query: string): string {
+  return query.trim();
+}
+
 function getUserIdLabel(item: unknown): string {
   if (typeof item !== "string") {
     return String(item);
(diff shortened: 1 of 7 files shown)


packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -776,10 +776,228 @@ describe("DropdownField", () => {
       expect(onChange).not.toHaveBeenCalled();
     });
   });
+
+  describe("creatable", () => {
+    // Coerces the query to upper case so the test can prove the committed value
+    // is what `createNewItemFromQuery` returned, not the raw query text.
+    const coerceToUpper = (query: string): string => query.trim().toUpperCase();
+
+    it("renders a searchable combobox when createNewItemFromQuery is provided without isSearchable", async () => {
```
nfdiop, line 817: We need to allow the user to type to be able to create an item

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -508,6 +508,38 @@ describe("DropdownField", () => {
       });
     });
 
+    it("preserves the last controlled query when returning to uncontrolled mode", async () => {
```
nfdiop, line 511: I was debating not supporting the transition since it is generally not recommended at all but supporting

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -776,10 +808,396 @@ describe("DropdownField", () => {
...
+      await openAndType("Dana");
+
+      await vi.waitFor(() => {
+        expect(
+          screen.getByRole("option", { name: 'Create "Dana"' })
+        ).toBeDefined();
+      });
+    });
+
+    it("does not show a create item for an empty or whitespace query", async () => {
```
nfdiop, line 852: like in Blueprint

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -776,10 +808,396 @@ describe("DropdownField", () => {
...
+
+      await vi.waitFor(() => {
+        expect(
+          screen.getByRole("option", { name: "Dana Reviewer" })
+        ).toBeDefined();
+      });
+      expect(screen.getAllByRole("option")).toHaveLength(1);
+    });
+
+    it("does not surface selected-but-absent values when not creatable", async () => {
```
nfdiop, line 1148: decision I made to keep the existing behaviour. But maybe we should always show the values selected in the list?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -275,13 +287,124 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
 }: InnerComboboxProps<V, Multiple>): React.ReactElement {
   const [open, setOpen] = useState(false);
   const isOpen = !disabled && open;
 
+  // Creatable works for single- and multi-select. `internalQuery` tracks the
+  // search text when the query isn't controlled, so we can derive the
+  // synthetic item.
+  const isCreatable = createNewItemFromQuery != null;
+  const [internalQuery, setInternalQuery] = useState(query ?? "");
+  React.useEffect(() => {
```
nfdiop, line 310: Decided to always keep track of an internalQuery and make sure it is in sync. If I get a chance we can add a hook like this one https://www.npmjs.com/package/@radix-ui/react-use-controllable-state

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -243,6 +243,33 @@ export interface DropdownFieldProps<
...
+   *
+   * Coerce/convert the raw query into an item value here (e.g. via
+   * `coerceFieldValue`). Return `undefined` to reject the query, in which case
+   * no create option is shown.
+   *
+   * Providing this callback implies a searchable dropdown. Works for both
+   * single-select and multi-select — multi-select behaves like a creatable tag
+   * input, appending each created value to the selection.
+   */
+  createNewItemFromQuery?: (query: string) => V | undefined;
```
nfdiop, line 261: In blueprint, the user needs to define `createNewItemRenderer` for the creation to be allowed. Here I am using createNewItemFromQuery because that is a simpler interface

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -275,13 +293,118 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
+    if (query !== undefined) {
+      setInternalQuery(query);
+    }
+  }, [query]);
+  const currentQuery = query ?? internalQuery;
```
ziyunp: Query is always set to internalQuery, why do we need currentQuery here?

nfdiop: Thanks for catching. Let me use `internalQuery`: it was present because I didn't have the useEffect before and missed updating this

packages/react-components-storybook/src/stories/ActionForm/BaseForm.stories.tsx
```diff
@@ -1330,6 +1330,91 @@ export const WithMultiSelectDropdown: Story = {
   },
 };
 
+const creatableDropdownFormContent: ReadonlyArray<FormContentItem> = [
+  field({
+    fieldKey: "department",
```
ziyunp, line 1335: Why is this story added to BaseForm and not the ActionForm story? Prefer to have all feature examples in ActionForm.stories.tsx and only keep a few examples of using BaseForm as a custom form that does not use OSDK Actions

packages/react-components/src/action-form/FormFieldApi.ts
ziyunp: We can also propose this change to Blueprint

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -97,38 +110,43 @@ export const DropdownField: <V, Multiple extends boolean = false>(
     [itemToKey, resolvedItemToStringLabel]
   );
 
+  const resolvedIsItemEqual = isItemEqual ?? defaultIsItemEqual;
```
ziyunp, line 113: Wrap this in useCallback just like getKey above

ziyunp (commented):
.

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -450,6 +477,7 @@ export type ObjectSelectFieldProps<
   | "onQueryChange"
   | "disableClientSideFiltering"
   | "renderItemList"
+  | "createNewItemFromQuery"
```
ziyunp, line 480: Can we omit createNewItemRenderer as well?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -275,13 +293,117 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
+    const lowerQuery = trimmedQuery.toLowerCase();
+    const matchesExisting = dropdownItems.some(
+      (item) => itemToStringLabel(item).trim().toLowerCase() === lowerQuery
+    );
+    // The query already names an existing option — nothing to create.
+    if (matchesExisting) {
+      return undefined;
+    }
+    const created = createNewItemFromQuery(trimmedQuery);
+    if (created === undefined) {
```
ziyunp, line 369: Nit: Should we do created == null for consistency?

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -275,13 +293,117 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
+
+  // The synthetic "Create …" item is injected into the item list, so base-ui's
+  // built-in filter would hide it whenever its (coerced) label doesn't contain
+  // the typed query — e.g. a numeric coercer where "4.9" becomes value 4 with
+  // label "4". Exempt it by identity so it stays visible whenever it exists,
+  // while real items still filter normally via base-ui's locale-aware matcher.
+  const { contains } = BaseUICombobox.useFilter({ multiple: isMultiple });
+  const creatableFilter = useCallback(
+    (item: V, nextQuery: string, itemToString?: (item: V) => string) =>
+      (syntheticItem !== undefined && Object.is(item, syntheticItem)) ||
```
ziyunp, line 402: Nit: Extract this `syntheticItem !== undefined && Object.is(item, syntheticItem` into a func as it's being used twice

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -373,10 +518,16 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
...
         itemToStringLabel={itemToStringLabel}
         isItemEqualToValue={isItemEqual}
-        items={items}
-        inputValue={query}
-        onInputValueChange={onQueryChange}
-        filter={disableClientSideFiltering ? null : undefined}
+        items={effectiveItems}
+        inputValue={internalQuery}
+        onInputValueChange={handleInputValueChange}
+        filter={
```
ziyunp, line 524: Can you add a comment / extract into a variable to make this condition easier to understand pls?

PR #3729, opened 2026-07-21
Fix EditBatch update rejecting edits with array-valued properties

nihalbhatnagar:
[TS Playground
](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgApQPYAdpgJ4AqeOyA3gFDLIDWoAJgFzIBEAzmFKAObPIA+LEAFcAtgCNovAczEYMAGwhwQzANyVkwVgEEoUOHiayFSkKuQB6C8jAALFMxFD5YYFnnAEwfLxjy4XOQAvuTkoJCwiCjo2Lh4AGpw8kIQAOrAUBAEGADCHhDgZBrsnCBcTCXc6lTC4tBMtRJQ1cjGispGcu1mwaH4JInJEADyMAA8ACLIEAAekCB0rGiYOFD4RDgAfMgAvMgTANrMWrr6eMwAutNzBYs2UCkaAPzIpwZjMav4gynpmdl5YAFMAHQ7MWgLS4XTYaJifOI-NIZLK5fLgUFHCF0KHqULhaDwJDIYZiABWEAQYA2EAmEBgoG8wAwICKVBEEDAcCeTFIyCwK1wQNYTAAShSMFA6GNKmUADTLWJrQjECDbILqELkfooACyHLgk2u8zuJPJlOptPpIEZzO2ewAcsz7c5-GJFJMjuzOZdNuptQqsABpCB4ViG2bGpamilUlWWhmuW27GghjAwZB6zmTTZHfmK1wQViXZAAMmQMq4uKsyDo0GAADcIHQWE3vHA3Sg818hb4Jcg4DW6QmmSAtSqA6xRuGbgso2SYxah9bEyA7Wg4GtgEkxhpeQdA5oWZ9g6HsxcmIip5mDRMc8wu4LC5d99DkCFfaFq9pkAhmQhMpAg5WjaLIAO7eLYyCZEk8oeBESTIF6cB0HAnKHgg8gAHT9iyG5nAAtPWSQpM2XaYWE4AElEyAAKIiO4GB4BAEAAOIFNAqFNpaRq3HOZqxjg8bLiOrKIfq3KiVQD6boWPIaFQNRwOyPI0PQTBsBw3BqJoOh6AYTDwPIrAoOq8kKZyXDCmQqkLOpFbaScemGPcKRvuY1ZjPh-ZOYRxFNmZplUKZmrVjkf4QFgYDIAATEwdgOJkv4gCUQiUiOvCACgEPGzsgABCcDGdxYhCFFdArEs8XIHBHHyGJnKyuQ1asBgyAAAbSQWrCtcghlGa0iDUDYLWVVw7GcAgbVir+krSppcr7IW-5uGAEqbK15H4pERIAJKUVtNJ0gAYpgIgADIYAgqF9hGvHEvO5pxkuIFkJq-phSA-4chAu0RISB3pnsnyTuMdEMUxrFjZxdCWh+1YKQAek8jXWFQgMbq4268iASkQBJFbmBZrB43NXAHFcQQwmOJBTcyKWUk2QPJkDU4-VRSCWsdGBnRdV1QLDKNUIjyMKWjm6Y8gBwzBUJPnuWJP8MgjTQArbSmG+2xUJ55aXf4UDIMy8h4J+1g0TMWDKHcK39vItVYBgrCsMAHZoIGrBU9EAoyawh0SgAqlgKGQB8a6oNldw08lHCpZAdCMy8ocNBAjbNKE7t5UIXBcHgTOe51PtQP7gcQGM72fZArP7TD6hw4LSPC1QoHKGAcLo1u8hjFjOPE6UlY2AERPSz3ZPq-XyBcBgUWJ8nCla7lvssUwrWE4P3Bk91WiKxP-YO8AXDY87VuVawOsboetYzKn1YEPYPXAFLNbMgA5FFiW01HlLYb7xk2DfSVlygtZgIrkfksDAoEWS1k4I2Ui5UGrVlArYTwkF+SFmgI2NqjkzhxQeBAdaacxTNXkI2Q6d8mzTkjHdfii4gEjhDuVKct4-TjhITMJsyYCEKGIaQqUoN5CMWYmxEAHEY5V3IEldgyAAD69JWF0GDBFXSWDkAsLYXsTuylBDMHlMvCWzA4BaJYGIYs6pLDWAwNQeUeEDB8kyMZKA0CgA)
Interface object updates for interface types that contained array properties failed to compile. Users saw 

```
Argument of type 'any' is not assignable to parameter of type 'never'.ts(2345)
```
for example code 
```
const interfaceObj = client(InterfaceType).fetchPage().data[0]
batch.update(interfaceObj, {anyProp: "hey"})
```

This is because of how we narrowed the scope of the edit batch to know which properties are valid for the locator. The type we used for this was `UpdatableObjectOrInterfaceLocatorProperties`. Inside, it contained a clause that checked `X extends UpdateObjectForInterface<InterfaceDefinitionForLocator<OL>>`, where X was the entire edit scope `X extends AnyEdit`.  

`InterfaceDefinitionForLocator<OL>>` created an interface definition type 

```
interface InterfaceDefinitionForLocator<
  IL extends ObjectLocator<any>,
> extends InterfaceDefinition {
  apiName: IL["$apiName"];
}
```

which resolved essentially to a very generic

```
{
 apiName: "SpecificApiName"
 DefinitionMetadata: {
      ...
      properties: record<any, ObjectMetadata.Property>
 }
}.
```

Then, when we pass that locator into UpdateObjectForInterface, we get

```
export interface UpdateObjectForInterface<S extends InterfaceDefinition> {
  type: "updateObjectForInterface";
  obj: InterfaceLocator<S>;
  properties: Partial<{
    [P in PropertyKeys<S>]: OsdkObjectCreatePropertyType<
      CompileTimeMetadata<S>["properties"][P]
    >;
  }>;
}
// resolves to for above locator
{
type: "updateObjectForInterface";
  obj: InterfaceLocator<S>;
  properties: Partial<{ [x: string]: string | number | boolean | CipherText | Media | … }>
}
```

Properties resolves to this Partial because it indexes into this type
```
export type OsdkObjectCreatePropertyType<
  STRICTLY_ENFORCE_NULLABLE extends boolean = true,
> = STRICTLY_ENFORCE_NULLABLE extends false
  ? MaybeArray<T, GetCreatePropertyValueFromWire<T["type"]>> | undefined
  : MaybeNullable<T, MaybeArray<T, GetCreatePropertyValueFromWire<T["type"]>>>;
```

and in the generic case, the MaybeArray case resolves to "boolean" instead of true for the array type. This means that it resolves to 
```
Partial<{property: PropertyValueWireToCreate[property]>
```
which is basically
```
Partial<{ [x: string]: string | number | boolean | CipherText | Media | … }>
``` 

almost ALL properties are included in the values of PropertyValueWireToCreate, including number[] (see vector type).

When the value of UpdateObjectForInterface is used in the original UpdatableObjectOrInterfaceLocatorProperties 

```
export type UpdatableObjectOrInterfaceLocatorProperties<
  X extends AnyEdit,
  OL extends ObjectLocator<any>,
> =
  X extends UpdateObject<ObjectTypeDefinitionForLocator<OL>>
    ? X["properties"]
    : X extends UpdateObjectForInterface<InterfaceDefinitionForLocator<OL>>
      ? X["properties"]
      : never;
 ```
the check in  X extends UpdateObjectForInterface<InterfaceDefinitionForLocator<OL>> basically resolves to

```
does {type: "updateObjectForInterface";
  obj: InterfaceLocator<S>;
  properties: Partial<{ fooIdp: string | undefined; fooSpt: string[] | undefined }>
 }
extends {type: "updateObjectForInterface";
  obj: InterfaceLocator<S>;
  properties: Partial<{ [x: string]: string | number | boolean | CipherText | Media | … }>
}
```

and because string[] isn't in the generic union, the type collapses to never.

 **Why objects weren't affected**

  UpdateObject excludes the primary key from its property keys; UpdateObjectForInterface doesn't:
```
  // UpdateObject
  properties: Partial<{
    [P in Exclude<PropertyKeys<S>, CompileTimeMetadata<S>["primaryKeyApiName"]>]: ...
  }>;

  // UpdateObjectForInterface
  properties: Partial<{ [P in PropertyKeys</>]: ... }>;   // no exclusion
```
  On the generic object:
  - PropertyKeys<SynObj> = keyof Record<any, Property> & string = string
  - primaryKeyApiName = keyof this["properties"] = keyof Record<any, Property> = string | number
  - Exclude<string, string | number> = never → Partial<{ [P in never]: … }> = Partial<{}> = {}

  So the object synthetic's properties is {}, and everything extends {} — the props part of the object gate always passes, arrays included, and it returns the concrete
  props.
  
 **Why create still worked**
Create still worked because it captures the entire concrete definition from the first arguement 
```
  create: <OI extends CreatableObjectOrInterfaceTypes<X>>(
    objectOrInterfaceType: OI,           // ← you pass FooInterface (the definition, full metadata)
    properties: CreatableObjectOrInterfaceTypeProperties<X, OI>
  ) => void;
  ```
  whereas our update update takes a locator
```
  update: <OL extends UpdatableObjectOrInterfaceLocators<X>>(
    obj: OL,                              // ← you pass { $apiName, $objectType, $primaryKey } — identity only, NO schema
    properties: UpdatableObjectOrInterfaceLocatorProperties<X, OL>
  ) => void;
  ```
  
The current fix makes what we're trying to narrow explicit -> find the right objects properties instead of trying to match the whole objects edit type.

### DIFF
--- a/packages/client/src/object/convertWireToOsdkObjects.test.ts
+++ b/packages/client/src/object/convertWireToOsdkObjects.test.ts
@@ -716,246 +716,256 @@ describe("convertWireToOsdkObjects", () => {
           employee.$__EXPERIMENTAL__NOT_SUPPORTED_YET__metadata.ObjectMetadata
         ).toMatchInlineSnapshot(
           `
-            {
-              "apiName": "Employee",
-              "description": "A full-time or part-time 
+          {
+            "apiName": "Employee",
+            "description": "A full-time or part-time 
 
-             employee of our firm",
-              "displayName": "Employee",
-              "icon": {
-                "color": "blue",
-                "name": "person",
-                "type": "blueprint",
-              },
-              "implements": [
-                "FooInterface",
-              ],
-              "interfaceImplementations": {
-                "FooInterface": {
-                  "fooIdp": {
-                    "propertyApiName": "office",
-                    "type": "localProperty",
-                  },
-                  "fooSpt": {
-                    "propertyApiName": "fullName",
-                    "type": "localProperty",
-                  },
+           employee of our firm",
+            "displayName": "Employee",
+            "icon": {
+              "color": "blue",
+              "name": "person",
+              "type": "blueprint",
+            },
+            "implements": [
+              "FooInterface",
+            ],
+            "interfaceImplementations": {
+              "FooInterface": {
+                "fooIdp": {
+                  "propertyApiName": "office",
+                  "type": "localProperty",
+                },
+                "fooSpt": {
+                  "propertyApiName": "fullName",
+                  "type": "localProperty",
                 },
               },
-              "interfaceMap": {
-                "FooInterface": {
-                  "fooIdp": "office",
-                  "fooSpt": "fullName",
-                },
+            },
+            "interfaceMap": {
+              "FooInterface": {
+                "fooIdp": "office",
+                "fooSpt": "fullName",
               },
-              "inverseInterfaceMap": {
-                "FooInterface": {
-                  "fullName": "fooSpt",
-                  "office": "fooIdp",
-                },
+            },
+            "inverseInterfaceMap": {
+              "FooInterface": {
+                "fullName": "fooSpt",
+                "office": "fooIdp",
               },
-              "links": {
-                "lead": {
-                  "multiplicity": false,
-                  "targetType": "Employee",
-                },
-                "officeLink": {
-                  "multiplicity": false,
-                  "targetType": "Office",
-                },
-                "peeps": {
-                  "multiplicity": true,
-                  "targetType": "Employee",
-                },
-                "visitedOffices": {
-                  "multiplicity": true,
-                  "targetType": "Office",
-                },
+            },
+            "links": {
+              "lead": {
+                "multiplicity": false,
+                "targetType": "Employee",
               },
-              "pluralDisplayName": "Employees",
-              "primaryKeyApiName": "employeeId",
-              "primaryKeyType": "integer",
-              "properties": {
-                "class": {
-                  "description": "",
-                  "displayName": undefined,
-                  "multiplicity": false,
-                  "nullable": true,
-                  "type": "string",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "employeeId": {
-                  "description": undefined,
-                  "displayName": undefined,
-                  "multiplicity": false,
-                  "nullable": false,
-                  "type": "integer",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "employeeLocation": {
-                  "description": "Geotime series reference of the location of the employee",
-                  "displayName": undefined,
-                  "multiplicity": false,
-                  "nullable": true,
-                  "type": "geotimeSeriesReference",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "employeeProfile": {
-                  "description": "Employee profile with main value being the bio",
-                  "displayName": undefined,
-                  "mainValue": {
-                    "fields": [
-                      "bio",
-                    ],
-                  },
-                  "multiplicity": false,
-                  "nullable": true,
-                  "type": {
-                    "bio": "string",
-                    "yearsExperience": "integer",
-                  },
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "employeeSensor": {
-                  "description": "TimeSeries sensor of the status of the employee",
-                  "displayName": undefined,
-                  "multiplicity": false,
-                  "nullable": true,
-                  "type": "stringTimeseries",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "employeeStatus": {
-                  "description": "TimeSeries of the status of the employee",
-                  "displayName": undefined,
-                  "multiplicity": false,
-                  "nullable": true,
-                  "type": "stringTimeseries",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "favoriteRestaurants": {
-                  "description": undefined,
-                  "displayName": undefined,
-                  "hasReducers": false,
-                  "multiplicity": true,
-                  "nullable": true,
-                  "type": "string",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "fullName": {
-                  "description": undefined,
-                  "displayName": undefined,
-                  "multiplicity": false,
-                  "nullable": true,
-                  "type": "string",
-                  "valueFormatting": undefined,
-                  "valueTypeApiName": undefined,
-                },
-                "office": {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/functions/src/edits/EditBatch.ts [hunk 1]
```diff
@@ -133,12 +127,11 @@ export type UpdatableObjectOrInterfaceLocators<X extends AnyEdit> =
 export type UpdatableObjectOrInterfaceLocatorProperties<
   X extends AnyEdit,
   OL extends ObjectLocator<any>,
-> =
-  X extends UpdateObject<ObjectTypeDefinitionForLocator<OL>>
+> = X extends UpdateObject<any> | UpdateObjectForInterface<any>
```
nihalbhatnagar, line 130: We're basically saying now "Instead of trying to narrow this type down to one of the interfaces or objects by matching the entire edit shape, narrow the entire edit scope to just that of the API name of the object locator"

packages/functions/src/edits/EditBatch.ts
```diff
@@ -41,12 +41,6 @@ interface ObjectTypeDefinitionForLocator<
   apiName: OL["$apiName"];
 }
 
-interface InterfaceDefinitionForLocator<
-  IL extends ObjectLocator<any>,
-> extends InterfaceDefinition {
```
nihalbhatnagar, line 46: InterfaceDefinitionForLocator extends InterfaceDefinition but only overrides apiName, so it basically becomes `{apiName: "myFoo", "definitionMetadata: {props: {type: WirePropTypes}`. Then when picked up in `UpdateObjectForInterface<InterfaceDefinitionForLocator<OL>>` it becomes `Partial<{ [x: string]: string | number | boolean | CipherText | Media | … | number[] /*vector*/ | … }>`

nihalbhatnagar: Because vector (number[]) was in this type, number arrays worked but string arrays didn't satisfy the union

packages/functions/src/edits/EditBatch.ts
bryantpark04: Let's include more of these verbal explanations as comments in the type definitions moving forward.

packages/client/src/object/convertWireToOsdkObjects.test.ts
```diff
@@ -716,246 +716,256 @@ describe("convertWireToOsdkObjects", () => {
...
-                "FooInterface": {
-                  "fooIdp": {
-                    "propertyApiName": "office",
-                    "type": "localProperty",
-                  },
-                  "fooSpt": {
-                    "propertyApiName": "fullName",
-                    "type": "localProperty",
-                  },
+          {
```
bryantpark04, line 719: How did this snapshot change? the main changes were in edits, is this a result of adding the reducers?

bryantpark04 (approved):

packages/client/src/object/convertWireToOsdkObjects.test.ts
nihalbhatnagar: Its cause I added the `fooArray` prop and I don't think the rest is deterministic so it looks like a lot of churn

PR #3733, opened 2026-07-21
Implement new seed builder with validation

esinx:
## Summary

Reworks the `@osdk/seed-helpers` `SeedBuilder` with a new validation layer.

- New `SeedBuilder` implementation with input validation
- Adds `validation.ts` (+ tests), `schema.ts`, `linkTypes.ts`, and a dedicated `SeedError` type
- Reworked `SeedBuilder` test suite


packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,290 @@
...
+import { validateSeedObjects } from "./validation.js";
 
-/** Internal representation of a resolved link before validation. */
-interface ResolvedLink {
-  sourceObjectType: string;
-  sourceKey: string;
-  linkType: string;
-  targetObjectType: string;
-  targetKey: string;
+interface SeedLinkRecord {
```
esinx, line 64: Not in `linkTypes.ts` since this is private

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,290 @@
...
 
-/** Checks whether a value is a {@link SeedRef} returned by `SeedBuilder.add()`. */
-function isSeedRef(value: unknown): value is SeedRef {
-  return (
-    typeof value === "object" &&
-    value != null &&
-    "__objectTypeApiName" in value &&
-    "__primaryKey" in value
-  );
+function linkIdentity(
```
esinx, line 70: A string rep to act as a key representing each link

esinx:
`@osdk/seed-compiler` will be unnecessary after this change. We can keep it, but other use cases should refrain from using it.

packages/seed-helpers/src/linkTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+  CompileTimeMetadata,
+  InterfaceDefinition,
+  ObjectTypeDefinition,
+} from "@osdk/api";
+
+import type { SeedRef } from "./types.js";
+
+/** The names of the links declared on object or interface type `Q`. */
+export type LinkApiNames<Q extends ObjectTypeDefinition | InterfaceDefinition> =
+  keyof CompileTimeMetadata<Q>["links"] & string;
```
nihalbhatnagar: Why does & sting help here

packages/seed-helpers/src/linkTypes.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * The object or interface type on the far side of `Q`'s `A` link.
+ */
+export type LinkTargetType<
```
nihalbhatnagar, line 33: Don't we export some helpers to do these already from @osdk/api? I might be wrong

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,377 @@
...
+    if (!this.#objectMap.has(objectTypeApiName)) {
+      this.#objectMap.set(objectTypeApiName, new Map());
+    }
+    return this.#objectMap.get(objectTypeApiName)!;
   }
 }
 
+export type SeedFunction<T> = (seed: SeedBuilder) => T;
+
+export type SeedClient = {
```
esinx: This is for integration testing; not too sure if this should live here but the types are identical to what SeedBuilder exposes, so I placed it here.

packages/seed-helpers/src/validation.ts
esinx, line 1: Almost exact dupe of the seed compiler's validation logic.

packages/seed-helpers/src/validation.ts
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * Expected runtime JS type for each cataloged wire type.
+ *
+ * Used to fail fast on `as any` callers that pass a value of the wrong shape
+ * (e.g., `age: "30"` when `age` is an integer). Wire types not listed here
+ * are not strictly typed by this validator — currently `attachment`,
+ * `mediaReference`, `geopoint`, `geoshape`, `vector`, `array`, `struct`,
+ * which have non-primitive runtime shapes that need bespoke validation.
+ */
+const EXPECTED_JS_TYPE: Record<string, "string" | "number" | "boolean"> = {
```
nihalbhatnagar, line 29: Can't we use WirePropertyType here or PropertyValueClientToWire?

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,393 @@
...
-}
-
-/**
- * A lightweight, type-safe builder for constructing seed data.
- *
- * TypeScript catches everything expressible at the call site: property
- * names and value types, required primary keys, link names valid on the
- * source type, and link target types matching what the link points to.
- */
 export class SeedBuilder {
```
nihalbhatnagar, line 84: Can we make this API an explicit interface/type and put it in a separate file or at the top level? hard to understadn waht the user will be interacting with here

packages/seed-helpers/src/SeedBuilder.ts
nihalbhatnagar: Ah ok this is the type I mean in my earlier comment

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,393 @@
...
+ * Utility handle for building seeds from a metadata
+ * @param ontologyMetadata Ontology metadata to instantiate seed builder from
+ * @param fn handle to create seeds, may return an arbitrary value
+ * @returns Tuple of [SeedOutput, T]: Seed output and arbitrary value returned from fn
  */
-export function createSeed(fn: (seed: SeedBuilder) => void): SeedOutput {
-  const builder = new SeedBuilder();
-  fn(builder);
-  return builder.build();
+export function createSeed<T>(
```
nihalbhatnagar: How is superrepo ging to pass in the metadata now?

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,393 @@
...
-    value != null &&
-    "__objectTypeApiName" in value &&
-    "__primaryKey" in value
-  );
+function linkIdentity(
+  source: SeedRef<ObjectTypeDefinition>,
+  apiName: string,
+  target: SeedRef<ObjectTypeDefinition>
+): string {
+  return [
```
nihalbhatnagar, line 75: Just curous, why did you choose a concat string instead of nested maps

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,393 @@
...
+   * @param primaryKey Primary key value of the object
+   * @returns Reference to the object, or `undefined` if not found
+   */
+  ref<Q extends ObjectTypeDefinition>(
+    o: Q,
+    primaryKey: PrimaryKeyType<Q>
+  ): SeedRef<Q> | undefined {
+    const schema = this.#schemaMap.objects.get(o.apiName);
+    if (typeof schema === "undefined") {
+      throw new SeedError("Object not found in metadata");
```
nihalbhatnagar: The paradigm we use in the repo is invariant(), can we migrate these custom errors to that?

packages/seed-helpers/src/schema.ts
```diff
@@ -0,0 +1,80 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type * as Ontology from "@osdk/foundry.ontologies";
+
+import { SeedError } from "./SeedError.js";
+
+/**
+ * Minimal schema for one object type — only what the {@link SeedBuilder} needs
```
nihalbhatnagar: What was your thought process around keeping a schema instead of indexing directly into the metadata every time?

packages/seed-helpers/src/validation.ts
nihalbhatnagar: Can you walk me through what you intend to replace in seed compiler here, and also why these checks are better suited for the build step instead of immediate validation

packages/seed-helpers/src/SeedBuilder.test.ts
```diff
@@ -14,346 +14,519 @@
...
+    it("returns the same ref it was given", () => {
+      const sb = newBuilder();
+      const ref = sb.create(Employee, { employeeId: 1 });
+      expect(sb.update(ref, { fullName: "Alice" })).toBe(ref);
     });
 
-    it("stores objects of different types separately", () => {
-      builder.add(Employee, { employeeId: "emp-001" });
-      builder.add(Department, { departmentId: "dept-001" });
+    it("keeps the primary key even if props tries to change it", () => {
```
nihalbhatnagar: Should we throw here?

packages/seed-helpers/src/SeedBuilder.test.ts
```diff
@@ -14,346 +14,519 @@
...
+      expect(sb.build().objects.Employee).toEqual([
+        { employeeId: 1, fullName: "Alicia" },
+      ]);
+    });
 
-      const output = builder.build();
-      expect(Object.keys(output.objects)).toHaveLength(2);
-      expect(output.objects.Employee).toHaveLength(1);
-      expect(output.objects.Department).toHaveLength(1);
+    it("throws a SeedError when the object type is not in the schema", () => {
```
nihalbhatnagar: Yeah I don't think we should introduce new error types, they do't add much value

packages/seed-helpers/src/SeedBuilder.test.ts
```diff
@@ -14,346 +14,519 @@
...
+  describe("delete", () => {
+    it("removes an existing object", () => {
+      const sb = newBuilder();
+      const ref = sb.create(Employee, { employeeId: 1 });
+      sb.delete(ref);
+      expect(sb.build().objects.Employee).toEqual([]);
+    });
 
-      props.firstName = "Mutated";
+    it("is a no-op for a non-existent object and does not throw", () => {
```
nihalbhatnagar: This makes sense for merge but would't we want to flag extraneous operations to users?

packages/seed-helpers/src/SeedBuilder.test.ts
```diff
@@ -14,346 +14,519 @@
...
-      (builder.link as any)(
-        "prod-dept",
-        Product,
-        42,
-        "department",
-        Department,
-        "dept-001",
-        "products"
-      );
+    it("omits links whose target object was not created", () => {
```
nihalbhatnagar: Should we throw here? I'm leaning towards defensively throwing for most things and expanding only where necessary. I think with the example we saw today of users accidentally doing link and modify FK property at the same time that code can be written that doesn't make sense and will never be caught unless we throw. Especially with AI clop

packages/seed-helpers/src/SeedBuilder.ts
esinx: more convenient for accessing them all at once, and I wouldn't ever need to stride through a specific nested map (i.e. I never have to get a list of all links from one object to another)

packages/seed-helpers/src/SeedBuilder.ts
esinx: Yup we can turn this into an interface

packages/seed-helpers/src/SeedBuilder.ts
esinx: 👍

packages/seed-helpers/src/SeedBuilder.ts
esinx: Superrepo also triggers an OSDK generation, so it should naturally emit metadata as a part of the codegen if it uses the correct version of OSDK

packages/seed-helpers/src/validation.ts
esinx: Will look into this, sounds right

packages/seed-helpers/src/validation.ts
esinx: I actually don't think there's much of a benefit here for not validating them immediately. This was a result of porting the compiler.

The seed compiler's job was to:

- Inject the metadata
- Validate it against the built seed output
- Emit result as json

We are essentially replacing all 3 steps.

Since there's the added benefit of directly pointing at the call where the error occurred, I think we could move this validation to the mutation call itself.

packages/seed-helpers/src/SeedBuilder.test.ts
esinx: I do agree with the idea, but maybe we can make the warning more obvious instead since foundry allows links with dangling ends; I actually might need to change the behavior so that we just let these links get generated regardless

packages/seed-helpers/src/SeedBuilder.test.ts
esinx: This gets propagated as a warning

packages/seed-helpers/src/SeedBuilder.test.ts
esinx: 👍

packages/seed-helpers/src/SeedBuilder.test.ts
esinx: Good call. We should throw here.

packages/seed-helpers/src/schema.ts
esinx: This was copied from the approach that the seed compiler took; we can use the direct metadata approach regardless.

packages/seed-helpers/src/validation.ts
esinx: Unfortunately we don't have other places that map runtime types with property types as an object; both suggested types don't have an "instance" version of it so we cannot provide runtime validation

packages/seed-helpers/src/validation.ts
nihalbhatnagar: yeah sounds good

packages/seed-helpers/src/SeedBuilder.ts
nihalbhatnagar: I mean moreso what their user code look like

packages/seed-helpers/src/types.ts
```diff
@@ -14,51 +14,32 @@
...
-  Q extends ObjectTypeDefinition = ObjectTypeDefinition,
-> {
-  readonly __objectTypeApiName: string;
-  readonly __primaryKey: string | number;
-  /** @internal Brand field — never set at runtime. Enforces nominal typing. */
-  readonly [SEED_REF_BRAND]: Q;
-}
+export type SeedRef<Q extends ObjectTypeDefinition> = Readonly<
+  OsdkBase<Q> & {
+    readonly [SEED_REF_BRAND]: Q;
```
nihalbhatnagar, line 40: Do we really need this brand if it satisfies OsdkBase?

packages/seed-helpers/src/SeedBuilder.ts
esinx:
```typescript
import ontologyMetadata from "@ontology/sdk/metadata" with { type: "json" }

export default createSeed(ontologyMetadata, seed => {
   ...
})
```

or

```typescript
import. ontologyMetadata from "@ontology/sdk/metadata" with { type: "json" }

const integrationClient = createIntegrationClient({ ontologyMetadata })
...
```

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,408 @@
...
+    if (!this.#objectMap.has(objectTypeApiName)) {
+      this.#objectMap.set(objectTypeApiName, new Map());
+    }
+    return this.#objectMap.get(objectTypeApiName)!;
   }
 }
 
+export type SeedFunction<T> = (seed: SeedBuilder) => T;
+
+export type SeedClient = {
```
nihalbhatnagar: Can we move this to the top of the class?

packages/seed-helpers/src/types.ts
esinx: Just to prevent arbitrary instantiation of the ref; The invariant is that any `SeedRef<Q>` that is reachable from the user's end are valid objects that exists as a seed entry

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -14,275 +14,408 @@
...
+ * Utility handle for building seeds from a metadata
+ * @param ontologyMetadata Ontology metadata to instantiate seed builder from
+ * @param fn handle to create seeds, may return an arbitrary value
+ * @returns Tuple of [SeedOutput, T]: Seed output and arbitrary value returned from fn
  */
-export function createSeed(fn: (seed: SeedBuilder) => void): SeedOutput {
-  const builder = new SeedBuilder();
-  fn(builder);
-  return builder.build();
+export function createSeedWithMetadata<T>(
```
esinx, line 414: renamed for clarity

nihalbhatnagar (approved):

PR #3735, opened 2026-07-22
Use action parameter displayName for ActionForm labels

nfdiop:
## What

`ActionForm` labeled each parameter with its raw parameter key/id (e.g. `primary_office_id`). Foundry already provides a human-friendly `displayName` on every action parameter, so this surfaces it and uses it for the default field labels.

## Changes

- `@osdk/api` — add `displayName?: string` to `ActionMetadata.Parameter`.
- `@osdk/generator-converters` — `wireActionParameterV2ToSdkParameterDefinition` now copies `displayName` from the wire response.
- `@osdk/react-components` — `getDefaultFieldDefinitions` uses `param.displayName ?? key`, so labels prefer the display name and fall back to the key when it's absent. Caller-supplied `formFieldDefinitions` are unchanged — callers still own their labels there.

## Testing

- `getDefaultFieldDefinitions` unit tests cover the displayName-present and fallback-to-key cases.
- `typecheck`, `check-api`, and `lint` pass for the touched packages.

### Before
Screenshot 2026-07-22 at 15 13 41

### After
Screenshot 2026-07-22 at 15 13 35

### DIFF
--- a/examples-extra/docs_example/src/generatedNoCheck/ontology/actions/completeTodo.ts
+++ b/examples-extra/docs_example/src/generatedNoCheck/ontology/actions/completeTodo.ts
@@ -14,12 +14,14 @@ export namespace completeTodo {
   export type ParamsDefinition = {
     is_complete: {
       description: undefined;
+      displayName: undefined;
       multiplicity: false;
       nullable: false;
       type: 'boolean';
     };
     Todo: {
       description: 'A todo Object';
+      displayName: undefined;
       multiplicity: true;
       nullable: false;
       type: ActionMetadata.DataType.Object<Todo>;
--- a/packages/api/src/ontology/ActionDefinition.ts
+++ b/packages/api/src/ontology/ActionDefinition.ts
@@ -50,6 +50,7 @@ export namespace ActionMetadata {
       | DataType.Interface<any>
       | DataType.Struct<any>;
     description?: string;
+    displayName?: string;
     multiplicity?: boolean;
     nullable?: boolean;
   }
--- a/packages/generator-converters/src/wireActionTypeV2ToSdkActionMetadata.ts
+++ b/packages/generator-converters/src/wireActionTypeV2ToSdkActionMetadata.ts
@@ -20,6 +20,7 @@ import type {
   ActionParameterV2,
   ActionTypeV2,
 } from "@osdk/foundry.ontologies";
+
 import { GeneratorError } from "./GeneratorError.js";
 import { getModifiedEntityTypes } from "./getEditedEntities.js";
 import {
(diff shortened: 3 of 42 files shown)


examples-extra/docs_example/src/generatedNoCheck/ontology/actions/completeTodo.ts
nfdiop, line 1: I needed to regenerate all these files to fix CI

packages/api/src/ontology/ActionDefinition.ts
```diff
@@ -50,6 +50,7 @@ export namespace ActionMetadata {
       | DataType.Interface<any>
       | DataType.Struct<any>;
     description?: string;
+    displayName?: string;
```
nfdiop, line 53: This is one of the main changes

packages/generator-converters/src/wireActionTypeV2ToSdkActionMetadata.ts
```diff
@@ -67,6 +69,7 @@ function wireActionParameterV2ToSdkParameterDefinition(
     ),
     nullable: !value.required,
     description: value.description,
+    displayName: value.displayName,
```
nfdiop, line 72: This is one of the main changes

packages/react-components/src/action-form/utils/__tests__/getDefaultFieldDefinitions.test.ts
```diff
@@ -19,12 +19,14 @@ import { describe, expect, it } from "vitest";
 
 import { getDefaultFieldDefinitions } from "../getDefaultFieldDefinitions.js";
 
-function makeMetadata(parameter: ActionMetadata.Parameter): ActionMetadata {
+function makeMetadata(
```
nfdiop, line 22: I changed this so that it is clearer in the final test that "tags" is the key

bryantpark04:
Was there a reason to do this in codegen instead of with dynamically-loaded Action metadata? Usually our codegen is only for typechecking, not runtime behavior.

nfdiop:
> Was there a reason to do this in codegen instead of with dynamically-loaded Action metadata? Usually our codegen is only for typechecking, not runtime behavior.

Thanks! My understanding was that we already load the data but we were ignoring it (I am already loading the metadata but I wouldn't get back the displayName). So this is the main change I made https://github.com/palantir/osdk-ts/pull/3735/changes#diff-82ff19b8463a8e85505564d18f700857e811e5c1e5f4398dc4572600db2b2847. 

Then CI started failing and I add to codegen (I noticed this also happened here https://github.com/palantir/osdk-ts/pull/2805/changes)

Happy to do the right thing :)

bryantpark04 (approved):
Right I should have read more carefully, sorry

PR #3738, opened 2026-07-23
Fix interface link pivotTo on hydrated object sets

tompersona:


tompersona:
A different fix is coming soon.

PR #3739, opened 2026-07-23
Add useDebouncedCallback

nfdiop:
## Summary

Adds internal utility hooks to `@osdk/react-components` under `src/shared/hooks/`:

- `useDebouncedCallback` — debounces a callback, cleaning up its pending timer on unmount
- `useGatedValue` — holds a value until a gating condition allows it through
- `useDeepEqual` — returns a referentially stable value that only changes on deep inequality
- `useOnUnmount` — runs a callback once when the component unmounts

These are foundational hooks not yet wired into any component or exported from a public entrypoint, so there is no public API change.

## Test plan

- Unit tests added for `useDebouncedCallback` and `useGatedValue`
- `pnpm turbo test --filter=@osdk/react-components`

## Notes
I'm adding this hook because I need a `useDebouncedCallback` that supports `leading: true`. I started off by extending the one from `osdk/react` https://github.com/palantir/osdk-ts/pull/3727. But after chatting with @ziyunp, I decided to add the hook directly the components because:
- I wanted to use lodash-es instead of coming up with my own implementation (and I cannot add that dependency to osdk/react)
- The hook inside packages/react is not that used but it is exported for users. The next step after extracting these hooks into their own package is to deprecate that one (will add the context to the ticket).

### DIFF
--- a/packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
+++ b/packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { renderHook } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+
+import { useDebouncedCallback } from "../useDebouncedCallback.js";
+
+describe("useDebouncedCallback", () => {
+  it("calls the callback provided callback once when called multiple times with `wait` ms", () => {
+    vi.useFakeTimers();
+    const cb = vi.fn();
+    const { result } = renderHook(() =>
+      useDebouncedCallback(cb as (n: number) => void, 10)
+    );
+    const f = result.current;
+    f(1);
+    f(2);
+    f(3);
+    vi.runAllTimers();
+
+    expect(cb).toHaveBeenCalledTimes(1);
+    expect(cb).toHaveBeenCalledWith(3);
+  });
+});
--- a/packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
+++ b/packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
@@ -35,4 +35,29 @@ describe("useDebouncedCallback", () => {
     expect(cb).toHaveBeenCalledTimes(1);
     expect(cb).toHaveBeenCalledWith(3);
   });
+
+  it("cancels a pending invocation when the component unmounts", () => {
+    vi.useFakeTimers();
+    const cb = vi.fn();
+    const { result, unmount } = renderHook(() =>
+      useDebouncedCallback(cb as (n: number) => void, 10)
+    );
+    result.current(1);
+    unmount();
+    vi.runAllTimers();
+
+    expect(cb).not.toHaveBeenCalled();
+  });
+
+  it("returns a stable function reference across re-renders", () => {
+    const { result, rerender } = renderHook(
+      ({ cb }) => useDebouncedCallback(cb, 10),
+      { initialProps: { cb: (_n: number) => {} } }
+    );
+    const first = result.current;
+    rerender({ cb: (_n: number) => {} });
+    rerender({ cb: (_n: number) => {} });
+
+    expect(result.current).toBe(first);
+  });
 });
--- a/packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
+++ b/packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
@@ -15,11 +15,15 @@
  */
 
 import { renderHook } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 import { useDebouncedCallback } from "../useDebouncedCallback.js";
 
 describe("useDebouncedCallback", () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
   it("calls the callback provided callback once when called multiple times with `wait` ms", () => {
     vi.useFakeTimers();
     const cb = vi.fn();
--- a/packages/react-components/src/shared/hooks/lifecycle/useOnUnmount.ts
+++ b/packages/react-components/src/shared/hooks/lifecycle/useOnUnmount.ts
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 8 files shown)


packages/react-components/src/shared/hooks/lifecycle/useOnUnmount.ts
nfdiop, line 1: Used inside useDebouncedCallback; kept because it's a useful hook to have. It's not used yet so also happy to inline inside useDebouncedCallback

packages/react-components/src/shared/hooks/variables/useDeepEqual.ts
nfdiop, line 1: used by useDebouncedCallback

packages/react-components/src/shared/hooks/variables/useGatedValue.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as React from "react";
+
+/**
+ * Lets through @param value when @param isGateOpen returns true, otherwise it returns
+ * the last value for which @param isGateOpen was true (or the very first value if it was
+ * never true).
+ *
+ * @see useInverseGatedValue for the opposite function
+ */
+export function useGatedValue<T>(
```
nfdiop, line 26: Used by useDebouncedCallback and I kept it vs inline because this can replace https://github.com/palantir/osdk-ts/blob/main/packages/react-components/src/filter-list/base/inputs/useStableData.ts

packages/react-components/src/shared/hooks/variables/useGatedValue.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/**
+ * Blocks the @param value when @param isGateClosed returns true, returning the last value for which
+ *
+ * @param isGateClosed was false (or the very first value if it was never false).
+ *
+ * @see useGatedValue for the opposite function
+ */
+export function useInverseGatedValue<T>(
```
ziyunp, line 50: I was actually looking to move this from Forge before, and thinking if we could rewrite this in a more readable / intuitive way. Something to make it easier to understand what value gets returned when the provided "isGateOpen" condition is true. Proposal:

useStableValue(value, isValueStable) -> returns value when isValueStable is true, else return ref.current. (just a proposal)

packages/react-components/src/shared/hooks/lifecycle/useOnUnmount.ts
olliejenningsPT: I would rather keep it separate, nice functional style

packages/react-components/src/shared/hooks/variables/useGatedValue.ts
olliejenningsPT: Same as above, l love the functional style for re-use down the line

packages/react-components/src/shared/hooks/variables/useGatedValue.ts
nfdiop: Good suggestion. Since this is a simple copy/paste, I prefer to not make any functional change to avoid regressions. Happy to track in an issue

packages/react-components/src/shared/hooks/useDebouncedCallback.ts
```diff
@@ -0,0 +1,68 @@
...
+ * The function returned will be stable, suitable for passing to memoized components or as the dependency to useEffect.
+ *
+ * Note that you _do not_ need to memoize the function passed to `useDebouncedCallback` by first using `useCallback` or `useEventCallback`, you can pass an inline function.
+ *
+ * @param {Function} callback The function to debounce.
+ * @param {number} wait The number of milliseconds to delay
+ * @param options Additional configuration options. See https://lodash.com/docs/4.17.15#debounce for more details
+ *
+ * @example
+ * const handleScroll = useDebouncedCallback((event: React.UIEvent<HtmlDivElement>) => {
```
olliejenningsPT: Small typo: `HtmlDivElement` should be `HTMLDivElement`

packages/react-components/src/shared/hooks/useDebouncedCallback.ts
```diff
@@ -0,0 +1,68 @@
...
+export function useDebouncedCallback<T extends (...args: any[]) => unknown>(
+  callback: HasExplicitParams<T>,
+  wait: number,
+  options?: DebounceOptions
+): DebouncedFunc<T> {
+  const f = useEventCallback(callback as T);
+  const deepOptions = useDeepEqual(options);
+  const debouncedCallback: DebouncedFunc<T> = React.useMemo(
+    () => debounce(f, wait, deepOptions),
+    [f, wait, deepOptions]
```
olliejenningsPT, line 64: Any changing of the `wait` or `deepOptions` would re-create the function and not cancel one in flight, it can be left as is (these things will rarely change) but just wanting to flag

packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { renderHook } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+
+import { useDebouncedCallback } from "../useDebouncedCallback.js";
+
+describe("useDebouncedCallback", () => {
```
olliejenningsPT, line 22: Only testing the trailing part, could we potentially add tests for the `cancel` unmount and function reference stability across multiple re-renders

packages/react-components/src/shared/hooks/lifecycle/useOnUnmount.ts
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+import * as React from "react";
+
+/**
+ * Executes the `onUnmount` callback when the component unmounts. This tracks the value of  the `onUnmount` argument
+ * so that it has the correct value when the component unmounts.
+ */
+export function useOnUnmount(onUnmount: () => void): void {
```
olliejenningsPT, line 23: Should we add some tests for this?

packages/react-components/src/shared/hooks/useDebouncedCallback.ts
nfdiop: Thanks for flagging! Let me keep the current logic since I want to make as few changes as possible

packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
nfdiop: Sounds good! I considered it initially because I copied over the tests (and we only had this one enabled) but discarded since I assumed it was working as expected. Let me add tests

packages/react-components/src/shared/hooks/lifecycle/useOnUnmount.ts
nfdiop: happy to add

packages/react-components/src/shared/hooks/__tests__/useDebouncedCallback.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    f(2);
+    f(3);
+    vi.runAllTimers();
+
+    expect(cb).toHaveBeenCalledTimes(1);
+    expect(cb).toHaveBeenCalledWith(3);
+  });
+
+  it("cancels a pending invocation when the component unmounts", () => {
+    vi.useFakeTimers();
```
olliejenningsPT, line 44: We probably want to an afterEach to reset the fakeTimers since they do not get reset normally.

`afterEach(() => vi.useRealTimers()) / vi.restoreAllMocks()`

olliejenningsPT (commented):
One question, what about the current `useDebouncedCallback` hook in the `@osdk/react` package: https://github.com/palantir/osdk-ts/blob/main/packages/react/src/utils/useDebouncedCallback.ts

nfdiop:
> One question, what about the current `useDebouncedCallback` hook in the `@osdk/react` package: https://github.com/palantir/osdk-ts/blob/main/packages/react/src/utils/useDebouncedCallback.ts

Full context: 
I'm adding this hook because I need a `useDebouncedCallback` that supports `leading: true`. I started off by extending the one from `osdk/react` https://github.com/palantir/osdk-ts/pull/3727. But after chatting with @ziyunp, I decided to add the hook directly the components because:
- I wanted to use lodash-es instead of coming up with my own implementation (and I cannot add that dependency to osdk/react)
- The hook inside packages/react is not that used but it is exported for users. The next step after extracting these hooks into their own package is to deprecate that one (will add the context to the ticket).

packages/react-components/src/shared/hooks/variables/useGatedValue.ts
ziyunp: I'm suggesting to rename the parameters like this and not functional change

Screenshot 2026-07-27 at 10 42 06

ziyunp (dismissed):

ziyunp (approved):

PR #3745, opened 2026-07-24
Note ObjectTable headless hooks in react-components AGENTS.md

nfdiop:
Updates the ObjectTable entry in `@osdk/react-components` `AGENTS.md` so agents know the doc also covers the headless hooks for building custom tables.

The hooks were exported and documented in #3669, but `AGENTS.md` only advertised `docs/ObjectTable.md` as covering props, column definitions, examples, theming, and troubleshooting — an agent triaging by keyword had no signal the hooks lived there.

Docs-only change; the hook list stays in `docs/ObjectTable.md` to avoid a second place to keep in sync.


packages/react-components/AGENTS.md
```diff
@@ -86,7 +86,7 @@ Components are imported from their individual entry points under `@osdk/react-co
 Before using any component, read the relevant doc from this package:
 
 - **Setup & installation**: Read [README.md](./README.md) for provider, CSS layers, and peer dependencies
-- **ObjectTable**: Read [docs/ObjectTable.md](./docs/ObjectTable.md) for props, column definitions, examples, theming, and troubleshooting
+- **ObjectTable**: Read [docs/ObjectTable.md](./docs/ObjectTable.md) for props, column definitions, examples, theming, troubleshooting, and hooks to build custom tables
```
ziyunp, line 89: Did we test if this improves the output?

ziyunp (approved):

packages/react-components/AGENTS.md
nfdiop: Not tested because I tried to match the same mention of hooks used in the PdfViewer

PR #3746, opened 2026-07-24
Grouping for ActionForm and BaseForm storybook stories

waynecai1213:
## Summary

Splits the two large monolithic ActionForm story files into focused, per-topic story files and extracts shared setup into helper modules. No behavior or component changes — this is a Storybook organization refactor.

- `ActionForm.stories.tsx` → `ActionForm` (Default) + `ActionForm/Features/{Fields, Title, States & Submission}`
- `BaseForm.stories.tsx` → `BaseForm` (Default) + `BaseForm/{Fields, Validation, In a Dialog, States & Submission, Title & Sections}`
- Extracted `actionFormStoryHelpers.tsx` and `baseFormStoryHelpers.tsx` for shared story scaffolding

The result is a navigable, grouped sidebar instead of one long flat list of ~40 stories dumped under a single `BaseForm` node.

## Sidebar: before / after

| Before | After |
| --- | --- |
|  |  | 

## Notes

- `@osdk/react-components-storybook` is `private: true`, so no changeset is required.
- Pure story-file reorganization; the underlying `ActionForm` / `BaseForm` components are untouched.


packages/react-components-storybook/src/stories/ActionForm/BaseForm.Dialog.stories.tsx
```diff
@@ -0,0 +1,265 @@
...
+  DEPARTMENT_ITEMS,
+  DROPDOWN_ITEMS,
+  field,
+  handleSubmit,
+  TAG_ITEMS,
+} from "./baseFormStoryHelpers.js";
+
+const meta: Meta<BaseFormStoryProps> = {
+  ...baseFormMeta,
+  title: "Components/ActionForm/BaseForm/In a Dialog",
```
ziyunp: Can we move BaseForm into Building Blocks section pls? Similar to ObjectTable

Image

nfdiop (commented):
Thanks for simplifying. The result looks smoother. 
Small notes:
- I wonder if we should avoid too much nesting to avoid extra clicks. For example Controlled Form State should be quickly accessible
- Iknow it is the convention to move BaseX inside building blocks but in the case of ActionForm, I didn't do it because having the extra click was a bit tedious

ziyunp:
> Thanks for simplifying. The result looks smoother. Small notes:
> 
> * I wonder if we should avoid too much nesting to avoid extra clicks. For example Controlled Form State should be quickly accessible
> * Iknow it is the convention to move BaseX inside building blocks but in the case of ActionForm, I didn't do it because having the extra click was a bit tedious

This is the list of stories under ActionForm now. We want to restructure stories similar to how ObjectTable stories were restructured. Can you propose how you would organise them?

Screenshot 2026-07-28 at 16 00 16

PR #3753, opened 2026-07-27
Export metadata as json module

esinx:
## What

Generated SDKs now ship the ontology metadata they were built from, as a typed subpath export.

New `generateOntologyMetadata.ts` writes two files into the generated package:

- `UNSTABLE_DO_NOT_USE/ontology-metadata.json` — `ontologyInfo.requestedMetadata`, i.e. the exact (filtered) `OntologyFullMetadata` the code generator itself consumed, pretty-printed
- `UNSTABLE_DO_NOT_USE/ontology-metadata.d.ts` — an `export =` declaration typing that JSON as `OntologyFullMetadata`

`generatePackageJson` adds the matching export condition:

```json
"./UNSTABLE_DO_NOT_USE/ontology-metadata": {
    "types": "./UNSTABLE_DO_NOT_USE/ontology-metadata.d.ts",
    "default": "./UNSTABLE_DO_NOT_USE/ontology-metadata.json"
}
```

## Why `UNSTABLE_DO_NOT_USE`

The name is a deliberate warning label. The payload is whatever `OntologyFullMetadata` happens to be today, after whatever filtering `ontologyMetadataResolver` applies — not a surface we want consumers pinning behavior to yet. Renaming or reshaping it later should stay cheap.

## Consumption

```ts
// CJS / bundlers / TS with esModuleInterop
import metadata from "@my-generated-sdk/UNSTABLE_DO_NOT_USE/ontology-metadata";
```

Note for reviewers: under native Node ESM, importing a subpath that resolves to `.json` requires an import attribute (`with { type: "json" }`). Generated packages are `type: "commonjs"`, so `require` and bundler paths are unaffected.

## Open question: resolving the type import

The emitted `.d.ts` does `import type { OntologyFullMetadata } from "@osdk/foundry.ontologies"`, but generated packages declare no dependency on that package (`dependencies: []`) and it is not in `betaPeerDependencies` either — only `@osdk/client` is. In a consumer project that does not independently install `@osdk/foundry.ontologies`, the `types` condition will fail to resolve.

An earlier revision of this branch pinned it as a real dependency, resolving the concrete installed version at generation time; that was reverted in ef253f2. Options if we want the `types` condition to stand on its own:

- add `@osdk/foundry.ontologies` as a peerDependency of generated packages, or
- inline a structural type into the emitted `.d.ts` so it has no imports

Happy to do either in this PR — flagging rather than deciding.

## Testing

- `__tests__/generateOntologyMetadata.test.ts` — generates into a tmpdir and covers: the emitted JSON (inline snapshot, pins pretty-printing), the emitted `.d.ts` (inline snapshot, pins `export =`), a verbatim `JSON.parse` round-trip against a fixture with an object type, and the subpath export itself — asserted against literal paths, with `Object.keys` pinning `default` last in the conditions block, and `access()` confirming both conditions point at files that were actually written.
- `__tests__/generatePackageJson.test.ts` — inline snapshot of the whole generated `package.json`, so the full exports map and its condition ordering are reviewable.

## Notes

- The `--ontologyJsonOnly` path is unchanged; it already writes a standalone `ontology.json` and returns before package generation.
- Adds a `minor` changeset for `@osdk/foundry-sdk-generator`.

### DIFF
--- a/packages/foundry-sdk-generator/src/generate/betaClient/generateOntologyMetadata.ts
+++ b/packages/foundry-sdk-generator/src/generate/betaClient/generateOntologyMetadata.ts
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { OntologyFullMetadata } from "@osdk/foundry.ontologies";
+import { mkdir, writeFile } from "fs/promises";
+import { dirname } from "path";
+
+export const ONTOLOGY_METADATA_JSON_PATH =
+  "UNSTABLE_DO_NOT_USE/ontology-metadata.json";
+export const ONTOLOGY_METADATA_DTS_PATH =
+  "UNSTABLE_DO_NOT_USE/ontology-metadata.d.ts";
+
+const typeFileContent =
+  `import type { OntologyFullMetadata } from "@osdk/foundry.ontologies";
+
+declare const ontologyFullMetadata: OntologyFullMetadata;
+
+export = ontologyFullMetadata;
+`;
+
+export async function generateOntologyMetadata(options: {
+  metadata: OntologyFullMetadata;
+  path: string;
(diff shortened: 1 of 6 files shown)


packages/foundry-sdk-generator/src/generate/betaClient/generateOntologyMetadata.ts
```diff
@@ -0,0 +1,43 @@
...
+import { mkdir, writeFile } from "fs/promises";
+import { dirname } from "path";
+
+export const ONTOLOGY_METADATA_JSON_PATH =
+  "UNSTABLE_DO_NOT_USE/ontology-metadata.json";
+export const ONTOLOGY_METADATA_DTS_PATH =
+  "UNSTABLE_DO_NOT_USE/ontology-metadata.d.ts";
+
+const typeFileContent =
+  `import type { OntologyFullMetadata } from "@osdk/foundry.ontologies";
```
esinx, line 27: generated clients have peer dep on `@osdk/client`, so this resolves automatically

nihalbhatnagar (approved):

PR #3757, opened 2026-07-28
Fix ObjectTable last-row border landing on an off-screen virtualized row

ziyunp:
## Problem

The last row of the table body looked visually unterminated. A `tr:last-child` rule already existed to draw the terminating border, but `TableBody` virtualizes rows — so the DOM's last child is normally a trailing overscan row sitting below the fold, not the final data row.

Measured in a browser (500 rows, 400px viewport, `overscan: 5`):

| scroll state | rendered | row that got the border |
| --- | --- | --- |
| at top | Row 0–14 | Row 14 (below the fold) |
| mid-scroll | Row 95–114 | Row 114 (below the fold) |
| scrolled to bottom | Row 486–499 | Row 499 (correct, by luck) |
| full-table loading skeleton | 9 rows | **all 9** |

So the border only landed correctly when scrolled fully to the bottom, and the full-table skeleton drew one under every row.

## Fix

Mark the visually last row explicitly with `data-last-row`, derived from the **data index** rather than DOM position, and key the CSS off that attribute. This mirrors the precedent already established for zebra striping one line above in the same file:

```css
/* Zebra rows — keyed off data index, not DOM position, because rows are virtualized */
```

While fetching more, the marker goes to the `LoadingRow` instead of the last data row — marking both would butt the data row's 1px `border-bottom` against the skeleton's 1px `border-top` and render a doubled 2px line.

Reuses the existing `--osdk-table-border` token. No hardcoded colors, no new CSS variables, no public API change (`check-api` reports no delta; `TableRow`/`TableBody`/`LoadingRow` are internal).

## Verification

After the fix, exactly one row carries the border in every state — 0 at top and mid-scroll (nothing to terminate there; separators come from each row's `border-top`), 1 on Row 499 at the bottom, 1 on the skeleton, 0 in the empty state (header keeps its own border).

## Regression test

`__tests__/TableBody.test.tsx` (new — no `TableBody` test existed) pins all four cases:

- the final data row is marked
- the last *rendered* row is not marked when it is only an overscan row
- the loading row takes the border while fetching more
- the full-table loading skeleton gets exactly one border, not one per row

The suite stubs `offsetHeight`, since happy-dom reports every element as 0x0 and the virtualizer would otherwise render nothing.

### DIFF
--- a/.changeset/objecttable-last-row-border.md
+++ b/.changeset/objecttable-last-row-border.md
@@ -1,0 +1,5 @@
+---
+"@osdk/react-components": patch
+---
+
+Fix `ObjectTable` bottom border landing on an off-screen virtualized row instead of the last row
--- a/packages/react-components/src/object-table/LoadingRow.tsx
+++ b/packages/react-components/src/object-table/LoadingRow.tsx
@@ -28,6 +28,12 @@ interface LoadingRowProps<TData extends RowData> {
   translateY: number;
   rowHeight?: number;
   columnWidth?: number;
+  /**
+   * Whether this row sits at the visual bottom of the table body and should
+   * carry the terminating border. `LoadingStateTable` renders a whole column
+   * of skeleton rows, so only its final row sets this.
+   */
+  isLastRow?: boolean;
 }
 
 export function LoadingRow<TData extends RowData>({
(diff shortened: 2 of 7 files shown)


ziyunp:
not an issue

PR #3765, opened 2026-07-29
ObjectTable sort icon on the column's property type

ziyunp:
Render sort icon based on property type.

Added `getSortIcons(dataType)` to map the column's property type to a Blueprint icon pair:

| property type | ascending | descending |
| --- | --- | --- |
| `string`, `cipherText`, `marking` | `sort-alphabetical` | `sort-alphabetical-desc` |
| `byte` `decimal` `double` `float` `integer` `long` `short` | `sort-numerical` | `sort-numerical-desc` |
| `datetime`, `timestamp` | `sort-asc` | `sort-desc` |
| `boolean`, anything else, unknown | `sort-asc` | `sort-desc` |

`getDefaultColumns` also never populated `meta.dataType`, so tables rendered without explicit `columnDefinitions` had no type info to key off. It now surfaces it the same way the `columnDefinitions` path does.

Screenshot 2026-07-29 at 15 05 25
Screenshot 2026-07-29 at 15 05 17
Screenshot 2026-07-29 at 15 05 02


packages/react-components-storybook/src/stories/ObjectTable/objectTableStoryHelpers.tsx
```diff
@@ -450,11 +450,31 @@ export async function clickHeaderMenuItem(label: string): Promise<void> {
...
+];
+
+function hasSortIcon(th: HTMLElement, iconNames: readonly string[]): boolean {
+  const selector = iconNames.map((name) => `[data-icon="${name}"]`).join(", ");
+  return th.querySelector(selector) != null;
+}
+
 function getSortDirection(th: HTMLElement): "asc" | "desc" | "none" {
-  if (th.querySelector('[data-icon="sort-alphabetical-desc"]') != null) {
+  if (hasSortIcon(th, DESCENDING_SORT_ICONS)) {
```
olliejenningsPT, line 474: Love this, one thing, is we have `MultiColumnSortDialog` and right now that still only has alphabetical sorting. So whilst the header for numeric columns will have this nice icon display, in that multi-sort they will still show the alphabet ones.

Ref: https://github.com/palantir/osdk-ts/blob/main/packages/react-components/src/object-table/MultiColumnSortDialog.tsx#L170

Do we want to also change that in a FLUP?

ziyunp: Good call. I will change that in a FLUP

packages/react-components/src/object-table/hooks/useColumnDefs.ts
olliejenningsPT, line 118: One tiny optional thing, since this is very similar to: https://github.com/palantir/osdk-ts/pull/3765/changes#diff-c8f35a1a53868ff27a297b7af50bcf8a852a56f6749302430db7420696d8f7ebR196.

Would it be worth having them in a tiny help function for re-use?

olliejenningsPT (approved):

PR #3769, opened 2026-07-29
type media transformation language params as closed enums

lucy-k-wang:
## BLUF

Type the OCR and transcription language parameters as closed enums so invalid codes fail at build time instead of server-side.

## Changes

- Narrow four language and script parameters to the enums declared by the platform.
- Mirror the enums in `@osdk/api` and check them against `@osdk/foundry.mediasets` during client typecheck, since `@osdk/api` does not depend on platform SDK packages.
- Export the archive, document, imagery and video encoding types already referenced by `MediaTransformation`.


packages/api/src/experimental/MediaTransformationLanguages.ts
```diff
@@ -0,0 +1,426 @@
...
+  | "ML"
+  | "MN"
+  | "MR"
+  | "MS"
+  | "MT"
+  | "MY"
+  | "NE"
+  | "NL"
+  | "NN"
+  | "false"
```
lucy-k-wang: api-gateway pr to fix: https://github.palantir.build/foundry/api-gateway/pull/16936

nihalbhatnagar (approved):

PR #3770, opened 2026-07-29
Setup @osdk/integration-testing package

esinx:
## What

Adds `@osdk/integration-testing` as an empty but **publishable** package. This is deliberately scope-limited: it exists to get the npm publishing step landed ahead of the integration-testing implementation itself, so that the follow-up PR is pure feature code with no release-plumbing mixed in.

The entry point is a placeholder:

```ts
export const packageName = "@osdk/integration-testing";
```

Extracted from `esinx/integration-testing`, which additionally carries the real implementation (Foundry CLI launcher, service discovery, status/ontology servers, and the conjure-generated `cli-api` client). Those follow separately.

## Package registration

A new oxc-toolchain library has to be listed in three places, all of which are enforced by CI:

| File | Change |
| --- | --- |
| `.monorepolint.config.mjs` | Added to the `oxc migrated libraries with check-api` archetype |
| `.lintstagedrc.mjs` | Added to the nested-oxlint-config map |
| `dprint.json` | Added to `excludes` — the package is oxfmt-formatted, not dprint-formatted |

Unlike the other members of that archetype, the nested `oxlint.config.ts` carries **no rule carve-outs**. Those packages predate the oxc migration and needed error-level Ultracite rules turned off to keep the migration a reformat rather than a rewrite. This package is new code authored against oxc, so there is nothing to grandfather in — the file exists only because the archetype points `lint` at a per-package config. The archetype comment was updated to say so.

## Notes for reviewers

- **Version `0.0.0`** with a `minor` changeset, so the first release cuts `0.1.0`. Matches how `@osdk/vite-plugin-code-workspace-preview` was introduced.
- **Not in the changesets `ignore` list** and not in a `fixed` group, so it publishes and versions independently.
- **No runtime dependencies.** The implementation branch's `dependencies` (`conjure-lite`, `consola`, `undici`, `ts-pattern`, …) and `peerDependencies` were intentionally dropped — shipping unused deps in a published package is real surface area, and they come back alongside the code that uses them.
- **The `postinstall` script was also dropped.** On the implementation branch it runs `node ./build/esm/scripts/postinstall.js` to fetch the Foundry CLI; with no such file present it would break `pnpm install` for every consumer of the published package.
- `src/index.test.ts` covers the placeholder export. It is not busywork: monorepolint pins the test script to `vitest run --pool=forks`, which exits `1` on "no test files found", so without a test the `test` task fails CI.
- Lockfile churn beyond the new importer block (`magic-string`, `expect-type`) is pnpm re-resolving a new `vitest` peer variant. Confirmed to be caused by this package rather than ambient drift, and `pnpm install --lockfile-only` was rejected as an alternative because it re-resolves far more (327 changed lines vs 93).

## Verification

All green:

- `pnpm turbo lint typecheck check-spelling check-attw check-api test --filter=@osdk/integration-testing --force` — 10/10 tasks
- `pnpm check-mrl` — exit 0
- `pnpm turbo transpile` (repo-wide, per the pre-push guidance) — 258/258 tasks
- `dprint check` on changed files — clean

`check-attw` passing is the one that matters most here, since it validates the published type-resolution surface across cjs/esm/browser conditions.

The one pre-existing `dprint check` failure in `benchmarks/tests/primary/src/helpers.js` (a `%DebugPrint` V8 intrinsic) is untouched by this PR.


esinx:
Merging #3787 instead

PR #3773, opened 2026-07-30
Set oxfmt `trailingComma` to "all"

nfdiop:
Override Ultracite's `trailingComma: "es5"` with `"all"` in `oxfmt.config.ts`, and reformat the oxfmt-migrated packages.

`"es5"` is the legacy Prettier default — both Prettier (v3+) and oxfmt now default to `"all"`. Our dprint-formatted packages already emit trailing commas on multi-line structures (dprint's `onlyMultiLine`), so this brings the oxfmt packages in line and keeps the whole repo consistent.

Trailing commas everywhere make lists nicer to work with: adding, removing, or reordering an item is a one-line diff instead of also touching the previous line's comma.

Formatting-only, no runtime impact — hence an empty changeset. Only oxfmt-owned packages changed; dprint packages are untouched.


nfdiop:
Closing in favour of https://github.com/palantir/osdk-ts/pull/3777/changes

PR #3775, opened 2026-07-30
Consolidate ObjectTable data fetching to use useObjectSet only

ziyunp:
ObjectTable was using two hooks to do data fetching. This PR consolidates the data fetching of ObjectTable into a single path, i.e. useObjectSet.

The function column data fetching returns [] previously when useObjectSet did not support interface type. Removed it now that useObjectSet added support for interface.

### DIFF
--- a/packages/react-components/src/object-table/hooks/useObjectTableData.ts
+++ b/packages/react-components/src/object-table/hooks/useObjectTableData.ts
@@ -25,6 +25,7 @@ import type {
 } from "@osdk/api";
 import {
   useObjectSet,
+  useOsdkClient,
   type UseOsdkListResult,
   useOsdkObjects,
 } from "@osdk/react";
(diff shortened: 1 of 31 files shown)


packages/react-components/src/object-table/hooks/useObjectTableData.ts
```diff
@@ -159,24 +161,34 @@ export function useObjectTableData<
...
-      orderBy,
-      pageSize,
-      enabled: shouldUseObjectSet,
-      dedupeIntervalMs,
-      streamUpdates,
-    }
-  );
+  /**
+   * Consolidating data fetching to use the useObjectSet only
+   * TODO: Remove shouldUseObjectSet and useOsdkObjects
```
ziyunp: This will be removed in a flup

nfdiop: Are there options only supported by useOsdkObjects?

ziyunp: You can see from the usage of the hooks that the options of useOsdkObjects is a subset of useObjectSet
 ```
 const objectSetResult = useObjectSet(os, {
    ...(objectSetOptions as ObjectSetOptions<Q>),
    withProperties: withProperties as WithProperties<Q, RDPs>,
    where: filter,
    orderBy,
    pageSize,
    enabled: shouldUseObjectSet,
    dedupeIntervalMs,
    streamUpdates,
  });

  const osdkObjectsResult = useOsdkObjects<Q, RDPs>(objectOrInterfaceType, {
    withProperties,
    pageSize,
    where: filter,
    orderBy,
    enabled: !shouldUseObjectSet,
    dedupeIntervalMs,
    streamUpdates,
  });
```

ziyunp:
Note: Should only merge if https://github.com/palantir/osdk-ts/pull/3781 gets in

PR #3779, opened 2026-07-30
metadata export ESM/CJS resolution + move to @osdk/generator

esinx:
## What

Two related changes to the `UNSTABLE_DO_NOT_USE/ontology-metadata` subpath export:

1. **Type it per resolution mode instead of per generated package type.** The single `.d.ts` shim used `export =`, which is only correct for `require()`. Under `import`, Node and bundlers hand back the JSON object as the *default* export, so consumers in ESM got the wrong type. The subpath now declares both conditions, sharing the one (format-neutral) `.json` file:

   ```jsonc
   "./UNSTABLE_DO_NOT_USE/ontology-metadata": {
     "import":  { "types": "…d.mts", "default": "…json" },
     "require": { "types": "…d.cts", "default": "…json" }
   }
   ```

2. **Move metadata JSON generation into `@osdk/generator`** so every generation path produces it, not just `foundry-sdk-generator`. `generateOntologyMetadataJSONFile` lives next to the existing type-file generator in `v2.0/generateMetadata.ts`, and the four path constants are exported from the package root for consumers that build a `package.json`.

## Details

- `MinimalFs` gains an optional `writeAsset`. Hosts that route `writeFile` through a compiler (foundry-sdk-generator's in-memory TS host) must write assets separately — a `.json` or a hand-written `.d.ts` has no compiler output and would otherwise be silently dropped. Hosts that write straight to disk can omit it and fall back to `writeFile`.
- `foundry-sdk-generator`: deleted `generateOntologyMetadata.ts` (now generator-owned), extracted the host fs into `createHostFs.ts` and `customNormalize` into its own file.
- `transpile2.mjs`: hand-written declaration files are copied rather than compiled (isolatedDeclarations would emit `foo.d.d.ts`), and `.json` is added to the copy list so the asset lands in `build/`.
- `.monorepolint.config.mjs`: added an `extraExports` escape hatch to `ourExportsConvention` for subpaths the convention can't derive from `src/public`, and used it to pin the metadata export on `@osdk/e2e.generated.catchall`.
- `@osdk/foundry.ontologies` is now a dependency of generated packages — the shims reference `OntologyFullMetadata`, and without it those types silently widen to `any` under the usual `skipLibCheck`.
- attw is run in two passes: everything except the metadata entrypoint fully checked, then the metadata entrypoint alone with `false-export-default` ignored. That rule fires on a `.d.mts` whose default export isn't a real ESM default, which is exactly the shape a JSON module has.

## Behavior change worth a look

`generatePackage` previously passed `"module"` as the package type to `generateClientSdkVersionTwoPointZero`; it now passes `undefined` (commonjs default). This is intentional given the per-condition typing above, but it's the one change here that isn't purely additive.

## Test plan

- `packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.test.ts` covers the new JSON + shim output.
- Deleted `generateOntologyMetadata.test.ts` — the code it tested no longer exists in that package.
- Regenerated fixtures under `packages/e2e.generated.*` and `packages/e2e.test.foundry-sdk-generator`.


packages/foundry-sdk-generator/src/generate/betaClient/__tests__/generatePackageJson.test.ts
```diff
@@ -36,10 +36,7 @@ describe("generatePackageJson", () => {
       packageName: "@my/generated-sdk",
       packageVersion: "1.2.3",
       packagePath,
-      dependencies: [{
-        dependencyName: "@osdk/foundry.ontologies",
-        dependencyVersion: "^2.63.0",
-      }],
+      dependencies: [],
```
esinx, line 39: Seems like the deps here were added in the previous PR but it shouldn't be there, `@osdk/foundry.ontologies` is a transitive dep

packages/foundry-sdk-generator/src/generate/betaClient/createHostFs.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import type { MinimalFs } from "@osdk/generator";
+import { mkdir, readdir, writeFile } from "fs/promises";
+import { dirname } from "path";
+import { customNormalize } from "./customNormalize.js";
+
+/**
+ * The filesystem handed to the generator for in-memory compilation
+ */
+export function createHostFs(
```
esinx, line 25: Move the host FS generation to a separate source for better modularity; doesn't need to be here if we want to keep it in generatePackage.ts

packages/foundry-sdk-generator/src/generate/betaClient/customNormalize.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { normalize } from "node:path/posix";
+
+export function customNormalize(pathName: string): string {
```
esinx, line 19: Exported for reuse in createHostFs

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
```diff
@@ -75,35 +71,21 @@ export async function generatePackage(
...
-  };
+  const hostFs = createHostFs(inMemoryFileSystem);
 
   await generateClientSdkVersionTwoPointZero(
     ontologyInfo.requestedMetadata,
     `typescript-sdk/${options.packageVersion} ${USER_AGENT}`,
     hostFs,
     packagePath,
-    "module",
+    undefined, // commonjs by default
```
esinx: Used to be "module", but this parameter was never read initially; setting it to `undefined` so that it gets defaulted to "commonjs", which is the desired behavior (and the default type that osdk generates)

packages/generator/src/v2.0/generateMetadata.ts
```diff
@@ -52,3 +53,52 @@ export async function generateOntologyMetadataFile(
...
+export async function generateOntologyMetadataJSONFile({
+  fs,
+  outDir,
+  ontology,
+}: GenerateContext): Promise<void> {
+  const writeAsset = fs.writeAsset ?? fs.writeFile;
+  await fs.mkdir(path.dirname(path.join(outDir, ONTOLOGY_METADATA_JSON_PATH)), {
+    recursive: true,
+  });
+  await writeAsset(
```
esinx, line 90: `writeAsset()` will bypass ts compiler and directly write to disk when using an in-memory FS for ts compilation

packages/generator/src/MinimalFs.ts
```diff
@@ -17,6 +17,18 @@
...
+  /**
+   * Writes a non-TypeScript asset, such as a JSON file or a hand written
+   * `.d.ts` shim.
+   *
+   * Hosts that route {@link writeFile} through a compiler must implement this
+   * separately, since assets have no compiler output and would otherwise be
+   * silently dropped. Hosts that write directly to disk can leave it undefined,
+   * in which case {@link writeFile} is used.
+   */
+  writeAsset?: WriteFileFn;
```
esinx, line 30: Added docs here

packages/tool.generate-with-mock-ontology/src/generateWithMockOntology.ts
```diff
@@ -89,13 +91,14 @@ export async function generateWithMockOntology(): Promise<void> {
...
-      "osdk",
-    )} --ignore-rules internal-resolution-error`;
+    const packageDir = path.join(testApp2Dir, "osdk");
+    const attwOpts = { stdout: "inherit", stderr: "inherit" } as const;
+    await $(
+      attwOpts,
+    )`attw --pack ${packageDir} --ignore-rules internal-resolution-error --exclude-entrypoints ${ONTOLOGY_METADATA_ENTRYPOINT}`;
+    await $(
+      attwOpts,
+    )`attw --pack ${packageDir} --ignore-rules internal-resolution-error false-export-default --entrypoints ${ONTOLOGY_METADATA_ENTRYPOINT}`;
```
esinx, line 102: Bypass check for json imports;

- Under ESM, Node synthesizes the default export of a JSON module as the parsed object, so `import om from "@test-app2/osdk/UNSTABLE_DO_NOT_USE/ontology-metadata"` genuinely works.
- attw's own node16 (from ESM) test agrees. Bundlers behave the same. The rule just doesn't respect the "dual export" of JSON modules having ESM/CJS variants.

The attw error is:
> ❗️ The resolved types use export default where the JavaScript file appears to use module.exports =. … These types should use export = instead of export default.

Which doesn't work for the `.mts` file.

esinx:
FYI the +8,152 is from `ontology-metadata.json` generated

packages/foundry-sdk-generator/src/generate/betaClient/generatePackage.ts
olliejenningsPT: How come we need commonjs, just curious since Node + JS world is slowly moving to a `module` only future

esinx: @olliejenningsPT 
Mostly because the generated OSDK has both cjs + esm exports, and the package.json also declares itself as a commonjs module by default:

https://github.com/palantir/osdk-ts/blob/main/packages/foundry-sdk-generator/src/generate/betaClient/generatePackageJson.ts#L78

esinx: Going back, this was an attempt to correct the module resolution with a single .ts file-- didn't work and this variable is not being used anywhere. Reverting it, but still would worth flagging that this parameter is not being used.

packages/cli.cmd.typescript/src/generate/TypescriptGenerateArgs.ts
```diff
@@ -32,4 +32,5 @@ export interface TypescriptGenerateArgs {
   externalInterfaces: Map<string, string>;
   branch?: string;
   skipPackageJsonUpdate?: boolean;
+  generateOntologyMetadata?: boolean;
```
nihalbhatnagar: Talked offline but we're making this a "turn off flag"

nihalbhatnagar (commented):
seems fine to me but yeah would like to make this default with the ability to turn off

nihalbhatnagar (approved):

PR #3783, opened 2026-07-31
ci: restore Node-24 build cache on test-matrix legs so codegen is a hit

olliejenningsPT:
`typescript-sdk-docs-examples#codegen` runs on every test-matrix leg (Node 18-24): a generator depends on it, so it can't be filtered out per leg. Its script ends in `pnpm run format`.

The per-node Turbo cache step stops at the first matching `restore-key`, so a stale `test-node<N>-*` cache gets picked ahead of this SHA's build cache. When codegen's inputs change, its hash misses on the leg and it re-executes (re-running `format`) instead of being restored from the Node-24 build job.

Fix: add a restore-only overlay of `turbo-<os>-build-<sha>` after the warming cache. Turbo cache entries are hash-named, so this SHA's build artifacts (codegen included, and node-independent) are present regardless of which stale cache matched, making codegen a guaranteed hit. Per-node test warming still coexists, and the build cache holds no `test` entries so it can't mask a per-node test result.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,6 +301,12 @@ jobs:
             turbo-${{ runner.os }}-build-${{ github.sha }}
             turbo-${{ runner.os }}-build-
 
+      - name: Restore Node-24 build artifacts (make codegen a cache hit)
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: .turbo/cache
+          key: turbo-${{ runner.os }}-build-${{ github.sha }}
+
       - name: Install dependencies
         run: pnpm install
 


.github/workflows/ci.yml
```diff
@@ -301,6 +301,12 @@ jobs:
             turbo-${{ runner.os }}-build-${{ github.sha }}
             turbo-${{ runner.os }}-build-
 
+      - name: Restore Node-24 build artifacts (make codegen a cache hit)
```
olliejenningsPT, line 304: I have the comment (make codegen cache hit) to really signify why this is required, but am happy to remove it and just leave it as restore Node-24.

The reason being is because we already have `Turbo Cache` section, and so its to signify why this one is required

nfdiop: question: sorry if I missed this but the PR says "restore". Was this the behaviour before? If yes, should be safe to bring back

olliejenningsPT: Ah sorry, so what l mean by restore here is basically `restoring the cache` (because this stage doesn't have the right cache restored, basically we only restore the matrix cache)

nfdiop: I see. So should we use actions/cache/restore vs actions/cache?

olliejenningsPT: Its not an approved action, basically only actions/cache is approved and anything actions/cache/* is not

nfdiop: Thanks! I'm not so familiar with this by my understanding is that if there is a cache miss here (for example due to eviction or failed upload from build job), we are now gonna write results from Node 18 into this cache. If we later test with Node 24, we could read from that cache.

Curious why not remove https://github.com/palantir/osdk-ts/pull/3783/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL300?

olliejenningsPT: So we can remove L300 and it would give almost the same effect as the cache overlay i am adding here. The part that L300 gives us is shared caching across previous runs, so unchanged test tasks stay cached across commits.

nfdiop: Thanks. That was my understanding too, but I've been bitten by this kind of caching before, where tests pass because they didn't actually run and then fail on the main branch. 
Happy to approve and will defer to you for the tradeoffs

nfdiop (approved):

PR #3784, opened 2026-07-31
Add @osdk/agents package with AgentConfig

mhgbrg:
Library for publishing utils for agents. Very similar to `@osdk/functions`. Currently only exports `AgentConfig` (the equivalent to `FunctionConfig` but for agents), but we expect to add more utils over time.

Generated with Claude Code. I asked it to copy the `@osdk/functions` setup. I reviewed the files and everything looks reasonable, but please let me know if anything looks off.


packages/agents/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "fix-lint": "oxlint -c ./oxlint.config.ts --fix . && oxfmt -c ../../oxfmt.config.ts .",
+    "lint": "oxlint -c ./oxlint.config.ts . && oxfmt -c ../../oxfmt.config.ts --check .",
+    "transpileBrowser": "monorepo.tool.transpile -f esm -m normal -t browser",
+    "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
+    "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
+    "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
+    "typecheck": "tsc --noEmit --emitDeclarationOnly false"
+  },
+  "peerDependencies": {
+    "@osdk/functions": "workspace:^"
```
ziyunp: Do you want to have dependency on @osdk/function? This means that whenever a user wants to install this package, they also need to install functions package

mhgbrg: The `AgentConfig` type extends the `FunctionConfig` type. I can make it a true dependency instead of a peer dependency if that makes more sense?

ziyunp: Is FunctionConfig a real dependency? Do you want your AgentConfig shape to change whenever FunctionConfig changes?

If not, maybe you can re-define the FunctionConfig type within this repo, so it is not coupled to @osdk/functions.

mhgbrg: Yes, at least for now. We generate functions for agents and need to forward `FunctionConfig` to the generated function.

ziyunp (dismissed):

packages/agents/src/scope/types.ts
mhgbrg, line 1: Copied from functions instead of pulling in with a dep so that we have flexibility to diverge from functions in the future if we want to.

ziyunp (dismissed):

.monorepolint.config.mjs
```diff
@@ -1413,7 +1388,9 @@ function standardPackageRules(shared, options) {
     }),
     fileContents({
       ...shared,
-      excludePackages: ["@osdk/maker"],
+      // @osdk/agents ships a single type and has no tests, so it has no test
```
nfdiop: curious why not let it have the vitest config. Ideally we want all packages to follow the same structure and want to avoid edge cases

packages/agents/oxlint.config.ts
```diff
@@ -14,21 +14,19 @@
...
+// `extends` only carries `rules`/`plugins`/`overrides`, so the root's
+// `ignorePatterns` are re-applied explicitly (otherwise generated/ignored files
+// would start being linted).
+export default defineConfig({
+  extends: [root],
+  ignorePatterns: root.ignorePatterns,
 
-  return resolvedAliases.streams[alias];
-}
+  rules: {},
```
nfdiop, line 31: Noticed this is different from other packages: we will likely need to be consistent. No actions to take

.monorepolint.config.mjs
mhgbrg: Figured it was unnecessary since there were no tests. I don't have strong opinions here. If I need to re-add the config to get this PR merged I'm fine with that.

mhgbrg: Done.

ssanjay1 (approved):

mhgbrg:
Closing in favor of https://github.com/palantir/osdk-ts/pull/3811.

PR #3786, opened 2026-07-31
Add struct action parameter support to Maker

colin4554:
Struct params previously weren't supported in actions


packages/client.unstable/src/generated/ontology-metadata/api/__components.ts
```diff
@@ -8835,6 +8841,10 @@ export interface OntologyIrConditionalOverride {
 export interface OntologyIrConditionalValidationBlock {
   conditionalOverrides: Array<OntologyIrConditionalOverride>;
   defaultValidation: OntologyIrParameterValidationBlock;
+  structFieldValidations?: Record<
```
ethanai22, line 8844: we need an oms change in order for this field to get picked up

davidpaquin (approved):

PR #3787, opened 2026-07-31
Local ontology integration testing

esinx:
# Local ontology integration testing

Adds `@osdk/integration-testing`, a package for running OSDK tests against a **real local ontology** rather than a mocked client. It launches the Foundry CLI's local ontology service as a subprocess, waits for it to become healthy, seeds it, and hands back an OSDK `Client` pointed at it.

## Usage

```ts
const server = await createIntegrationServer({
  metadata: om,                                       // ontology metadata from the generated SDK
});
await server.start();
const { client, seed } = await server.createClient();

await seed.set(baseSeed);                             // replace ontology contents
await client(Book).fetchOne(bookRef.id);              // real request against the local ontology
```

See `packages/integration-testing/examples/example.ts` for the full vitest fixture pattern (worker-scoped server, per-test seed reset).

## Included

**Public API** — `createIntegrationServer` / `createIntegrationClient`, plus the `IntegrationServer` and `IntegrationClient` interfaces. `IntegrationClient` bundles:

- `client` — a real `@osdk/client` `Client` for objects, interfaces, and actions. The local ontology can't execute functions, so queries route to `@osdk/unit-testing`'s mock client; the escape hatch is exposed as `client.whenQuery(...)`.
- `seed` — a `SeedClient` from `@osdk/seed-helpers` that pushes each mutation to the live ontology's seeding endpoint.

**Service orchestration** (internal):

- `FoundryCliService` — one `foundry` subprocess: spawn, stderr capture, health, `SIGINT`-then-`SIGKILL` stop.
- `CliServiceLauncher` — dependency-ordered start/stop with cycle detection, memoized starts so a shared dependency starts once, and readiness polling with a deadline. On any startup failure it stops everything already up and rethrows.
- `ServiceDiscoverer` — reads the discovery files services publish into `.palantir/`, and treats an entry whose PID is dead as absent.
- `StatusServer` / `OntologyServer` — the two concrete services; the status server is the health authority for the others and coalesces concurrent status fetches.

Each server gets its own `.test-run-*` directory under `projectPath`, removed on `stop()`, so parallel vitest workers don't collide.

**Foundry CLI bootstrap** — a `postinstall` that checks for `foundry` on `PATH`, enforces a minimum version (`0.200.0`), and otherwise fetches and runs the installer the stack serves.

**Generated code**: for the CLI's status and seeding APIs, wired into `scripts/generateConjure.sh` (`com.palantir.foundry-cli:cli-api`).

## Seed Helper Changes

- **`@osdk/seed-helpers`**: adds `SeedClient.set(seed?)`, which *replaces* seeded state instead of appending. This is what makes a fixture reusable across tests — the same primary keys can be re-applied rather than having to be unique per test.
- **`createSeedWithMetadata`** now returns `{ output, context }` instead of the positional tuple `[SeedOutput, T]`. **Breaking** for existing callers; destructuring by name reads far better at the call site given `context` is usually a bag of refs.
- Repo config: dprint/lintstaged/monorepolint entries for the new package. Its nested oxlint config re-includes the generated `src/generated/cli` tree the way `@osdk/client.unstable` does.

## Testing

Unit tests cover the launcher (ordering, cycles, failure cleanup), service discovery (stale PIDs, malformed files), status server coalescing, and the CLI service lifecycle, using a stub service and a no-spawn status server.

`IntegrationServer.e2e.test.ts` exercises the real CLI end to end — starts an ontology, asserts discovery, reads the published CA cert, and round-trips an empty seed through the live seeding endpoint. It self-skips where `foundry` isn't installed.


packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -414,8 +419,14 @@ export class SeedBuilder {
 export function createSeedWithMetadata<T>(
   ontologyMetadata: Ontology.OntologyFullMetadata,
   fn: SeedFunction<T>,
-): [SeedOutput, T] {
+): {
+  output: SeedOutput;
+  context: T;
+} {
```
esinx, line 390: Updated `createSeedWithMetadata` return value format as a object instead of tuple, for clarity

packages/integration-testing/bin/postinstall.mjs
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+// @ts-check
+
+const main = async () => {
+  try {
+    const { postinstall } = await import("../build/esm/scripts/postinstall");
+    await postinstall();
+  } catch {
+    // no-op
```
esinx, line 10: Soft fail on import failure pre-build. This is because `postinstall` is also executed during osdk repo installation

packages/integration-testing/examples/example.ts
esinx, line 1: Will be moved into e2e tests with real objects / ignore for now

packages/integration-testing/src/scripts/download.ts
```diff
@@ -0,0 +1,125 @@
...
+import consola from "consola";
+import invariant from "tiny-invariant";
+
+import { resolveFoundryHost } from "./foundry-host.js";
+import {
+  getFoundryToken,
+  type ResolveBearerTokenOptions,
+} from "./foundry-token.js";
+
+const INSTALL_SCRIPT_PATH = "/code/api/extension/install-script";
```
esinx, line 31: Reuse the foundry cli script to install

packages/integration-testing/src/scripts/postinstall.ts
```diff
@@ -0,0 +1,41 @@
...
+  MIN_FOUNDRY_CLI_VERSION,
+  versionIsAtMinimum,
+} from "../utils/version.js";
+import { installFoundryCli } from "./download.js";
+
+export const postinstall = async (): Promise<void> => {
+  const foundryVersion = await getFoundryVersion();
+  if (typeof foundryVersion === "string") {
+    invariant(
+      versionIsAtMinimum(foundryVersion, MIN_FOUNDRY_CLI_VERSION),
```
esinx: Throw if failed to let user know that they should update; This is the only way to print anything during `postinstall`

packages/integration-testing/src/utils/empty-ontology-block.ts
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+import type { PreviewOntologyIrConverter } from "@osdk/generator-converters.preview";
+
+type OntologyBlockDataV2 = Parameters<
+  typeof PreviewOntologyIrConverter.getPreviewFullMetadataFromBlockData
+>[0];
+
+export const EMPTY_ONTOLOGY_BLOCK_DATA: OntologyBlockDataV2 = {
```
esinx, line 23: For `PreviewOntologyIrConverter` to take in an empty block + real ontology data as "imported metadata"

packages/integration-testing/src/utils/foundry-cli.ts
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { exec } from "node:child_process";
+import { promisify } from "node:util";
+
+import { gte, valid } from "semver";
```
esinx, line 20: Would rather use `semver` than implement own

packages/integration-testing/src/cli-service/CliServiceLauncher.ts
```diff
@@ -0,0 +1,281 @@
...
+      }
+      ordered.push(service);
+    };
+    for (const service of this.#services.values()) {
+      visit(service);
+    }
+    return ordered;
+  }
+
+  #assertAcyclic(): void {
```
esinx, line 239: Not exactly required since foundry-cli services have a grace period that waits for its dependencies to come up. Just good to ensure this safety.

packages/integration-testing/src/createIntegrationClient.ts
```diff
@@ -0,0 +1,199 @@
...
+  whenQuery: MockClient["whenQuery"];
+};
+
+export interface IntegrationClient {
+  client: LocalOntologyClient;
+  seed: SeedClient;
+}
+
+/** The definition kinds accepted by {@link Client}'s call signatures. */
+type ClientArg =
```
esinx, line 33: For properly typing the client wrapper

packages/integration-testing/src/createIntegrationClient.ts
```diff
@@ -0,0 +1,199 @@
...
+};
+
+export async function createIntegrationClient(
+  config: IntegrationClientConfig
+): Promise<IntegrationClient> {
+  const { baseUrl, metadata, caCertPath } = config;
+  const agent = new Agent({
+    connect: caCertPath
+      ? {
+          ca: await fs.readFile(caCertPath),
```
esinx, line 48: foundry-cli publishes a custom cert, if we cannot find it we should still prevent unauthorized requests from throwing

packages/integration-testing/src/IntegrationClient.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (def.type === "action") {
+      return realClient(def);
+    }
+    if (def.type === "interface") {
+      return realClient(def);
+    }
+    if (def.type === "object") {
+      return realClient(def);
+    }
+    return realClient(def);
```
esinx: The if branches are here because typescript does not want us to un-discriminate the union

packages/integration-testing/bin/generateConjure.mjs
esinx, line 1: Claude-generated conversion of `genereateConjure.sh` into `.mjs`

packages/integration-testing/package.json
```diff
@@ -0,0 +1,92 @@
...
+  "dependencies": {
+    "@osdk/api": "workspace:~",
+    "@osdk/client": "workspace:~",
+    "@osdk/foundry.ontologies": "catalog:foundry-platform-typescript",
+    "@osdk/unit-testing": "workspace:~",
+    "conjure-lite": "^0.7.3",
+    "consola": "^3.4.2",
+    "semver": "^7.7.2",
+    "tiny-invariant": "^1.3.3",
+    "undici": "^6.28.0"
```
esinx, line 55: undici is required for calling `fetch` with custom ca certs.

undici v7 and v8 are not supported by node 18 which is our min target; pinning a lower version.

packages/integration-testing/src/scripts/download.ts
```diff
@@ -0,0 +1,125 @@
...
+    await run("cmd.exe", ["/d", "/s", "/c", batPath], env);
+  } finally {
+    await rm(batPath, { force: true });
+  }
+};
+
+/**
+ * Installs the Foundry CLI by running the installer the stack serves, with the
+ * resolved stack url and token injected as `FOUNDRY_URL` and `TOKEN`. Warns
+ * when the installed CLI does not end up on PATH.
```
ssanjay1: Hmm may be missing something, but where is the warning happening?

packages/integration-testing/src/scripts/foundry-host.ts
```diff
@@ -0,0 +1,107 @@
...
+};
+
+export const resolveFoundryHost = async (
+  options: ResolveBearerTokenOptions = {},
+): Promise<string> => {
+  const env = options.env ?? process.env;
+  const envValue = env.FOUNDRY_EXTERNAL_HOST ?? env.FOUNDRY_HOSTNAME;
+  if (envValue) {
+    const configured = parseConfiguredHost(envValue);
+    if (typeof configured !== "undefined") {
```
ssanjay1: nit: do we need typeof here or can we just check undefined?

packages/integration-testing/src/scripts/foundry-token.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const parseGitRemoteToken = (gitRemote: string): string | undefined => {
+  let url: URL;
+  try {
+    url = new URL(gitRemote);
+  } catch {
+    return undefined;
+  }
+  return url.password || url.username || undefined;
```
ssanjay1: Does it matter if these are url encoded?

packages/integration-testing/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
+    "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
+    "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
+    "typecheck": "tsc --noEmit --emitDeclarationOnly false"
+  },
+  "dependencies": {
+    "@osdk/api": "workspace:~",
+    "@osdk/client": "workspace:~",
+    "@osdk/foundry.ontologies": "catalog:foundry-platform-typescript",
+    "@osdk/unit-testing": "workspace:~",
```
ssanjay1, line 50: Since these are both still in beta, should this be a ~ dep or a fixed * dep?

packages/integration-testing/src/createIntegrationClient.ts
```diff
@@ -0,0 +1,199 @@
...
+    );
+    return res;
+  };
+  const seedClientUtils: SeedClientProps = {
+    ref: (o, pk) => builder.ref(o, pk),
+    addAll: async (output) => {
+      builder.addAll(output);
+      await applySeed(builder.build());
+    },
+    set: async (output) => {
```
ssanjay1, line 68: Is there a way for people to batch these calls?

packages/integration-testing/src/IntegrationClient.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/** The definition kinds accepted by {@link Client}'s call signatures. */
+type ClientArg =
+  | ObjectTypeDefinition
+  | InterfaceDefinition
+  | ActionDefinition
+  | QueryDefinition
+  | Experiment<"2.0.8">
+  | Experiment<"2.1.0">
+  | Experiment<"2.8.0">
```
ssanjay1: You're missing experiment 2.2.0 right? Is there a way we can just derive this from the client type or did you want explicit control of it?

packages/integration-testing/src/IntegrationClient.ts
ssanjay1: Sure but maybe use a switch statement to be more idiomatic

packages/integration-testing/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "postinstall": "node ./bin/postinstall.mjs",
+    "test": "vitest run --pool=forks",
+    "transpileBrowser": "monorepo.tool.transpile -f esm -m normal -t browser",
+    "transpileCjs": "monorepo.tool.transpile -f cjs -m bundle -t node",
+    "transpileEsm": "monorepo.tool.transpile -f esm -m normal -t node",
+    "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
+    "typecheck": "tsc --noEmit --emitDeclarationOnly false"
+  },
+  "dependencies": {
+    "@osdk/api": "workspace:~",
```
ssanjay1: Do you need api and client as concrete deps? You also have them in dev deps, should this just be peer +dev

ssanjay1: Like the client should match whatever the user is using in functions right, rather than a separate version

packages/integration-testing/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "@osdk/foundry.ontologies": "catalog:foundry-platform-typescript",
+    "@osdk/unit-testing": "workspace:~",
+    "conjure-lite": "^0.7.3",
+    "consola": "^3.4.2",
+    "semver": "^7.7.2",
+    "tiny-invariant": "^1.3.3",
+    "undici": "^6.28.0"
+  },
+  "peerDependencies": {
+    "@osdk/generator-converters.ontologyir": "workspace:^",
```
ssanjay1: Is this used?

packages/integration-testing/package.json
```diff
@@ -0,0 +1,92 @@
...
+    "@osdk/generator-converters.preview": "workspace:^",
```
ssanjay1: Do these need to be dev deps too?

packages/monorepo.cspell/cspell.config.js
```diff
@@ -131,7 +129,7 @@ const cspell = {
     "oauth",
     "css",
   ],
-  words: ["todoapp", "borderless", "overfetching"],
+  words: ["todoapp", "borderless", "overfetching", "envrc"],
```
ssanjay1, line 132: Why did you add a word here?

packages/seed-helpers/src/SeedBuilder.ts
```diff
@@ -32,11 +32,16 @@ import { validateSeedObject } from "./validation.js";
 export type SeedFunction<T> = (seed: SeedBuilder) => T;
 
 export type SeedClient = {
-  <T = void>(seed: SeedFunction<T> | SeedOutput): Promise<T>; // equivalent to 'SeedBuilder.set'
+  /**
+   * Transactional interface to SeedBuilder
+   * Applies changes if function is provided, calls .addAll(seed) if SeedOutput is provided.
+   */
+  <T = void>(seed: SeedFunction<T> | SeedOutput): Promise<T>;
```
ssanjay1: Ah so this is the way you can batch things?

ssanjay1 (commented):
Took a quick first pass

packages/integration-testing/src/scripts/download.ts
esinx: Ah sorry this was a comment left when we had the warning before, but I migrated off to the foundry-cli provided installation script and this warning is no longer there.

packages/integration-testing/src/scripts/foundry-host.ts
esinx: We can check undefined

packages/integration-testing/src/scripts/foundry-token.ts
esinx: Good flag, we may want to decode it

packages/integration-testing/package.json
esinx: Should be a fixed dep; but I think your other comment about the api and the client package makes sense-- we should have them as peer dep (since the types are still user surfaced)

packages/integration-testing/package.json
esinx: Undici is used to provide fetch calls with custom CA certs which is what lohi requires

packages/monorepo.cspell/cspell.config.js
esinx: For user-facing warning in the postinstall script; there wasn't an mjs filter where it's more appropriate, but we could make a separate entry for this.

packages/integration-testing/package.json
esinx: The `.preview` actually should be a direct dep since we use the converter itself. `.ontologyir` is not used. Will remove.

packages/seed-helpers/src/SeedBuilder.ts
esinx: Yes, this will mutate the internal state then "sync" its value to the backend

packages/integration-testing/package.json
ssanjay1: Ah I meant "@osdk/generator-converters.ontologyir

packages/integration-testing/src/IntegrationClient.ts
esinx: Maybe we could do `Experiment<any>` and see if that works; although we would lose property typechecking there. This was because typescript will only infer from the last overload signature in the list and `Client` is heavily overloaded.

[TS playground example](https://www.typescriptlang.org/play/?#code/C4TwDgpgBA8gbhATgGwPYEMAmFMDECuAdgMbACWqhUAvFAN4BQUUAFAM7CIBcUHiZhAOYBKHnFRlMTVoXwBbHrLkAjJKKjjJDAL4MGoSFAAK6ROjk1jp8xGBI2AHnhI0WHARLlKAPgYB6PygDaFQAMyszCzI2KCVVRAAaWNRgXk4BQSgAH1j5eIYgA)

packages/integration-testing/package.json
esinx: Oh that makes more sense, I removed it since it was unused. Thanks for the flag.

packages/seed-helpers/src/SeedBuilder.ts
nihalbhatnagar: Agreed this is cleaner

packages/e2e.generated.catchall/.mdcheck/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
```
nihalbhatnagar: Why did this file get added

packages/integration-testing/src/utils/foundry-cli.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const execAsync = promisify(exec);
+
+const FOUNDRY_VERSION_PATTERN =
+  /cli (\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?)/u;
```
nihalbhatnagar, line 25: If a version of foundry cli that changes the way it outputs its version to not fit this regex anymore, wouldn't all prev versions of this library be bricked? Maybe what we should do is if the version parsed isnt semver compat we still let it through

packages/integration-testing/src/utils/version.ts
```diff
@@ -0,0 +1,60 @@
...
+  stdout.trim().match(FOUNDRY_VERSION_PATTERN)?.[1];
+
+export const getFoundryVersion = (): Promise<string | undefined> => {
+  return execAsync("which foundry")
+    .then(() => execAsync("foundry --version"))
+    .then(({ stdout }) => parseFoundryVersion(stdout))
+    .catch(() => undefined);
+};
+
+export const MIN_FOUNDRY_CLI_VERSION = "0.200.0";
```
nihalbhatnagar: Will we ever be bumping the version we need for the foundry cli?

packages/e2e.sandbox.catchall/src/integration-tests/e2e.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  FoundryCLIInstallation,
+} from "@osdk/integration-testing";
+import { expect } from "vitest";
+
+import { createSeed, test } from "./test.fixture.js";
+
+const foundryCliInstalled = await checkFoundryCli();
+
+const baseSeed = createSeed((seed) => {
+  const alice = seed.create(Person, {
```
nihalbhatnagar, line 20: This is not about this specific line but can we get this Intellisense to show something cleaner from the seed builder?

Image

packages/integration-testing/src/scripts/download.ts
nihalbhatnagar: I'm sorry can you elaborate here in which environments will these scripts exist

packages/integration-testing/oxlint.config.ts
```diff
@@ -0,0 +1,59 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { defineConfig } from "oxlint";
+
+import root from "../../oxlint.config.ts";
+
+// Nested oxlint config for @osdk/integration-testing. It inherits the whole
```
nihalbhatnagar: is there anyw way we can reduce these comments to just whats important

nihalbhatnagar (commented):
my morning questions but still digesting

packages/e2e.sandbox.catchall/src/integration-tests/test.fixture.ts
```diff
@@ -0,0 +1,52 @@
...
+    ),
+  ),
+  actionTypes: {},
+  interfaceTypes: {},
+  queryTypes: {},
+  valueTypes: {},
+  sharedPropertyTypes: {},
+};
+
+export const test = baseTest
```
nihalbhatnagar, line 42: small comment explaining why we have this setup (to narrow metadata) cause I already forgot myself

packages/e2e.generated.catchall/.mdcheck/tsconfig.json
esinx: Looks like an artifact from build, shouldn't be here

packages/integration-testing/src/createIntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export interface IntegrationServer {
+  start(): Promise<void>;
+  stop(): Promise<void>;
+  getOntologyUrl(): string | undefined;
+  getOntologyCaCertPath(): string | undefined;
+  createClient(): Promise<IntegrationClient>;
+}
+
+export async function createIntegrationServer(
```
nihalbhatnagar, line 60: Can we rename this file to createIntegrationServer? I feel that having the name be the main export from the file makes my searchability in the repo better, and its generally what we use as precedent in client (applYQuery, toDataValue, etc). Maybe a types directory with the main type shapes would be more helpful (see unit-testing) package

packages/integration-testing/src/IntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+  type IntegrationClient,
+} from "./IntegrationClient.js";
+import { OntologyServer } from "./OntologyServer.js";
+import { EMPTY_ONTOLOGY_BLOCK_DATA } from "./utils/empty-ontology-block.js";
+
+export type IntegrationServerConfig = {
+  /** Path to a prebuilt ontology metadata JSON. */
+  metadata: OntologyFullMetadata;
+  /** Path to the `foundry` binary. Defaults to `foundry` on `PATH`. */
+  foundryCliPath?: string;
```
nihalbhatnagar: Do we really need this? Who will be providing their own path to the exe

packages/integration-testing/src/IntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Directory to create this server's run directory in. Each server gets its
+   * own `.test-run-*` beneath it, removed on {@link IntegrationServer.stop}, so
+   * servers sharing a `projectPath` stay isolated. Must already exist; defaults
+   * to the working directory.
+   */
+  projectPath?: string;
+  /** How long each service may take to become ready. Defaults to 30_000ms. */
+  readyTimeoutMs?: number;
```
nihalbhatnagar: Have you run into this? I'd opt against introducing preemptive options until they actually become necessary/we get an ask for it.

esinx: This was added with preventing CI from hanging in mind, but not absolutely necessary for v0

packages/integration-testing/src/IntegrationServer.ts
esinx: We can remove this, I added this for my own use (needed to test multiple version of foundry cli pre/post lohi library version bump)

packages/integration-testing/src/utils/version.ts
esinx: If we add a new feature & need to regenerate the conjure generated source, yes.

packages/integration-testing/src/IntegrationServer.ts
nihalbhatnagar: Also what is a "service"? If I'm a user I should be completely transparent that anything is launching behind the scenes

packages/integration-testing/src/IntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+  metadata: OntologyFullMetadata;
+  /** Path to the `foundry` binary. Defaults to `foundry` on `PATH`. */
+  foundryCliPath?: string;
+  /**
+   * Directory to create this server's run directory in. Each server gets its
+   * own `.test-run-*` beneath it, removed on {@link IntegrationServer.stop}, so
+   * servers sharing a `projectPath` stay isolated. Must already exist; defaults
+   * to the working directory.
+   */
+  projectPath?: string;
```
nihalbhatnagar: If I'm not in a superrepo scaffolded repository, should this really be launching by default in the home dir? I feel like this should probably be in `node_modules/.cache/<tool>` which seems like prior art from some initial research

packages/integration-testing/src/createIntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+  config: IntegrationServerConfig,
+): Promise<IntegrationServer> {
+  const {
+    metadata,
+    projectPath = process.cwd(),
+    foundryCliPath,
```
nihalbhatnagar: Do superrepo/functions always have their ci test runner run from the root dir? Should we use `findup` to search for the .palantir dir instead?

packages/integration-testing/src/cli-service/CliServiceLauncher.ts
```diff
@@ -0,0 +1,281 @@
...
+  async dependenciesReady(service: FoundryCliService): Promise<boolean> {
+    const healths = await this.checkDependencies(service);
+    return healths.every(({ ready }) => ready);
+  }
+
+  /**
+   * Stop every service, dependents before their dependencies, waiting for each
+   * process to be gone before the next is signalled.
+   */
+  async stop(): Promise<void> {
```
nihalbhatnagar, line 207: Will this affect already running foundry servers spun up by superrepo?

packages/integration-testing/src/CliServiceLauncher.ts
```diff
@@ -0,0 +1,281 @@
...
+export interface CliServiceLauncherConfig {
+  /** Directory the services run in; `.palantir/` hangs off it. */
+  projectDir?: string;
+  /** Path to the `foundry` binary. */
+  foundryCliPath?: string;
+  /** Status server to use. One is created if not supplied. */
+  statusServer?: StatusServer;
+}
+
+const formatError = (
```
nihalbhatnagar: nit: helper at bottom of file

packages/integration-testing/src/cli-service/CliServiceLauncher.ts
esinx: No, this will only work on the services spawned by this class

packages/integration-testing/src/IntegrationServer.ts
esinx: A service is a single process ran by `foundry`, so in this case it could be:
`foundry start status-server` or `foundry start ontology`

packages/integration-testing/src/createIntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+    projectDir: testPath,
+    foundryCliPath,
+  });
+  // write metadata to metadataPath for foundry cli to read from
+  await writeMetadata({
+    previewMetadata,
+    metadataPath,
+  });
+
+  const ontology = serviceLauncher.register(
```
nihalbhatnagar, line 79: Could you spin up a draft PR that highlights a very very basic way you think function backed actions would be added here (doesn't need to work, but will be very valuable as a reference)

packages/integration-testing/src/IntegrationServer.ts
esinx: This is what vitest does and I agree, initially followed the superrepo approach but this seems more generic

packages/integration-testing/src/createIntegrationServer.ts
esinx: The `.palantir` directory is always populated at `cwd` unless a `foundry.yml` exists

packages/integration-testing/src/createIntegrationServer.ts
```diff
@@ -0,0 +1,133 @@
...
+    }),
+  );
+
+  return {
+    start: async () => {
+      await serviceLauncher.start();
+    },
+    stop: async () => {
+      await serviceLauncher.stop();
+      await fs.rm(testPath, { recursive: true, force: true });
```
nihalbhatnagar, line 98: Whats being written to this dir?

packages/integration-testing/src/cli-service/FoundryCliService.ts
```diff
@@ -0,0 +1,240 @@
...
+  getReadyTimeoutMs(): number {
+    return this.#readyTimeoutMs;
+  }
+
+  getProjectPath(): string {
+    return this.#projectPath;
+  }
+
+  getCaCertPath(): string | undefined {
+    return this.#context?.discoverer.get(this.name)?.caCertPath ?? undefined;
```
nihalbhatnagar, line 101: Does the foundry cli create its own cert?

esinx: Yes it does, it usually is cached (irrelevant to this code though).

packages/integration-testing/src/createIntegrationServer.ts
esinx: The modified ontology metadata after conversion and `.palantir` path for service discovery

packages/integration-testing/src/createIntegrationClient.ts
nihalbhatnagar: Why wouldn't we want to throw?

packages/integration-testing/src/IntegrationClient.ts
```diff
@@ -0,0 +1,199 @@
...
+  type SeedOutput,
+} from "@osdk/seed-helpers";
+import { createMockClient, type MockClient } from "@osdk/unit-testing";
+import { fetch as undiciFetch, Agent } from "undici";
+
+import { OntologySeedingService } from "./generated/cli/index.js";
+
+type LocalOntologyClient = Client & {
+  // local ontology does not have the ability to support function queries at the moment.
+  whenQuery: MockClient["whenQuery"];
```
nihalbhatnagar: Shouln't we expose clear all stubs? Or should the user create a new client

packages/e2e.sandbox.catchall/src/integration-tests/e2e.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    expect(Number(page.totalCount)).toBe(2);
+    expect(
+      page.data.find((v) => v.id === baseSeed.context.todoItem1.id),
+    ).toBeDefined();
+    expect(
+      page.data.find((v) => v.id === baseSeed.context.todoItem2.id),
+    ).toBeDefined();
+  });
+  test("Alice and Charlie are friends", async ({ integration }) => {
+    const friends = await integration
```
nihalbhatnagar: Can we add an interface test?

packages/integration-testing/src/scripts/foundry-token.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const resolveFoundryToken = async (
+  options: ResolveBearerTokenOptions = {},
+): Promise<string | undefined> => {
+  const env = options.env ?? process.env;
+  if (env.JEMMA) {
+    const jobToken = env.JOB_TOKEN;
+    invariant(jobToken, "JOB_TOKEN: unbound variable");
+    return jobToken;
+  }
```
nihalbhatnagar: Dont need to support code repos for now I'd say

packages/integration-testing/src/scripts/foundry-host.ts
```diff
@@ -0,0 +1,107 @@
...
+  | "macos-amd64"
+  | "macos-arm64"
+  | "windows-amd64";
+
+export interface FoundryCliInstallPathsOptions {
+  env?: NodeJS.ProcessEnv;
+  platform?: NodeJS.Platform;
+}
+
+export const getArtifactPlatform = (
```
nihalbhatnagar: Talked offline, remove

packages/integration-testing/src/IntegrationServer.ts
esinx: Talked offline, it's non-trivial to get the `node_modules` folder from the package path. Will keep cwd but allow user to set projectPath for anomalies.

packages/integration-testing/src/IntegrationClient.ts
esinx: Agreed / exposing clearStubs()

packages/e2e.sandbox.catchall/package.json
```diff
@@ -43,6 +44,8 @@
     "@osdk/e2e.generated.catchall": "workspace:~",
     "@osdk/foundry": "catalog:foundry-platform-typescript",
     "@osdk/functions": "workspace:~",
+    "@osdk/integration-testing": "workspace:~",
+    "@osdk/seed-helpers": "workspace:~",
```
nihalbhatnagar, line 48: Maybe not relevant to this PR but is this the best name for this package? Maybe @osdk/ontology-seeding

packages/integration-testing/src/types.ts
```diff
@@ -0,0 +1,56 @@
...
+  foundryCliPath?: string;
+  /**
+   * Directory to create this server's run directory in. Each server gets its
+   * own `.test-run-*` beneath it, removed on {@link IntegrationServer.stop}, so
+   * servers sharing a `projectPath` stay isolated. Must already exist; defaults
+   * to the working directory.
+   */
+  projectPath?: string;
+  /** How long each service may take to become ready. Defaults to 30_000ms. */
+  readyTimeoutMs?: number;
```
nihalbhatnagar: I thought we were removing these?

packages/integration-testing/src/createSeedClient.ts
```diff
@@ -0,0 +1,99 @@
...
+  metadata: Ontology.OntologyFullMetadata;
+  fetchFn?: typeof fetch;
+};
+
+type PropsOf<T> = { [K in keyof T]: T[K] };
+type SeedClientProps = PropsOf<SeedClient>;
+
+export function createSeedClient(config: SeedClientConfig): SeedClient {
+  const { baseUrl, metadata, fetchFn } = config;
+  const builder = new SeedBuilder(metadata);
```
nihalbhatnagar, line 38: Should the seed builder expose a constructor that accepts handlers for things we want to executer for each method instead of us monkeypatching it on?

packages/integration-testing/src/utils/foundry-cli.ts
esinx: Done, now it ignores the check and allows execution

packages/integration-testing/src/utils/version.ts
esinx: Will be doing this for the new discovery flag

packages/e2e.sandbox.catchall/src/integration-tests/e2e.test.ts
esinx: Unfortunately this might be a typescript limitation. We can FLUP on this later maybe?

packages/integration-testing/src/createIntegrationServer.ts
esinx: Will do!

packages/integration-testing/src/scripts/foundry-token.ts
esinx: Can remove this

packages/integration-testing/src/createSeedClient.ts
esinx: Talked offline, moved SeedClient into integration testing

packages/e2e.sandbox.catchall/package.json
esinx: We can make a FLUP on this

packages/integration-testing/src/createIntegrationClient.ts
esinx: We are running in a local environment so an absent CA cert shouldn't matter as much, and the untrusted cert is still being served on the server since it only talks in https (not http). Not something we must flag to the user-- so we can silence the error and ignore.

nihalbhatnagar (dismissed):

nihalbhatnagar (approved):

PR #3788, opened 2026-07-31
Fix storybook/no-uninstalled-addons false positives when linting from the repo root

esinx:
## Problem

Linting `packages/react-components-storybook/.storybook/main.ts` from the repo root — which is what `lint-staged`, and therefore the pre-commit hook, does — fails with seven `storybook(no-uninstalled-addons)` errors:

```
.storybook/main.ts:24:5: error storybook(no-uninstalled-addons): The @storybook/addon-a11y is not installed
in osdk-main/. Did you forget to install it or is your package.json in a different location?
... (same for addon-docs, addon-links, addon-mcp, addon-vitest, msw-storybook-addon, storybook-addon-tag-badges)
```

Every one of those addons *is* declared in `packages/react-components-storybook/package.json`. The rule resolves its package.json relative to the process working directory:

```js
// eslint-plugin-storybook/dist/index.js
let packageJsonPath = resolve(packageJsonLocation || "./package.json");
```

`lint-staged` invokes `oxlint` from the repo root, so the rule read the root `package.json`, where none of the Storybook addons are declared. This left the pre-commit hook unable to pass on any commit touching `.storybook/main.ts`.

## Fix

Pass an absolute `packageJsonLocation` derived from `import.meta.url`, so resolution no longer depends on the directory `oxlint` happens to run from (`resolve()` returns an absolute input unchanged). This mirrors the existing `import.meta.url` pattern used for the license-header plugin path in the root `oxlint.config.ts`.

## Verification

- `npx lint-staged` from the repo root: passes (previously failed).
- `npx oxlint -c packages/react-components-storybook/oxlint.config.ts packages/react-components-storybook/.storybook/main.ts`, run from both the repo root and the package directory: no errors.
- Counterfactual — the rule is still enforcing, not silenced. Adding a bogus addon to the `addons` array still errors, and now names the right file:

```
.storybook/main.ts:27:5: error storybook(no-uninstalled-addons): The @storybook/addon-definitely-not-installed
is not installed in osdk-main/packages/react-components-storybook/package.json.
```

## Notes

No changeset: `@osdk/react-components-storybook` is `private: true` and publishes nothing.

Out of scope, noted while here — `.storybook/main.ts:67` carries a `no-shadow` warning (the `viteFinal(config)` parameter shadows the module-level `const config`). It is warning-level repo-wide and pre-existing, so it is left alone.


esinx:
[.lintstagedrc.mjs](https://github.com/palantir/osdk-ts/blob/main/.lintstagedrc.mjs) is ran as a part of the pre-commit hook. It includes the snippet below to execute `oxlint -c packages/react-components-storybook/oxlint.config.ts --fix packages/react-components-storybook/.storybook/main.ts` which is done at the root of the package, not the package dir; this is why `fix-lint` or `lint` would not catch this, but the pre-commit hook version of the linter would catch this.

```typescript
  // Same as the OXC_PACKAGE_GLOB handler above, but `-c` points at each
  // package's nested oxlint config (which extends the root). Mirrors those
  // packages' fix-lint scripts.
  ...Object.fromEntries(
    Object.entries(OXC_NESTED_CONFIG_PACKAGES).map(([pkg, config]) => [
      `packages/${pkg}/**/*.{js,jsx,ts,tsx,mjs,cjs}`,
      (files) => {
        const match = micromatch.not(
          files,
          IGNORED_FILE_GLOBS,
          MICROMATCH_OPTS,
        );
        if (match.length === 0) return [];
        return [
          `oxlint -c ${config} --fix ${match.join(" ")}`,
          `oxfmt -c oxfmt.config.ts ${match.join(" ")}`,
          `${CSPELL_CMD} ${match.join(" ")}`,
        ];
      },
    ]),
  ),
```

olliejenningsPT (approved):

packages/react-components-storybook/oxlint.config.ts
```diff
@@ -85,7 +85,13 @@ export default defineConfig({
         "**/.storybook/main.mjs",
       ],
       rules: {
-        "storybook/no-uninstalled-addons": "error",
+        "storybook/no-uninstalled-addons": [
+          "error",
+          {
+            packageJsonLocation: new URL("package.json", import.meta.url)
```
ssanjay1, line 91: You probably want to do something like `fileURLToPath(new URL("package.json", import.meta.url)),` since I think new URL will try and URL escape things like spaces and stuff which are valid in filenames but not URLs

ssanjay1: Probably worth checking other spots in the repo that do this too

esinx: Talked offline, will FLUP with repo wide check on URL -> filePath sanity

PR #3789, opened 2026-08-02
Support synchronous token providers

cacabo:
Currently `createPlatformClient` and related constructors only support `() => Promise<string>` token providers. This is a bit annoying when you already have a token handy, like when you already got an OAuth token or have a long-lived user token.

https://github.com/palantir/osdk-ts/blob/fd00f48adf2cf8cfc916d9ecb942bd3795ef7931/packages/client/src/createPlatformClient.ts#L34

This PR widens the type to `() => string | Promise<string>` to support both sync and async token providers.

### DIFF
--- a/packages/shared.client.impl/src/createSharedClientContext.ts
+++ b/packages/shared.client.impl/src/createSharedClientContext.ts
@@ -28,7 +28,7 @@ export const USER_AGENT_HEADER = "Fetch-User-Agent";
 
 export function createSharedClientContext(
   baseUrl: string,
-  tokenProvider: () => Promise<string>,
+  tokenProvider: () => string | Promise<string>,
   userAgent: string,
   fetchFn: typeof globalThis.fetch = fetch,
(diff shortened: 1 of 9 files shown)


packages/shared.client.impl/src/createSharedClientContext.ts
```diff
@@ -54,7 +56,7 @@ export function createSharedClientContext(
         }
       }
 
-      const token = await tokenProvider();
+      const token = await asyncTokenProvider();
```
nfdiop, line 59: I think we can keep using tokenProvider  like this (const token = await tokenProvider()) and return it below (so we don't return a new function provider to the user)

nfdiop (commented):
Thanks for this contribution! Left a comment

cacabo:
Closing based on discussion: we don't want users to await an oauth provider and thus not have tokens auto-refresh

PR #3791, opened 2026-08-03
Bump foundry platform SDK catalog to 2.70.0

braeden:
Moves `@osdk/foundry.*` and `@osdk/internal.foundry.ontologies` from `2.63.0` to `2.70.0`. Stubs and snapshots pick up `aliases` / `datasources`, now required on `ObjectTypeV2`.

### `streamingExecute`

`@osdk/foundry.functions` stopped generating a binding for `/v2/functions/queries/{queryApiName}/streamingExecute` in `2.67.0`, but the endpoint is still served and `StreamingExecuteQueryRequest` is still published — only the binding went away. Holding the whole package back at `2.66.0` would pull a second copy of `foundry.core` / `foundry.ontologies` into the tree and block future bumps, so instead `packages/client/src/queries/streamingExecuteQuery.ts` declares that one endpoint locally as a verbatim copy of the `2.66.0` generated binding, and `applyStreamingQuery` calls it. Delete the file if the endpoint is published again.

The existing `executeStreamingFunction` tests cover it unchanged — they mock the real URL, so URL construction, auth, and ndjson handling are all exercised.

### Why not 2.71.0

`2.71.0` adds a required `actionTypes` to `ObjectTypeInterfaceImplementation`; `2.70.0` is the highest version whose changes are all optional additions.


braeden:
Superseded by #3750, which landed the same 2.70.0 catalog bump and the same `aliases`/`datasources` stub fixes.

The one difference was `streamingExecute`: this PR kept the endpoint working via a local binding, while #3750 removed streaming query support outright. Deferring to that decision. The local binding is preserved on the `braedens/bump-foundry-platform-sdks-2.70` branch if it's ever wanted back.

PR #3792, opened 2026-08-03
Migrate typescript-sdk-docs-examples to oxlint/oxfmt

olliejenningsPT:
Migrates `@osdk/typescript-sdk-docs-examples` to oxlint + oxfmt, the last package still on ESLint + dprint from #3031. Rebased on latest main.

It was the final holdout because its `codegen` step reformats freshly-generated examples via `pnpm run format`, and that runs on every CI test-matrix leg (Node 18-24), since a generator depends on its codegen output. oxfmt can't run on the old legs (its native binding isn't installed on Node 18, and it can't load the TypeScript config on Node 20), so codegen must never re-run there. #3783 plus the transpile-hash-stabilization commit below keep codegen a cache hit on those legs, so the format step never runs.

## Commits
1. **turbo: exclude generatedNoCheck from transpile inputs (stabilize hash)** — prerequisite. `src/generatedNoCheck/**` (codegen's gitignored output) was in the transpile task inputs, so its appearance between the build-materialize and test steps flipped transpile hashes, which cascaded into codegen's hash and re-ran it on the test legs. Excluding it keeps codegen's hash stable, so #3783's cache restore actually holds and oxfmt never runs on Node 18/20.
2. **Migrate typescript-sdk-docs-examples to oxlint/oxfmt** — the migration:
   - Adds `oxc: true` + a nested `oxlint.config.ts` (extends root, ignores the generated `src/examples/`) to the package's monorepolint archetype.
   - Switches `lint`/`fix-lint`/`format` from eslint+dprint to oxlint+oxfmt, and drops the ESLint devDeps.
   - Registers it in `.lintstagedrc` `OXC_PACKAGES` and `dprint.json` excludes.
   - Reformats sources with oxfmt via `codegen`; verified codegen is idempotent.
   - Adds an empty changeset (the package is private, so there's no published change to release).

## How to review
The diff is ~118 files, but the reviewable surface is small: the 8-line `turbo.json` change, the 5 config changes, and the new `oxlint.config.ts`. Everything else (`src/examples/` snippets, `src/typescriptOsdk*.ts`, `ontology.json`) is the mechanical dprint to oxfmt reformat of generated output. Rather than reading those, regenerate and confirm they are a stable fixed point:

```
pnpm turbo codegen --filter=@osdk/typescript-sdk-docs-examples --force
git diff --exit-code   # expect no changes
```


packages/typescript-sdk-docs-examples/package.json
```diff
@@ -27,11 +27,10 @@
     "check-spelling": "cspell --quiet .",
     "clean": "rm -rf lib dist types build tsconfig.tsbuildinfo",
     "codegen": "rm -rf src/generatedNoCheck/* && osdk-unstable-typescript generate --outDir src/generatedNoCheck --ontologyPath ontology.json --version 0.1.0 --packageType module --internal --skipPackageJsonUpdate --disableOntologyMetadataExport && rm -rf src/examples/* && generate-examples && pnpm run format",
-    "fix-lint": "eslint . --fix && dprint fmt",
-    "format": "npx dprint fmt",
+    "fix-lint": "oxlint -c ./oxlint.config.ts --fix . && oxfmt -c ../../oxfmt.config.ts .",
+    "format": "oxfmt -c ../../oxfmt.config.ts .",
     "generateExamples": "rm -rf src/examples/* && generate-examples && pnpm run format",
-    "lint": "eslint . && dprint check",
-    "lint:fix": "eslint src --fix",
```
nfdiop, line 34: assuming this is not used anywhere

.monorepolint.config.mjs
```diff
@@ -464,15 +464,13 @@ const archetypeRules = archetypes(
...
-      // and that step runs on every CI test-matrix leg (Node 18-24) because a
-      // generator package, @osdk/osdk-docs-context-generator, depends on this
-      // package's codegen output. oxfmt cannot load the repo's TypeScript
-      // oxfmt.config.ts on Node < 22.18, so the format step must stay on dprint
-      // (which is Node-version-independent). Migrate this package once codegen is
-      // restored from the Node-24 build cache on the matrix (see #3031 follow-up)
-      // rather than re-executed per leg.
+      // Migrated to the oxc toolchain (oxlint + oxfmt). Its `codegen` step
+      // reformats freshly-generated documentation examples via `pnpm run format`
+      // (now oxfmt). This is safe on the CI test matrix because codegen is
```
nfdiop: nit: maybe I would reword into "This is safe on CI because the test step is the only step that runs on Node 18; every other job runs on Node 24. And when the test runs on Node 18, it is safe because codegen ..."

nfdiop (dismissed):

packages/typescript-sdk-docs-examples/package.json
olliejenningsPT: Yeah this seems to be from some legacy, nothing calls or references this

.monorepolint.config.mjs
olliejenningsPT: Will reword based on suggestion

nfdiop (approved):

PR #3803, opened 2026-08-05
Run the pull-request CI on CircleCI alongside GitHub Actions

olliejenningsPT:
The Claude PR description was not very clear, so will try and be as concise as possible.

This PR introduces CircleCI for running our CI checks. It is a faithful mirror of what GitHub CI does, but we some changes that l will outline.

### Not Changed

- **Release workflow** - this stays the same
- **Current CI** - this is additive to our current CI

### Why / How this is this faster

- **Dedicated machines** - GitHub Runners used shared VMs which are 4 core machines, Circle gives us dedicated machines with us to 16 cores (if we require, based on runner size)
- **More Cores** - Turbo + other rust based tooling greatly benefit from more cores, so they can parallelise, they spin up 1 thread per CPU (1 task per CPU in turbo case).
- **Parallelism** - similar to above, we even have the ability to have multiple machines per stage, we do this for certain tests (`react-components`) because those tests are very heavy, and splitting them out so they run individually gives a big speed up
- **Better caching / splitting of stages** - arguably we could do this on GH CI, but also addressed an issue where storybook gets built multiple times, instead we build this once.

### Benefits

- **browsable storybook per PR** - CircleCI artifacts are browsable, and so per PR we can output these and have a full interactive storybook to see how PR changes that affect any `react-components` look. (was previously available, but quite hacky to get it working, and not perfect)
- **Proper test result reporting** - We are now storing proper test artifacts, which means if there is a failing test, you don't need to scroll through logs, instead you get a nice view in the circle UI to see which tests are failing, and the logs related to those tests only.
- **Faster** - this drops the average CI build times from ~11mins on GitHub to ~4mins on Circle

### Follow Ups

- **RAM Cache** - it requires some re-wiring for pnpm to work out the box, but it will shave off around 30s-45s per stage where we restore from cache for node_modules
- **GitHub PR comments** - right now we do not have these running, we would need to add a `GITHUB_TOKEN` to Circle in order to get this running (blueprint does this), but l didn't want multiple identical comments in this initial PR
- **Proper Code Coverage** - this was attempted, but due to a PR this broke, completely stripping this out and putting in a FLUP.
- **Chromatic** - this enables visual diffs for storybook, we should be careful in only enabling this for PRs that actually touch or trigger changes in the components.


.monorepolint.config.mjs
```diff
@@ -1406,7 +1406,21 @@ function standardPackageRules(shared, options) {
...
               : ""
           }
+              // JUnit XML for CI. Deliberately not passed as a turbo \`--\` arg:
+              // turbo folds passthrough args into its *global* hash, so doing
+              // that makes every task in the run a cache miss, not just \`test\`.
+              reporters: process.env.CI ? ["default", "junit"] : ["default"],
+              outputFile: { junit: "reports/junit.xml" },
               coverage: {
+                // Switched on by env, not a turbo \`--\` arg, for the same reason
+                // as the reporter above. CI sets it on the Node 24 leg only;
```
ssanjay1: I don't think the node 24 piece is true here

.circleci/scripts/collect-test-results.sh
```diff
@@ -0,0 +1,47 @@
...
+  # safe here in a way it would not be for hand-written XML. Inputs go through
+  # the environment rather than argv, because `node -e` shifts positional
+  # arguments down by one and that is an easy thing to get subtly wrong.
+  # The `$1` below is a regex backreference for String.replace, not a shell
+  # positional, so the single quotes are deliberate.
+  # shellcheck disable=SC2016
+  REPORT="$report" PKG="$pkg" node -e '
+    const fs = require("node:fs");
+    const xml = fs.readFileSync(process.env.REPORT, "utf8");
+    process.stdout.write(
```
ssanjay1: Do we need to do this? Doesn't vitest's junit reporter option have a className field you can use instead? https://vitest.dev/guide/reporters.html#junit-reporter Then the collecting into one directory could just go in config.yml right?

.monorepolint.config.mjs
```diff
@@ -1406,7 +1406,21 @@ function standardPackageRules(shared, options) {
...
+                // \`pnpm run coverage\` passes its own CLI flags, which win.
+                enabled: process.env.COVERAGE === "true",
+                provider: "v8",
```
ssanjay1: Is this not the default, why specify?

.circleci/config.yml
```diff
@@ -0,0 +1,766 @@
...
+      Set $TURBO_FORCE so a run cannot rely on stale cached artifacts and mask a
+      broken task dependency. The list is shorter than ci.yml's because turbo
+      already tracks the rest; see README "Why the infra list is short".
+    steps:
+      - run:
+          name: Detect infra changes
+          command: |
+            force=""
+            if files=$(.circleci/scripts/changed-files.sh); then
+              if printf '%s\n' "$files" | grep -qE '^(pnpm-lock\.yaml|pnpm-workspace\.yaml)$'; then
```
ssanjay1: I'm confused why we need this, doesn't the turbo cache already catch these

.circleci/config.yml
```diff
@@ -0,0 +1,766 @@
...
+                echo "Infrastructure changed; forcing a clean build."
+                force="--force"
+              fi
+            else
```
ssanjay1: Also not sure if this is right if you're leaning on the turbo cache

.circleci/config.yml
```diff
@@ -0,0 +1,766 @@
...
+  # PR-only reporting and guards
+  # ---------------------------------------------------------------------------
+
+  bundle-size:
+    executor:
+      name: node
+      resource-class: large
+    steps:
+      - checkout
+      - halt-unless-pull-request
```
ssanjay1: This isn't in ci-all, can we just like have a filter to ignore certain branches?

.circleci/config.yml
```diff
@@ -0,0 +1,766 @@
...
+            - build
+            - build-apps
+            - build-storybook
+            - typecheck
+            - lint
+            - cspell
+            - test
+            - e2e
+            - fork-guard
+            - changesets
```
ssanjay1: Does this need to be in ci:all?

.circleci/config.yml
```diff
@@ -0,0 +1,766 @@
...
+              exit 1
+            fi
+
+  halt-unless-pull-request:
+    description: End the job successfully when this is not a pull-request build.
+    steps:
+      - run:
+          name: Skip unless this is a pull request
+          command: |
+            case "${CIRCLE_BRANCH:-}" in
```
ssanjay1: why can't you use CIRCLE_PULL_REQUEST

.circleci/scripts/pack-packages.sh
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env bash
+#
+# Pack every publishable workspace package into a directory of .tgz files, plus a
```
ssanjay1: This is all net new stuff for the testing out unreleased PRs right? Mind keeping it out of this PR for now

.circleci/scripts/pr-comment.sh
```diff
@@ -0,0 +1,80 @@
...
+set -uo pipefail
+
+MARKER="${1:?usage: pr-comment.sh <marker> <body-file>}"
+BODY_FILE="${2:?usage: pr-comment.sh <marker> <body-file>}"
+
+REPO="${CIRCLE_PROJECT_USERNAME:-}/${CIRCLE_PROJECT_REPONAME:-}"
+API="https://api.github.com"
+
+if [ -z "${GITHUB_TOKEN:-}" ]; then
+  echo "GITHUB_TOKEN is not set; skipping PR comment. Body would have been:"
```
ssanjay1, line 23: Is this normal to have set?

.monorepolint.config.mjs
```diff
@@ -1406,7 +1406,21 @@ function standardPackageRules(shared, options) {
               }\n            },`
               : ""
           }
+              // JUnit XML for CI. Deliberately not passed as a turbo \`--\` arg:
+              // turbo folds passthrough args into its *global* hash, so doing
+              // that makes every task in the run a cache miss, not just \`test\`.
+              reporters: process.env.CI ? ["default", "junit"] : ["default"],
+              outputFile: { junit: "reports/junit.xml" },
               coverage: {
```
ssanjay1: Honestly, do we need any of these args by default here if you already passed them in your collect script here? https://github.com/palantir/osdk-ts/blob/9d44daea1bac0736fba4a210e0dd287e48fc0f92/scripts/coverage/collect.sh#L19

ssanjay1:
Some of these comments may be out of date/redundant as well, I would heavily reduce if possible

.circleci/scripts/pack-packages.sh
olliejenningsPT: Yep happy to keep out for a FLUP

.circleci/scripts/pr-comment.sh
olliejenningsPT: Yes blueprint has it set, but l didn't want to set it and trigger double PR comments in this PR

.circleci/config.yml
olliejenningsPT: No really, we can remove

.monorepolint.config.mjs
olliejenningsPT: it is default, was just being explicit since in theory someone could install Istanbul in a package for coverage

.monorepolint.config.mjs
olliejenningsPT: your right, sorry left over comment from churn

.circleci/config.yml
olliejenningsPT: Yeah l can mirror the opposite of the release.yml workflow so we don't overlap

.circleci/config.yml
olliejenningsPT: We can

.circleci/scripts/collect-test-results.sh
olliejenningsPT: So partially right, yes the className means we don't need to do the re-write, but the collecting still needs to be here, as they are all called junit. Now l could use mrl generator to derive these names so they output to a unique name per package, but its not clean code, and so would keep the collect script for this

.monorepolint.config.mjs
olliejenningsPT: yeah fair, actually removed since coverage is now off, a recent PR brought in a vitest version that is incompatible with our current coverage version

packages/agents/vitest.config.mts
ericjeney, line 1: In the future, I find it useful to separate all generated files like this one into a separate commit so that reviewers can easily tell the difference between the human-authored changes and the repetitive boilerplate

.circleci/README.md
ericjeney, line 1: This file seems quite low signal/noise. If there are particular things you think are important to call out, I'd comment them in the circle config itself, but otherwise this seems unlikely to be read by anyone (and likely to be out-of-date in the future if it ever is).

.circleci/scripts/is-affected.sh
```diff
@@ -0,0 +1,62 @@
...
+      '
+)"
+status=$?
+
+# Any failure at all — turbo erroring, unparseable output — means we do not
+# know, and "do not know" has to mean "run it".
+if [ "$status" -ne 0 ]; then
+  echo "Could not determine the affected set (exit $status); treating $pkg as affected."
+  exit 0
+fi
```
ericjeney: I think it'd be better to fail the build with an error here instead of just swallowing it. Otherwise, this is likely to regress in the future and it'll take ~months for someone to notice

turbo.json
```diff
@@ -266,8 +266,24 @@
 
     "test": {
       "outputLogs": "new-only",
-      "env": ["FORCE_COLOR"],
-      "dependsOn": ["^transpileEsm"]
+      /* CI is hashed because each package's vitest config switches on it to add
+         the JUnit reporter. Without it, a cache entry created locally (no XML)
+         could be restored in CI and the report upload would silently be empty.
```
ericjeney: Is this true? Do we currently persist local caches somewhere that CI would be able to restore?

Does this also mean that test jobs run on CI will no longer be restorable from cache locally?

turbo.json
```diff
@@ -266,8 +266,24 @@
...
+         one. That cost the test job its entire build cache. */
+      /* COVERAGE is hashed for the same reason as CI: the vitest config
+         switches on it, so a cache entry from a non-coverage run must not be
+         restored into one that is meant to emit coverage. */
+      "env": ["FORCE_COLOR", "CI", "COVERAGE"],
+      "dependsOn": ["^transpileEsm"],
+      /* Declaring the outputs is what makes a cache hit restore these files;
+         without it turbo replays the logs and writes nothing. coverage/** being
+         declared is also what would let scripts/coverage/collect.sh drop its
+         --force, which only exists because the output was undeclared. */
```
ericjeney: Seems like an unnecessary comment given every other `outputs` in this file is uncommented 🙂

turbo.json
```diff
@@ -266,8 +266,24 @@
 
     "test": {
       "outputLogs": "new-only",
-      "env": ["FORCE_COLOR"],
-      "dependsOn": ["^transpileEsm"]
+      /* CI is hashed because each package's vitest config switches on it to add
```
ericjeney: nit: Should use a consistent comment pattern in this file

.gitignore
```diff
@@ -9,6 +9,11 @@ node_modules/
 /coverage/
 packages/*/coverage/
 
+# JUnit XML written by CI test runs. Must stay ignored: turbo folds untracked,
+# non-ignored files into a package's input hash, so leaving these visible makes
+# every test task a cache miss on the run after it emits a report.
+reports/
```
ericjeney: Should this be `packages/*/reports/` to mirror coverage above?

.circleci/config.yml
```diff
@@ -0,0 +1,704 @@
...
+          name: Collect JUnit reports
+          when: always
+          command: .circleci/scripts/collect-test-results.sh /tmp/test-results
+      - store_test_results:
+          path: /tmp/test-results
+      - run:
+          # Merged in-job because there is no separate coverage job, so each
+          # container reports only on the packages it ran. Non-fatal: a merge
+          # problem should not turn a passing test run red.
+          name: Merge coverage
```
ericjeney: Is COVERAGE ever set to true? This looks like dead code to me

.circleci/scripts/fork-guard.sh
ericjeney, line 1: Can you explain more here? My expectation is that Circle wouldn't inject credentials into builds from forks at all (and we should configure it that way if we haven't yet!)

I'm unclear on how a file like this can help given Circle runs builds with the config from the PR it's building, so someone could just delete this file if it were preventing anything.

.circleci/config.yml
```diff
@@ -0,0 +1,704 @@
...
+commands:
+  activate-pnpm:
+    description: Make the repo's pinned pnpm available on PATH.
+    steps:
+      - run:
+          name: Activate pnpm
+          command: |
+            # cimg/node's bundled corepack can be too old to verify the
+            # signatures of current pnpm releases.
+            sudo npm install --global --silent corepack@latest
```
ericjeney: Corepack seems a bit heavy to quietly take on just for this, vs. installing pnpm directly

But if we do pin Corepack, wan we pin it to a version? Otherwise builds won't be reproducible and we're susceptible to corepack breaking out from under us

ericjeney (commented):
I started taking a look but haven't finished. There's a lot going on here, so I think this is worth breaking this down into smaller chunks that are easier to review before we go further

Can you include more overall timing details in the PR description as well, as well as a summary on what makes it faster than the Github Actions alternative? I know you reported it's faster, but it's unclear to me what the key differences are that makes that so. I'd also be interested in how the timings compare for a PR that affects one package vs. one that affects all packages, and whether you've confirmed that a full build of all packages still works as expected.

packages/agents/vitest.config.mts
olliejenningsPT: 100% valid, i'm going to remove all of the test coverage stuff out of this, a recent PR introduced another version of `vitest` which completely broke all coverage since the `v8` coverage is not backwards compatible

.circleci/README.md
olliejenningsPT: Yeah true, the idea was for some information up until we do the final switch from GitHub to CircleCI then l wanted to delete this file

.circleci/scripts/is-affected.sh
olliejenningsPT: Will make it fail

turbo.json
olliejenningsPT: Fair, am too use to preferring explicit comments over none, willa ddress

.gitignore
olliejenningsPT: Will address, my mind set was in the future l wanted to move `apps` about of the `packages` folder into a separate `apps` or `e2e` folder to make the distinction clearer between things, and so l put it as `reports` for future

.circleci/config.yml
olliejenningsPT: It was, but due to a recent PR breaking it, l was slowly removing it last night, but l stopped. Since the last PR introduced another version of `vitest` that broke it, l then decided that its better to downgrade that version of vitest back to `3.x` and then re-introduce coverage here, so we have it stored on artifact. But now i actually want to remove coverage completely and do that in a FLUP to keep this smaller

.circleci/config.yml
olliejenningsPT: It is quite heavy, basically it was for adding reliability to the storybook interaction tests, somethign that blueprint had encountered and included in their CI.

I was going back and forth between keeping it and deleting it to be honest. So l might just delete it, and then if we can see if our storybook interaction tests are flakey or not.

.circleci/scripts/fork-guard.sh
olliejenningsPT: Your right, i will verify to make sure we do not have that turned on, was thinking more protecton, but yeah someone in theory could just deleting thi and it does nothing

olliejenningsPT: @ericjeney one thing we could do on-top, is something like `CODEOWNERS` whilst it does introduce some friction if there are certain files modified?

turbo.json
olliejenningsPT: Its a layer of safety incase in the future, so if we do have a shared cache, it still won't poison our cache on our CI. We also had a big issue with cache overlap in GH CI, so addition can ensure safety.

Yes it would also mean test jobs on CI would not be restorable locally, but l would consider that a fair trade off.

olliejenningsPT:
This should be replaced by the stacked PR: https://github.com/palantir/osdk-ts/pull/3861, will closed if we go foward with this new one

.circleci/scripts/fork-guard.sh
ericjeney: CODEOWNERS doesn't help here for the same reason this `fork-guard.sh` file wouldn't. The issue is when a build runs on the forked PR, not after the PR merges.

PR #3807, opened 2026-08-06
Populate ActionForm defaults from validation

nfdiop:
## Summary

- populate untouched fields from scalar defaults returned by action validation
- debounce validation and reject stale or submission-era responses
- preserve controlled values and explicit field defaults while disabling fields during submission

## Testing

- `pnpm turbo test --filter=@osdk/react-components`
- `pnpm turbo check-api --filter=@osdk/api --filter=@osdk/react-components`

### DIFF
--- a/packages/api/src/actions/ActionResults.ts
+++ b/packages/api/src/actions/ActionResults.ts
@@ -63,6 +63,7 @@ export interface ValidateActionResponseV2 {
       result: "VALID" | "INVALID";
       evaluatedConstraints: Array<ParameterEvaluatedConstraint>;
       required: boolean;
+      defaultValue?: any;
     }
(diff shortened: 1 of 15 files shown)


packages/api/src/actions/ActionResults.ts
```diff
@@ -63,6 +63,7 @@ export interface ValidateActionResponseV2 {
       result: "VALID" | "INVALID";
       evaluatedConstraints: Array<ParameterEvaluatedConstraint>;
       required: boolean;
+      defaultValue?: any;
```
nfdiop, line 66: staying consistent with the rest of the file but I'm tempted to use `unknown` instead

PR #3808, opened 2026-08-06
Generate widget set manifest authorizations defined in a resources.json file

gpirindev:
As part of the custom widget integration with PASS, users can define widges set authorizations in a resouces.json file in the widget set repo, e.g.

```json
{
  "authorizations": {
    "read": [["OrgA", "OrgB"], ["PII"]],
    "requiredRead": []
  },
  "version": 1
}

```

In this PR we are updating the widget set manifest build to also contain the defined authorizations.

### DIFF
--- a/packages/widget.vite-plugin/src/common/ResourcesJson.ts
+++ b/packages/widget.vite-plugin/src/common/ResourcesJson.ts
@@ -23,7 +23,7 @@ type DeclaredWidgetSetAuthorizations = Pick<
 
 interface ResourcesJsonV1 {
   authorizations: DeclaredWidgetSetAuthorizations;
-  version: number;
+  version: 1;
 }
 
 export type ResourcesJson = ResourcesJsonV1;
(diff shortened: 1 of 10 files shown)


packages/widget.vite-plugin/src/common/ResourcesJson.ts
```diff
@@ -14,15 +14,16 @@
...
-// They are illustrative code, not hand-written source, so they are exempt from
-// linting (the previous ESLint setup tolerated their patterns too). oxfmt still
-// formats them via `codegen`'s format step.
-export default defineConfig({
-  extends: [root],
-  ignorePatterns: [...(root.ignorePatterns ?? []), "src/examples/"],
-});
+interface ResourcesJsonV1 {
+  authorizations: DeclaredWidgetSetAuthorizations;
+  version: number;
```
amishshah:
```suggestion
  version: 1;
```

I think is what we want?

amishshah (approved):

PR #3815, opened 2026-08-07
Allow overriding labels in ActionForm

ziyunp:
Add a `labels` prop to `ActionForm`/`BaseForm` so every user-facing string can be overridden for localization or wording.

Add a shared createLabelsContext function to resolve labels through a React context (`LabelsProvider`). Updated ObjectTable to use this method.

Existing per-field text props (`submitButtonText`, `text`/`buttonText` on `FILE_PICKER`, `emptyMessage` on `OBJECT_SET`) still win over the corresponding label, so no behavior changes for current callers.

`ActionFormLabels` is exported from `@osdk/react-components/experimental`.


packages/react-components/src/action-form/ActionFormLabels.tsx
ziyunp, line 1: Review this file mainly to see if the label keys make sense

PR #3818, opened 2026-08-07
Remove dead and deprecated props from FilterList API

ziyunp:
Remove dead and deprecated props from FilterListApi. 

## filterClause - it is not used, no behavioral change

```ts
/** If provided, the filter clause is controlled. */
filterClause?: WhereClause<Q>;
```

Nothing read it. `useFilterListState` owns filter state unconditionally, so there is no controlled mode for the filter clause — a consumer who wired this up got a prop that silently did nothing, and the docs and stories actively taught the pattern. Removed, along with the "Required in controlled mode" claim on `onFilterClauseChanged` (which is the real, working way to read state out).

## Also removed, all unreferenced

- `FilterStatesMap` — zero usages, and its doc claimed object-identity keying while the runtime keys by string via `getFilterKey`
- `FilterState<Q>` and its `ExtractFilterState` helper — zero usages
- `FilterListItemProps` — exported publicly, but described props no component accepts; the real `FilterListItem` has its own local interface
- `BaseFilterListProps.hasVisibilityChanges` — already `@deprecated` in favour of `canReset`, along with its fallback branch in `FilterListHeader` and the `eslint-disable` that existed only to consume it

## `FilterStateType` alias collision

`FilterListApi` imported `FilterState as FilterStateType`, but `FilterListItemApi` exports a *real* `FilterStateType` — the state discriminator union (`"EXACT_MATCH" | "DATE_RANGE" | …`). So `onFilterStateChanged`'s signature rendered in hover text and in the generated props table as a type that resolves, for anyone who looks it up, to something entirely different. Now imported unaliased; the regenerated props table shows `FilterState`.

## Docs, e2e sandbox apps and stories

- The ObjectTable integration example no longer passes `filterClause`; state flows out through `onFilterClauseChanged` as it always actually did. It also gains the required `objectType` prop it was missing.
- The "use controlled mode for persistence" best practice pointed at a prop that never worked. Replaced with the mechanism that does: `onFilterStateChanged` + `getFilterKey` + `initialFilterStates`, with `serializeFilterStates` / `deserializeFilterStates` for the `Date` round-trip.
- Removed `filterClause` from 7 story render bodies, their displayed source snippets, and the argTypes table.
- Removed `filterClause` from sandbox apps

## Breaking

Removals from `@osdk/react-components/experimental/filter-list`. Anything passing `filterClause` or importing the removed types will need to drop them; no behavior changes, since none of it was wired to anything.


packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
olliejenningsPT, line 2225: the setter is never called now, so the json pane will always be empty

ziyunp: thanks for spotting this, fixed now

olliejenningsPT (approved):

PR #3820, opened 2026-08-10
Widen @osdk/integration-testing peer ranges to avoid a forced major bump

esinx:
Ran into an issue with peer dep `@osdk/seed-helper`'s range, bumping beta versions (`0.x.x`) with `workspace:*` triggers a major release which is unintended.

---

This is similar to what `@osdk/functions` does, but is not an issue since the version bump on `workspace:^` is on `@osdk/client`, which is already on a stable version (`2.x`). npm treats "magic numbered versions"(any version below 1, so `0.x`) differently. A caret dep on 0.1 would range from 0.1 to 0.2 whereas a caret dep on 1.0 would range from 1.0 to 2.0.

> semver.toComparators("^1.1")
[ [ '>=1.1.0', '<2.0.0-0' ] ]
> semver.toComparators("^0.1")
[ [ '>=0.1.0', '<0.2.0-0' ] ]

### DIFF
--- a/packages/integration-testing/package.json
+++ b/packages/integration-testing/package.json
@@ -55,9 +55,9 @@
     "undici": "^6.28.0"
   },
   "peerDependencies": {
-    "@osdk/api": "workspace:*",
-    "@osdk/client": "workspace:*",
-    "@osdk/seed-helpers": "workspace:*"
+    "@osdk/api": "workspace:^",
+    "@osdk/client": "workspace:^",
+    "@osdk/seed-helpers": ">=0.25.0 <1.0.0"
   },
   "devDependencies": {
     "@microsoft/api-documenter": "^7.26.32",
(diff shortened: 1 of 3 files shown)


packages/integration-testing/package.json
```diff
@@ -55,9 +55,9 @@
     "undici": "^6.28.0"
   },
   "peerDependencies": {
-    "@osdk/api": "workspace:*",
-    "@osdk/client": "workspace:*",
-    "@osdk/seed-helpers": "workspace:*"
+    "@osdk/api": "workspace:^",
+    "@osdk/client": "workspace:^",
+    "@osdk/seed-helpers": ">=0.25.0 <1.0.0"
```
esinx, line 60: This is necessary because semver resolves `^0.x` into minor-compatible, instead of major-compatible.

Example:
```
> semver.toComparators("^1.1")
[ [ '>=1.1.0', '<2.0.0-0' ] ]
> semver.toComparators("^0.1")
[ [ '>=0.1.0', '<0.2.0-0' ] ]
```

nihalbhatnagar (approved):

PR #3824, opened 2026-08-10
Fix ObjectTable dropping renderCell in edit mode for non-editable rows

ziyunp:
## Problem
The design of the API is to ignore renderCell when a cell is editable. This is to allow for the built-in editable components to be rendered, e.g. a dropdown in the cell. 

However, the cells are conditionally editable even when the column is configured to be editable. `useColumnDefs` passed the column's raw `editable` value straight to `shouldShowEditableCell`, which only tested `editable != null && editable !== false`. The two other call sites, `TableCell` and `DefaultCellRenderer`, already resolved the predicate correctly via `isCellEditable`.

## Fix

`shouldShowEditableCell` now takes the row and resolves `editable` itself:

```ts
export function shouldShowEditableCell<TData>(
  editable: EditablePredicate<TData> | undefined,
  object: TData,
  onCellEdit: unknown,
  isInEditMode: boolean | undefined,
): boolean {
  return isCellEditable(editable, object)
    && onCellEdit != null
    && isInEditMode === true;
}
```
## Test evidence
Senior Product row is editable and the other rows are not. The non-editable rows are rendered with renderCell() where the text is italic and grey

Screenshot 2026-08-19 at 18 59 56


packages/react-components-storybook/src/stories/ObjectTable/Features/Editing.stories.tsx
```diff
@@ -758,6 +758,17 @@ export const PerRowEditableAndFieldConfig: Story = {
...
         },
+        cellValueType: "string",
+        renderCell: (obj, _, value) => (
+          <span
+            style={{
+              fontStyle: "italic",
+              color: "grey",
+            }}
+          >
+            {value as string}
```
KuberSethi: could we avoid `value as string` here? `renderCell` does not guarantee that the value is a string, so this hides the mismatch instead of narrowing or formatting it safely.

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
```diff
@@ -766,6 +766,135 @@ describe(useColumnDefs, () => {
       expect(customRenderCell).not.toHaveBeenCalled();
     });
 
+    it("uses custom renderCell in edit mode for rows the editable predicate rejects", async () => {
+      const deferred = pDefer();
+      const fakeClient = {
+        fetchMetadata: vitest.fn(() => deferred.promise),
+      } as unknown as Client;
```
KuberSethi: could we avoid the new `as unknown as` casts here?

packages/react-components/src/object-table/components/NonEditableCellInEditMode.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Pads a read-only cell so it lines up with the editor cells beside it while
+ * the table is in edit mode. Without it, a column that mixes the two — an
+ * `editable` predicate that accepts some rows and rejects others — staggers,
+ * because `.osdkEditableCell` carries horizontal padding the raw content does
+ * not.
+ */
+export function NonEditableCellInEditMode({
+  children,
+}: React.PropsWithChildren): React.ReactElement {
+  return <span className={styles.nonEditableCellInEditMode}>{children}</span>;
```
KuberSethi: `renderCell` can return a `div`, so this wrapper can produce invalid `<span><div>…</div></span>`. could we use a `div` here instead?

packages/react-components/src/object-table/hooks/useColumnDefs.tsx
```diff
@@ -179,12 +180,23 @@ function getColumnsFromColumnDefinitions<
         > = cellContext.row.original;
 
         const meta = cellContext.table.options.meta;
-        const isEditable = shouldShowEditableCell<
-          Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>
-        >(editable, meta?.onCellEdit, meta?.isInEditMode);
+        const isEditable = shouldShowEditableCell(
```
KuberSethi: could we only check whether the cell is editable when `renderCell` is present, and check edit mode and the handler first? right now the same callback can run two or three times for one cell render.

packages/react-components-storybook/src/stories/ObjectTable/Features/Editing.stories.tsx
```diff
@@ -758,6 +758,17 @@ export const PerRowEditableAndFieldConfig: Story = {
           const jobTitle = rowData.jobTitle ?? "";
           return jobTitle === "Senior Product Manager";
         },
+        cellValueType: "string",
+        renderCell: (obj, _, value) => (
+          <span
+            style={{
```
KuberSethi, line 764: could we move this new inline style into the story styles?

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
```diff
@@ -766,6 +766,135 @@ describe(useColumnDefs, () => {
       expect(customRenderCell).not.toHaveBeenCalled();
     });
 
+    it("uses custom renderCell in edit mode for rows the editable predicate rejects", async () => {
```
KuberSethi: could these two new cases share one setup or use `it.each`? nearly all 129 lines are identical apart from the row and final expect.

packages/react-components-storybook/src/stories/ObjectTable/Features/Editing.stories.tsx
ziyunp: This is explicitly added here to show custom cell styling override in the code example

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
ziyunp: This is a partial implementation of the mock Client, should be fine to cast it like this in tests

packages/react-components/src/object-table/hooks/useColumnDefs.tsx
ziyunp: Moved this under if (renderCell) { ... }, also re-ordered the conditions in shouldShowEditableCell so isCellEditable is the last condition.

KuberSethi (approved):

PR #3828, opened 2026-08-10
FilterList: fire onFilterVisibilityChange in controlled mode; deprecate addFilterMode and the onFilterAdded newDefinitions arg

ziyunp:
## Summary

- `onFilterVisibilityChange` now fires on drag-reorder in controlled mode. Previously a drag in `addFilterMode="controlled"` fired no callback at all and the row snapped back, so reordering was impossible to observe or persist.
- `addFilterMode` is now **deprecated**. It is going away, and visibility will always be managed internally — today's `"uncontrolled"` behavior. To migrate off `"controlled"`: seed the initial visibility with `isVisible` on each `filterDefinitions` entry, and persist `onFilterVisibilityChange` to observe what the user shows, hides, and reorders.
- The second `newDefinitions` argument of `onFilterAdded` is **deprecated but still passed**. It is the caller's own unchanged `filterDefinitions`, never the post-add state. Read `onFilterVisibilityChange` instead, which reports visible-then-hidden filters in display order.
- Callback docstrings now describe the behavior directly instead of splitting it per mode, since the modes are going away. Reset is documented as a third `onFilterVisibilityChange` trigger, and `enableSorting` is documented `@default false`.

## API change

**No breaking changes.** Two deprecations, both still fully functional:

```diff
+ /** @deprecated Going away; visibility will always be managed internally. */
  addFilterMode?: "controlled" | "uncontrolled"

  onFilterAdded?: (
    filterKey: FilterKey<Q>,
+   /** @deprecated Use `onFilterVisibilityChange`. */
    newDefinitions: Array<FilterDefinitionUnion<Q>>,
  ) => void
```

`newDefinitions` is kept **required** on purpose. Making it optional would reject an existing `(key, defs: Array<…>) => void` handler under `strictFunctionTypes`. As written, both one-arg and two-arg handlers compile and the runtime value is unchanged.

The param-level `@deprecated` tag is intent-signalling only. TypeScript strikes a deprecated parameter through solely where it is referenced by name inside a function body, so a consumer writing `(key, defs) => …` — their own symbol — gets no diagnostic, and the tag does not reach their hover either. The `@param newDefinitions Deprecated. …` line above the signature is what consumers actually see.

## Tests

Three cases covering the callbacks, none of them mode-specific: add reports the shown filter visible (asserting `onFilterAdded` alongside `onFilterVisibilityChange`), remove reports it hidden (asserting `onFilterRemoved` too), and a drag reports the new order.

Also removes a test-ordering workaround whose diagnosis was wrong. The culprit was not base-ui but dnd-kit: `PointerSensor` suppresses the post-drag click with a capture-phase `document` listener torn down on a 50ms timer, so a synchronous test swallowed every later click in the file. The drag helper now waits it out.


packages/react-components/src/filter-list/__tests__/FilterList.test.tsx
```diff
@@ -160,4 +160,155 @@ describe("FilterList", () => {
...
+
+      fireEvent.click(screen.getByRole("button", { name: /add filter/iu }));
+      fireEvent.click(screen.getByRole("menuitem", { name: "dept" }));
+
+      expect(onFilterVisibilityChange).toHaveBeenCalledWith([
+        { filterKey: "dept", isVisible: true },
+      ]);
+    });
+
+    // The drag cases come last on purpose: synthetic pointer events leave
```
ziyunp: suggestion (non-blocking): good that you tracked this down and wrote it up, but it bakes in a declaration-order dependency — the next `it` someone appends to this describe breaks for a reason the comment above it doesn't explain. Could the leaked state be cleared in `afterEach` instead (a `pointercancel` on document, or whatever base-ui is holding), so ordering stops mattering?

packages/react-components/src/filter-list/FilterList.tsx
```diff
@@ -100,10 +101,7 @@ export function FilterList<Q extends ObjectTypeDefinition>(
     reorderVisible,
     hasVisibilityChanges,
     resetVisibility,
-  } = useFilterVisibility(
-    filterDefinitions,
-    uncontrolledAddFilterMode ? handleVisibilityChange : undefined,
```
ziyunp, line 105: changed to always call onFilterVisibilityChange regardless of modes

olliejenningsPT (approved):

PR #3830, opened 2026-08-10
Fix interface to object searcharound

ssanjay1:
Since we construct object sets synchronously, interface links that hop to objects are tricky. We have no knowledge that we our now in an object context, so when we construct another link, we type it as `interfacelinksearcharound` rather than just a `searchAround`. We don't have the data in codegen to disambiguate this, and we can't really add it without breaking use cases that rely on dynamic behavior, so we have to normalize this before we run any fetches instead.


packages/client/src/util/normalizeInterfaceLinkSearchArounds.wiring.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  link: "peeps",
+};
+
+describe("normalizeInterfaceLinkSearchArounds wiring", () => {
+  let client: Client;
+  let apiServer: SetupServer;
+
+  beforeAll(() => {
+    const testSetup = startNodeApiServer(new LegacyFauxFoundry(), createClient);
+    ({ client, apiServer } = testSetup);
```
esinx, line 67: nit: why not `const {client, apiServer} = startNodeApiServer(...)`

packages/client/src/util/normalizeInterfaceLinkSearchArounds.ts
```diff
@@ -0,0 +1,217 @@
...
+    // derived property (`MethodObjectSet` is an alias for `ObjectSet`).
+    case "withProperties": {
+      const child = await normalizeInterfaceLinkSearchArounds(
+        clientCtx,
+        objectSet.objectSet,
+      );
+      const derivedProperties = await normalizeDerivedProperties(
+        clientCtx,
+        objectSet.derivedProperties,
+      );
```
esinx, line 107: Nit: can we parallel await?
```typescript
const [child, derivedProperties] = await Promise.all([
      normalizeInterfaceLinkSearchArounds(
        clientCtx,
        objectSet.objectSet,
      ),
      normalizeDerivedProperties(
        clientCtx,
        objectSet.derivedProperties,
      )
])
```

packages/client/src/util/normalizeInterfaceLinkSearchArounds.ts
```diff
@@ -0,0 +1,217 @@
...
+          objectSet: child,
+          // `InterfaceLinkTypeApiName` and `LinkTypeApiName` are distinctly
+          // branded strings, but the underlying value is the same: the caller
+          // passed an object-type link name to `pivotTo` and it was placed in
+          // the wrong field. Only the field is wrong, not the name.
+          link: objectSet.interfaceLink as string,
+        };
+      }
+
+      return child === objectSet.objectSet
```
esinx, line 77: Why not always return `{ ...objectSet, objectSet: child }`?

packages/client/src/util/normalizeInterfaceLinkSearchArounds.ts
```diff
@@ -0,0 +1,217 @@
...
+    case "reference":
+    case "methodInput":
+      return objectSet;
+
+    case "interfaceLinkSearchAround": {
+      const child = await normalizeInterfaceLinkSearchArounds(
+        clientCtx,
+        objectSet.objectSet,
+      );
+      const sourceDef = await extractObjectOrInterfaceType(clientCtx, child);
```
williamhe723: NIT: `sourceDef` was confusing me for a while because I couldn't figure out what it was describing. IMO something like `producedType` makes more sense since it's describing what the inner object set produces

packages/client/src/util/normalizeInterfaceLinkSearchArounds.ts
```diff
@@ -0,0 +1,217 @@
...
+    // Leaves: no nested object sets.
+    case "base":
+    case "interfaceBase":
+    case "static":
+    case "reference":
+    case "methodInput":
+      return objectSet;
+
+    case "interfaceLinkSearchAround": {
+      const child = await normalizeInterfaceLinkSearchArounds(
```
williamhe723: nit: `child` is not very descriptive and took me a while to figure out exactly what it meant. Something more descriptive like `innerObjectSet` may be clearer to someone reading in the future

packages/client/src/util/normalizeInterfaceLinkSearchArounds.ts
williamhe723: This would create a new copy every single time, which I don't think is what we would want. Besides allocating memory where we don't need, in `fetchPage`, we check `objectSet === withArgs.objectSet` and if we had made a new copy for new reason, this would always return false

williamhe723 (commented):
nit: I feel like some of the comments are very long and can either be removed or shortened. For example, the one at the top of `normalizeInterfaceLinkSearchAround.ts` is very long and then every single place it is called has the exact same comments

PR #1021, opened 2024-11-26
Add expo template to create-app

ewitkon:
This PR Adds support for expo template to create-app.

Open issues to resolve 
- [x] We need the build command to work on the resulting template and mapped to `"build": "expo export -p web"`
- [x] Build is failing on the generated example
- [x] When we run create-app using this template the resulting @osdk/client is set to `"^2.1.0-beta.14"` and not to the one matching the client.
- [x] Test are failing as this is only supported in 2.x and tests for 1.x are failing

PR #3831, opened 2026-08-11
chore: migrate create-app/create-widget scaffold templates to oxlint/oxfmt (oxc increment 13)

olliejenningsPT:
## Summary

Completes the shipped-template half of the oxc migration (#3031). Every React and
TypeScript scaffold that `create-app` / `create-widget` emits now lints and formats
with oxlint + oxfmt via ultracite instead of ESLint and Prettier, matching the
`typescript-library` template that already shipped this way.

This is **user-facing**: it changes the tooling inside projects people scaffold.

Supersedes the stale draft #3604, which was written before #3431 refactored the
React beta template onto the shared `@osdk/create-app.react.beta.common` package.
That work is rebased here with the drift resolved.

## What changed

**8 template packages migrated** (each drops `templates/eslint.config.mjs`, gains
`templates/oxlint.config.ts` + `templates/oxfmt.config.ts`):

- `create-app.react.beta.common` (also drops `prettier.config.js`; serves the React beta template)
- `create-app.template.react`
- `create-app.template.tutorial-todo-app{,.beta}`
- `create-app.template.tutorial-todo-aip-app{,.beta}`
- `create-widget.template.react.v2`
- `create-widget.template.minimal-react.v2`

Generated `package.json` now exposes `lint` (`ultracite check`), `fix`
(`ultracite fix`) and `format` (`oxfmt .`), which is exactly the script set the
`typescript-library` template already ships, so all 10 templates are consistent.
The 9 corresponding apps under `examples/` are regenerated to match.

Because `lint` is now `ultracite check` rather than a bare `oxlint`, the examples'
`turbo lint` gained an `oxfmt --check` gate it did not have before. All of them pass it.

**React coverage preserved.** Ultracite's react preset does not enable everything the
old scaffold config did, so the template `oxlint.config.ts` re-enables
`react-hooks/rules-of-hooks` (error), `react-hooks/exhaustive-deps` (warn) and
`react/only-export-components` (warn, `allowConstantExport`).

**`rules-of-hooks` enabled repo-wide** at error in the root `oxlint.config.ts`. The
prior ESLint config had it at warn and the oxc migration had dropped it, so this is
stricter than before. Every package is already clean. Story files in
`react-components-storybook` are carved out in that package's existing story override
block: Storybook mounts `render: (args) => {...}` as a component, but the rule's
heuristic sees an anonymous arrow on an object property and reports 51 false positives.
Real components in that package still get the rule.

**Expo drops its dead ESLint tree.** `create-app.template.expo.v2` shipped 14 ESLint
packages to every scaffolded project while having no lint config and no lint script,
and never having had either (verified with `git log --diff-filter=AD` and `git log -S`).
They arrived in #1021, whose own commit message includes the step "remove lint from
expo". 11 of the 14 were copy-pasted from the React web template
(`eslint-plugin-react-refresh` is a Vite HMR plugin; `jsx-a11y` is DOM accessibility).
`eslint-plugin-prettier` shipped without `prettier`, so its required peer was
permanently unmet.

Expo is not migrated to oxlint in this PR, but that is a scheduling choice rather than
a technical blocker: with the same carve-outs the other templates ship plus the root
config's cosmetic policy, its scaffold measures 46 findings (about 25 needing hand
edits), comparable to what the other 8 templates required. Only 14 findings were
genuinely React Native specific (`require()` for Metro assets). Worth a follow-up.

**Repo wiring:**

- `create-app.react.beta.common` added to `dprint.json` excludes and
  `.lintstagedrc.mjs` `OXC_PACKAGES`. It was in an oxc monorepolint archetype but
  missing from both, so lint-staged would run `dprint fmt` on a package whose CI runs
  `oxfmt --check`.
- Template oxc pins aligned to root (`oxfmt ^0.53.0`, `oxlint ^1.68.0`,
  `ultracite ^7.8.2`). All 8 were on `0.52.0`/`1.67.0`/`7.8.0`, which split the
  lockfile across two oxfmt versions. That is the cross-platform nondeterminism that
  broke CI during increment 12, since pnpm can hoist a different oxfmt binary on Linux
  than on macOS. The lockfile now resolves a single `oxfmt@0.53.0`.

## User-facing impact

- Scaffolded React projects get `pnpm lint` / `pnpm fix` / `pnpm format` backed by oxc
  and no longer install the ESLint or Prettier dependency trees. `pnpm lint` now checks
  formatting as well as lint rules.
- Scaffolded Expo projects no longer ship 14 unused ESLint packages. `npx expo lint`
  still works and will now prompt to install ESLint rather than finding it
  pre-installed. That is the CLI's intended flow, and the pre-installed copy was
  unusable anyway: Expo requires `eslint@^8.57.0` and the template pinned `^9.35.0`, so
  its own dependency check rejected it.

## Notable decisions

- **Vue templates unchanged.** oxlint's Vue support is still in early stages (the `vue`
  plugin is off by default and covers only rules that work with script tags), so those
  scaffolds stay as they are for now.
- **oxc configs for React beta live in the shared common package**, not in
  `create-app.template.react.beta`, mirroring where the eslint/prettier configs lived.
  The packager collects shared templates first and lets the consumer override, so any
  future template on that base inherits them.

## Test plan

- [x] `example-generator --check`: all examples byte-match their templates
- [x] `ultracite check` in each of the 10 migrated scaffolds: exit 0
- [x] `turbo build` for all examples (`tsc && vite build`): 69/69. This is the gate that
      matters, since lint autofixes can emit APIs newer than the ES2020 target
- [x] Expo builds: 43/43, confirming nothing consumed the removed deps
- [x] `turbo lint` whole repo: 226/226, with examples now on the stricter `ultracite check`
- [x] `turbo typecheck` for expo template and both scaffolds: 27/27
      (`@types/eslint-plugin-jsx-a11y` was among the removals)
- [x] `turbo test` for example-generator, create-app, create-widget: 32/32
- [x] `mrl check`: clean
- [x] `cspell` on all changed files: clean
- [x] Single `oxfmt` version in the lockfile
- [x] `rules-of-hooks` verified live with a deliberate conditional-hook probe, and
      verified still active for non-story files in storybook
- [x] Rebased onto current `origin/main` (2.56.0 version bump) and fully re-verified,
      since PR CI builds the merge commit

## Out of scope

Remaining for #3031: the global config flip, dprint removal, and the 9 generator
packages plus `examples-extra/docs_example`. The generators are blocked on type-aware
linting (`restrict-template-expressions` guards their template-literal codegen), which
would also close the repo-wide type-aware gap.


nfdiop (commented):
Can we use oxlint/oxfmt without any code changes (for example the `rules-of-hooks`) and we can add it later to minimize changes?
We also want to have `trailingComma: all` in the config

examples/example-react-sdk-2.x-no-osdk/.vscode/settings.json
```diff
@@ -1,7 +1,6 @@
 {
   "editor.formatOnSave": true,
-  "editor.defaultFormatter": "esbenp.prettier-vscode",
```
nfdiop, line 3: this is nice!

examples/example-expo-sdk-2.x-no-osdk/package.json
```diff
@@ -55,31 +55,17 @@
   },
   "devDependencies": {
     "@babel/core": "^7.28.4",
-    "@eslint/compat": "^1.3.2",
```
nfdiop, line 58: nice that we get to drop so many packages

examples/example-tutorial-todo-aip-app-sdk-2.x/src/DeleteProjectButton.tsx
```diff
@@ -1,7 +1,8 @@
 import { useCallback, useState } from "react";
 import css from "./DeleteProjectButton.module.css";
 import DeleteProjectDialog from "./DeleteProjectDialog";
-import useProjects, { IProject } from "./useProjects";
+import type { IProject } from "./useProjects";
```
nfdiop: Can we have a single import line? You can use:

```
"typescript/consistent-type-imports": [
        "error",
        { fixStyle: "inline-type-imports" },
      ],
```

examples/example-react-sdk-1.x/package.json
```diff
@@ -10,7 +10,9 @@
   "scripts": {
     "build": "tsc && vite build",
     "dev": "vite",
-    "lint": "eslint . --report-unused-disable-directives --max-warnings 0",
+    "fix": "ultracite fix",
+    "format": "oxfmt .",
+    "lint": "ultracite check",
```
nfdiop: Do we want to keep `--report-unused-disable-directives-severity` or not required here?

nfdiop (commented):
Thanks for doing this! Left some small comments/questions.

Assuming this is also fine to go ahead with since we don't support node 18 anymore (since we added Vite 7)

examples/example-tutorial-todo-aip-app-sdk-2.x/src/DeleteProjectButton.tsx
olliejenningsPT: The reason for splitting them up to enabled better dead code elimination / tree shaking, it makes it easier for the bundler this way, and is recommended these days.

examples/example-react-sdk-1.x/package.json
olliejenningsPT: Yeah we can add it with as an error, we have zero overrides in the templates so didn't bother with this, but let me add it

examples/example-tutorial-todo-aip-app-sdk-2.x/src/DeleteProjectButton.tsx
nfdiop: Hm, I am not sure it is recommended anymore. 
- TS explicitly added support for this https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names 
- Internally we also recommend a single import. 
- I would expect this to work fine with Vite

examples/example-react-sdk-2.x-no-osdk/package.json
```diff
@@ -10,9 +10,9 @@
   "scripts": {
     "build": "tsc && vite build",
     "dev": "vite",
-    "format": "prettier --write src/",
-    "lint": "eslint . --report-unused-disable-directives --max-warnings 0",
-    "lint:fix": "eslint --report-unused-disable-directives --max-warnings 0 --fix",
+    "fix": "ultracite fix",
+    "format": "oxfmt .",
+    "lint": "ultracite check --report-unused-disable-directives-severity=error",
```
nfdiop, line 15: Should it be --report-unused-disable-directives to keep it as a warning like before?

examples/example-tutorial-todo-aip-app-sdk-2.x/src/DeleteProjectButton.tsx
olliejenningsPT: It still is recommended:

* TS added this in 5.0: https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax
* There are 3 different ways of doing type imports mixed with normal imports, whilst bundlers can handle this most of the time, you do risk having side-effects as not even all bundlers can ensure stripping this perfectly
* Modern bundlers such as rolldown have this defaulted on https://rolldown.rs/reference/Interface.RolldownOptions

Further reading links:

* https://betterstack.com/community/guides/scaling-nodejs/ts-verbatimmodulesyntax/#prerequisites
* https://mergify.com/blog/typescripts-import-type-the-flag-that-makes-builds-honest
* https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/

examples/example-react-sdk-2.x-no-osdk/package.json
olliejenningsPT: it was never a warning before, so i've kept the behaviour identical

examples/example-tutorial-todo-aip-app-sdk-2.x/src/DeleteProjectButton.tsx
nfdiop: Wait, `verbatimModuleSyntax` supports `import { useProjects,  type IProject } from "./useProjects"`: not sure why it is mentioned?

nfdiop: If we worry about side-effects of `import {type IProject} from "..."` we can enable "typescript/no-import-type-side-effects": "error"

olliejenningsPT: True it does, but since we don't even use it, we have 2 choices:

* We can keep the separate line for type imports (as eslint recommends) so it keeps type imports clear, 
* or back to single line imports, enable flag above, and enable eslint rules to catch it on one line (but since we are moving to oxc, it made sense to just keep this as separate imports - following eslint + oxc default for this).

If you really want to keep the single line imports (but its not the recommended or standard way i've seen it in industry), then we go with option 2

examples/example-react-sdk-2.x-no-osdk/package.json
nfdiop: My bad. Assumed it meant warning

nfdiop (dismissed):
Approving. Before we go ahead, let's make sure to generate an example with one of the templates and test it works as expected

PR #3833, opened 2026-08-11
Add getCellValue to ObjectTable column definitions

ziyunp:
## Problem

For a `type: "custom"` column, it names nothing on the object, so its value is always `undefined`. Consumers work around this with `renderCell`, which paints something on screen but leaves the column with no *value* — so the cell editor has nothing to seed from, `CellEditInfo.oldValue` is `undefined`, and `renderCellContextMenu` receives `undefined`.

## Change

**`getCellValue?: (object, locator) => unknown`** on the column definition, wired as tanstack's `accessorFn` instead of `accessorKey`:

```ts
const accessor = getCellValue
  ? { accessorFn: (object) => getCellValue(object, locator) }
  : { accessorKey: colKey };
```

**`renderCell` gains a third `value` argument**, so a custom renderer can use what the table already derived instead of recomputing it. Additive and positional — existing two-arg callbacks are unaffected.

**`cellValueType`** is added to column definition so users can pass in the data type for custom column's value

**Default RDPs to {}** in the object-table types, matching @osdk/api; Record<string, never> is unassignable to it, which broke columnDefinitions + filter once ColumnDefinition became a union.

### Function columns are excluded at the type level

Function-backed columns already merge an `AsyncCellData` wrapper (`{ isLoading, data, error }`) onto the row and have their own `locator.getValue` for extracting from the query result while preserving those states. A column-level `getCellValue` would replace the wrapper outright and silently drop loading and error rendering.

`ColumnDefinition` is now a four-member union pairing editability against value source, with `getCellValue?: never` on the function variants:

```ts
export type ColumnDefinition<Q, RDPs, FunctionColumns> =
  | EditableColumnDefinition<Q, RDPs, FunctionColumns>          // derivable locator + getCellValue
  | EditableFunctionColumnDefinition<Q, RDPs, FunctionColumns>  // getCellValue?: never
  | ReadonlyColumnDefinition<Q, RDPs, FunctionColumns>
  | ReadonlyFunctionColumnDefinition<Q, RDPs, FunctionColumns>;
```

### Test evidence

Reports To and Contact are custom columns

https://github.com/user-attachments/assets/05b69583-fa7d-4cf3-90d6-ce742003e969


packages/react-components-storybook/src/stories/ObjectTable/Features/Editing.stories.tsx
```diff
@@ -815,3 +816,69 @@ return (
     );
   },
 };
+
+export const CustomEditableColumns: Story = {
```
ziyunp: Instead of a new story, add a custom column to the existing editable stories

packages/react-components/src/object-table/hooks/__tests__/useColumnDefs.test.tsx
```diff
@@ -420,10 +423,227 @@ describe(useColumnDefs, () => {
...
+      deferred.resolve(mockMetadata);
+
+      await waitFor(() => {
+        expect(result.current.loading).toBe(false);
+      });
+
+      expect(result.current.columns[0]?.meta?.dataType).toBe("integer");
+    });
+
+    it("prefers an explicit dataType over the property's metadata type", async () => {
```
nfdiop, line 537: nice test

packages/react-components/src/object-table/ObjectTableApi.ts
```diff
@@ -75,6 +118,14 @@ interface SharedColumnDefinition<
   resizable?: boolean;
   orderable?: boolean;
 
+  /**
+   * The value's wire type. Picks the default editor and how committed values
+   * are parsed. Property columns read it from ontology metadata; custom and
+   * derived columns have none, so without it they get a text input that
+   * commits strings.
+   */
+  dataType?: BaseWirePropertyTypes;
```
nfdiop: Instead of having this type here, why not add it to the custom locator type since it is only needed for those cases?

ziyunp: Good suggestion. I put it here because having getCellValue now makes it possible to return another data type. For example, I have a status where the data type is number (0, 1), but getCellValue returns it as (OFF, ON). If I want to make this cell editable, I will probably want the data type to be string. 

Maybe I should rename this as cellValueType to match getCellValue

nfdiop: Makes sense. What if you tried something like this so users have to pass the valueType when they give us getCellValue?

```

  type CellValueOverride<Q> = {
    [T in BaseWirePropertyTypes]: {
      cellValueType: T;
      getCellValue: (object: Osdk.Instance<Q>) =>
        | PropertyValueWireToClient[T]
        | null
        | undefined;
    };
  }[BaseWirePropertyTypes];
```

nfdiop (approved):

PR #3799, opened 2026-08-04
Use spreadsheet instead of Excel

nfdiop:
### What is done in this PR
- Rename ExcelViewer to SpreadsheetViewer (+ all Excel occurences)

### Impact
This is a breaking change (unavoidable).

### Test
Storybook still working https://palantir.github.io/osdk-ts/storybook/pr-3799/c23e08252e68b7b14717bf8d6fe11cc124ec403f/?path=/story/components-documentviewer--spreadsheet

PR #3834, opened 2026-08-11
Normalize viewer component naming

olliejenningsPT:
Renames the TIFF and markdown viewers so every viewer follows the `<X>Viewer` / `Base<X>Viewer` pattern already used by the other document viewers, and aligns the pdf-viewer file names with the same shape. Follows the same approach as #3799 (`ExcelViewer` to `SpreadsheetViewer`).

## Import paths are unchanged

`experimental/markdown-renderer` and `experimental/tiff-renderer` keep their current names, now exporting the renamed components. They get renamed when the `experimental/` prefix is dropped, so consumers change import paths once rather than twice, per @ziyunp's suggestion.

Each of those subpaths also keeps `@deprecated` aliases of the old component names, pointing at the exact same values, so no component import needs to change yet:

```
MarkdownRenderer      -> BaseMarkdownViewer
MarkdownViewerMedia   -> MarkdownViewer
MarkdownRendererProps -> BaseMarkdownViewerProps
TiffRenderer          -> BaseTiffViewer
TiffViewerMedia       -> TiffViewer
TiffRendererProps     -> BaseTiffViewerProps
```

Those aliases are also reachable from the aggregate `@osdk/react-components/experimental` entry point, so both naming generations work there too.

`DocumentViewer`'s `tiffRendererProps` and `usePdfViewerInstance`'s `highlightEnabled` are likewise kept as deprecated aliases.

## Breaking changes

**Prop types drop the `Media` suffix, with no aliases.** All eight: `EmailViewerMediaProps`, `ImageViewerMediaProps`, `MarkdownViewerMediaProps`, `PdfViewerMediaProps`, `SpreadsheetViewerMediaProps`, `TiffViewerMediaProps`, `VideoViewerMediaProps`, `XmlViewerMediaProps` become `<X>ViewerProps`. This is deliberately uniform rather than partially deprecated, because the pdf one cannot be aliased.

**`PdfViewerProps` changes meaning.** It used to describe `BasePdfViewer` (taking `src`) and now describes `PdfViewer` (taking `media`); the old type is `BasePdfViewerProps`. Pdf is the only viewer affected, because it was the only one whose base props were not already `Base`-prefixed, so its old name is the only one the `Media` drop recycles. There is no way to alias it, since the name is occupied. Old usage fails to compile rather than silently changing behavior, since `src` and `media` are not interchangeable.

**Four `DocumentViewer` prop bags are removed:** `markdownViewerProps`, `spreadsheetViewerProps`, `emailViewerProps`, `xmlViewerProps`. Each resolved to `{}` and could never carry a value, because those Base props consist only of the primary input plus `className`, both of which `DocumentViewer` omits when forwarding.

**CSS tokens** `--osdk-markdown-renderer-*` become `--osdk-markdown-viewer-*`, with no fallback, so consumer themes need updating. Component names reported to metrics change from `TiffViewerMedia` and `MarkdownViewerMedia` to `TiffViewer` and `MarkdownViewer`, so dashboards keyed on the old names need updating.

## Structure

Each viewer now matches the layout the other viewers already use, with both prop interfaces living in `<X>ViewerApi.ts`:

```
markdown-viewer/    BaseMarkdownViewer.tsx  MarkdownViewer.tsx  MarkdownViewerApi.ts
images/tiff-viewer/ BaseTiffViewer.tsx      TiffViewer.tsx      TiffViewerApi.ts
pdf-viewer/         BasePdfViewer.tsx       PdfViewer.tsx       PdfViewerApi.ts
```

`pdf-viewer` previously had `BasePdfViewer` living in `PdfViewer.tsx` and the media wrapper in `PdfRenderer.tsx`, so those two swapped contents. Git renders that as a large add plus delete rather than a rename, but no behaviour changed: each moved file is byte-identical to its previous source once the renames are applied.

`PdfViewerInstanceOptions` is now derived from `BasePdfViewerProps` rather than restating it, so the two cannot drift. They already had: the options type spelled the highlight flag `highlightEnabled` and omitted `downloadFileName` entirely. `enableHighlight` is now the spelling, with `highlightEnabled` kept as a deprecated alias, and `downloadFileName` stays excluded because the hooks tier never implemented it.

## Docs

`docs/PdfViewer.md` picks up fixes for pre-existing inaccuracies the rename surfaced:

- `annotations` was documented as `Record<number, PdfAnnotation[]>`. It is a flat `PdfAnnotation[]` where each annotation carries its own `page`, so the usage example in that doc would not have compiled.
- The `BasePdfViewerProps` table covered 10 of the 19 props. It now covers all 19.
- `AnnotationType` omitted `"custom"`, hiding the custom-renderer path behind `PdfCustomAnnotation`.

`CLAUDE.md` and `CONTRIBUTING.md` no longer claim base props live inline in `Base<Name>.tsx`, since every viewer declares them in `<X>ViewerApi.ts`.

Converting the hand-rolled viewer props tables to `AUTOGEN:props` is deferred to a follow-up, as agreed with @ziyunp.

## Testing

Rebased onto main. `typecheck`, `test` and `lint` pass for `@osdk/react-components` and `@osdk/react-components-storybook` (1433 tests). `check-gen-props`, `cspell`, `mrl check`, dprint and oxfmt are clean.

New tests cover the deprecated `tiffRendererProps` alias, including through the `enableTiffToPdf` path, which the compiler cannot check because the original destructured prop stays in scope. Verified separately that both subpaths resolve to real build output for all four export conditions and hand out object-identical components for the old and new names, that the renamed CSS tokens reach the bundled `styles.css`, and that each moved component file is byte-identical to its pre-PR source modulo the intended renames.


packages/react-components/src/public/experimental/pdf-viewer.ts
```diff
@@ -110,10 +110,10 @@ export {
...
 // PdfViewer (Media wrapper)
-import { PdfViewer as _PdfViewer } from "../../pdf-viewer/PdfRenderer.js";
+import { PdfViewer as _PdfViewer } from "../../pdf-viewer/PdfViewer.js";
 import { withOsdkMetrics } from "../../util/withOsdkMetrics.js";
 export const PdfViewer: typeof _PdfViewer = withOsdkMetrics(
   _PdfViewer,
   "PdfViewer",
 );
-export type { PdfViewerMediaProps } from "../../pdf-viewer/PdfRenderer.js";
+export type { PdfViewerMediaProps } from "../../pdf-viewer/PdfViewerApi.js";
```
ziyunp: Do we want the Media suffix here?

packages/react-components/src/document-viewer/DocumentViewerApi.ts
```diff
@@ -48,20 +48,32 @@ export interface DocumentViewerProps {
...
    * @default undefined */
   videoViewerProps?: Partial<Omit<BaseVideoViewerProps, "src" | "className">>;
-  /** Props forwarded to TiffRenderer when rendering TIFF
+  /** Props forwarded to BaseTiffViewer when rendering TIFF
    * @default undefined */
-  tiffRendererProps?: Partial<Omit<TiffRendererProps, "content" | "className">>;
-  /** Props forwarded to MarkdownRenderer when rendering markdown
+  tiffViewerProps?: Partial<Omit<BaseTiffViewerProps, "content" | "className">>;
+  /** Alias for `tiffViewerProps`, which takes precedence if both are set.
+   * @deprecated Use `tiffViewerProps` instead.
```
lucy-k-wang, line 58: these two are the only ones marked `@deprecated` in the package, and [#3799](https://github.com/palantir/osdk-ts/pull/3799) renamed `excelViewerProps` to `spreadsheetViewerProps` on this same component with no alias.

do we want aliases here? I'm slightly toward dropping them directly because of the above precedent, but can be missing context here.

packages/react-components/src/document-viewer/DocumentViewerApi.ts
```diff
@@ -48,20 +48,32 @@ export interface DocumentViewerProps {
...
   /** Props forwarded to BaseVideoViewer when rendering video
    * @default undefined */
   videoViewerProps?: Partial<Omit<BaseVideoViewerProps, "src" | "className">>;
-  /** Props forwarded to TiffRenderer when rendering TIFF
+  /** Props forwarded to BaseTiffViewer when rendering TIFF
    * @default undefined */
-  tiffRendererProps?: Partial<Omit<TiffRendererProps, "content" | "className">>;
-  /** Props forwarded to MarkdownRenderer when rendering markdown
+  tiffViewerProps?: Partial<Omit<BaseTiffViewerProps, "content" | "className">>;
+  /** Alias for `tiffViewerProps`, which takes precedence if both are set.
```
lucy-k-wang: nit
```suggestion
  /** Alias for `tiffViewerProps`, which takes precedence when set to a non-nullish value.
```

packages/react-components/src/document-viewer/DocumentViewerApi.ts
```diff
@@ -48,20 +48,32 @@ export interface DocumentViewerProps {
...
+   * @deprecated Use `tiffViewerProps` instead.
+   * @default undefined */
+  tiffRendererProps?: Partial<
+    Omit<BaseTiffViewerProps, "content" | "className">
+  >;
+  /** Props forwarded to BaseMarkdownViewer when rendering markdown
+   * @default undefined */
+  markdownViewerProps?: Partial<
```
lucy-k-wang: `BaseMarkdownViewerProps` only has `content` and `className` and we omit both, so this resolves to `{}`.  the doc says "Props forwarded to BaseMarkdownViewer when rendering markdown" which overpromises. let's drop both markdown props until `BaseMarkdownViewer` actually has something forwardable.

packages/react-components/docs/PdfViewer.md
```diff
@@ -73,22 +71,31 @@ import { BasePdfViewer } from "@osdk/react-components/experimental/pdf-viewer";
...
-| `initialSidebarOpen` | `boolean`                                      | `false`        | Whether the sidebar is initially open                           |
-| `sidebarMode`        | `SidebarMode`                                  | `"thumbnails"` | Which sidebar panel to show: `"thumbnails"` or `"outline"`      |
-| `outlineIcons`       | `Partial<Record<number, React.ComponentType>>` | —              | Custom icon components for each outline depth level (0-indexed) |
-| `enableDownload`     | `boolean`                                      | `false`        | Whether the download button is shown in the toolbar             |
-| `className`          | `string`                                       | —              | CSS class applied to the root element                           |
+Plus all props from `BasePdfViewerProps` except `src`.
+
+### BasePdfViewerProps
+
+| Prop                 | Type                                                    | Default        | Description                                                                                                           |
```
lucy-k-wang, line 78: this table is hand-rolled and can drift easily. `DocumentViewer.md`,  `ActionForm.md` etc. the generator, but none of the viewer docs do. Here or in a flup when we split the docs out, can we wrap it in a marker block to autogen here as well?

packages/react-components/src/images/tiff-viewer/TiffViewerApi.ts
```diff
@@ -14,9 +14,21 @@
...
+import type { Media } from "@osdk/api";
+
+export interface BaseTiffViewerProps {
   /** TIFF bytes to render */
   content: Uint8Array;
   /** Callback fired when rendering fails */
   onError?: () => void;
 }
+
+export interface TiffViewerMediaProps extends Omit<
```
lucy-k-wang: should we drop `Media` from []Props naming too? can be a flup

packages/react-components/AGENTS.md
```diff
@@ -42,8 +42,8 @@ Components are imported from their individual entry points under `@osdk/react-co
 - `@osdk/react-components/experimental/filter-list` — FilterList, BaseFilterList
 - `@osdk/react-components/experimental/action-form` — ActionForm, BaseForm, and form field definitions
 - `@osdk/react-components/experimental/pdf-viewer` — PdfViewer, BasePdfViewer, and building blocks/hooks
-- `@osdk/react-components/experimental/tiff-renderer` — TiffRenderer
-- `@osdk/react-components/experimental/markdown-renderer` — MarkdownRenderer
+- `@osdk/react-components/experimental/tiff-viewer` — TiffViewer, BaseTiffViewer
```
lucy-k-wang: while we're here: CONTRIBUTING.md:142 and CLAUDE.md:96 both still say "Base props live inline in `Base<Name>.tsx`", but all viewers put them in `<X>ViewerApi.ts` now. let's fix those two lines too for consistency

packages/react-components/src/document-viewer/DocumentViewerApi.ts
olliejenningsPT: Am happy to drop them, l think its part of us to trying to avoid to much drift / change from now until GA

packages/react-components/src/document-viewer/DocumentViewerApi.ts
olliejenningsPT: Happy to, we actually have this for a couple more (email viewer, xml viewer etc). The idea was to leave a path for adding props to, but we can just remove them for now completely

packages/react-components/docs/PdfViewer.md
olliejenningsPT: So me and @ziyunp agreed to auto gen all in a FLUP

packages/react-components/src/public/experimental/pdf-viewer.ts
olliejenningsPT: We don't and we can remove

packages/react-components/src/document-viewer/DocumentViewerApi.ts
ziyunp: Yes I've been removing dead props in prep for GA as it would be harder to remove props later

packages/react-components/src/document-viewer/DocumentViewerApi.ts
ziyunp: We were assuming there is less usage of spreadsheet viewer, so we got the breaking change in. There are many planned API changes for GA, and we want to do it in the least disruptive way, especially for props. We plan to mark old props with @deprecated first, but we won't keep it for as long as 6 months. We will prepare a codemod script to help with the migration, and then only we clear all deprecated props in one go.

.monorepolint.config.mjs
```diff
@@ -583,11 +583,13 @@ const archetypeRules = archetypes(standardPackageRules, {
         "./experimental/filter-list",
         "./experimental/image-viewer",
         "./experimental/markdown-renderer",
+        "./experimental/markdown-viewer",
```
ziyunp: Question: What do you think if we leave the import path as-is, and change the import path when we remove the experimental?

olliejenningsPT: We can, my concern would be having different exports from old paths, e.g. it would feel weird (change the export name but path doesn't change, then another follow up where the path name changes - yes we are removing experimental but also more).

.changeset/normalize-viewer-component-naming.md
```diff
@@ -0,0 +1,13 @@
+---
+"@osdk/react-components": minor
+---
+
+Normalize viewer component naming so every viewer follows the `<X>Viewer` / `Base<X>Viewer` pattern: `MarkdownViewerMedia`/`TiffViewerMedia` are now `MarkdownViewer`/`TiffViewer`, `MarkdownRenderer`/`TiffRenderer` are now `BaseMarkdownViewer`/`BaseTiffViewer`, `MarkdownRendererProps`/`TiffRendererProps` are now `BaseMarkdownViewerProps`/`BaseTiffViewerProps`, and the `experimental/markdown-viewer` and `experimental/tiff-viewer` subpaths replace `experimental/markdown-renderer` and `experimental/tiff-renderer`.
+
+The old subpaths keep working and are marked deprecated. `experimental/markdown-renderer` and `experimental/tiff-renderer` re-export the very same component instances under their previous names, so existing imports from those paths continue to work unchanged. `DocumentViewer`'s `tiffRendererProps` is likewise kept as a deprecated alias for `tiffViewerProps`, taking precedence when set to a non-nullish value.
```
ziyunp: nit: Can we make a list of migration instructions here, something like old -> new:
```
EmailViewerMediaProps -> EmailViewerProps
DocumentViewer's markdownViewerProps -> delete markdownViewerProps

import { MarkdownRenderer } from "@osdk/react-components/experimental" -> import { MarkdownViewer } from "@osdk/react-components/experimental"
```

packages/react-components/src/public/experimental.ts
```diff
@@ -20,11 +20,11 @@ export * from "./experimental/document-viewer.js";
 export * from "./experimental/email-viewer.js";
 export * from "./experimental/filter-list.js";
 export * from "./experimental/image-viewer.js";
-export * from "./experimental/markdown-renderer.js";
+export * from "./experimental/markdown-viewer.js";
```
ziyunp: I think this breaking change is avoidable too. Let's duplicate the exports so it's consistent

ziyunp (commented):
.

packages/react-components/src/pdf-viewer/PdfViewerApi.ts
```diff
@@ -117,7 +118,7 @@ export interface PdfViewerHandle {
 
 /**
  * Options for {@link usePdfViewerInstance}.
- * Equivalent to {@link PdfViewerProps} minus the `className` rendering concern.
+ * Equivalent to {@link BasePdfViewerProps} minus the `className` rendering concern.
```
ziyunp: It says `Equivalent to {@link BasePdfViewerProps}` but they are maintained separately. Seeing differences like highlightEnabled vs enableHighlight. Can we do Omit<BasePdfViewerProps, "className">?

ziyunp (approved):

PR #3839, opened 2026-08-11
Support dynamic query mocks in unit-testing

esinx:
Adds a dynamic-implementation overload to `MockClient.whenQuery`:

```ts
client.whenQuery(myQuery, (params) => ({ result: params.a + params.b }));
```

The implementation receives the caller's params and computes the result; it may be async. The existing `whenQuery(query, params).thenReturn(...)` / `.thenThrow(...)` form is unchanged, and query stubs still resolve last-registered-first. `IntegrationClient.whenQuery` inherits the overload.

### DIFF
--- a/etc/unit-testing.report.api.md
+++ b/etc/unit-testing.report.api.md
@@ -65,9 +65,9 @@ export interface MockClient extends Client {
     	>(objectSet: ObjectSet<Q>, callback: ObjectSetStubCallback<Q, T>): StubBuilderFor<T>;
     	// Warning: (ae-forgotten-export) The symbol "QueryParamsFromDef" needs to be exported by the entry point index.d.ts
     // Warning: (ae-forgotten-export) The symbol "QueryReturnTypeFromDef" needs to be exported by the entry point index.d.ts
-    //
-    // (undocumented)
     whenQuery<Q extends QueryDefinition>(query: Q, params?: QueryParamsFromDef<Q>): QueryStubBuilder<QueryReturnTypeFromDef<Q>>;
+    	// Warning: (ae-forgotten-export) The symbol "QueryImplFromDef" needs to be exported by the entry point index.d.ts
+    whenQuery<Q extends QueryDefinition>(query: Q, impl: QueryImplFromDef<Q>): void;
 }
 
 // @public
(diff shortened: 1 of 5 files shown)


etc/unit-testing.report.api.md
```diff
@@ -65,9 +65,9 @@ export interface MockClient extends Client {
     	>(objectSet: ObjectSet<Q>, callback: ObjectSetStubCallback<Q, T>): StubBuilderFor<T>;
     	// Warning: (ae-forgotten-export) The symbol "QueryParamsFromDef" needs to be exported by the entry point index.d.ts
     // Warning: (ae-forgotten-export) The symbol "QueryReturnTypeFromDef" needs to be exported by the entry point index.d.ts
-    //
-    // (undocumented)
     whenQuery<Q extends QueryDefinition>(query: Q, params?: QueryParamsFromDef<Q>): QueryStubBuilder<QueryReturnTypeFromDef<Q>>;
+    	// Warning: (ae-forgotten-export) The symbol "QueryImplFromDef" needs to be exported by the entry point index.d.ts
```
esinx, line 69: Not sure if we really want to do this, surfacing this on an unstable API might be undesirable

packages/unit-testing/src/mock/createMockClient.ts
```diff
@@ -167,27 +167,37 @@ export function createMockClient(): MockClient {
...
+      queryStubs.push({
+        queryApiName: query.apiName,
+        type: "impl",
+        impl: params as (params: unknown) => unknown,
+      });
+      return;
+    }
     return {
-      thenReturn: (result: QueryReturnTypeFromDef<Q>) => {
+      thenReturn: (result: unknown) => {
```
nihalbhatnagar: Im confused why is this unknown now

nihalbhatnagar (dismissed):

packages/unit-testing/src/mock/createMockClient.ts
esinx: The function implementation signature is modified to be no longer generic since it's not possible to resolve the type of a generic function 😞 I can post a typescript playground repro, but I think we could also add a `as` cast

esinx: nvm we are already doing the `as` cast, we can make it generic

nihalbhatnagar (approved):

PR #3843, opened 2026-08-12
Add labels to customise text

nfdiop:
### What is done in this PR
- Allow the user to customise the labels used inside ActionForm/BaseForm

### Next
- Support this also for date pickers

### Tests
https://palantir.github.io/osdk-ts/storybook/pr-3843/2e474ed852844d5c819d0543d280d3c058977221/?path=/story/components-actionform-usage--with-labels


packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -630,6 +716,11 @@ export type RendererFieldDefinition = {
     fieldType?: FieldType;
     label: string;
     isRequired?: boolean;
+    /**
+     * User-facing strings rendered by FormFieldRenderer.
+     * Supply values resolved by your i18n library to customize this copy.
+     */
+    labels?: Partial<FormFieldLabels>;
```
nfdiop: I am not happy with the fact that this is named `labels` and we have another `label` above but I couldn't find a better name

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -34,6 +34,26 @@ import type { PortalContainer } from "../shared/PortalDismissLayer.js";
...
+export interface FormFieldLabels {
+  /** @default "required" */
+  requiredIndicatorAriaLabel: string;
+  /** @default "Edited" */
+  editedLabel: string;
+  /** @default (label) => label != null ? `Info about ${label}` : "More information" */
+  renderInfoTipAriaLabel: (label: string | undefined) => string;
+}
+
+export const DEFAULT_FORM_FIELD_LABELS: FormFieldLabels = {
```
nfdiop, line 46: colocating default values and types so they don't diverge

packages/react-components/src/action-form/__tests__/ObjectSetField.test.tsx
```diff
@@ -251,6 +251,44 @@ describe("ObjectSetField", () => {
       expect(screen.getByText("1,000 Employees")).toBeDefined();
     });
 
+    it("preserves precision for counts above Number.MAX_SAFE_INTEGER", () => {
```
nfdiop, line 254: Fixing how we converted the objects count for big numbers

packages/react-components/src/action-form/FormFieldApi.ts
ziyunp: What about labelsOverride or componentLabels?

packages/react-components/src/action-form/FormFieldApi.ts
ziyunp: Can we keep default values and types in a separate file from FormFieldApi?

packages/react-components-storybook/src/stories/ActionForm/ActionForm.stories.tsx
```diff
@@ -802,6 +810,40 @@ export const WithFieldOverrides: Story = {
   },
 };
 
+export const WithLabels: Story = {
+  args: {
+    formFieldDefinitions: actionFormOverrideFields,
+    onSubmit: handleSlowSubmit,
+    labels: {
```
ziyunp: Looks good!

packages/react-components/src/action-form/BaseFormLabels.ts
```diff
@@ -0,0 +1,76 @@
...
+  /** @default "Submission failed" */
+  submissionFailed: string;
+  /** @default "Invalid" */
+  validationError: string;
+  /** @default "Some fields are invalid" */
+  invalidFields: string;
+  /** @default "Submitting…" */
+  submitting: string;
+  /** @default (count) => count === 1 ? "1 issue" : `${count} issues` */
+  renderIssueCount: (count: number) => string;
```
ziyunp: Nit: Instead of render-, should we use get...Label, e.g. getIssueCountLabel ? To me render sounds like it should return ReactNode

nfdiop: You are right. I was hesiting too but I decided to keep it because it sounded better. Let me make the change

packages/react-components/src/action-form/FormFieldApi.ts
nfdiop: Yes, I am tracking a FLUP for this to have every field props to be in a separate file like for the DatePicker (but not for now)

packages/react-components/src/action-form/__tests__/ActionForm.test.tsx
```diff
@@ -51,29 +51,11 @@ interface TestActionDef extends ActionDefinition<unknown> {
   };
 }
 
-interface BooleanActionDef extends ActionDefinition<unknown> {
```
nfdiop, line 54: was not used

packages/react-components/docs/ActionForm.md
```diff
@@ -51,6 +51,8 @@ Type parameters: `Q extends ActionDefinition<unknown>`
...
 | `formTitle`            | `string`                                                                                                                                    | Title shown in the form header. Omit to hide the header.                                                                                                             |
 | `isSubmitDisabled`     | `boolean`                                                                                                                                   | Whether the submit button is disabled. Default `false`.                                                                                                              |
+| `submitButtonText`     | `string`                                                                                                                                    | Label for the submit button. Default `"Submit"`.                                                                                                                     |
```
nfdiop, line 54: FLUP to order these

PR #3844, opened 2026-08-12
FilterList: make collapse support uncontrolled and controlled modes

ziyunp:
FilterList: support both controlled and uncontrolled collapse. 

- `onCollapsedChange` no longer gates the collapse control, it is now purely an event listener that fires in both modes. 
- Pass `collapsed` for controlled mode
- Added a new `defaultCollapsed` to seed uncontrolled mode; `collapsed` takes precedence when both are given. - - - Added a new `enableCollapse` prop defaults to `true`, so existing collapse setups keep working and lists that had no collapse control gain one; pass `enableCollapse={false}` to opt out.


packages/react-components/src/filter-list/base/FilterListHeader.tsx
```diff
@@ -44,8 +46,6 @@ function FilterListHeaderInner({
   activeFilterCount = 0,
   canReset,
 }: FilterListHeaderProps): React.ReactElement {
-  const showCollapseButton = onCollapsedChange != null;
```
ziyunp, line 47: Main fix to not rely on onCollapsedChange to show collapse button

packages/react-components/src/filter-list/base/BaseFilterList.tsx
```diff
@@ -33,7 +33,9 @@ export function BaseFilterList<D extends FilterDefinitionControls>(
   const {
     title,
     titleIcon,
-    collapsed = false,
+    enableCollapse = true,
+    collapsed,
```
olliejenningsPT, line 37: Does having collapse mean that if we pass `collapsed` only and then no `onComplassedChange` function, nothing will happen when the user clicks the button, because handle collapase change will call nothing since it thinks its controlled?

ziyunp: Yes, this is consistent with all other components' controlled states

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -207,12 +205,26 @@ const meta: Meta<EmployeeFilterListProps> = {
       control: "boolean",
       table: { defaultValue: { summary: "false" } },
     },
+    enableCollapse: {
```
nfdiop, line 208: Maybe in the next version of the components we could have something like in here https://blueprintjs.com/docs/#core/components/section?
- collapseProps?: (default, collapsed, default, onCollapsedChange)
- collapsible =collapseProps != null

nfdiop (approved):

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
ziyunp: Most of the props are flat on the components, that's the decision we made for other components. Section component in blueprint is different because collapse props is forwarded to the Collapse component

PR #3849, opened 2026-08-12
[1/3] create-widget react template support no osdk

tzyl:
This PR is the first step to unifying what is currently the minimal react (no osdk) template with the react (with osdk template).

It follows the create-app structure to start supporting the react template to be rendered both with and without osdk support from the same files using handlebars helpers but does not actually change the user facing template contents.

The next step will be making the react template have the minimal template code for the basic counter widget and remove the minimal template with a compatibility option for the CLI to render out the react template with no osdk.

**Before this PR**
- minimal template (no osdk, counter example)
- react template (with osdk, todo list example)

**After this PR**
- minimal template (no osdk, counter example)
- react template (with osdk/no osdk, todo list example)

**Future goal state**
- react template (with osdk/no osdk, counter example)
- alias for compatibility minimal -> react no osdk


examples/example-widget-minimal-react-sdk-2.x-no-osdk/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@osdk/examples.widget-minimal-react-sdk-2.x",
+  "name": "@psdk/examples.widget-minimal-react-sdk-2.x",
```
tzyl, line 2: This is a little bit weird but I think is harmless and reuses the convention from the create-app example changes to make rendering with no osdk deconflict the package name with `@psdk` rather than `@osdk`

amishshah: 🤔

examples/example-widget-minimal-react-sdk-2.x-no-osdk/README.md
```diff
@@ -5,7 +5,7 @@ This project was generated with [`@osdk/create-app`](https://www.npmjs.com/packa
 To quickly create your own version of this template run the following command and answer the prompts based on your Developer Console application:
 
 ```
-npm create @osdk/app@latest -- --template widget-minimal-react
+npm create @osdk/app@latest -- --template widget-minimal-react --skipOsdk
 ```
 
 Alternatively check out the Developer Console docs for a full guide on creating and deploying frontend applications with the Ontology SDK.
```
amishshah, line 11: This README seems completely off for widgets I'm very confused 😅 But I guess it's always been off in that case? Why does it reference Developer Console and `@osdk/create-app`?

tzyl: Yes this is inherited from the setup from example-generator. I can track cleaning this up alongside some other weird inherited dead code after these PRs. We mostly use these examples as a way to test that the rendered templates compile anyway so the README / package name is not super important given we provide prefilled snippets as part of the create flow anyway

amishshah (dismissed):

amishshah (approved):

PR #3850, opened 2026-08-12
Add experimental $defaultLoadLevel fetch arg, reflected in fetchPageByRid result types

nihalbhatnagar:
Standalone version of #3790 + #3838, rebased onto `main` so it can merge on its own. Same content as those two PRs combined; if #3790 merges first, this can be dropped (or vice versa).

## What

`$defaultLoadLevel` applies reducers and struct main values to every property without having to list property IDs, as an alternative to naming each one in `$applyModifiers`.

**From #3790 (@braedens):**
- `$defaultLoadLevel` on `FetchPageArgs`, carried on the wire through the object and static-rid load paths as `defaultLoadLevel`. `applyMainValue` maps to the `extractMainValue` wire discriminant.

**From #3838:**
- `fetchPageByRid` previously hardcoded the `FetchPageResult` modifiers slot to `{}` and had no load-level slot, so neither `$applyModifiers` nor `$defaultLoadLevel` showed up in the returned object types — you'd pass a load level, the request would carry it, and the result would still be typed as the unmodified struct/array.
- `FetchPageResult` now takes a `DEFAULT_LOAD_LEVEL` parameter and resolves the selected keys through it: reducers and struct main values where a property supports them, with `applyReducersAndExtractMainValue` degrading to whichever half a property supports, and properties supporting neither left untouched. A per-property `$applyModifiers` entry wins over the default level.
- `FetchPageArgs` gains a trailing `DEFAULT_LOAD_LEVEL` slot so the literal can be inferred; its default stays `PropertyModifierValue` so existing unparameterized uses still accept any level.
- New `ResolveDefaultLoadLevelForProp` / `ApplyDefaultLoadLevelToKeys` helpers in `PropertyModifiers.ts`.

The object-set `fetchPage` / `fetchPageWithErrors` signatures are deliberately untouched, so regular `fetchPage` result types are unchanged. `fetchPageByRidNoType` is untouched too — `Q` is the base union there, so the helpers have nothing to resolve against.

## Tests

`packages/client/src/objectSet/ObjectSet.test.ts` covers the wire body (omitted by default, all three levels, the `extractMainValue` mapping, the static-rid path) and the result-type pass-through on `fetchPageByRid` / `fetchPageByRidNoType` against the real generated `Employee` (`employeeProfile` has `mainValue: ['bio']`, `performanceScores` has reducers). `runReducerAndMainValueTest.ts` exercises it against a real stack.

### DIFF
--- a/packages/api/src/object/FetchPageArgs.ts
+++ b/packages/api/src/object/FetchPageArgs.ts
@@ -139,7 +139,7 @@ export interface FetchPageArgs<
    *
    * @experimental
    */
-  $defaultLoadLevel?: DEFAULT_LOAD_LEVEL;
+  $EXPERIMENTAL_defaultLoadLevel?: DEFAULT_LOAD_LEVEL;
   /**
    * Ensures paging consistency by freezing the view at the time of query to prevent duplicate or missing items. Setting $snapshot to false ensures that you will always get the latest results.
    * @default false
--- a/packages/api/src/object/FetchPageArgs.ts
+++ b/packages/api/src/object/FetchPageArgs.ts
@@ -19,7 +19,10 @@ import type {
   PropertyKeys,
 } from "../ontology/ObjectOrInterface.js";
 import type { CompileTimeMetadata } from "../ontology/ObjectTypeDefinition.js";
-import type { ApplyModifiersArg } from "../ontology/PropertyModifiers.js";
+import type {
+  ApplyModifiersArg,
+  PropertyModifierValue,
+} from "../ontology/PropertyModifiers.js";
 
 export type NullabilityAdherence = false | "throw" | "drop";
 export namespace NullabilityAdherence {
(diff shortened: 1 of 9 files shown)


packages/api/src/object/FetchPageArgs.ts
```diff
@@ -127,6 +132,14 @@ export interface FetchPageArgs<
...
   $applyModifiers?: ApplyModifiersArg<Q> &
     MODIFIERS & { [P in Exclude<keyof MODIFIERS, PropertyKeys<Q>>]: never };
+  /**
+   * Best-effort load level applied to every property without listing them:
+   * reducers and struct main values where defined, other properties unchanged.
+   * A per-property `$applyModifiers` entry wins.
+   *
+   * @experimental
+   */
+  $defaultLoadLevel?: DEFAULT_LOAD_LEVEL;
```
ssanjay1: Should we explicitly name it experimental as well?

ssanjay1 (approved):

PR #3852, opened 2026-08-13
Bump @osdk/foundry to 2.73.0 + add OTDP flag

williamhe723:
Bump the @osdk/foundry version to 2.73.0 and fix all the issues that come with it. These issues included
1. media set endpoints went GA so had to drop `preview = True `from all of them
2. `ObjectTypeInterfaceImplementation` gained a required `actionTypes` field that had to be added to use cases

Also added the plumbing needed for `loadOntologyDefinedDerivedProperties` flag


packages/api/src/object/FetchPageArgs.ts
```diff
@@ -82,6 +82,7 @@ export interface SelectArg<
   $select?: readonly L[];
   $includeRid?: R;
   $loadPropertySecurityMetadata?: PROPERTY_SECURITIES;
+  $loadOntologyDefinedDerivedProperties?: boolean;
```
nihalbhatnagar: Should this be generic and then ` L extends string = PropertyKeys<Q>` removes OTDPs from the key set by default? Do we have that metadata at compile time

nihalbhatnagar: NVM looks like we don't know at compile time

ssanjay1: We should probably make this experimental btw `$UNSTABLE_loadOntologyDefinedDerivedProperties

bryantpark04 (dismissed):

bryantpark04:
@williamhe723 could you add the OTDP flag to the PR title?

bryantpark04 (approved):

PR #3854, opened 2026-08-13
Preserve action parameter types in FormFieldDefinition

nfdiop:
## Summary

- derive action-form component props from the selected action parameter
- preserve typed boolean and string values for dropdowns, radio buttons, and custom fields
- preserve concrete object and generated object-set target types without widening
- add public type-contract coverage for valid and invalid field configurations

## Next
Proper support for params with `multiple = true` or mention it as a limitation

## Testing

- `pnpm --filter @osdk/react-components exec vitest run src/action-form/__tests__/ObjectSetField.test.tsx src/action-form/__tests__/ActionForm.test.tsx`
- `pnpm turbo check-api --filter=@osdk/react-components`
- `pnpm --filter @osdk/react-components lint`
- `pnpm --filter @osdk/react-components check-spelling`
- `pnpm turbo transpile`


packages/react-components/src/action-form/__tests__/FormFieldDefinition.test-d.ts
```diff
@@ -0,0 +1,382 @@
...
+>;
+
+const invalidSingleSelectForRepeatedString: FormFieldDefinition<
+  UpdateProfileAction,
+  "tags"
+> = {
+  fieldKey: "tags",
+  fieldComponent: "DROPDOWN",
+  label: "Tags",
+  // @ts-expect-error Repeated action parameters require multi-select dropdowns
```
ziyunp, line 302: Why is this a constraint? a multiplicity field type can also have one item in the array no?

packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -157,6 +160,35 @@ export interface FormFieldPropsByType {
   UNSUPPORTED: UnsupportedFieldProps;
 }
 
+/**
+ * Maps the components supported by an action parameter to props that preserve
```
nfdiop, line 164: Fixing the types to:
- use the correct value type instead of unknown for fields like DropdownField
- use the correct objectType vs accepting all object types

packages/react-components-storybook/src/stories/ActionForm/ActionForm.stories.tsx
ziyunp: Nice!

ziyunp (dismissed):

ziyunp (approved):

PR #3827, opened 2026-08-10
[FilterList] unify every definition kind on defaultFilterState; rename linkedFilterComponent to filterComponent

ziyunp:
Problem:
Previously PROPERTY and STATIC_VALUES required filterState, HAS_LINK and CUSTOM required a filterState that was never read, and LINKED_PROPERTY needed three fields (linkedFilterState, defaultLinkedFilterState, filterState) to express one value. 

Changes:
- use filterState, deprecate defaultFilterState, linkedFilterState defaultLinkedFilterState
- id and isVisible moved to the shared FilterDefinitionControls mix-in (not a shape change).

PR #3856, opened 2026-08-13
[FilterList] drop the linked- prefix from LINKED_PROPERTY fields

ziyunp:
**Stack: 1 of 3** — splits #3827. Merge in order; 2 and 3 are stacked on this.

## Problem

A `LINKED_PROPERTY` definition needed four fields to express one value:

```ts
{
  type: "LINKED_PROPERTY",
  linkedPropertyKey: "fullName",
  linkedFilterComponent: "MULTI_SELECT",
  linkedFilterState: { type: "SELECT", selectedValues: [] },        // never read
  defaultLinkedFilterState: { type: "SELECT", selectedValues: [] }, // the actual seed
  filterState: { type: "linkedProperty", linkedFilterState: { ... } }, // never read
}
```

Only `linkedFilterComponent` and `defaultLinkedFilterState` were ever read. The other two were required by the type but ignored at runtime, so every call site carried redundant, misleading boilerplate.

## Changes

| Before | After |
| --- | --- |
| `linkedFilterComponent` | `filterComponent` |
| `defaultLinkedFilterState` | `defaultFilterState` |
| `linkedFilterState` | deprecated no-op |
| `filterState` | deprecated no-op |

The same definition now reads:

```ts
{
  type: "LINKED_PROPERTY",
  linkedPropertyKey: "fullName",
  filterComponent: "MULTI_SELECT",
  defaultFilterState: { type: "SELECT", selectedValues: [] },
}
```

**Nothing is removed.** Every old field is optional and `@deprecated`. `getLinkedFilterComponent` resolves `filterComponent ?? linkedFilterComponent`, and the seed reads `defaultFilterState ?? defaultLinkedFilterState`.

`linkedFilterState` and `filterState` are deprecated as no-ops rather than renamed. Both were required, so they're set on essentially every existing definition — reading them now would silently activate filters that were never active.

## Notes for review

- `filterComponent` is optional only so the deprecated spelling keeps compiling. A definition setting neither renders as unsupported and does not seed, so it can't narrow the object set off a filter the user can neither see nor clear. Both `LinkedPropertyInput`'s `case undefined` branch and the `TODO`s go away when `linkedFilterComponent` is removed.
- In the next [PR](https://github.com/palantir/osdk-ts/pull/3857), we will convert all filterState to defaultFIlterState


packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -89,7 +90,15 @@ function buildInitialStates<Q extends ObjectTypeDefinition>(
...
       case "LINKED_PROPERTY": {
-        const innerState = definition.defaultLinkedFilterState;
+        // No component renders as unsupported, so seeding would narrow
+        // `objectSet` off a filter the user can neither see nor clear.
+        if (getLinkedFilterComponent(definition) == null) {
+          break;
+        }
+        const innerState =
+          definition.defaultFilterState ??
+          // eslint-disable-next-line @typescript-eslint/no-deprecated -- pre-rename fallback
```
nfdiop, line 100: nit: maybe don't disable so devs can know quickly?

packages/react-components/src/filter-list/utils/__tests__/getLinkedFilterComponent.test.ts
nfdiop, line 1: nice set of tests!

nfdiop (approved):

packages/react-components/src/filter-list/hooks/useFilterListState.ts
ziyunp: I left it in as it's just a nit :D

PR #3857, opened 2026-08-13
[FilterList] unify definition seeds on defaultFilterState; move id/isVisible to the shared mix-in

ziyunp:
**Stack: 2 of 3** — splits #3827. #3856 is merged, so this now targets `main`. #3858 follows.

## Problem

Seeding a filter meant a different field depending on the kind, and none of the spellings matched the `default<Feature>` convention the rest of the library uses for an uncontrolled seed:

| Kind | Seed field | Required? | Actually read? |
| --- | --- | --- | --- |
| `PROPERTY` | `filterState` | yes | yes |
| `STATIC_VALUES` | `filterState` | yes | yes |
| `CUSTOM` | `defaultFilterState` | no | `filterState` is required but only pre-fills `renderInput` |
| `HAS_LINK` | `defaultFilterState` | no | yes |
| `KEYWORD_SEARCH` | `defaultFilterState` | no | yes |

`PROPERTY` and `STATIC_VALUES` called an uncontrolled seed `filterState`, which reads as controlled state — but FilterList owns the state after mount and there is no controlled mode. `HAS_LINK`, `KEYWORD_SEARCH` and `CUSTOM` additionally required a `filterState` that never seeded anything.

`id` and `isVisible` were declared separately on five definition kinds with drifting doc comments.

## Changes

- `PROPERTY`, `STATIC_VALUES`: `filterState` → `defaultFilterState`.
- `HAS_LINK`, `KEYWORD_SEARCH`: `filterState` deprecated as a no-op.
- `CUSTOM`: `filterState` becomes optional and `@deprecated`, with **no change to what it does** — it still only pre-fills `renderInput`, and `defaultFilterState` is still the only field that seeds. #3858 is where `CUSTOM` starts filtering off its seed. Because the field is now optional, a `CUSTOM` definition may seed nothing at all, so `renderInput` gets an empty custom state instead of `undefined`.
- Top level: `initialFilterStates` → `defaultFilterStates`.
- `id` and `isVisible` move onto the shared `FilterDefinitionControls` mix-in. Not a shape change for the kinds that already declared them — but every kind now accepts both, and the docs are written once.
- Seeding moves out of the `buildInitialStates` switch into `getSeedFilterState`, which owns the per-kind deprecated fallbacks in one place.

**Nothing is removed.** Every old field is optional, `@deprecated`, and honoured as a fallback (`defaultFilterState ?? filterState`, `defaultFilterStates ?? initialFilterStates`).

## Stories and docs

Now that the seed is optional, a definition carrying an empty state (`{ type: "EXACT_MATCH", values: [] }`, `{ type: "SELECT", selectedValues: [] }`, a bare `CONTAINS_TEXT`/`DATE_RANGE`/`NUMBER_RANGE`, or `HAS_LINK`'s `{ hasLink: false }`) are deleted.

## Notes for review

- `filterState` on `HAS_LINK` / `KEYWORD_SEARCH` / `CUSTOM` is deprecated as a no-op rather than renamed, for the same reason as in PR 1: it's set on nearly every existing definition, so reading it would activate filters nobody asked for. `getSeedFilterState` groups the kinds by that distinction, with the reasoning at the switch.


packages/react-components/src/filter-list/utils/getSeedFilterState.ts
ziyunp, line 1: Check the old implementation in packages/react-components/src/filter-list/hooks/useFilterListState.ts

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1867,7 +1867,7 @@ function WithLinkedPropertyFiltersStory(
         type: "HAS_LINK",
         linkName: "lead",
         label: "Has Manager",
-        filterState: { type: "hasLink", hasLink: false },
+        defaultFilterState: { type: "hasLink", hasLink: false },
```
olliejenningsPT: Should these ones be simply deleted? the changeset says to delete rather than rename them? or is the defaultFilterState fine and jsut for filterState we delete them?

packages/react-components/src/filter-list/FilterListItemApi.ts
```diff
@@ -50,6 +49,26 @@ export interface FilterDefinitionControls {
    * @default true
    */
   searchField?: boolean;
+
+  /**
+   *
```
olliejenningsPT: should we delete this extra line?

packages/react-components/src/filter-list/utils/__tests__/getSeedFilterState.test.ts
```diff
@@ -0,0 +1,187 @@
...
+import type { FilterDefinitionUnion } from "../../FilterListApi.js";
+import { getSeedFilterState } from "../getSeedFilterState.js";
+
+type Def = FilterDefinitionUnion<typeof MockObjectType>;
+
+const EXACT_A = { type: "EXACT_MATCH", values: ["a"] } as const;
+const EXACT_B = { type: "EXACT_MATCH", values: ["b"] } as const;
+const EMPTY_CUSTOM = { type: "custom", customState: {} } as const;
+
+function def(fields: Record<string, unknown>): Def {
```
olliejenningsPT: should this be more of a type guard, understood its a test, but sometimes having good type checking in the test can be helpful

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
ziyunp: I can clean it up so filterState that are not seeding anything can be removed

packages/react-components/src/filter-list/FilterListItemApi.ts
ziyunp: Removed

olliejenningsPT (approved):

PR #3858, opened 2026-08-13
[FilterList] seed CUSTOM filters from defaultFilterState; fix their active filter count

ziyunp:
**Stack: 3 of 3** — splits #3827. Based on #3857; review that first. **Retarget to `main` once #3857 merges.**

## Problem

**1. `filterState` was used for FilterInput component rendering ONLY. It's not passed to `toWhereClause` and active filter count is not shown.

Screenshot 2026-08-13 at 15 59 33

**2. `defaultFilterState` seeds the filter state and rendering FilterInput rendering, but it does not clear activeCount when the filter is cleared.

https://github.com/user-attachments/assets/7ec296a9-58f0-47f7-99ce-66458b1bab8c

## Changes

https://github.com/user-attachments/assets/06e299ef-eaf5-4587-bcef-75c53bfb43f4

- `filterState` → `defaultFilterState`, joining the group that seeds the state map. The value now reaches `toWhereClause` and the active filter count, not just `renderInput`. Both props behave the same now.
- `FilterInput` no longer reads the definition's seed fields at all. An untouched custom filter gets `EMPTY_CUSTOM_STATE`, so `renderInput` still always receives a state object; `defaultFilterState` reaches it through the state map like every other kind.
- `activeFilterCount` asks a custom filter's own `toWhereClause` whether its state filters anything, matching the clause `buildWhereClause` actually keeps.

## Note

`getCustomRenderInputState` — the helper that existed only to keep the old field feeding `renderInput` without seeding — is gone, since the two paths are now one.


packages/react-components/src/filter-list/utils/getSeedFilterState.ts
```diff
@@ -40,6 +40,7 @@ export function getSeedFilterState<Q extends ObjectTypeDefinition>(
     // `filterState` is the pre-rename seed here.
     case "PROPERTY":
     case "STATIC_VALUES":
+    case "CUSTOM":
```
ziyunp, line 43: Previously, it was only seeding from defaultFilterState, but filterState is used to seed the rendered FilterInput component

packages/react-components/src/filter-list/FilterInput.tsx
```diff
@@ -120,10 +120,7 @@ function FilterInputInner<Q extends ObjectTypeDefinition>({
         );
       }
       const customFilterState =
-        filterState?.type === "custom"
-          ? filterState
-          : // eslint-disable-next-line @typescript-eslint/no-deprecated -- pre-rename fallback
-            (definition.filterState ?? EMPTY_CUSTOM_STATE);
+        filterState?.type === "custom" ? filterState : EMPTY_CUSTOM_STATE;
```
ziyunp, line 123: filterState is seeded with `definition.defaultFilterState ?? definition.filterState` already

olliejenningsPT (approved):

PR #3859, opened 2026-08-13
Add onLoadedObjectsChanged callback to ObjectTable

ziyunp:
## What

Adds an `onLoadedObjectsChanged` listener to `ObjectTable` so callers can observe which rows the table currently holds, plus the object set's total count.

```tsx
<ObjectTable
  objectType={Employee}
  onLoadedObjectsChanged={({ loadedObjects, totalCount }) => {
    console.log(`Loaded ${loadedObjects.length} of ${totalCount} employees`);
  }}
/>
```

## API

New `LoadedObjectsChange<Q, RDPs>` payload, exported from `@osdk/react-components/experimental/object-table`:

| Field | Type | Notes |
| --- | --- | --- |
| `loadedObjects` | `Osdk.Instance<Q, "$allBaseProperties", PropertyKeys<Q>, RDPs>[]` | Rows loaded so far, in display order. Unscrolled pages are absent. Each row also carries its function-backed column values. |
| `totalCount` | `string \| undefined` | Total matching the object set; `undefined` when the API reports none. String, consistent with `getSnapshot()`'s `totalCount`. |

A payload object was chosen over positional params to match the existing `RowSelectionChange` shape and stay extensible.

https://github.com/user-attachments/assets/9780c976-7981-4b2a-866e-c588490168cb


packages/react-components/src/object-table/hooks/useLoadedObjectsChanged.ts
```diff
@@ -0,0 +1,79 @@
...
+const EMPTY_ROWS: never[] = [];
+
+const DEBOUNCE_MS = 50;
+
+/**
+ * Function-backed columns and streamed updates may result in a burst of changes.
+ * Setting leading so the first report is immediate,
+ * and trailing to report the latest state.
+ */
+const DEBOUNCE_OPTIONS = { leading: true, trailing: true };
```
olliejenningsPT: So the only concern here would be starvation, could we add in a `maxWait`. e.g. if the stream is continuous, the debounce never fires, might be worth having like a `200ms` max wait.

packages/react-components/src/object-table/hooks/useLoadedObjectsChanged.ts
```diff
@@ -0,0 +1,79 @@
...
+  SimplePropertyDef,
+} from "@osdk/api";
+import { useEffect } from "react";
+
+import { useDebouncedCallback } from "../../shared/hooks/useDebouncedCallback.js";
+import type { LoadedObjectsChange } from "../ObjectTableApi.js";
+
+const EMPTY_ROWS: never[] = [];
+
+const DEBOUNCE_MS = 50;
```
olliejenningsPT: would 50ms be slightly too small? am just thinking about the factors in this for concurrent requests etc

packages/react-components/docs/ObjectTableOverview.md
olliejenningsPT, line 138: Do we want to add this new hook here?

olliejenningsPT (approved):

PR #3861, opened 2026-08-13
Run the pull-request CI on CircleCI alongside GitHub Actions

olliejenningsPT:
The Claude PR description was not very clear, so will try and be as concise as possible.

This PR introduces CircleCI for running our CI checks. It is a faithful mirror of what GitHub CI does, but we some changes that l will outline.

### Not Changed

- **Release workflow** - this stays the same
- **Current CI** - this is additive to our current CI

### Why / How this is this faster

- **Dedicated machines** - GitHub Runners used shared VMs which are 4 core machines, Circle gives us dedicated machines with us to 16 cores (if we require, based on runner size)
- **More Cores** - Turbo + other rust based tooling greatly benefit from more cores, so they can parallelise, they spin up 1 thread per CPU (1 task per CPU in turbo case).
- **Parallelism** - similar to above, we even have the ability to have multiple machines per stage, we do this for certain tests (`react-components`) because those tests are very heavy, and splitting them out so they run individually gives a big speed up
- **`e2e` waits on `build`** - it used to start alongside `build` and transpile the whole workspace itself, losing that race two thirds of the time (~250s instead of ~130s). Waiting costs ~40s of serialisation, removes ~105s of duplicated work, and is deterministic either way.
- **Better caching / splitting of stages** - arguably we could do this on GH CI, but also addressed an issue where storybook gets built multiple times, instead we build this once.

### Benefits

- **browsable storybook per PR** - CircleCI artifacts are browsable, and so per PR we can output these and have a full interactive storybook to see how PR changes that affect any `react-components` look. (was previously available, but quite hacky to get it working, and not perfect)
- **Proper test result reporting** - We are now storing proper test artifacts, which means if there is a failing test, you don't need to scroll through logs, instead you get a nice view in the circle UI to see which tests are failing, and the logs related to those tests only. This is enabled in the prior PR stack.
- **Faster** - this drops the average CI build times from ~11mins on GitHub to ~4mins on Circle

### Follow Ups

- **RAM Cache** - it requires some re-wiring for pnpm to work out the box, but it will shave off around 30s-45s per stage where we restore from cache for node_modules
- **`e2e` publish step** - 71-73s of that job is `pnpm publish -r` pushing ~47 packages into Verdaccio. Dropping Verdaccio's npmjs uplink, `--workspace-concurrency` (rejected outright as an unknown option) and a bigger box were all measured and none moved it, the limit is pnpm's own fixed publish concurrency. Publishing through turbo would let us set it, but needs a `publish` script in every manifest and an uncached task.
- **GitHub PR comments** - right now we do not have these running, we would need to add a `GITHUB_TOKEN` to Circle in order to get this running (blueprint does this), but l didn't want multiple identical comments in this initial PR
- **Proper Code Coverage** - this was attempted, but due to a PR this broke, completely stripping this out and putting in a FLUP.
- **Chromatic** - this enables visual diffs for storybook, we should be careful in only enabling this for PRs that actually touch or trigger changes in the components.


.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+    environment: *ci-environment
+
+commands:
+  activate-pnpm:
+    description: Make PNPM available on PATH.
+    steps:
+      - run:
+          name: Install pnpm
+          command: |
+            sudo npm install --global --silent pnpm@10
```
ericjeney: As far as I can tell, the `pnpm/action-setup` in our Github Actions today leverages the pnpm version we pin in `package.json`, whereas this will just install the latest v10 of pnpm.

We should make this more similar to the Github Actions version. At minimum, it needs to be pinned to ensure that our builds are reproducible and to insulate us from any pnpm compromise, but ideally we'd also have the version number configured in just a single place.

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+            auth:
+              htpasswd:
+                file: /verdaccio/storage/htpasswd
+                max_users: 1000
+            packages:
+              '**':
+                access: $all
+                publish: $all
+                unpublish: $all
+            YAML
```
ericjeney, line 55: What makes this config necessary on Circle but not on Github Actions?

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+            - v1-pnpm-store-{{ arch }}-{{ checksum "pnpm-lock.yaml" }}
+            - v1-pnpm-store-{{ arch }}-
+      - run:
+          name: Install dependencies
+          command: pnpm install --frozen-lockfile --store-dir "$PNPM_STORE_DIR"
+
+  attach-deps:
+    description: >
+      Take node_modules from the workspace instead of reinstalling. Jobs
+      downstream of `build` pick up its turbo cache this way too.
```
ericjeney: Should rename the task for clarity if the workspace is being used for more than just attaching deps

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+            - docs/node_modules
+
+  changesets:
+    executor:
+      name: node
+      resource-class: medium
+    steps:
+      - checkout
+      - activate-pnpm
+      - attach-deps
```
ericjeney:
```suggestion
      - prepare
```

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+            exit "$status"
+
+  bundle-size:
+    executor:
+      name: node
+      resource-class: large
+    steps:
+      - checkout
+      - activate-pnpm
+      - attach-deps
```
ericjeney:
```suggestion
      - prepare
```

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+  activate-pnpm:
+    description: Make PNPM available on PATH.
+    steps:
+      - run:
+          name: Install pnpm
+          command: |
+            sudo npm install --global --silent pnpm@10
+            pnpm --version
+
+  install-deps:
```
ericjeney, line 71: Would be helpful to put the `install-deps` and `save-pnpm-store` tasks next to each other since they both reference the same cache keys and are intertwined

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+        type: string
+    steps:
+      - restore_cache:
+          name: Restore turbo cache (<< parameters.scope >>)
+          keys:
+            - turbo-{{ arch }}-<< parameters.scope >>-{{ .Environment.CIRCLE_NODE_INDEX }}-{{ .Revision }}
+            - turbo-{{ arch }}-<< parameters.scope >>-{{ .Environment.CIRCLE_NODE_INDEX }}-
+            # `build` is never parallel, so its cache is always at index 0.
+            - turbo-{{ arch }}-build-0-{{ .Revision }}
+            - turbo-{{ arch }}-build-0-
```
ericjeney: Feels weird to hardcode this since we may rename the job in the future or split it into more containers

Is `CIRCLE_NODE_INDEX` not set when a job doesn't use parallelism? The docs don't make this clear, but I don't see how `save-turbo-cache` would work as written for the `build` step if that's the case

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+      scope:
+        type: string
+    steps:
+      - save_cache:
+          name: Save turbo cache (<< parameters.scope >>)
+          key: turbo-{{ arch }}-<< parameters.scope >>-{{ .Environment.CIRCLE_NODE_INDEX }}-{{ .Revision }}
+          paths:
+            - .turbo/cache
+
+  verify-working-tree-clean:
```
ericjeney, line 140: Should you run this on all of the jobs below? I only see it on a few of them

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+  install:
+    executor:
+      name: node
+      resource-class: large
+    steps:
+      - checkout
+      - activate-pnpm
+      - install-deps
+      - save-pnpm-store
+      - persist_to_workspace:
```
ericjeney, line 167: Does this actually save time? It looks like persisting the workspace takes far longer than just running pnpm install, and it doesn't look like re-attaching the workspace is particularly fast either 🤔 

Image

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+      is a silent no-op: one key shared across a job's parallel containers would
+      mean only one container's results ever persist.
+    parameters:
+      scope:
+        type: string
+    steps:
+      - restore_cache:
+          name: Restore turbo cache (<< parameters.scope >>)
+          keys:
+            - turbo-{{ arch }}-<< parameters.scope >>-{{ .Environment.CIRCLE_NODE_INDEX }}-{{ .Revision }}
```
ericjeney: Does it work to replace `parameters.scope` with `CIRCLE_JOB`? Or is there a reason you want this to explicitly allow cross-job restores?

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+            - .turbo/cache
+
+  typecheck:
+    executor:
+      name: node
+      resource-class: xlarge
+    steps:
+      - prepare
+      - restore-turbo-cache:
+          scope: typecheck
```
ericjeney, line 222: Is this just overwriting the cache that was restored as part of the workspace in `prepare`? Or is it somehow merging them?

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+      - run:
+          name: Restore build artifacts from cache
+          command: pnpm exec turbo run transpile transpileTypes
```
ericjeney, line 225: This drops a comment from `ci.yml` that includes context on why this is needed as a separate step:
```yaml
# Materialize build/{esm,cjs,browser,types} from the build cache (cache
# hits) so check-attw (which packs them) finds everything, avoiding a race
# with the tasks that produce them.
```

I recognize this predates your PR, but I'm pretty skeptical of this reasoning. Does this mean we haven't defined inputs/outputs correctly somewhere? Or is Turbo just not good enough as a build system?

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+    executor:
+      name: node
+      resource-class: xlarge
+    steps:
+      - prepare
+      - restore-turbo-cache:
+          scope: lint
+      - run:
+          name: Restore build artifacts from cache
+          command: pnpm exec turbo run transpile transpileTypes
```
ericjeney, line 243: This one also has a comment in the `ci.yml` that I find surprising and would think we'd be able to express via build inputs instead

```yaml
# lint dependsOn ^transpileTypes (deps' .d.ts for type-aware rules); restore
# the build artifacts from cache so those are present.
```

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+          name: Run this container's share of the tests
+          command: |
+            set +e
+            case "$CIRCLE_NODE_INDEX" in
+              0)
+                pnpm exec turbo run test --filter='!@osdk/react-components' $TURBO_FILTERS
+                ;;
+              1)
+                pnpm exec turbo run test --filter=@osdk/react-components
+                ;;
```
ericjeney, line 307: This breakdown seems fine for now, but longer term we can probably have Circle intelligently split packages between different parallel containers based on the expected time they'll take

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+              *)
+                # A config error rather than a test failure, so it fails here
+                # instead of being deferred.
+                echo "No tests assigned to container $CIRCLE_NODE_INDEX." >&2
+                echo "\`parallelism\` and this case statement have to agree." >&2
+                exit 1
```
ericjeney: Appreciate the defensiveness 🙂

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+                ;;
+            esac
+            status=$?
+            echo "$status" > /tmp/test-exit-code
+            if [ "$status" -ne 0 ]; then
+              echo "Tests exited $status. Reports upload first; the job fails at the end."
+            fi
```
ericjeney: If you add `when: always` on `save-turbo-cache` then I don't think this is necessary, and it'd be clearer to fail the job on the step that actually failed

.circleci/config.yml
ericjeney: Should include a version at the start of the key to be consistent with the other keys and to future proof

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+          scope: apps
+      - run:
+          name: Build apps and sites
+          command: pnpm exec turbo run build --filter='!@osdk/react-components-storybook'
+      - verify-working-tree-clean
+      - save-turbo-cache:
+          scope: apps
+
+
+  e2e:
```
ericjeney, line 343: What actually performs tests in here? Is it just confirming we can publish?

.circleci/config.yml
```diff
@@ -0,0 +1,592 @@
...
+        default: large
+    docker:
+      - image: cimg/node:<< parameters.node-version >>
+    resource_class: << parameters.resource-class >>
+    environment: *ci-environment
+
+  node-with-verdaccio:
+    docker:
+      - image: cimg/node:24.19
+      - image: verdaccio/verdaccio:6
```
ericjeney: Can we pin a version here too?

We may want to consider pinning Node further as well, since we're just specifying major + minor right now, but especially for just a major like this one

ericjeney (commented):
Thanks for cleaning this up

.circleci/config.yml
olliejenningsPT: Its less around the inputs/outputs but more of a dependsOn issue, where we are have a lot of gaps / not correctly defining the right set.

E.g. check-attw can run before its own `build/types` exist, and if you run the check-api task in parallel which triggers `transpileTypes` which would build the check-attw ones it could finish after the check-attw (race condition).

Therefore the separate build step prevents this, granted by masking, but l would say the follow up should be tighten the dependsOn gaps so we can then remove this separate build step.

.circleci/config.yml
olliejenningsPT: Yep, l know it takes a few runs on main at least for CircleCI's own intelligent splitting of tests to become worthwhile, so a good follow up to do.

.circleci/config.yml
olliejenningsPT: Similar to the above comment: https://github.com/palantir/osdk-ts/pull/3861#discussion_r3987133296 with the dependsOn gap being the problem here.

.circleci/config.yml
olliejenningsPT: Yep, it is literally confirming that we can publish. Kept for parity with the GH actions.

.circleci/config.yml
olliejenningsPT: Yeah l think thats reasonable, pining down to the patch provides extra level of security, will apply to both.

.circleci/config.yml
olliejenningsPT: Let me add

.circleci/config.yml
olliejenningsPT: I will see if l can add it to more of them, but we don't want it on all, e.g. bundle-size switches commits, and e2e actually versions the packages.

.circleci/config.yml
olliejenningsPT: Yep let me do that

.circleci/config.yml
olliejenningsPT: Yep, it merges into the existing .turbo/cache directory. 

Turbo stores results by task hash, so the build cache from the workspace and the restored typecheck cache can sit alongside each other.

turbo then uses the matching hash for the task, or runs it if there isn't a match, files with same hash overwrite

.circleci/config.yml
olliejenningsPT: The config is to prevent verdaccio from actually hitting the real npm registry, as verdaccio by default config acts as a proxy to the real npm registry.

I can make these changes to the GH actions as a follow up, but l wanted a strict not reaching out on the network rule here.

.circleci/config.yml
olliejenningsPT: Yep let me make it identical, installing pnpm@10 does initially install the latest 10, but then pnpm underneath switches to the version listed in our packageManager field, but l agree its better for the initial install to be explicitly identical.

.circleci/config.yml
olliejenningsPT: So this is something l actually tested alot with a few different approache (prior commits such as: 2f7f21135a).

One strict rule l gave myself was to install once, then future stages purely use this set of deps, and if some future stage had missing deps, then l would fix those install / missing deps issue, rather than masking through future installs.

So l went for consistency over timing here, but IMO it should help prevent broken installs in the future.

.circleci/config.yml
olliejenningsPT: Will do

.circleci/config.yml
olliejenningsPT: it does get set as 0 if no parallelism, but this `build-0` is in reference to the actual `build` job container rather than the current job, e.g. test container-1 needs to restore from build-0, if we used the node_index it would try and hit build-1 which doesn't exist.

.circleci/config.yml
olliejenningsPT: Yes l want to allow cross-job restores, the scope is just allowing us to control which jobs share. E.g. for the tests we put in the Node version, so make sure different versions don't share same test cache.

.circleci/config.yml
olliejenningsPT: Yep if you mean on the `save_cache` in side of the save-turbo-cache

.circleci/config.yml
ericjeney: +1 that we should tighten up the `dependsOn` later in that case. Our task dependencies should definitely be correct!

I think it's worth preserving the comments around this in this file in the meantime to give context on why it's written the way it is.

.circleci/config.yml
ericjeney: Agree that's desirable! Can you add a comment to explain it?

.circleci/config.yml
ericjeney: When would `test` restore from `build-0`? Wouldn't it hit the line prior to this one and restore the `test` cache?

ericjeney (dismissed):
This generally looks good to me now, pending just a couple comments. Thank you for the iteration here!

.circleci/config.yml
olliejenningsPT: So basically if the test misses its cache (both for the commit, and for the node version, arch / container index), then it fallback to the build, this is to always make sure we never re-build in the test stages.

ericjeney (approved):

PR #3863, opened 2026-08-13
Revert @osdk/e2e.sandbox.catchall from vitest v4 to v3.x

olliejenningsPT:
Puts `@osdk/e2e.sandbox.catchall` back on the vitest v3 line. #3787 added it as a new
`vitest: ^4.1.10` devDependency, making it the only package pulling a v4 install while
the other ~47 sit on `^3.2.4`.

### Why

- **One package held a second vitest tree** - a duplicate `@vitest/*` set plus `vite@8.0.8` in the lockfile to serve one sandbox package, 4 distinct vitest versions down to 3.
- **v4 drops Node 18** - engines `^20 || ^22 || >=24` against v3's `^18 || ^20 || >=22`, which is why #3787 had to exclude this package from the Node 18 CI leg.
- **Not a deliberate upgrade** - the pin rode along with #3787's integration-testing work, going to v4 deserves its own PR.

### Changed

- **Version pin** - `^4.1.10` to `^3.2.4`, matching root and the `@osdk/integration-testing` package #3787 added on `^3.2.4`.
- **`test.extend` shape** - v4's `(name, options, fn)` with `onCleanup` becomes v3's `({ name: [fn, options] })` with teardown after `await use(value)`, and `scope: "worker"` already works in 3.2.x.
- **Cleanup ordering** - a `try/finally` preserves #3787's `stop()`-even-if-`start()`-throws behaviour.
- **Hooks** - `test.describe` / `test.beforeEach` become top-level `describe` / `beforeEach`, and `beforeEach<IntegrationFixtures>` lets the fixture set drop its `as TestAPI<...>` cast.
- **Node 18 CI exclusion removed** - restores that `ci.yml` block to its pre-#3787 state.

### Not Changed

- **`examples/example-typescript-library-sdk-2.x`** - still `^4.1.7`, predates #3787, so 4.1.9 stays in the lockfile.
- **`@storybook/addon-vitest`** - still pairs `@vitest/runner@4.1.10` with `vitest@3.2.6`, unmet-optional-peer behaviour unrelated to this.
- **No changeset** - `@osdk/e2e.sandbox.*` is in the changesets ignore list.

### Follow Ups

- **Decide on v4 properly** - one PR moving every package, which needs the Node 18 matrix leg dropped first.
- **Catalog entry for vitest** - 48 independent pins is how single-package outliers like this go unnoticed.

### DIFF
--- a/packages/e2e.sandbox.catchall/src/integration-tests/test.fixture.ts
+++ b/packages/e2e.sandbox.catchall/src/integration-tests/test.fixture.ts
@@ -11,7 +11,7 @@ import {
   type SeedFunction,
   createSeedWithMetadata,
 } from "@osdk/seed-helpers";
-import { test as baseTest, type TestAPI } from "vitest";
+import { test as baseTest } from "vitest";
 
 const filteredObjectTypes = new Set(["Person", "Todo", "Game", "Book"]);
 const filteredInterfaceTypes = new Set(["LibraryItem"]);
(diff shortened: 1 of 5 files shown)


packages/e2e.sandbox.catchall/src/integration-tests/test.fixture.ts
```diff
@@ -34,35 +34,44 @@ const modifiedMetadata: Ontologies.OntologyFullMetadata = {
...
-    onCleanup(async () => await server.stop());
-    await server.start();
-    return server;
-  })
-  .extend(
-    "seed",
+export const test = baseTest.extend<IntegrationFixtures>({
+  server: [
+    // eslint-disable-next-line no-empty-pattern
+    async ({}, use) => {
```
olliejenningsPT, line 51: scope worker doesn't exist on `v3` so had to resolve to using finally

esinx (approved):

PR #3868, opened 2026-08-13
[2/3] Use minimal react widget template structure for react widget template

tzyl:
Requires https://github.com/palantir/osdk-ts/pull/3849

This PR moves the react widget template over to the minimal react widget template structure of an example counter widget for consistency. It mirrors create-app to optionally include the OSDK client and @osdk/react bootstrapping and renders an extra tag when an OSDK package is used.

Aside from the OSDK branching there are two minor changes made to the template:

- Apply CSS gap to container rather than header margin to provide OSDK section spacing when it is rendered
- Use blueprint design tokens which are now exposed in CSS variables since [6.7.0](https://github.com/palantir/blueprint/releases/tag/%40blueprintjs/core%406.7.0) which are also used in create-app

The next step will be to delete the minimal react widget template completely and provide a hidden compatibility alias for it to render this react template without osdk instead.

| Light | Dark |
|-|-|
Screenshot 2026-08-13 at 16 25 09 | Screenshot 2026-08-13 at 16 50 17

### DIFF
--- a/packages/create-widget.template.react.v2/templates/src/Widget.tsx.hbs
+++ b/packages/create-widget.template.react.v2/templates/src/Widget.tsx.hbs
@@ -1,180 +1,67 @@
 {{#if osdkPackage}}
-import { $Actions, $Objects, $Queries } from "{{osdkPackage}}";
+import { Button, Card, Classes, H4, Tag } from "@blueprintjs/core";
+{{else}}
+import { Button, Card, Classes, H4 } from "@blueprintjs/core";
 {{/if}}
 import {
-  Box,
-  Button,
-  Checkbox,
-  Container,
-  Flex,
-  Heading,
-  Skeleton,
-  Table,
-{{#if osdkPackage}}
-  Text,
-{{/if}}
-  TextField,
-  Theme,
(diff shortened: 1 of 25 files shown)


packages/create-widget.template.react.v2/templates/src/Widget.tsx.hbs
```diff
@@ -1,180 +1,67 @@
...
 
+const useWidgetContext: () => FoundryWidgetClientContext<typeof MainConfig> =
+  useFoundryWidgetContext.withTypes<typeof MainConfig>();
+
 export const Widget: React.FC = () => {
 {{#if osdkPackage}}
-  // See Ontology and Platform SDK docs in Developer Console on how to
-  // use the client object to access Ontology resources and platform APIs
+  // View the API documentation for your widget set to learn how to use the Ontology SDK.
+  // {{foundryUrl}}/workspace/custom-widgets/widget-set/{{widgetSet}}/sdk/docs
```
tzyl, line 25: Using redirect route to show docs link in code rather than rendering it to avoid issues with opening urls out of the box

amishshah (approved):

PR #3873, opened 2026-08-13
Document the viewer content prop convention

olliejenningsPT:
Layer 1 of 4 unifying the primary input prop across the base viewers. Docs only, no code changes.

### The convention

- **`src`** - the binary source to render from, in whatever forms that renderer supports: a URL, raw bytes, or both
- **`content`** - the already-decoded payload: text, or a parsed object

### Why two names and not one

A single name would make `content={someString}` compile against both `BaseImageViewer` (which wants a URL) and `BaseXmlViewer` (which wants text), with opposite meanings and no type error. Keeping the split means `src: string` is always a URL and `content: string` is always text, so passing document text to a URL-based viewer fails to compile instead of rendering a broken image.

### Where it lands

- **`CLAUDE.md` / `CONTRIBUTING.md`** - the rule itself, added to the existing API Design sections
- **`docs/PdfViewer.md`** - `BasePdfViewer` supports every source form, so it is the reference for `src`
- **`AGENTS.md`** - also corrects a stale line describing `BasePdfViewer` as accepting only a URL or `ArrayBuffer`, omitting `Uint8Array` and `Blob`

### Follow ups

- **Layers 2-4** - `BaseEmailViewer.email`, `BaseSpreadsheetViewer.spreadsheet` and `BaseTiffViewer.content` move onto the convention, each keeping a deprecated alias

image


packages/react-components/CLAUDE.md
```diff
@@ -94,6 +94,10 @@ For every state change with a built-in default behavior (sort, filter, select, e
...
 - **Document defaults inline** with `@default` JSDoc on every optional prop. Use `@param` / `@returns` for callbacks
 - **Define the API in its own file**: `<Name>Api.ts` co-located with the component, exporting the base props, the OSDK-aware outer-component props, and any public sub-types
+- **Name a viewer's primary input `src` or `content`, by form.** `src` is the binary source to render from, in whatever forms that renderer supports — a URL, raw bytes, or both. `content` is the already-decoded payload — text, or a parsed object. Never overload one name across both categories, and never name the prop after the file type (`email`, `spreadsheet`); that does not generalise to the next viewer. The split is what buys type safety: `src: string` is always a URL and `content: string` is always text, so passing document text to a URL-based viewer is a compile error rather than a silently broken render. `BasePdfViewer` is the reference for `src` — `src: PdfSource = string | ArrayBuffer | Uint8Array | Blob`, a renderer that supports every source form; `BaseXmlViewer` is the reference for `content`.
```
ziyunp: Nit - I don't understand this line, maybe it can be removed. The other references are descriptive enough.

`The split is what buys type safety: `src: string` is always a URL and `content: string` is always text, so passing document text to a URL-based viewer is a compile error rather than a silently broken render. `

packages/react-components/CLAUDE.md
```diff
@@ -94,6 +94,10 @@ For every state change with a built-in default behavior (sort, filter, select, e
...
+
+  The base viewers are mid-migration onto this rule. `BaseTiffViewer.content`, `BaseEmailViewer.email` and `BaseSpreadsheetViewer.spreadsheet` are still the old names; each keeps a deprecated alias as it moves.
```
ziyunp: Remove this line, we don't want to expose implementation details here

ziyunp (approved):

PR #3874, opened 2026-08-13
BaseEmailViewer: rename email to content

olliejenningsPT:
Layer 2 of 4. `BaseEmailViewer` takes its parsed email as `content`. `email` named the prop after the file type, which does not generalise to the next viewer.

### Not Changed

- **Nothing is removed** - `email` is still accepted and `@deprecated`; `content` wins when both are set
- **`EmailViewer`** - the OSDK wrapper still takes `media`

### Worth a look

- **`content` has to be optional** while the alias exists, so the viewer now renders its existing `No content` empty state when neither prop is passed. That is a strict improvement on today, where a missing prop throws on `email.html`
- **First tests for this component** - it had none. Covers both prop names, their precedence, and the empty case

### DIFF
--- a/packages/react-components/src/email-viewer/EmailViewer.tsx
+++ b/packages/react-components/src/email-viewer/EmailViewer.tsx
@@ -31,7 +31,7 @@ export function EmailViewer({
   ...emailViewerProps
 }: EmailViewerProps): React.ReactElement {
   const {
-    data: email,
+    data: content,
     loading,
     error,
   } = useMediaContents<ParsedEmail>(media, parseEmailFromResponse);
@@ -52,7 +52,9 @@ export function EmailViewer({
           Failed to load email: {error.message}
         </div>
       )}
-      {email != null && <BaseEmailViewer email={email} {...emailViewerProps} />}
+      {content != null && (
+        <BaseEmailViewer content={content} {...emailViewerProps} />
+      )}
     </div>
   );
 }
--- a/packages/react-components/src/email-viewer/__tests__/EmailViewer.test.tsx
+++ b/packages/react-components/src/email-viewer/__tests__/EmailViewer.test.tsx
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


packages/react-components/src/email-viewer/EmailViewer.tsx
```diff
@@ -52,7 +52,9 @@ export function EmailViewer({
           Failed to load email: {error.message}
         </div>
       )}
-      {email != null && <BaseEmailViewer email={email} {...emailViewerProps} />}
+      {email != null && (
```
ziyunp: Nit: rename email to content to avoid confusion

packages/react-components/src/email-viewer/__tests__/EmailViewer.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+      data: PARSED,
+      loading: false,
+      error: undefined,
+    });
+
+    render(<EmailViewer media={{} as unknown as Media} />);
+
+    const props = mockedBaseEmailViewer.mock.calls[0]?.[0];
+    expect(props).toMatchObject({ content: PARSED });
+    expect(props).not.toHaveProperty("email");
```
ziyunp: Nit - we can remove this assertion, it will be irrelevant once we remove deprecated props
`expect(props).not.toHaveProperty("email");`

packages/react-components/src/email-viewer/__tests__/EmailViewer.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+afterEach(() => {
+  cleanup();
+  vi.clearAllMocks();
+});
+
+// Pins the prop name the wrapper forwards under. The deprecated `email` alias
+// behaves identically, so without this a revert to the old name would leave
+// every other test green.
+describe("EmailViewer", () => {
+  it("should forward the fetched email as content, not as the deprecated email prop", () => {
```
ziyunp, line 55: Nit - remove `not as the deprecated email prop`

ziyunp (dismissed):

ziyunp (approved):

olliejenningsPT:
closed by https://github.com/palantir/osdk-ts/pull/3885

PR #3875, opened 2026-08-13
BaseSpreadsheetViewer: rename spreadsheet to content

olliejenningsPT:
Layer 3 of 4, structurally identical to layer 2. `BaseSpreadsheetViewer` takes its parsed workbook as `content`.

### Not Changed

- **Nothing is removed** - `spreadsheet` is still accepted and `@deprecated`; `content` wins when both are set
- **`SpreadsheetViewer`** - the OSDK wrapper still takes `media`

### Worth a look

- **`EMPTY_SHEETS` constant** - the sheet list is read through a module-level constant so the `useMemo` keeps a stable reference when neither prop is passed, rather than allocating a new `[]` per render. Falls through to the existing `No sheets` state
- **First tests for this component** - it had none. Covers both prop names, their precedence, and the empty case

### DIFF
--- a/packages/react-components/src/spreadsheet-viewer/__tests__/BaseSpreadsheetViewer.test.tsx
+++ b/packages/react-components/src/spreadsheet-viewer/__tests__/BaseSpreadsheetViewer.test.tsx
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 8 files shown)


packages/react-components/src/spreadsheet-viewer/__tests__/SpreadsheetViewer.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      data: PARSED,
+      loading: false,
+      error: undefined,
+    });
+
+    render(<SpreadsheetViewer media={{} as unknown as Media} />);
+
+    const props = mockedBaseSpreadsheetViewer.mock.calls[0]?.[0];
+    expect(props).toMatchObject({ content: PARSED });
+    expect(props).not.toHaveProperty("spreadsheet");
```
ziyunp: Nit - no need to assert that deprecated prop is not used

packages/react-components/src/spreadsheet-viewer/__tests__/BaseSpreadsheetViewer.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+        ["Region", "Total"],
+        ["Europe", "42"],
+      ],
+    },
+  ],
+};
+
+const TWO_SHEETS: ParsedSpreadsheet = {
+  sheets: [
+    { name: "Revenue", rows: [["Region"], ["Europe"]] },
```
ziyunp: Europe is in both test spreadsheets, can we make the rows different in the two examples

ziyunp (dismissed):

olliejenningsPT:
closed by https://github.com/palantir/osdk-ts/pull/3885

PR #3876, opened 2026-08-13
BaseTiffViewer: rename content to src

olliejenningsPT:
Layer 4 of 4, and the one worth arguing about. `BaseTiffViewer` takes its bytes as `src`, not `content`.

### Why src and not content

TIFF bytes are a binary source, so they belong under `src` alongside `BasePdfViewer` (`src: string | ArrayBuffer | Uint8Array | Blob`). `content` is reserved for already-decoded payloads, which is what `BaseXmlViewer` and `BaseMarkdownViewer` take.

The counter-argument, stated plainly: in HTML `src` is always a locator and never bytes, so `src: Uint8Array` reads oddly. That is the cost `BasePdfViewer` already pays today, and being consistent inside the package seemed worth more than matching HTML.

### Not Changed

- **Nothing is removed** - `content` is still accepted and `@deprecated`; `src` wins when both are set
- **Its deprecation note says why, not just what** - unlike `email` and `spreadsheet`, `content` is not an obviously wrong name here, so the tag explains the distinction

### Also in here

- **A dead omit key** - `DocumentViewerApi` omitted `"className"` from `BaseTiffViewerProps` twice, a member that type never had
- **`docs/DocumentViewer.md` regenerated** - its generated table prints the `tiffViewerProps` type text, so `check-gen-props` fails without a re-run


packages/react-components/src/images/tiff-viewer/__tests__/TiffViewer.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+afterEach(() => {
+  cleanup();
+  vi.clearAllMocks();
+});
+
+// Pins the prop name the wrapper forwards under. The deprecated `content` alias
+// behaves identically, so without this a revert to the old name would leave
+// every other test green.
+describe("TiffViewer", () => {
+  it("should forward the fetched bytes as src, not as the deprecated content prop", () => {
```
ziyunp: Nit - remove `not as the deprecated content prop`

ziyunp (dismissed):

olliejenningsPT:
https://github.com/palantir/osdk-ts/pull/3885

PR #3878, opened 2026-08-14
Run the CI jobs on the ramdisk, and drop the pnpm store cache

olliejenningsPT:
This delivers the **RAM Cache** follow up from the previous PR in the stack.

One correction up front: that PR estimated 30s-45s per stage. Measured, it is closer to 7s-8s per stage, so the optimisation is smaller than hoped.

### Not Changed

- **Release workflow** - untouched
- **Current CI** - still additive, GitHub CI is unaffected
- **Every job's resource class** - unchanged except `build-apps`, see below

### What changed

- **Every job runs on `/mnt/ramdisk`** - CircleCI's docker executor pre-mounts a tmpfs there. It is the executor default rather than a per job override, because it has to be the same path everywhere
- **`install` moved too** - not just the jobs that restore node_modules from the workspace
- **pnpm store cache removed** - it cost time rather than saving it
- **`build-apps` moved to `2xlarge`** - the only lasting cost in this PR
- **Turbo cache keys bumped to v3** - cache paths are recorded absolutely, so the old entries point at `/home/circleci/project`

### Why it has to be every job

This is the "re-wiring for pnpm" the previous PR mentioned, and it is worth writing down because the failure is very misleading.

pnpm bakes an absolute `NODE_PATH` into every `node_modules/.bin` shim at install time, ending in `.pnpm/node_modules`, and Babel resolves undeclared presets through it. Move only the downstream jobs and every one of those baked paths points at a directory that does not exist, so you get `ERR_MODULE_NOT_FOUND` from a tree that is otherwise completely correct. Nothing to do with tmpfs, mount boundaries or the resolvers.

### The numbers

- **Workspace attach** - 22s-25s down to 14s-18s, across the 13 jobs that attach it
- **Measured directly** - the same 2.3 GB workspace attached twice in one container, disk first so tmpfs got no warm-path advantage: 23.4s against 16.0s
- **pnpm store cache** - restoring 449 MiB took 8.6s and only moved the install from 25.9s to 23.2s, so it cost about 6s a run, plus a 17.8s save whenever the lockfile changed
- **Memory** - peaks run 2.6 GiB to 11.6 GiB, with `cspell` tightest at 3.16 of 4 GiB

### Why `build-apps` needs 2xlarge

It peaked at 17.56 GiB against xlarge's 16, and 2.3 GB of that is the tmpfs itself. Three alternatives were each measured on a forced build and all were worse:

- `--concurrency=6` fits xlarge, but takes the build step from 53s to 162s
- `parallelism: 2` over two xlarge containers costs the same as one 2xlarge and finishes in 154s, gated by the heavier half
- `--concurrency=100%` takes it to 104s, because these tasks already spawn their own worker pools and oversubscribe the cores

Worth knowing: turbo's `--concurrency` defaults to 10 rather than the core count, so the extra vCPUs give those 10 tasks more room rather than running more of them.

### Also fixed

Three things this work turned up, all reproduced locally before fixing:

- **`transpileEsm` now cascades** - it never declared `^transpileEsm` while `transpileCjs` always has, so building a package's ESM did not guarantee the ESM of what it imports at runtime. `e2e.generated.catchall#codegen` could start before the CLI it invokes had an importable dependency closure and fail with `ERR_MODULE_NOT_FOUND`. A cached codegen normally hid it, which is why it has never been seen on GitHub CI
- **Truncated JUnit reports are skipped** - `store_test_results` hard fails on malformed XML, so a vitest killed mid-write turned a test failure into `failed uploading test results: end of token stream without closing testsuites element`


turbo.json
nihalbhatnagar: does this need to depend on ^transpileBrowser too?

turbo.json
```diff
@@ -200,6 +200,7 @@
       "dependsOn": [
         "//#babel-config",
         "@osdk/monorepo.tsconfig#typecheck",
+        "^transpileEsm",
```
nihalbhatnagar, line 204: This seems right according to the original [PR](https://github.com/palantir/osdk-ts/pull/1100) that added these tasks. Seems like after removing we did run into these issues https://github.com/palantir/osdk-ts/pull/3565. Can we add a comment about why this cant be parallel here so we dont remove

nihalbhatnagar (dismissed):

nihalbhatnagar (approved):

PR #3880, opened 2026-08-14
BaseTiffViewer: accept className

olliejenningsPT:
Layer 5 of the stack. `BaseTiffViewer` accepts `className`, like every other base viewer.

### The inconsistency

- **`BaseTiffViewerProps` had no `className`** while the other seven base viewers all do
- **It had no root element either** — it returned `null`, a bare error `div`, or a bare `canvas`, so there was nowhere to put one
- **`TiffViewerProps` re-declared `className` by hand** to compensate
- **`DocumentViewerApi` omitted a `className` key the interface never had.** That omit was not a mistake: `pdfViewerProps`, `imageViewerProps` and `videoViewerProps` all omit `className` too, so the TIFF bag was written for the shape TIFF should have had

### What changed

- **Root element in every state** carrying `classnames(styles.container, className)`, matching `BaseImageViewer`
- **`.container` is layout-only** (`width`/`height: 100%`). Rendering is unchanged: the canvas was already `100%`/`100%` and now resolves against the same grandparent as before, and the shared error styles are plain flex with no positioning or child selectors
- **`TiffViewerProps` drops its hand-rolled `className`** and inherits it, exactly as `ImageViewerProps` does. No change for consumers

### Not Changed

- **No new CSS custom properties.** `BaseImageViewer`'s container also carries a background, border and radius from `--osdk-image-viewer-*`. Giving TIFF the same would be a visible change and needs new tokens plus `CSSVariables.md` entries, so it is left for its own PR

### Worth a look

- **`BaseTiffViewer` no longer early-returns `null`** before decoding finishes. That is required for `className` to apply in every state, and it brings the component in line with the package rule about rendering during loading and error states. The one existing test that asserted this is retitled from "renders nothing" to "renders no canvas"
- **Visual check belongs in CI.** The container is layout-only by construction, but the browsable storybook artifact is the place to confirm the TIFF stories are unchanged

### DIFF
--- a/.changeset/tiff-viewer-classname-prop.md
+++ b/.changeset/tiff-viewer-classname-prop.md
@@ -1,0 +1,7 @@
+---
+"@osdk/react-components": minor
+---
+
+`BaseTiffViewer` accepts `className`, like every other base viewer. It renders a root element in every state to carry it, where before it rendered nothing until decoding finished.
+
+`TiffViewerProps` is unchanged for consumers: it now inherits `className` from `BaseTiffViewerProps` instead of declaring its own. `DocumentViewer`'s `tiffViewerProps` omits `className`, matching `pdfViewerProps`, `imageViewerProps` and `videoViewerProps`.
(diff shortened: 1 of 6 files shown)


ziyunp (dismissed):

ziyunp (approved):

olliejenningsPT:
Closed by https://github.com/palantir/osdk-ts/pull/3885

PR #3886, opened 2026-08-14
Add experimental link subscriptions

jbusa22:
## Summary

- add an experimental `subscribeToLinks` API for subscribing to directed link changes from selected OSDK objects
- infer the object type from the first object and require every remaining object to have that same concrete type
- infer valid directed link API names and each update target type from the generated ontology types
- deliver backend update batches over the ontology subscriptions WebSocket
- reuse the existing object-set subscription retry and WebSocket test infrastructure

## Usage

This example watches two managers for changes to their direct reports. `fetchOne` returns real `Osdk.Instance<Employee>` values, and `peeps` is the generated directed link API name from a manager to their reports.

```ts
import {
  __EXPERIMENTAL__NOT_SUPPORTED_YET__linkSubscriptions,
} from "@osdk/api/unstable";
import { Employee } from "./generated/ontology/objects/Employee.js";

const [managerOne, managerTwo] = await Promise.all([
  client(Employee).fetchOne(managerOneId),
  client(Employee).fetchOne(managerTwoId),
]);

# Still need to get the links from a separate service
const directReportsLink = "peeps"

const subscription = client(
  __EXPERIMENTAL__NOT_SUPPORTED_YET__linkSubscriptions,
).subscribeToLinks({
  objects: [managerOne, managerTwo],
  links: [directReportsLink],
  listener: {
    onSuccessfulSubscription() {
      console.log("Listening for direct-report changes");
    },
    onChange({ updates }) {
      for (const update of updates) {
        // source is the manager; target is the added or removed report.
        console.log(update.state, update.source, update.target);
      }
    },
    onOutOfDate({ links }) {
      // Re-fetch any local state derived from these links.
      console.log("Links need refreshing", links);
    },
    onError({ error, subscriptionClosed }) {
      console.error({ error, subscriptionClosed });
    },
  },
});

subscription.unsubscribe();
```


packages/client/src/objectSet/LinkSubscriptionWebsocket.ts
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  subscribe(): { readonly unsubscribe: () => void } {
+    void this.#prepare();
+    return { unsubscribe: () => this.#close() };
+  }
+
+  async #prepare(): Promise<void> {
+    try {
+      const [firstObject] = this.#args.objects;
```
jbusa22: I think ts should enforce that all objects are the same type so this is safe

packages/client/src/createClient.ts
```diff
@@ -165,6 +166,7 @@ export function createClientFromContext(clientCtx: MinimalClient) {
       | QueryDefinition<any>
       | Experiment<"2.0.8">
       | Experiment<"2.1.0">
+      | Experiment<"2.17.0">
```
nihalbhatnagar: This probably needs to be `2.59` and updated to the version at which we ship osdk/client. @ssanjay1 we def need to review this paradigm since we're shipping minor versions rapidly now

pnpm-workspace.yaml
```diff
@@ -29,6 +29,6 @@ catalogs:
     "@osdk/foundry.geo": 2.70.0
     "@osdk/foundry.admin": 2.70.0
     "@osdk/foundry.mediasets": 2.70.0
-    "@osdk/foundry.ontologies": 2.70.0
+    "@osdk/foundry.ontologies": 2.73.0
```
nihalbhatnagar: Lets go ahead and bump the rest to keep in sync

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -40,12 +40,12 @@ import type { SubscriptionConnection } from "../SubscriptionConnection.js";
 import { ExponentialBackoff } from "../util/exponentialBackoff.js";
 
 const MINIMUM_RECONNECT_DELAY_MS = 5 * 1000;
-const EXPONENTIAL_BACKOFF_INITIAL_DELAY_MS = 1000;
-const EXPONENTIAL_BACKOFF_MAX_DELAY_MS = 60000;
-const EXPONENTIAL_BACKOFF_MULTIPLIER = 2;
-const EXPONENTIAL_BACKOFF_JITTER_FACTOR = 0.3;
-const WEBSOCKET_IDLE_DISCONNECT_DELAY_MS = 15000;
-const WEBSOCKET_HEARTBEAT_INTERVAL_MS = 45 * 1000;
+export const EXPONENTIAL_BACKOFF_INITIAL_DELAY_MS: number = 1000;
```
nihalbhatnagar: Probably should move these to a shared util file instead of exporting from here

packages/client/src/objectSet/LinkSubscriptionWebsocket.ts
```diff
@@ -0,0 +1,413 @@
...
+} {
+  return subscription.request != null;
+}
+
+function isDone(subscription: Subscription<any, any>): boolean {
+  return subscription.status === "done" || subscription.status === "error";
+}
+
+/** @internal */
+export class LinkSubscriptionWebsocket {
```
nihalbhatnagar: Can we create a super class that both of these websockets now use for websocket connection/retry/heartbeat logic? I'd imagine that we can basically reuse onMessage, onOpen, initiateSubscribe, all the way upto the specific message ser/de and request payloads?

packages/client/src/objectSet/ObjectSetListenerWebsocket.ts
```diff
@@ -810,9 +810,39 @@ export class ObjectSetListenerWebsocket {
 
 /** @internal */
 export function constructWebsocketUrl(baseUrl: string, ontologyRid: string) {
```
nihalbhatnagar: Yeah I think keeping these in the same utils file as above will be helpful

packages/api/src/experimental/subscribeToLinks.ts
```diff
@@ -0,0 +1,90 @@
...
+  > {
+    readonly links: readonly [L, ...ReadonlyArray<NoInfer<L>>];
+    readonly listener: Listener<NoInfer<Q>, NoInfer<L>>;
+    readonly objects: readonly [
+      Osdk.Instance<Q>,
+      ...ReadonlyArray<Osdk.Instance<NoInfer<Q>>>,
+    ];
+  }
+
+  export interface ChangeEvent<
```
nihalbhatnagar: Maybe LinkUpdate to match ObjectSetUpdate?

nihalbhatnagar: Can we flatten this out too? We did the same for object set updates where youc an get multiple object updates back on one message, but the handler should be triggered once per specific update/object

packages/api/src/experimental/subscribeToLinks.ts
```diff
@@ -0,0 +1,90 @@
...
+import type { ObjectIdentifiers } from "../OsdkBase.js";
+import type { Osdk } from "../OsdkObjectFrom.js";
+import type { Experiment } from "./Experiment.js";
+
+export namespace LinkSubscription {
+  export interface Args<
+    Q extends ObjectTypeDefinition,
+    L extends LinkTypeApiNamesFor<Q>,
+  > {
+    readonly links: readonly [L, ...ReadonlyArray<NoInfer<L>>];
```
nihalbhatnagar: Can we leave this as a regular array? (NoInfer is fine). We generally try to avoid non-empty array type in this codebase

packages/api/src/experimental/subscribeToLinks.ts
```diff
@@ -0,0 +1,90 @@
...
+  > = L extends LinkTypeApiNamesFor<Q> ? {
+      readonly linkType: L;
+      readonly source: ObjectIdentifiers<Q>;
+      readonly state: "ADDED" | "REMOVED";
+      readonly target: ObjectIdentifiers<LinkedObjectType<Q, L>>;
+    }
+    : never;
+}
+
+type SubscribeToLinks = <
```
nihalbhatnagar, line 69: Can we add javadoc for this method. Also, will this work for updates to one to many link types backed by a FK? If not, we should document that here

packages/api/src/experimental/subscribeToLinks.test.ts
```diff
@@ -0,0 +1,52 @@
...
+>["subscribeToLinks"];
+
+describe("subscribeToLinks", () => {
+  const employeeOne = {} as Osdk.Instance<EmployeeApiTest>;
+  const employeeTwo = {} as Osdk.Instance<EmployeeApiTest>;
+  const subscribeToLinks: SubscribeToLinks = () => ({
+    unsubscribe: () => {},
+  });
+
+  it("infers links and updates from the first object", () => {
```
nihalbhatnagar: Thank you for the type test

packages/client/src/objectSet/LinkSubscriptionWebsocket.test.ts
```diff
@@ -0,0 +1,231 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("multiplexes subscriptions and routes update batches", async () => {
```
nihalbhatnagar: Could you make this description a bit more clear, like "uses one WebSocket for multiple subscriptions and sends updates to the correct listener", I was a bit confused by the wording

packages/api/src/experimental/subscribeToLinks.ts
```diff
@@ -28,26 +28,16 @@ export namespace LinkSubscription {
...
-      ...ReadonlyArray<Osdk.Instance<NoInfer<Q>>>,
-    ];
-  }
-
-  export interface ChangeEvent<
-    Q extends ObjectTypeDefinition,
-    L extends LinkTypeApiNamesFor<Q>,
-  > {
-    readonly updates: ReadonlyArray<Update<Q, L>>;
+    readonly objects: ReadonlyArray<Osdk.Instance<Q>>;
```
nihalbhatnagar: We should probably have this mimic the queries and actions API that take objects like here https://github.com/palantir/osdk-ts/blob/830d7919c09d774762ce9e02f8f8d85cfb955c76/packages/api/src/queries/Queries.ts#L57 , which probably means just using that same union

packages/client/src/objectSet/LinkSubscriptionWebsocket.ts
```diff
@@ -385,19 +290,19 @@ export class LinkSubscriptionWebsocket {
   }
 
   #fail(subscription: Subscription<any, any>, error: unknown): void {
-    if (isDone(subscription)) return;
+    if (isSubscriptionDone(subscription)) return;
     try {
       subscription.listener.onError?.({ error, subscriptionClosed: true });
```
nihalbhatnagar, line 297: Should we try catch this error call too?

packages/api/src/experimental/subscribeToLinks.test.ts
nihalbhatnagar: Nitty but I think the API should be (Employee, {listener, objects,links})

nihalbhatnagar (dismissed):

nihalbhatnagar (dismissed):

nihalbhatnagar (dismissed):

nihalbhatnagar (dismissed):

nihalbhatnagar (approved):

PR #3892, opened 2026-08-17
[FilterList] add onFilterListChanged callback

ziyunp:
Consumers currently have to wire up onFilterStateChanged, onFilterClauseChanged and onEffectiveObjectSet separately to react to one filter change.

## Changes:
- Added `onFilterListChanged` that reports every filter-state change once, with everything derived from it in a single call.
- Marked the 3 existing callbacks as deprecated.
- Changed FilterKey type to `string` as the previous union resolves to string anyway.

## onFilterListChanged

Signature: `(event: FilterChangeEvent<Q>) => void`

The event has two fields:

- `snapshot: FilterChangeSnapshot<Q>` — everything derived from the resulting state:
  - `filterClause` — combined clause for all active filters (`HAS_LINK` and `LINKED_PROPERTY` not represented here)
  - `filteredObjectSet` — the `objectSet` prop filtered by all active filters, `HAS_LINK` and `LINKED_PROPERTY` included; `undefined` when no `objectSet` prop was supplied
- `reason: FilterChangeReason` — a discriminated union describing what changed:

| `event.reason.type` | Extra fields | Fires when |
| --- | --- | --- |
| `FILTER_LIST_INITIALIZED` | — | The list mounted, reporting the state it started with |
| `FILTER_STATE_CHANGED` | `filterKey`, `newState` | A filter's state was set — selected values changed, include/exclude mode toggled, or filter cleared |
| `FILTER_REMOVED` | `filterKey` | A filter was removed from the list |
| `FILTER_LIST_RESET` | — | Every filter was restored to the state it mounted with |

## Note:
After discussion with @nfdiop, we decided to keep the snapshot and reason together in a single event object rather than splitting them into two callback params. This gives us more flexibility to add or reshape fields on the event without breaking the callback signature.


packages/react-components/src/filter-list/types/LinkedFilterTypes.ts
```diff
@@ -37,7 +37,7 @@ import type {
  */
 export type LinkedFilter<Q extends ObjectTypeDefinition> = {
   [L in LinkNames<Q>]: {
-    /** Stable identity for the filter */
+    /** The filter definition's `getFilterKey` value. */
```
nfdiop, line 40: is it expected?

packages/react-components/src/filter-list/FilterListApi.ts
```diff
@@ -56,17 +57,46 @@ export type FilterDefinitionUnion<Q extends ObjectTypeDefinition> =
...
+   * The `objectSet` prop filtered by all active filters, including `HAS_LINK`
+   * and `LINKED_PROPERTY`, or `undefined` when no `objectSet` was supplied.
+   */
+  filteredObjectSet: ObjectSet<Q> | undefined;
+
+  /**
+   * Every active filter in `filterDefinitions` order, each tagged with the
+   * `kind` of its definition.
+   */
+  activeFilters: ReadonlyArray<ActiveFilter<Q>>;
```
nfdiop: Could you expand on why we need to give users the `activeFilters`?

packages/react-components/src/filter-list/FilterListApi.ts
```diff
@@ -115,16 +147,30 @@ export interface FilterListProps<Q extends ObjectTypeDefinition> {
...
   ) => void;
 
+  /**
+   * Called whenever filter state changes — a filter set, a filter cleared, or a
+   * reset — with the resulting clause, filtered `ObjectSet` and active filters
+   * in a single payload.
+   *
+   * @param change What changed and the filter state it produced
+   */
+  onFilterChanged?: (change: FilterChangeEvent<Q>) => void;
```
nfdiop: comments:
- Should we have  `onFiltersChanged` instead? 
- What if we have something like this (that is the genea
```
export interface FilterSnapshot<Q extends ObjectTypeDefinition> {
    readonly filterStates: ReadonlyMap<string, FilterState>;
    readonly filterClause: WhereClause<Q>;
    readonly filteredObjectSet: ObjectSet<Q> | undefined;
  }

  export type FilterChangeDetails =
    | {
        readonly reason: "filterStateChange";
        readonly filterKey: string;
      }
    | {
        readonly reason: "filterRemoved";
        readonly filterKey: string;
      }
    | { readonly reason: "reset" }
    | { readonly reason?: undefined };

  export interface FilterListProps<Q extends ObjectTypeDefinition> {
    onFilterChange?: (
      snapshot: FilterSnapshot<Q>,
      details: FilterChangeDetails,
    ) => void;
  }
```
Do you have a sense how people are using this API?

packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -126,52 +183,92 @@ export function useFilterListState<Q extends ObjectTypeDefinition>(
     () => new Map(initialFilterStatesSnapshot),
   );
 
+  // Back-to-back writes can occur before React commits, so each transition must
+  // build on the latest eagerly computed state rather than the render snapshot.
+  const filterStatesRef = useRef(filterStates);
+  filterStatesRef.current = filterStates;
+
+  const applyChange = useCallback(
```
nfdiop, line 202: note: we no longer call the new callback on mount like we did before

packages/react-components/src/filter-list/types/LinkedFilterTypes.ts
ziyunp: Yes, it's more clear to explain this cause we also expose getFilterKey function

packages/react-components/src/filter-list/FilterListApi.ts
ziyunp: There was a few use cases - one is our users have been tracking all active filters to persist the state in the client app. The other use case is client app may want to render all active filters somewhere.

packages/react-components/src/filter-list/FilterListApi.ts
ziyunp: Yes, I analysed our early adopter's source code and came up with the list of params that would be helpful to them.

I intentionally name it onFilterChanged because the callback is triggered on a single filter change, and the change is represented by filterKey and newState. The other params are expressing the current global state of FilterList.

The structure you proposed here restricts the params positioning and will be a breaking change later if we decide to add or remove the params. I did a single object param so it's easy to change later. 

I don't see a strong reason to rewrite the callback, maybe can you explain why you proposed this API?

packages/react-components/src/filter-list/hooks/useFilterListState.ts
ziyunp: I initially think we should not call it on mount because nothing has changed yet. But maybe we need to, when user passed default filter states and they want to get back the filter clause, object set etc. Let me fix.

packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -69,28 +78,74 @@ function buildInitialStates<Q extends ObjectTypeDefinition>(
...
   const filterDefinitionsRef = useRef(filterDefinitions);
   filterDefinitionsRef.current = filterDefinitions;
+  const objectSetRef = useRef(objectSet);
+  objectSetRef.current = objectSet;
+
+  const emitFilterChanged = useDebouncedCallback(
+    (change: FilterChangeEvent<Q>) => {
+      onFilterChangedRef.current?.(change);
+    },
+    ON_FILTER_CHANGED_DEBOUNCE_MS,
```
ziyunp: Reviewing whether to keep this debounce - maybe to debounce at the relevant input components

packages/react-components/src/filter-list/hooks/useFilterListState.ts
nfdiop: Since this is a refactor, we need to keep the existing behaviour

packages/react-components/src/filter-list/FilterListApi.ts
nfdiop: Proposed this API since I found it more readable to have:
- the reason for the change saved in another param.  
- all the snapshot data grouped together

packages/react-components/src/filter-list/FilterListApi.ts
nfdiop: Do we need to support this in this refactoring PR? My worry is that this PR is extending the API when we use these activeFilters.

ziyunp: I will remove activeFilters

packages/react-components/src/filter-list/hooks/useFilterListState.ts
ziyunp: I checked in the storybook on main and only onFilterClauseChanged is emitted on mount, and that behaviour is the same in this PR. Which callback exactly is it that was called on mount and no longer called?
 
I'm adding an INIT event to call onFilterChanged so user can get the initial filter clause and filteredObjectSet.

packages/react-components/src/filter-list/types/LinkedFilterTypes.ts
nfdiop: Hum, it is surprising because this file doesn't reference  `getFilterKey` unless I'm missing something. So it is weird to mention here

packages/react-components/src/filter-list/hooks/useFilterListState.ts
nfdiop: I am seeing these two effects here. Can you elaborate on why it is OK to drop them? 

```
 useEffect(() => {
    onFilterClauseChangedRef.current?.(whereClause);
  }, [whereClause]);

  useEffect(() => {
    if (effectiveObjectSet == null) {
      return;
    }
    onEffectiveObjectSetRef.current?.(effectiveObjectSet);
  }, [effectiveObjectSet]);
```

packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -126,54 +191,88 @@ export function useFilterListState<Q extends ObjectTypeDefinition>(
...
-        ? undefined
-        : narrowObjectSet(objectSet, whereClause, linkedFilters),
-    [objectSet, whereClause, linkedFilters],
+      deriveSnapshot(filterDefinitions, filterStates, propertyTypes, objectSet),
+    [filterDefinitions, filterStates, propertyTypes, objectSet],
   );
 
+  const hasEmittedInit = useRef(false);
+  useEffect(() => {
+    // `propertyTypes` is empty until metadata resolves, and a keyword search
```
nfdiop: it looks surprising to mention `propertyTypes` here when they are not immediately used. Do we need to keep (and do we need to also wait for metadata loading)?

packages/react-components/src/filter-list/FilterListApi.ts
```diff
@@ -115,16 +136,32 @@ export interface FilterListProps<Q extends ObjectTypeDefinition> {
...
 
+  /**
+   * Called on filter init and when filter state changes.
+   *
+   * @param event What changed
+   * @param snapshot The filter state the change produced
+   */
+  onFilterChanged?: (
+    event: FilterChangeEvent,
+    snapshot: FilterChangeSnapshot<Q>,
```
nfdiop: nit: general convention I have seen is to have the state first in case the user want to ignore the event, they don't have to do `onFilterChanged(_, snapshot)`

packages/react-components/src/filter-list/types/index.ts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+export * from "./ActiveFilterTypes.js";
```
nfdiop: Sorry if I missed this but are we still using this?

packages/react-components/src/filter-list/FilterListApi.ts
ziyunp: I will keep this order. Here the event is the main focus as that is what changed. It contains info about the change as well. Snapshot contains additional info. I think our users will need to read from both fields to get new filter state to persist in the client app state, filterClause, filteredObjectSet anyway.

packages/react-components/src/filter-list/types/index.ts
ziyunp: Good catch, cleaning up

packages/react-components/src/filter-list/hooks/useFilterListState.ts
ziyunp: The comment is there to explain your question on why we need to wait for metadata loading. I will rephrase it to make it clearer

packages/react-components/src/filter-list/hooks/useFilterListState.ts
ziyunp: They are not dropped in this PR. If you mean to drop them after deprecated, it is ok to drop because onFilterChanged now emits with an INIT event on mount, and are called with whereClause and effectiveObjectSet

packages/react-components/src/filter-list/hooks/useFilterListState.ts
```diff
@@ -126,54 +192,87 @@ export function useFilterListState<Q extends ObjectTypeDefinition>(
...
-        ? undefined
-        : narrowObjectSet(objectSet, whereClause, linkedFilters),
-    [objectSet, whereClause, linkedFilters],
+      deriveSnapshot(filterDefinitions, filterStates, propertyTypes, objectSet),
+    [filterDefinitions, filterStates, propertyTypes, objectSet],
   );
 
+  const hasEmittedInit = useRef(false);
+  useEffect(() => {
+    // We need to wait until metadata is loaded because we need
```
nfdiop, line 271: we can keep but I don't like this type of comment since if:
- we don't need to wait for metadataLoading
- we need to wait for something else
this comment can be outdated. Whenever a value is referenced, it should be visibly used nearby.

packages/react-components/src/filter-list/FilterListApi.ts
```diff
@@ -56,17 +56,36 @@ export type FilterDefinitionUnion<Q extends ObjectTypeDefinition> =
...
+   * and `LINKED_PROPERTY`, or `undefined` when no `objectSet` was supplied.
+   */
+  filteredObjectSet: ObjectSet<Q> | undefined;
+}
+
+/**
+ * What the user did to the filter state.
+ */
+export type FilterChangeEvent =
+  /** The list mounted, reporting the state it started with. */
```
nfdiop: I've been thinking more about this. Let's keep the API simple/clear and just send

```onFilterChanged(FilterChangeSnapshot & { filterStates: ....})```. Users should be able to figure out what changed.

ziyunp: I do not agree with letting user figure out what changed. If we send the full filter states, user will have to diff it with the previous state to find out what changed. It might be a nested key that has changed, e.g. the values array within the newState object. User will have to do a deep comparison to figure out what changed. The point of this callback is to tell user the exact filter that has changed. 

Image


If you are thinking that this onFilterChanged is giving too much info, I'm happy to keep the multiple callback approach and user listens to different callback depending on what they need.

onFilterStateChanged - to get the diff of what changed
onFilterClauseChanged - to get the full filter clause
onEffectiveObject - maybe rename to onFilteredObjectSetChanged

ziyunp: Resolved this with the new structure where we have two fields in event - snapshot and reason. Renamed the reasons to be clearer.

export type FilterChangeReason =
  /** A filter's state was set — selected values changed, include/exclude mode toggled, or filter cleared. */
  | { type: "FILTER_STATE_CHANGED"; filterKey: string; newState: FilterState }
  /** A filter was removed from the list. */
  | { type: "FILTER_REMOVED"; filterKey: string }
  /** Every filter was restored to the state it mounted with. */
  | { type: "FILTER_LIST_RESET" }
  /** The list mounted, reporting the state it started with. */
  | { type: "FILTER_LIST_INITIALIZED" };

nfdiop (approved):

PR #3893, opened 2026-08-17
catalog version bump to 2.73.0

lucy-k-wang:
## BLUF

bump catalog entries from `2.70.0` -> `2.73.0` to pick up api-gateway change (fix `TranscriptionLanguage` to be `"NO"` instead of the  incorrect `"false"` YAML coercion. 

## Changes in this PR
bump catalog entries to 2.73.0 to pick up api-gateway change

## FLUP

pre-work for media transformation param improvement


bryantpark04 (approved):
Mostly same as #3852

PR #3895, opened 2026-08-17
Prefetch interface metadata during page loads

nihalbhatnagar:
## Summary
- prefetch interface and implementing object metadata while loading interface pages
- reuse in-flight cached metadata during object conversion
- verify metadata requests start before object-set loads and occur only once

## Testing
- `pnpm exec vitest run src/object/fetchPage.test.ts`
- pre-commit lint, format, and spelling checks

### DIFF
--- a/packages/client/src/object/fetchPage.test.ts
+++ b/packages/client/src/object/fetchPage.test.ts
@@ -25,7 +25,8 @@ import type {
 } from "@osdk/api";
 import { Employee, FooInterface, Todo } from "@osdk/client.test.ontology";
 import type { SearchJsonQueryV2 } from "@osdk/foundry.ontologies";
-import { describe, expect, expectTypeOf, it } from "vitest";
+import { LegacyFauxFoundry, startNodeApiServer } from "@osdk/shared.test";
+import { describe, expect, expectTypeOf, it, vi } from "vitest";
 
 import { createMinimalClient } from "../createMinimalClient.js";
 import {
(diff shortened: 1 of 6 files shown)


packages/client/src/object/fetchPage.test.ts
```diff
@@ -588,4 +589,83 @@ describe(fetchPage, () => {
...
+        expect(result.data).not.toHaveLength(0);
+        expect(result.data[0].$apiName).toBe("Employee");
+        expect(result.data[0].$objectType).toBe("Employee");
+        expect(employeeMetadataRequests).toHaveLength(1);
+      } finally {
+        testSetup.apiServer.close();
+      }
+    });
+
+    it("prefetches interface and implementing object metadata before conversion", async () => {
```
nihalbhatnagar: Test fails without this new change which does mean we were waiting on metadata AFTER we fired the fetch page

packages/client/src/object/fetchPage.ts
```diff
@@ -329,6 +329,18 @@ async function fetchInterfacePage<
       interfaceType.apiName,
     );
     allProperties = ifaceDef ? Object.keys(ifaceDef.properties) : undefined;
+  } else {
+    void client.ontologyProvider
+      .getInterfaceDefinition(interfaceType.apiName)
+      .then((def) => {
+        def.implementedBy?.forEach((implementedBy) =>
+          // oxlint-disable-next-line promise/no-nesting
```
nihalbhatnagar: I prefer the inline rather than extracting to a helper here

esinx: Maybe `Promise.allSettled`?
```typescript
Promise.allSettled(
  def.implementedBy?.map((implementedBy) =>
    client.ontologyProvider.getObjectDefinition(implementedBy),
  ) ?? [],
),
```

nihalbhatnagar: Good suggestion, updated to use Promise.allSettled in 0e58164eeb. This also removes the lint suppression.

packages/client/src/object/fetchPage.test.ts
```diff
@@ -588,4 +589,83 @@ describe(fetchPage, () => {
...
+        );
+        const interfaceMetadataRequests = fetchFn.mock.calls.filter(([input]) =>
+          String(input).includes("interfaceTypes/FooInterface"),
+        );
+        const employeeMetadataRequests = fetchFn.mock.calls.filter(([input]) =>
+          String(input).includes("objectTypes/Employee/fullMetadata"),
+        );
+
+        expect(interfaceMetadataRequestIndex).toBeGreaterThanOrEqual(0);
+        expect(fetchPageRequestIndex).toBeGreaterThan(
```
ssanjay1: Shouldn't we also check that the object metadata load request index is less than the fetch page one?

nihalbhatnagar: Updated using pdefer to instead make sure the request goes out before fetch page is resolved

ssanjay1 (approved):

PR #3900, opened 2026-08-18
Fix `HAS_LINK` and `LINKED_PROPERTY` to filter by aggregated count

ziyunp:
- Fix FilterList's `HAS_LINK` filter, which sent an unsupported `$isNotNull` filter. 
- Fix LINKED_PROPERTY filter excluding mode to keep rows with no link.
- `HAS_LINK` and `LINKED_PROPERTY` now filter by deriving a count of matching linked objects. 
- LINKED_PROPERTY filter is no longer using object sets intersect operation.
- reverseLinkName is no longer required for linked property filters.
- Remove bad documentation `$isNotNull` in `useOsdkObjects` snippets.

https://github.com/user-attachments/assets/76fd8fc9-6c73-4f55-8224-7f9375a897b0


packages/react-components/src/filter-list/utils/filterStateToWhereClause.ts
```diff
@@ -219,7 +221,7 @@ export function buildPropertyKeyClause(
   key: string | number | symbol,
   state: FilterState,
   propertyType?: string,
-): Record<string, unknown> | undefined {
+): WhereClauseFragment | undefined {
```
ziyunp, line 224: Type fix to prevent a no-op like isNotNull to be defined

packages/react-components/src/filter-list/utils/narrowObjectSet.ts
```diff
@@ -14,35 +14,75 @@
...
+    return hasWhereClause ? objectSet.where(whereClause) : objectSet;
+  }
+
+  const creators: Record<string, DerivedProperty.Creator<Q, LinkCountDef>> = {};
+  const conditions: WhereClauseFragment[] = hasWhereClause
+    ? [whereClause as WhereClauseFragment]
+    : [];
+
   for (const filter of linkedFilters) {
-    narrowed = narrowed.intersect(applyLinkedFilter(narrowed, filter));
```
ziyunp, line 31: Main changes of this PR - change the filter method from intersection to chaining of where clauses on derived properties. Intersection is limiting as it does not work when the objectSet passed to FilterList contains RDPs (unsupported by OSDK)

packages/react-components/src/filter-list/utils/narrowObjectSet.ts
```diff
@@ -14,35 +14,75 @@
...
+type LinkCountRdps = Record<string, LinkCountDef>;
+
+/**
+ * Derived-property names share a namespace with the object type's own
+ * properties, so the prefix keeps these from colliding with a real property.
+ */
+const LINK_COUNT_PREFIX = "osdkFilterListLinkCount_";
+
+function linkCountPropertyName(id: string): string {
+  return `${LINK_COUNT_PREFIX}${id.replaceAll(/[^A-Za-z0-9_]/gu, "_")}`;
```
ziyunp: suggestion (non-blocking): filter ids are unique, but the sanitised names aren't. Two definitions with `id: "office-city"` and `id: "office.city"` both land on `osdkFilterListLinkCount_office_city`. The second creator quietly overwrites the first, but both conditions still get pushed, so a keep on one plus an exclude on the other produces `{X: {$gt: 0}}` AND `{X: 0}` — matches nothing, no error.

Since the only property you need is stability for a given filter list, would folding the loop index in be enough?

```suggestion
function linkCountPropertyName(id: string, index: number): string {
  return `${LINK_COUNT_PREFIX}${index}_${id.replaceAll(/[^A-Za-z0-9_]/gu, "_")}`;
}
```

Unlikely to bite with the default ids, so happy for this to be a FLUP.

packages/react-components/src/filter-list/utils/filterStateToWhereClause.ts
```diff
@@ -464,27 +449,53 @@ export function getActiveLinkedFilters<Q extends ObjectTypeDefinition>(
...
+
+    if (state.type !== "linkedProperty") {
       continue;
     }
     const innerWhere = buildLinkedInnerWhere(definition, state);
     if (innerWhere === undefined) {
       continue;
     }
     result.push({
+      id: key,
```
ziyunp: question: is it deliberate that the two filter types express "excluding" differently now? HAS_LINK gets `isExcluding`, which `narrowObjectSet` turns into count `=== 0`. LINKED_PROPERTY keeps folding it into `innerWhere` as `$not`, which becomes "count of *non-matching* linked objects > 0".

On a many-cardinality link those differ: excluding "Alice" on `peeps` keeps an employee whose reports are Alice and Bob, because Bob makes the count non-zero. I read that as matching the pre-PR intersect behaviour, so nothing regressed — but `isExcluding` now makes the other reading ("has no report named Alice") expressible, and having two mechanisms for one UI toggle seems like a trap for whoever touches this next. Would it make sense to route both through `isExcluding` and leave `innerWhere` un-negated?

.changeset/reject-invented-where-clause-operators.md
```diff
@@ -0,0 +1,6 @@
+---
+"@osdk/react-components": patch
+"@osdk/react-sdk-docs": patch
+---
+
+Fix FilterList's `HAS_LINK` filter, which sent an unsupported `$isNotNull` filter. `HAS_LINK` and `LINKED_PROPERTY` now filter by deriving a count of matching linked objects and not using intersect. Remove bad documentation `$isNotNull` in `useOsdkObjects` snippets.
```
ziyunp: suggestion (blocking): I think this is more than a patch, and the summary undersells what consumers have to do. Three things land here that someone will hit:

- HAS_LINK no longer appears in `onFilterClauseChanged`, so anyone feeding only `filter={clause}` to a table loses the filter silently — exactly the rewiring `page.tsx` needed in this PR.
- `LinkedFilter` is public via `@osdk/react-components/experimental/filter-list`; it gains a required `id` and `innerWhere` becomes optional, so existing constructions stop compiling.
- The `reverseLinkName == null` skip is gone, so LINKED_PROPERTY definitions that deliberately omitted `reverseLinkName` to stay UI-only now narrow `objectSet`. That was documented behaviour (#3287), and there's no opt-out left.

Would you bump this to `minor` and add a line telling people to wire `onEffectiveObjectSet`?

Related: a FilterList with a HAS_LINK filter and no `objectSet` now renders the filter, counts it in `activeFilterCount`, and filters nothing, with no signal at all. `getActiveLinkedFilters` already has the dev-only warn pattern a few lines up — worth reusing it here so this shows up in the console rather than in a support thread.

packages/react-components-storybook/src/stories/FilterList/FilterList.stories.tsx
```diff
@@ -1852,7 +1852,6 @@ function WithLinkedPropertyFiltersStory(
       {
         type: "LINKED_PROPERTY",
         linkName: "lead",
-        reverseLinkName: "peeps",
         linkedPropertyKey: "department",
         filterComponent: "MULTI_SELECT",
         label: "Manager Department",
```
ziyunp: issue (non-blocking): this story's note (further down, around line 1885) still reads "Note: LINKED_PROPERTY filter is not applied through where clause" — but the story also has the "Has Manager" HAS_LINK filter, which after this PR isn't in the clause either. So ticking it shows `(no active filters)` with nothing explaining why.

The story also doesn't pass `onEffectiveObjectSet`, so neither of these two filters has anywhere to land. Would be a good place to demo the new wiring — worth updating the note either way.

packages/react-components/src/filter-list/utils/filterStateToWhereClause.ts
```diff
@@ -290,30 +292,10 @@ export function buildWhereClause<Q extends ObjectTypeDefinition>(
...
-        }
-        const hasLinkClause = { [definition.linkName]: { $isNotNull: true } };
-        // "Excluding" keeps objects that do NOT have the link.
-        clauses.push(
-          state.isExcluding ? { $not: hasLinkClause } : hasLinkClause,
-        );
-        break;
-      }
-
+      case "HAS_LINK":
```
ziyunp: nit: `buildWhereClause`'s doc a bit above (lines 253-256) still says only "LINKED_PROPERTY filters are excluded" — HAS_LINK belongs in that sentence now. `useFilterListState.ts:42` has the same drift ("Active linked-property records", though the array carries HAS_LINK entries too). The `FilterListApi.ts` and `getActiveLinkedFilters` docs in this PR both got it right, so it's just these two that got left behind.

packages/react-components/src/filter-list/utils/narrowObjectSet.ts
```diff
@@ -14,35 +14,81 @@
...
+        : { [propertyName]: { $gt: 0 } },
+    );
   }
-  return narrowed;
+
+  const countClause: WhereClauseFragment =
+    conditions.length === 1 ? conditions[0] : { $and: conditions };
+
+  // Cast needed because the RDPs are created at runtime.
+  return objectSet
```
KuberSethi, line 74: i think this new `withProperties().where()` bit exposes a bug in the clients `createObjectSet.where` as the count comes from `withProperties()` but sends its name to the server as a normal ontology field instead of a derived property.

i tested it myself by adding a client test for this and confirmed that it produces `field: "derivedPropertyName"` instead of a `propertyIdentifier`. I think basically in packages/client/src/objectSet/createObjectSet.ts before we serialize the `.where()` clause, we need createObjectSet to traverse the wire object set, collect the prop names added by preceding `withProperties`, and wire those names into `modernToLegacyWhereClause`.

KuberSethi (commented):
Generally looks good just one bit on withProperties probably not working as we expect

packages/react-components/src/filter-list/utils/narrowObjectSet.ts
```diff
@@ -14,35 +14,81 @@
...
+  // see them.
+  // If we see an issue with it, we can look into stripping any property whose name begins with
+  // `LINK_COUNT_PREFIX`.
   for (const filter of linkedFilters) {
-    narrowed = narrowed.intersect(applyLinkedFilter(narrowed, filter));
+    const propertyName = linkCountPropertyName(filter.id);
+    creators[propertyName] = createLinkCount(filter);
+    conditions.push(
+      filter.isExcluding === true
+        ? { [propertyName]: 0 }
```
ziyunp: Possibly do 
```
{
  $not: { [propertyName]: { $gt: 0 }  
}
```

ziyunp (commented):
.

packages/react-components/src/filter-list/utils/narrowObjectSet.ts
KuberSethi: On second thought I think this actually isn't a problem. I think I just got tripped up that internally when building RDPs we serialize using `propertyIdentifier`, but here with `withProperties().where()` we'd use `field`. Both work, so this isn't really a bug just an inconsistency. I think it's fine to leave that out of this PR though!

packages/react-components/src/filter-list/types/LinkedFilterTypes.ts
```diff
@@ -33,15 +33,22 @@ import type {
...
- * against the linked object type, so `narrowObjectSet` can pivot type-safely.
+ * Runtime representation of an active link-traversing filter
  */
 export type LinkedFilter<Q extends ObjectTypeDefinition> = {
   [L in LinkNames<Q>]: {
+    /** Stable identity for the filter */
+    id: string;
+
     linkName: L;
-    reverseLinkName: LinkNames<LinkedType<Q, L>>;
```
KuberSethi, line 43: This is technically a break right? I think it's fine as we're still beta but just calling out

KuberSethi (approved):

packages/react-components/src/filter-list/types/LinkedFilterTypes.ts
ziyunp: Yes, but LinkedFilter is exported as a type only, the fields here are populated internally. The reverseLinkName on LinkedPropertyFilterDefinition is actually used by users and is marked as deprecated.

PR #3905, opened 2026-08-19
Moved aip-core to direct dependency

ziyunp:
`@osdk/aip-core` is now a direct dependency of `@osdk/react` and `@osdk/react-components` instead of an optional peer dependency, so consumers of `AipAgentChat` and `useChat` no longer have to install it themselves. 

AipAgentChat component and useChat hook are experimental and the dependencies will likely change. Making `@osdk/aip-core` a direct dependency so it is not surfaced to users now.

### DIFF
--- a/packages/aip-core/package.json
+++ b/packages/aip-core/package.json
@@ -51,7 +51,7 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
-    "@osdk/language-models": "workspace:~"
+    "@osdk/language-models": "workspace:*"
   },
   "peerDependencies": {
     "@ai-sdk/provider": "^3.0.0",
--- a/packages/aip-core/package.json
+++ b/packages/aip-core/package.json
@@ -50,10 +50,12 @@
     "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
+  "dependencies": {
+    "@osdk/language-models": "workspace:~"
+  },
   "peerDependencies": {
     "@ai-sdk/provider": "^3.0.0",
-    "@osdk/client": "workspace:^",
-    "@osdk/language-models": ">=0.6.0 <1.0.0"
+    "@osdk/client": "workspace:^"
   },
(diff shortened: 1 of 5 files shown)


nihalbhatnagar (dismissed):

packages/aip-core/package.json
```diff
@@ -50,10 +50,12 @@
     "transpileTypes": "monorepo.tool.transpile -f esm -m types -t node",
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
+  "dependencies": {
+    "@osdk/language-models": "workspace:~"
```
ssanjay1: Language-models is in beta right? Is tilde dep really the right move, or should we do exact dep?

packages/react-components/package.json
```diff
@@ -202,6 +202,7 @@
     "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/sortable": "^10.0.0",
     "@dnd-kit/utilities": "^3.2.2",
+    "@osdk/aip-core": "workspace:~",
```
ssanjay1: Same question here

packages/aip-core/package.json
ziyunp: I will fix this to exact dep.

ssanjay1 (changes requested):
Saw nihal approved, but think we should fix dep issues

packages/react/package.json
```diff
@@ -96,10 +96,10 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
+    "@osdk/aip-core": "workspace:*",
     "fast-deep-equal": "^3.1.3"
   },
   "peerDependencies": {
-    "@osdk/aip-core": ">=0.5.0 <1.0.0",
```
esinx, line 102: FYI this might cause an issue when making a release, see #3820

packages/aip-core/package.json
ziyunp: addressed

packages/react-components/package.json
ziyunp: addressed

ssanjay1 (approved):

PR #3908, opened 2026-08-19
Update example-generator no osdk package name and widget readmes

tzyl:
- example-generator uses `@osdk/examples...-no-osdk` package names rather than `@psdk/examples...` which matches their directory name and better reflects widgets which don't have psdk
- Widget examples use a more accurate readme rather than referencing create-app
- Drop create-widget dead code from initial create-app fork


packages/create-widget/codegen.mjs
```diff
@@ -26,14 +26,12 @@ export const TEMPLATES = [
   {
     id: "react",
     label: "OSDK React",
-    envPrefix: "VITE_",
```
amishshah, line 29: Why do we drop the envPrefix 🤔

tzyl: This is actually dead code in create-widget as can be seen from removing it from the typescript type in the file above. It was originally used in create-app for writing .env files which don't exist in the widget templates:
https://github.com/palantir/osdk-ts/blob/192d1cbb45d8f6b34fc65e84f5f9a0f03f4f0427/packages/create-app/src/run.ts#L216-L242

Rather than the copy of this it'd be nice to have create-app / create-widget share more of this code but for now in this create-widget copy I think it's more confusing that these references like beta and envPrefix are here because they don't do anything

amishshah (dismissed):

amishshah (dismissed):

amishshah (dismissed):

amishshah (approved):

PR #3912, opened 2026-08-20
support portable generated sdks

KuberSethi:
generated SDKs can now omit installed ontology and branch identities

• add an opt-in generator setting for portable SDK output
• omit ontology and branch identity exports when enabled
• keep existing generated SDK output unchanged by default


KuberSethi:
closing this draft because portable identity belongs with the oac generation command that actually uses and tests it

PR #3913, opened 2026-08-20
preserve interface link actions in oac conversion

KuberSethi:
interface link actions were omitted from converted sdk action metadata

• convert supported create and delete interface link operations
• resolve interface and interface link api names from block data
• keep preview action operations alongside full logic rules


KuberSethi:
closing this draft because the platform metadata type cannot represent these interface-link operations without unsafe casts; the change needs an upstream type contract before it can be split safely

PR #3922, opened 2026-08-21
[Branching] Add new vite branch plugin

zeyadkhaled:
## Summary

  - Add @osdk/vite-plugin-branch to inject the current Git branch into VITE_FOUNDRY_BRANCH_RID during local Vite development.
  - Add @osdk/shared.branch with shared Git branch detection and normalization utilities.
  - Update the branch CLI to use the shared utilities while preserving existing behavior.
  - Respect branch values supplied through .env, CI, or Foundry runtimes, and skip injection for default or detached branches.
  - Add tests, package metadata, monorepo configuration, lockfile updates, and changesets.


packages/cli/src/commands/branch/utils/getGitBranch.ts
```diff
@@ -16,17 +16,12 @@
...
-export async function getGitBranch(): Promise<string | undefined> {
+/** The current git branch, or `undefined` if git fails. */
+export async function getGitBranch(cwd?: string): Promise<string | undefined> {
   try {
-    const { stdout } = await execa("git", [
-      "rev-parse",
-      "--abbrev-ref",
-      "HEAD",
-    ]);
+    const { stdout } = await execa("git", ["branch", "--show-current"], {
```
zeyadkhaled, line 22: osdk unstable branch sync now
  detects the branch on an unborn branch (fresh git init,
  no commits), where rev-parse --abbrev-ref HEAD failed
  outright. Detached HEAD now reports "" instead of "HEAD";
  both already normalized to "no branch", so nothing else
  shifts.

packages/cli/src/commands/branch/utils/getGitBranch.ts
```diff
@@ -16,17 +16,12 @@
 
 import { execa } from "execa";
 
-/**
- * The current git branch (`git rev-parse --abbrev-ref HEAD`), or `undefined` if
- * git fails. Detached HEAD returns the literal "HEAD".
- */
-export async function getGitBranch(): Promise<string | undefined> {
+/** The current git branch, or `undefined` if git fails. */
+export async function getGitBranch(cwd?: string): Promise<string | undefined> {
```
amishshah, line 20: I wonder if there's a way to avoid making this an actual npm package would be nice if it's just some code that gets written once but can be imported and built into the package builds?

.changeset/vite-plugin-branch.md
```diff
@@ -0,0 +1,9 @@
+---
```
amishshah: Two changesets

packages/vite-plugin-branch/src/branchPlugin.ts
```diff
@@ -0,0 +1,97 @@
...
+import { getGitBranch } from "@osdk/shared.branch";
+import { loadEnv, type Plugin } from "vite";
+
+import { resolveBranchToInject } from "./resolveBranchToInject.js";
+
+/**
+ * The environment variable a Foundry runtime sets to the branch the application
+ * is checked out on. `@osdk/client` reads it to scope every request.
+ */
+export const FOUNDRY_BRANCH_RID_ENV_VAR: string = "VITE_FOUNDRY_BRANCH_RID";
```
amishshah: Is this always a rid?

packages/vite-plugin-branch/src/branchPlugin.ts
```diff
@@ -0,0 +1,97 @@
...
+
+      const pinnedBranch = loadEnv(mode, envDir, "VITE_")[
+        FOUNDRY_BRANCH_RID_ENV_VAR
+      ];
+
+      injectedBranch = resolveBranchToInject(
+        pinnedBranch,
+        await readGitBranch(root),
+      );
+      if (injectedBranch === undefined) {
```
amishshah: `== null` is convention

packages/vite-plugin-branch/src/branchPlugin.ts
```diff
@@ -0,0 +1,97 @@
...
+ * branch, which holds for a branch checked out with `osdk unstable branch`.
+ *
+ * Nothing is injected when the variable already has a value, when git reports
+ * `main`/`master`/a detached HEAD, or when the directory is not a git
+ * repository — in each case the client falls back to the default branch.
+ *
+ * @example
+ * ```ts
+ * export default defineConfig({ plugins: [react(), branchPlugin()] });
+ * ```
```
amishshah, line 51: Can we word this such that it gets these points across:

- It determines the global branch a user is on and the rules it uses to do this. I think wording could be more concise here.
- How does it expose it I think we should write more like "make it available to the @osdk/client via import.meta.env.BLAH"

packages/vite-plugin-branch/src/branchPlugin.ts
```diff
@@ -0,0 +1,97 @@
...
+      const envDir =
+        config.envDir === false
+          ? false
+          : path.resolve(root, config.envDir ?? ".");
+
+      const pinnedBranch = loadEnv(mode, envDir, "VITE_")[
+        FOUNDRY_BRANCH_RID_ENV_VAR
+      ];
+
+      injectedBranch = resolveBranchToInject(
```
amishshah: It is not really obvious how pinnedBranch is different to injectedBranch?

packages/vite-plugin-branch/src/resolveBranchToInject.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { normalizeGitBranch } from "@osdk/shared.branch";
+
+export function resolveBranchToInject(
```
amishshah: It feels strange that this file reads so similar to the ones in the `shared.branch` package

packages/vite-plugin-branch/src/branchPlugin.ts
```diff
@@ -0,0 +1,90 @@
...
+      const root = path.resolve(config.root ?? process.cwd());
+      const envDir =
+        config.envDir === false
+          ? false
+          : path.resolve(root, config.envDir ?? ".");
+
+      const configuredBranch = loadEnv(mode, envDir, "VITE_")[
+        FOUNDRY_BRANCH_ENV_VAR
+      ];
+      if (configuredBranch?.trim()) {
```
amishshah:
```suggestion
      if (configuredBranch?.trim() != null) {
```

conventions

amishshah (dismissed):
Approved! I assume we're going to have to iterate on this as we test it out

.github/workflows/release.yml
```diff
@@ -9,6 +9,7 @@ on:
       - main
       - release/*
       - next
+      - zka/vite-plugin-branch
```
ssanjay1: Need to remove this please 😄

ssanjay1 (approved):
Stamping

PR #3923, opened 2026-08-21
protect oac generated output

KuberSethi:
regenerating OaC output should never expose a partially written directory or destroy the prior output on failure

• generate into a sibling temporary directory
• reject protected and existing output paths by default
• replace clean output atomically and restore it after failure


KuberSethi:
closing because the staging assembler already generates into a fresh isolated work directory, so output replacement safety is outside the RFC path

PR #3926, opened 2026-08-24
Add relative date option to DATE_RANGE

pmartin2309:
Add a relative date mode to the DATE_RANGE filter component. When enableRelativeMode: true is set on a property filter definition, users can toggle between absolute (calendar date pickers) and relative mode, where each bound (From/To) is configured with a count, unit (days/weeks/months/years), and direction (ago/from now). An "Indefinitely" option represents no bound.

- Types: New RelativeDateBound interface and DateRangeRelativeProps. DateRangeFilterState extended with optional isRelative, relativeMin, relativeMax fields. All backward compatible.
- Resolution: resolveRelativeDateBound() converts relative bounds to absolute dates at midnight (or end-of-day 23:59:59.999 for the upper bound). Resolution happens lazily in filterStateToWhereClause at query time, so saved relative state always anchors to "now".
- UI: Trigger + popover pattern matching the existing Select component styling. RelativeModeToggle switch, RelativeDateBoundInput (single bound editor), and RelativeDateRangeInput (From – To pair) as separate components following codebase conventions.
- Summary labels: summarizeFilterValue shows relative descriptions (e.g., "5 months ago – Indefinitely") when in relative mode.
- Serialization: Works out of the box — relative fields are plain JSON primitives.
- Story: Dedicated RelativeDateFilter story with three configurations (disabled, enabled, with defaults) and where clause JSON output.

### Test Plan

- 15 unit tests for resolveRelativeDateBound and formatRelativeBound (date arithmetic, midnight reset, end-of-day, month clamping)
- 55 existing filterStateToWhereClause tests pass unchanged (relative metadata ignored for absolute state)
- Typecheck passes (pnpm turbo typecheck --filter=@osdk/react-components)
- Storybook: Beta / FilterList / Relative Date Filter — toggle on/off, set bounds, verify where clause JSON

https://github.com/user-attachments/assets/3a87ec65-ec06-4887-b777-06e1a7085eb6


palantirtech:
Thanks for your interest in [palantir/osdk-ts](https://github.com/palantir/osdk-ts), @pmartin2309! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/react-components/src/filter-list/utils/resolveRelativeDate.ts
```diff
@@ -0,0 +1,82 @@
...
+      break;
+    case "weeks":
+      result.setDate(result.getDate() + sign * bound.count * 7);
+      break;
+    case "months":
+      result.setMonth(result.getMonth() + sign * bound.count);
+      break;
+    case "years":
+      result.setFullYear(result.getFullYear() + sign * bound.count);
+      break;
```
ziyunp, line 55: issue (blocking): `setMonth`/`setFullYear` don't clamp the day-of-month, so on the 29th–31st these roll over into the wrong month. Can you add tests for these cases to verify please?

| anchor | bound | actual | expected |
|---|---|---|---|
| Mar 31 2026 | 1 month ago | `Mar 3 2026` | Feb 28 |
| May 31 2026 | 1 month ago | `May 1 2026` | Apr 30 |
| Feb 29 2024 | 1 year ago | `Mar 1 2023` | Feb 28 2023 |
| Jan 31 2026 | 1 month from now | `Mar 3 2026` | Feb 28 |

packages/react-components/src/filter-list/utils/resolveRelativeDate.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  result.setHours(0, 0, 0, 0);
+
+  if (endOfDay) {
+    result.setDate(result.getDate() + 1);
+  }
```
ziyunp, line 62: issue (blocking): the `endOfDay` bump can produce unexpected value in the filter clause.

If I set "To = Today" on Aug 15, it will emit `$lte: "2026-08-16"`.

I would expect endOfDay to be `2026-08-15 23:59:59.999`.

packages/react-components/src/filter-list/__tests__/resolveRelativeDate.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    const result = resolveRelativeDateBound(
+      { count: 0, unit: "days", direction: "ago" },
+      true,
+      NOW,
+    );
+    // today = Aug 15, +1 day = Aug 16 at midnight
+    expectDate(result, 2026, 7, 16);
+  });
+
+  it("handles month day clamping (Jan 31 minus 1 month)", () => {
```
ziyunp: issue (blocking): this picks Jan 31 → Dec 31, which is the one anchor where no overflow happens (both months have 31 days). So the test name claims the clamping path is covered while the actual overflow case goes untested. Mar 31 or May 31 would exercise it.

packages/react-components/src/filter-list/inputs/DateRangeFilterInput.tsx
```diff
@@ -76,6 +89,50 @@ function DateRangeFilterInputInner<Q extends ObjectTypeDefinition>({
...
+    (
+      relativeMin: RelativeDateBound | undefined,
+      relativeMax: RelativeDateBound | undefined,
+    ) => {
+      const minValue = relativeMin
+        ? resolveRelativeDateBound(relativeMin)
+        : undefined;
+      const maxValue = relativeMax
+        ? resolveRelativeDateBound(relativeMax)
+        : undefined;
```
ziyunp: issue (blocking): this resolves the max bound without `endOfDay`, but `filterStateToWhereClause.ts:108` resolves the same `relativeMax` *with* `endOfDay=true.

packages/react-components/src/filter-list/utils/filterStateToWhereClause.ts
```diff
@@ -96,14 +97,25 @@ function filterStateToPropertyFilter(
...
+      // When in relative mode, resolve bounds to fresh absolute dates
+      // so the where clause is always anchored to "now".
+      const minDate =
+        state.isRelative && state.relativeMin
+          ? resolveRelativeDateBound(state.relativeMin)
+          : state.minValue;
+      const maxDate =
+        state.isRelative && state.relativeMax
+          ? resolveRelativeDateBound(state.relativeMax, true)
+          : state.maxValue;
```
ziyunp, line 107: Issue (blocking): when `isRelative` is `true` but `relativeMin` is absent, this falls through to `state.minValue`. This is inconsistent with the documented behaviour where the From bound is 'Indefinitely' when no lower bound is set. This would happen when a `defaultFilterState` with this value: `{ isRelative: true, maxValue: someDate }`

packages/react-components/src/filter-list/base/inputs/RelativeDateBoundInput.tsx
```diff
@@ -0,0 +1,296 @@
...
+                        <Select.Popup>
+                          {DIRECTION_OPTIONS.map((opt) => (
+                            <Select.Item key={opt.value} value={opt.value}>
+                              {opt.label}
+                            </Select.Item>
+                          ))}
+                        </Select.Popup>
+                      </Select.Positioner>
+                    </Select.Portal>
+                  </Select.Root>
```
ziyunp: issue (blocking): I think the unit and direction triggers render raw values rather than the labels. `Select.Root` has no `itemToStringLabel`, and per the Select's own test — `"falls back to String() when itemToStringLabel is not provided"` — `<Select.Item value="hello">Hello</Select.Item>` renders `hello`. So the trigger should read `days` / `fromNow`, not `day(s)` / `from now`.

The `placeholder={unitLabel}` fallback can't cover for it either: `Select.module.css:142` hides the placeholder whenever a value is set (`.osdkSelectValue:not([data-placeholder]) ~ .osdkSelectPlaceholder { display: none }`), and `value` is always set here. Which also means `unitLabel`/`directionLabel` and their two `useMemo`s are dead code. `itemToStringLabel` looks like the fix.

packages/react-components/src/filter-list/base/inputs/RelativeDateBoundInput.tsx
```diff
@@ -0,0 +1,296 @@
...
+          </Popover.Popup>
+        </Popover.Positioner>
+      </Popover.Portal>
+    </Popover.Root>
+  );
+}
+
+export const RelativeDateBoundInput = memo(
+  RelativeDateBoundInputInner,
+) as typeof RelativeDateBoundInputInner;
```
ziyunp, line 244: nit: two exported components in one file. Prefer to always put new components in their own file

packages/react-components/src/filter-list/base/inputs/RelativeDateBoundInput.module.css
```diff
@@ -0,0 +1,291 @@
...
+
+/* ---- Trigger (mirrors Select.Trigger styling) ---- */
+
+.trigger {
+  display: inline-flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: var(--osdk-select-spacing, var(--osdk-surface-spacing));
+  width: 100%;
+  min-height: var(--osdk-select-min-height, 30px);
```
ziyunp, line 25: suggestion (non-blocking): these CSS map 1:1 onto `Select.module.css` (`.trigger`↔`.osdkSelectTrigger`, `.triggerValue`↔`.osdkSelectValue`, `.popover`↔`.osdkSelectPopup`, `.positioner`↔`.osdkSelectPositioner`, …). Can you see if we can reuse the Select trigger styles or any common components in base-components?

packages/react-components/src/filter-list/FilterListItemApi.ts
```diff
@@ -202,6 +218,29 @@ export interface DateRangeFilterState extends BaseFilterState {
    * The latest date the user can select
    */
   maxValue?: Date;
+
+  /**
+   * When `true`, the filter UI renders in relative-date mode.
+   * The `minValue` / `maxValue` are computed from `relativeMin` /
+   * `relativeMax` and refreshed each time the component mounts.
```
ziyunp: Nit: remove "refreshed each time the component mounts", we don't want to expose implementation details in the comment

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
```diff
@@ -93,6 +94,15 @@ export function summarizeFilterValue<Q extends ObjectTypeDefinition>(
...
     }
     case "DATE_RANGE": {
+      if (state.isRelative === true) {
+        const minLabel = state.relativeMin
+          ? formatRelativeBound(state.relativeMin)
+          : "Indefinitely";
+        const maxLabel = state.relativeMax
+          ? formatRelativeBound(state.relativeMax)
+          : "Indefinitely";
+        return `${minLabel} – ${maxLabel}`;
```
ziyunp: issue (non-blocking): this branch returns early without checking `includeNull`. Reading L108, it also seems like includeNull is only read when both min and max values are unset. Can you check how the value of summarizeFilterValue render for different settings?

packages/react-components/src/filter-list/__tests__/testUtils.ts
```diff
@@ -195,14 +196,23 @@ export function createNumberRangeState(
...
 ): DateRangeFilterState {
   return {
     type: "DATE_RANGE",
     minValue,
     maxValue,
     includeNull: options?.includeNull,
     isExcluding: options?.isExcluding,
+    isRelative: options?.isRelative,
+    relativeMin: options?.relativeMin,
+    relativeMax: options?.relativeMax,
```
ziyunp: issue (non-blocking): these three options aren't used by any test. Can you add tests for the following:
- `filterStateToWhereClause.test.ts`'s `isRelative` case, so the `endOfDay` day-shift and the `state.minValue` fallback are both tested; 
- `summarizeFilterValue`'s relative branch;
- `PropertyFilterDateExtras.test-d.ts` - extend with an `@ts-expect-error` case for `enableRelativeMode`to test that the type is set to never when property is not datetime / timestamp

packages/react-components/src/filter-list/FilterListItemApi.ts
```diff
@@ -264,15 +303,35 @@ export interface DateFormattingProps {
...
+   * In relative mode the user builds each bound with a count, unit, and
+   * direction (e.g. "5 months ago") instead of picking calendar dates.
+   * The resulting `minValue` / `maxValue` are still absolute `Date` objects.
+   *
+   * Has no effect on filter components other than `DATE_RANGE`.
+   *
+   * @default false
+   */
+  enableRelativeMode?: boolean;
+}
```
ziyunp, line 311: Nit: shorten the comment. 

```suggestion
  /**
   * When `true`, the `DATE_RANGE` filter shows an Absolute / Relative toggle.
   * Has no effect on filter components other than `DATE_RANGE`.
   *
   * @default false
   */
  enableRelativeMode?: boolean;
}
```

packages/react-components/src/filter-list/FilterListItemApi.ts
```diff
@@ -264,15 +303,35 @@ export interface DateFormattingProps {
...
+
+/**
+ * Conditionally adds date-specific props to a property filter definition
```
ziyunp: Nit: Comment not needed

packages/react-components/src/filter-list/utils/resolveRelativeDate.ts
```diff
@@ -42,12 +42,24 @@ export function resolveRelativeDateBound(
     case "weeks":
       result.setDate(result.getDate() + sign * bound.count * 7);
       break;
-    case "months":
-      result.setMonth(result.getMonth() + sign * bound.count);
+    case "months": {
+      const targetMonth = result.getMonth() + sign * bound.count;
+      result.setMonth(targetMonth);
+      // Clamp day overflow to the last day of the target month.
+      const expectedMonth = ((targetMonth % 12) + 12) % 12;
```
ziyunp: Can we use date-fns subMonths / addMonths function here?

packages/react-components/src/filter-list/base/inputs/RelativeDateBoundInput.module.css
pmartin2309: So few things:
1. I can't reuse the Select component directly because it expects a dropdown of options rather than what happens on click. When we click here I want a popover with radio buttons but I needed the click of that to look like the Select component hence the copied css. (i.e. I needed Select.Trigger without the select)
2. Combox component also has it's own equivalent css duplicate
3. I didn't want the dependence just on css, since it is a bit convoluted and people might not realise that changing things there would change how the relative date range input looks. 

I don't know if it is worth it to have a common Trigger component or similar, but would say that's separate concern/refactor

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
pmartin2309: Not sure I fully understand the ask here. This behaviour is consistent with workshop, i.e. even if there are no values as input the relative filter stays rendered, cause we don't necessarily show any summary metrics like in absolute mode. 

https://github.com/user-attachments/assets/1b918b48-48c7-4722-ab7c-f8e4f567c9c1

packages/react-components/src/filter-list/__tests__/filterStateToWhereClause.test.ts
```diff
@@ -821,4 +821,95 @@ describe("getActiveLinkedFilters", () => {
...
+      const result = buildWhereClause(
+        [def],
+        filterStates,
+        mockPropertyTypes,
+      ) as Record<string, unknown>;
+      // Should produce $and with $gte and $lte
+      expect(result).toHaveProperty("$and");
+      const clauses = (result as { $and: unknown[] }).$and;
+      expect(clauses).toHaveLength(2);
+      expect(clauses[0]).toHaveProperty("createdAt");
```
ziyunp: Nit (not blocking): We are not asserting the dates values used in the filter here?

packages/react-components/src/filter-list/__tests__/filterStateToWhereClause.test.ts
```diff
@@ -821,4 +821,95 @@ describe("getActiveLinkedFilters", () => {
...
+      const filterStates = stateMap([def, state]);
+      const result = buildWhereClause(
+        [def],
+        filterStates,
+        mockPropertyTypes,
+      ) as Record<string, unknown>;
+      // Should only have $lte (no $gte), stale minValue must not leak through
+      expect(result).toHaveProperty("createdAt");
+      const clause = result.createdAt as Record<string, unknown>;
+      expect(clause).toHaveProperty("$lte");
```
ziyunp, line 916: Nit: should also assert the date values used in the filter clause

packages/react-components/src/filter-list/__tests__/summarizeFilterValue.test.ts
ziyunp, line 1: Nice!

packages/react-components/src/filter-list/base/inputs/RelativeDateBoundInput.tsx
```diff
@@ -0,0 +1,260 @@
...
+  const handleDirectionChange = useCallback(
+    (nextDirection: RelativeDateDirection | null) => {
+      if (nextDirection != null) {
+        onChange({ ...currentBound, direction: nextDirection });
+      }
+    },
+    [onChange, currentBound],
+  );
+
+  const unitLabel = useMemo(
```
ziyunp: Suggestion (not blocking): For finding labels logic, can we create a Map of value -> label instead so we can remove useMemo / .find() loop. 
```
const DIRECTION_OPTIONS_LABEL = {
  ago: "ago", 
  fromNow: "from now"
}
```

packages/react-components/src/filter-list/base/inputs/RelativeDateBoundInput.module.css
ziyunp: got it, agree, espeically on the 3rd point

packages/react-components/src/filter-list/utils/summarizeFilterValue.ts
ziyunp: Let's leave this for now as I checked that the function is only used in storybook. I have a follow up to remove this from the export, and remove the function completely.

ziyunp:
Lgtm

ziyunp (dismissed):

nfdiop (dismissed):

nfdiop (dismissed):

ziyunp (approved):

PR #3932, opened 2026-08-24
Bump @osdk/foundry.* to 2.75.0 + Add support for streaming queries w/ SSE

williamhe723:
This PR reimplements support for queries with streamed responses, which were previously implemented in #3105 and removed in #3750. This new reimplementation uses an SSE stream rather than NDJSON as it was previously. The platform SDK versions were bumped as well to get the support added in the TS platform SDK.

Unit tests were added in the same way they were before, just swapped out NDJSON for SSE. I also tested end-to-end locally with the executeStreamingFunction method and it returned correctly.

### DIFF
--- a/.changeset/four-parents-hear.md
+++ b/.changeset/four-parents-hear.md
@@ -1,0 +1,21 @@
+---
+"@osdk/api": minor
+"@osdk/client": minor
+"@osdk/faux": patch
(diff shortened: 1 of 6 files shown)


bryantpark04 (approved):
LGTM - the streaming queries were tested with multiple SSEs being sent right? Instead of batched

PR #3473, opened 2026-06-09
Default to Workshop Light theme with Blueprint-native box-shadow tokens

Leolide:
## Summary

- Align OSDK component input/button/select styling with Blueprint conventions: use `border: none` + `box-shadow` (inset) for visual borders instead of CSS borders
- Add proper dark mode token overrides for inputs (bg, color, shadow), buttons (bg, shadow), and surface border color in the component token layer
- Update base tokens to use Blueprint-matching values: `--bp-palette-light-gray-2` for border color, `--bp-typography-line-height-large` for line height, Blueprint-style inset shadows
- Group Storybook theme presets into "Built-in themes" (Workshop Light/Dark) and "Custom themes" (DevCon, Midnight Blue, Warm Sand, Royal Purple)
- Workshop Light/Dark use empty assignments — they render exactly what consumers get from the built-in tokens
- Fix theme switching bug where dark mode persisted when switching back to a light theme

## Test plan

- [ ] Open any form story (BaseForm) in Storybook — inputs should have subtle inset shadow border, not a CSS border
- [ ] Switch to Workshop Dark — inputs should have dark bg (`rgba(17,20,24,0.3)`), light text, and appropriate dark inset shadow
- [ ] Select dropdowns should have a single subtle border line, not doubled
- [ ] Buttons in dark mode should match Blueprint's dark button styling
- [ ] Switch between Workshop Light and Dark repeatedly — should always switch correctly
- [ ] Custom themes (DevCon, Midnight Blue, etc.) should appear under "Custom themes" section in the dropdown
- [ ] Dropdown/popover shadows should use elevation shadow, not inset

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3933, opened 2026-08-25
ActionForm: Accessibility fixes + move tokens

nfdiop:
### What is done in this PR
- Move combobox/select tokens into the tokens file
- Fix accessibility issues found in storybook
- Fix dark mode (tested in the storybook for ActionForm)

### Tests
- Clicked around the ActionForm in light/dark/warm sand theme (note for the warm sand, I couldn't get the theme to be always accessible. With the new blueprint tokens, we should make that easier)


packages/react-components-storybook/.storybook/addons/brand-theme-extractor/decorator.tsx
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import { Classes } from "@blueprintjs/core";
```
nfdiop, line 17: This is required only for Blueprint components we use (like the dialog). Not trying to scope to those stories since this will help whenever we use another component. We need to revisit this once we move to Blueprint.

Image

packages/react-components/src/action-form/FormField.module.css
```diff
@@ -74,6 +74,7 @@
 .osdkFormFieldInfoPopup {
   padding: var(--osdk-form-info-popup-padding);
   background-color: var(--osdk-form-info-popup-bg);
+  color: var(--osdk-form-info-popup-color);
```
nfdiop, line 77: **Before**

Image

**After**

Image

Note: I want to revisit the tokens when we move to Blueprint because it looks off to require the color for tooltips/popovers

packages/react-components/src/action-form/FormField.tsx
```diff
@@ -103,19 +103,23 @@ interface InfoTipProps {
...
 function InfoTip({ label, children }: InfoTipProps): React.ReactElement {
+  const ariaLabel = label != null ? `Info about ${label}` : "More information";
   return (
     <Popover.Root>
       <Popover.Trigger
         render={<span className={styles.osdkFormFieldInfoIcon} />}
         nativeButton={false}
         openOnHover={true}
-        aria-label={label != null ? `Info about ${label}` : "More information"}
+        aria-label={ariaLabel}
```
nfdiop, line 113: required for accessibility

packages/react-components/src/shared/calendar/__tests__/DatePicker.test.tsx
```diff
@@ -862,7 +862,7 @@ describe("DatePicker", () => {
 
       // Find the end-of-popover focus sentinel inside the dialog.
       const endSentinel = dialog.querySelector(
-        "[aria-label='End of date picker dialog']",
+        '[data-osdk-date-picker-focus-boundary="end"]',
```
nfdiop, line 865: The aria-label was not required and actually bad for accessibility

Image

packages/react-components/src/tokens/component-tokens/combobox.css
```diff
@@ -1,9 +1,86 @@
 :root {
-  /* Combobox popup: cap the dropdown's max-height so very long option lists
```
nfdiop, line 2: Dropped the comment since it should be obvious why we limit

packages/react-components/docs/CSSVariables.md
```diff
@@ -107,10 +107,13 @@ Control text appearance.
 | ------------------------------------------ | ---------------------------------------- | ---------------------------- |
 | `--osdk-typography-family-default`         | `--bp-typography-family-default`         | Default font family          |
 | `--osdk-typography-color-muted`            | `--bp-typography-color-muted`            | Muted/subtle text color      |
+| `--osdk-typography-color-placeholder`      | `--bp-typography-color-muted`            | Theme-aware placeholder text |
 | `--osdk-typography-color-default-rest`     | `--bp-typography-color-default-rest`     | Default text color           |
 | `--osdk-typography-color-default-hover`    | `--bp-typography-color-default-hover`    | Default hover text color     |
 | `--osdk-typography-color-default-active`   | `--bp-typography-color-default-active`   | Default active text color    |
 | `--osdk-typography-color-default-disabled` | `--bp-typography-color-default-disabled` | Default disabled text color  |
+| `--osdk-typography-color-error`            | `--bp-typography-color-danger-hover`     | Theme-aware error text       |
```
nfdiop: Not 100% sold on this but I have defined these tokens so we can use the same typography token in both light and dark mode

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -36,6 +36,7 @@
   /* Blueprint token mappings - Typography tokens */
   --osdk-typography-family-default: var(--bp-typography-family-default);
   --osdk-typography-color-muted: var(--bp-typography-color-muted);
+  --osdk-typography-color-placeholder: var(--bp-typography-color-muted);
```
nfdiop: token added because we need to fix the color in dark mode for select items. It is easier to fix it by having this token work in light/dark mode

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -46,6 +47,8 @@
   --osdk-typography-color-default-disabled: var(
     --bp-typography-color-default-disabled
   );
+  --osdk-typography-color-error: var(--bp-typography-color-danger-hover);
```
ziyunp: Prefer to have 1:1 mapping of osdk to bp tokens.


```suggestion
  --osdk-typography-color-danger-hover: var(--bp-typography-color-danger-hover);
```

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: Can we make this secondary mapping and keep it in another section outside of blueprint token mappings? We should try keeping blueprint token mappings section for blueprint 1-2-1 mappings only.

```suggestion
 --osdk-typography-color-placeholder: var(--osdk-typography-color-muted);
```

packages/react-components/src/tokens/base-tokens/base.css
nfdiop: I couldn't do that because in dark mode we use the token `--osdk-palette-red-4`: there is no equivalent for --bp-typography-color-danger-hover in dark mode. Ideally it would work in all cases. see https://github.com/palantir/osdk-ts/pull/3933#discussion_r3852409976

packages/react-components/src/action-form/fields/DropdownField.tsx
```diff
@@ -447,7 +447,7 @@ const ComboboxDropdown = typedReactMemo(function ComboboxDropdownFn<
             />
           )}
           <Combobox.Positioner>
-            <Combobox.Popup>
+            <Combobox.Popup aria-label={placeholder || "Options"}>
```
ziyunp: question: the placeholder is instructional text rather than the field's name, so SRs will announce "Choose an object, dialog" instead of the label the user actually sees. Would pointing `aria-labelledby` at the `FormField` label read better? Also curious why the `Select` branch's `Select.Popup` doesn't get the same treatment — is that because Base UI already labels `role="listbox"` from the trigger?

nit: `||` here vs `!= null` for `placeholder` everywhere else in this file (L204, L418).

packages/react-components/src/tokens/base-tokens/dark.css
```diff
@@ -41,7 +41,10 @@
   --osdk-dark-background-primary: #111418;
   --osdk-dark-background-secondary: #1c2127;
   --osdk-dark-background-tertiary: #1c2127;
+  --osdk-dark-typography-color-error: var(--osdk-palette-red-4);
   --osdk-dark-typography-color-muted: #8f99a8;
+  --osdk-dark-typography-color-placeholder: var(--osdk-palette-gray-4);
```
ziyunp, line 46: question: `--osdk-palette-gray-4` is `#abb3bf`, which lands around 8.6:1 on `#111418` — brighter than the default body text (`#a5aab3`, ~7.9:1). So in dark mode the placeholder ends up more prominent than actual entered text, which inverts the hierarchy. Was something dimmer ruled out for contrast reasons? The previous `--osdk-typography-color-muted` (`#8f99a8`) is ~6.1:1, so still comfortably AA.

packages/react-components/src/tokens/component-tokens/select.css
```diff
@@ -1,11 +1,49 @@
 :root {
-  /* Select trigger: border is none; visual border comes from box-shadow
-     (--osdk-button-shadow) to match Blueprint's convention */
+  /* Shared */
+  --osdk-select-spacing: var(--osdk-surface-spacing);
+  --osdk-select-min-height: 30px;
+  --osdk-select-border-radius: var(--osdk-surface-border-radius);
+  /* The visual border comes from --osdk-button-shadow to match Blueprint. */
   --osdk-select-border-width: none;
+  --osdk-select-border-color: var(--osdk-surface-border-color-default);
```
ziyunp, line 8: question: Not from this PR but`--osdk-select-border-width: none` sounds like it is to be used in a border style like this  `border: none solid <color>`. "none" isn't a valid line width, how is this used? Should we set it to 0px instead?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -46,6 +46,8 @@
...
```
ziyunp: suggestion (non-blocking): both this and `--osdk-typography-color-primary-rest` get re-pointed in `dark.css` (L44, L47), so they aren't plain 1:1 Blueprint mappings — but `CSSVariables.md` documents them in the Blueprint mapping table with only the light value. Worth a note in the docs (and maybe an inline `/* diverges in dark */`), since `dark.css`'s own header comment sets that expectation: "tokens that ... intentionally diverge from Blueprint's value". Related: `--osdk-typography-color-placeholder` moved into a semantic section after the same feedback, while these two stayed in the mapping block.

packages/react-components/src/tokens/component-tokens/combobox.css
```diff
@@ -1,9 +1,93 @@
...
+  --osdk-combobox-line-height: var(--osdk-typography-line-height-default);
+  --osdk-combobox-placeholder-color: var(--osdk-typography-color-placeholder);
+  --osdk-combobox-icon-color: var(--osdk-iconography-color-muted);
+  --osdk-combobox-transition-duration: var(--osdk-emphasis-transition-duration);
+  --osdk-combobox-transition-ease: var(--osdk-emphasis-ease-default);
+  --osdk-combobox-z-index: var(--osdk-surface-z-index-3);
+
+  /* Trigger */
+  --osdk-combobox-trigger-spacing: var(--osdk-surface-spacing);
+  --osdk-combobox-trigger-min-height: 30px;
```
ziyunp, line 19: Can we use a factor of --osdk-surface-spacing for hardcoded 30px here and --osdk-combobox-min-height above?

packages/react-components/src/action-form/__tests__/DropdownField.test.tsx
```diff
@@ -362,6 +362,25 @@ describe("DropdownField", () => {
       expect(screen.getByText("Pick one…")).toBeDefined();
     });
 
+    it("labels the popup with the field placeholder", async () => {
```
ziyunp: suggestion (non-blocking): good to see this covered — could we add the no-placeholder case too? The `|| "Options"` fallback is the branch most likely to regress silently.

packages/react-components/src/action-form/__tests__/FormField.test.tsx
```diff
@@ -86,6 +86,25 @@ describe("FormField", () => {
   });
 
   describe("tooltip helper text", () => {
+    it("labels the helper text dialog", () => {
```
ziyunp: suggestion (non-blocking): same thought as the dropdown test — the `label == null` -> `"More information"` fallback isn't exercised anywhere.

packages/react-components/src/shared/calendar/DatePicker.tsx
```diff
@@ -480,7 +480,7 @@ export const DatePicker: React.NamedExoticComponent<DatePickerProps> =
                 <div
                   onFocus={handleStartFocusBoundary}
                   tabIndex={0}
-                  aria-label="Start of date picker dialog"
+                  data-osdk-date-picker-focus-boundary="start"
```
ziyunp: thought: agreed the old label was misleading, but the sentinel is still `tabIndex={0}` with no role and no name, so tabbing onto it now announces nothing at all rather than something wrong. Have you considered dropping `tabIndex` and handling Tab via `onKeyDown` on the popup instead? Same applies to `DateRangePicker.tsx`. Happy for this to be a FLUP if it's a bigger change than it looks.

packages/react-components/src/base-components/combobox/Combobox.module.css
```diff
@@ -606,88 +398,46 @@
...
+    border-color: var(--osdk-combobox-checkbox-border-color-checked);
+    color: var(--osdk-combobox-checkbox-color-checked);
   }
 }
 
 .osdkComboboxTriggerChips {
   display: flex;
   flex-wrap: wrap;
-  gap: calc(var(--osdk-combobox-spacing, var(--osdk-surface-spacing)) * 1);
+  gap: calc(var(--osdk-combobox-spacing) * 1);
```
ziyunp: nit: `calc(var(--osdk-combobox-spacing) * 1)` is just `var(--osdk-combobox-spacing)` (also L433, L434).

packages/react-components/src/base-components/combobox/Combobox.module.css
ziyunp: Nice!

packages/react-components/src/action-form/FormField.module.css
ziyunp: I also want to try out to see if we can have only the core set of blueprint tokens, and only expose that set.

packages/react-components/src/shared/calendar/__tests__/DatePicker.test.tsx
ziyunp: Are we doing anything to address this? Why did we change aria-label to a data attr?

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: Can we move this to the same section under /* Typography colors */ as well pls? Maybe add a comment explaining why they are needed

packages/react-components/src/tokens/component-tokens/combobox.css
```diff
@@ -1,9 +1,93 @@
 :root {
-  /* Combobox popup: cap the dropdown's max-height so very long option lists
-     scroll inside the popup instead of forcing the popup to grow tall enough
-     to push other UI off-screen. The actual `max-height` rule uses
-     `min(this token, --available-height)` — `--available-height` is set
-     automatically by Base UI's `Combobox.Positioner` (same source as
-     `--anchor-width`) so the cap is automatically lowered on short viewports. */
+  /* Shared */
+  --osdk-combobox-spacing: var(--osdk-surface-spacing);
```
ziyunp, line 3: Do you think we need custom spacing variables for variables that use --osdk-surface-spacing directly? Maybe we can remove those spacing variables until we see a need to add them?

packages/react-components/src/tokens/base-tokens/dark.css
ziyunp: Note: AI review detected this, can u pls attach the dark mode placeholder vs input text where this token is used just to double check?

packages/react-components/src/tokens/component-tokens/combobox.css
nfdiop: Probably not needed. There are many component tokens I think we can drop. But I want us to focus on that separately. In this PR, I'm focusing on refactoring + a11y in this PR.

packages/react-components/src/tokens/component-tokens/combobox.css
nfdiop: I did hesitate but I didn't have a good rationale. Since this PR is focused on refactoring + a11y, I didn't want to introduce some coupling with the spacing. I prefer if we take some time and take a deeper look at our tokens

packages/react-components/src/tokens/component-tokens/select.css
nfdiop: not sure. It was introduced here https://github.com/palantir/osdk-ts/pull/3473 and we can look into it as a FLUP

packages/react-components/src/shared/calendar/__tests__/DatePicker.test.tsx
nfdiop: Because having an aria-label here was flagged an accessibility issue (see image attached).  And I couldn't think of a role since these divs are mainly used for focus management but happy to reconsider :)

packages/react-components/src/tokens/base-tokens/dark.css
nfdiop: Thanks a lot for catching! Let me check!

ziyunp (approved):

packages/react-components/src/tokens/base-tokens/dark.css
nfdiop: Screenshot 2026-08-28 at 14 22 59
Screenshot 2026-08-28 at 14 22 55

packages/react-components-storybook/src/styles/storybook.css
```diff
@@ -211,6 +211,5 @@ body,
 }
 
 .osdkRichDropdownDescription {
-  color: var(--osdk-typography-color-muted);
```
nfdiop, line 214: this wasn't accessible

PR #3945, opened 2026-08-27
Fix RDP Object Set loading states when combined with other query loads

nihalbhatnagar:
Users were running into issues where RDP-backed object-set queries would set their state as loaded even though the RDP load hadn't finished.

```typescript
const { data: oneData, isLoading: oneLoading, error: oneError } = useObjectSet(oneObjectSet);
    const {
        data: twoData,
        isLoading: twoLoading,
        error: twoError,
    } = useObjectSet(twoObjectSet, {
        withProperties: {
            purposeIds: base => base.pivotTo("purposes").aggregate("purposeId:collectList"),
        },
    });
```

What would happen here is that the second use object set would return a list of undefined.

There are two different stages:

  1. A cache update tells the object-set query: “Object Document/123 was added and definitely matches your filter.”
  2. The object-set result later resolves its stored keys by reading each exact cache entry.

With the example given, the first non-RDP query finished and wrote Document/123 to the shared cache. That write notified the second object-set query because it watches the same object type. That update flows to "#handleLocalUpdate". That then evaluated Document/123 against its filter and determined that it belonged in its result set, which it said "yes it did". This is because the only check it performs for the object being in the object set is based on the object set filter.

At stage 2, even though the object did belong to the object set, when we went to perform the lookup with the RDP key, we saw undefined and returned it. This is because the object itself is resolved via a different cache key.

The behavior to check "is this part of my object set" is still correct. However, we should also check that the cache key we resolve to actually exists, which is what this PR adds a check for.

you may ask, whats the point of checking object set scope and cache availability in two separate checks? 
 The checks answer different questions. The membership check tells us whether the object should appear in the result, while the cache check tells us whether we can return its requested RDP representation yet.

If the object belongs but its RDP variant is unavailable, omitting it would make the result look complete when it is not. Instead, we omit it temporarily, keep the query loading, and revalidate so the missing data is fetched. Once the exact cache variant is available, the object can safely be added to the result.

### DIFF
--- a/packages/client/src/observable/internal/object/ObjectsHelper.test.ts
+++ b/packages/client/src/observable/internal/object/ObjectsHelper.test.ts
@@ -269,6 +269,49 @@ describe("ObjectsHelper.propagateWrite RDP merge", () => {
     store.cacheKeys.release(queryWithRdp.cacheKey);
   });
 
+  it("loads an object with no RDPs from an RDP write without derived fields", () => {
+    const rdpConfig = createFakeRdpConfig("derivedAddress");
+    const queryWithRdp = store.objects.getQuery(
+      {
+        apiName: Employee,
+        pk: 1,
+      },
+      rdpConfig,
+    );
+    const queryNoRdp = store.objects.getQuery({
+      apiName: Employee,
+      pk: 1,
+    });
+    store.cacheKeys.retain(queryNoRdp.cacheKey);
+    const subscription = store.subjects
+      .get(queryNoRdp.cacheKey)
+      .subscribe(() => {});
+    store.batch({}, (batch) => {
+      queryNoRdp.writeToStore(emp as any, "loading", batch);
+    });
+    const rdpValue = emp.$clone({
+      fullName: "Bob",
+      derivedAddress: "123 Main St",
+    } as any);
+
+    store.batch({}, (batch) => {
+      queryWithRdp.writeToStore(rdpValue as any, "loaded", batch);
+    });
+
+    expect(store.getValue(queryNoRdp.cacheKey)).toMatchObject({
+      status: "loaded",
+      value: {
+        $primaryKey: 1,
+        fullName: "Bob",
+      },
+    });
+    expect(
+      (store.getValue(queryNoRdp.cacheKey)?.value as any)?.derivedAddress,
+    ).toBeUndefined();
+    subscription.unsubscribe();
+    store.cacheKeys.release(queryNoRdp.cacheKey);
+  });
+
   it("keeps the recomputed derived value on a $select refetch", () => {
     const rdpConfig = createFakeRdpConfig("derivedAddress");
     const queryB = store.objects.getQuery(
--- a/packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
+++ b/packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
@@ -405,17 +405,11 @@ export class ObjectSetQuery extends BaseListQuery<
       effectiveWhere,
     );
 
-    const status =
-      optimisticId ||
-      addedMatches.uncertain.size > 0 ||
-      modifiedMatches.uncertain.size > 0
-        ? "loading"
-        : "loaded";
-
     const { retVal: needsRevalidation } = this.store.batch(
       { optimisticId, changes },
       (batch) => {
-        const existingKeys = new Set(batch.read(this.cacheKey)?.value?.data);
+        const existingEntry = batch.read(this.cacheKey);
+        const existingKeys = new Set(existingEntry?.value?.data);
 
         const { newList, needsRevalidation } = reconcileListChanges(
           existingKeys,
@@ -425,9 +419,22 @@ export class ObjectSetQuery extends BaseListQuery<
           changes.deleted,
           batch.optimisticWrite,
           (obj) => this.#getObjectCacheKey(obj),
+          (key) => {
+            const value = batch.read(key)?.value;
+            return value != null && typeof value === "object";
+          },
         );
 
-        const existingTotalCount = batch.read(this.cacheKey)?.value?.totalCount;
+        const status =
+          (this.pendingFetch != null && existingEntry?.status === "loading") ||
+          needsRevalidation ||
+          optimisticId ||
+          addedMatches.uncertain.size > 0 ||
+          modifiedMatches.uncertain.size > 0
+            ? "loading"
+            : "loaded";
+
+        const existingTotalCount = existingEntry?.value?.totalCount;
         this._updateList(
           newList,
           status,
@@ -440,7 +447,7 @@ export class ObjectSetQuery extends BaseListQuery<
       },
     );
 
-    if (needsRevalidation) {
+    if (needsRevalidation && !optimisticId) {
       return this.revalidate(true);
     }
     return undefined;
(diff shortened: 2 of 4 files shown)


packages/client/src/observable/internal/object/ObjectsHelper.test.ts
```diff
@@ -269,6 +269,49 @@ describe("ObjectsHelper.propagateWrite RDP merge", () => {
     store.cacheKeys.release(queryWithRdp.cacheKey);
   });
 
+  it("loads an object with no RDPs from an RDP write without derived fields", () => {
```
nihalbhatnagar, line 272: I added this test even though this PR doesn't change its behavior for my own understanding. Basically I wanted to guarantee that a base object variant could still be satisfied from the RDP variant without waiting on a load, which this verified.

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -425,9 +419,22 @@ export class ObjectSetQuery extends BaseListQuery<
           changes.deleted,
           batch.optimisticWrite,
           (obj) => this.#getObjectCacheKey(obj),
+          (key) => {
+            const value = batch.read(key)?.value;
```
nihalbhatnagar, line 427: This new check actually makes sure that the key we claim to be part of the object set actually is in the cache and loaded

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -543,18 +550,30 @@ function reconcileListChanges(
...
+  hasCachedObject: (key: ObjectCacheKey) => boolean,
 ): { newList: ObjectCacheKey[]; needsRevalidation: boolean } {
-  const objectsToInsert = new Set<ObjectHolder | InterfaceHolder>(
-    addedDefiniteMatches,
-  );
+  const objectsToInsert = new Set<ObjectHolder | InterfaceHolder>();
   const keysToRemove = new Set<CacheKey>(deleted);
 
   let needsRevalidation = false;
+  const addIfAvailable = (obj: ObjectHolder | InterfaceHolder): void => {
```
nihalbhatnagar, line 612: They are two categories of cache changes, not necessarily two different kinds of objects:

  - addedDefiniteMatches: objects newly introduced to the cache that match the query.
  - modifiedObjects: objects already known to the cache whose properties changed.

We need to check addedDefiniteMatches for this specific example to make sure that when its loaded it doesn't trigger the RDP query.  For modified objects, we need the same check because an existing object may change and newly match the query before the RDP invalidaiton query has finished

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
nihalbhatnagar, line 399: This is where we're basically saying "This is part of my object set"

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -543,18 +550,30 @@ function reconcileListChanges(
...
+    const key = getObjectCacheKey(obj);
+    if (existingKeys.has(key)) {
+      return;
+    }
+    if (hasCachedObject(key)) {
```
nihalbhatnagar: This would now resolve to false for the case described in the description

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -440,7 +447,7 @@ export class ObjectSetQuery extends BaseListQuery<
       },
     );
 
-    if (needsRevalidation) {
+    if (needsRevalidation && !optimisticId) {
```
nihalbhatnagar, line 458: This avoids starting a request in the middle of an optimistic update

packages/client/src/observable/internal/object/ObjectsHelper.test.ts
nihalbhatnagar: Essentially what I think this means is that this change was scoped to object set/list vlaidations but object normalization still functioned correctly, which is why we were seeing undefined

KuberSethi: How did we actually test this btw? Did we make a repro / use one of the notional apps?

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -425,9 +419,22 @@ export class ObjectSetQuery extends BaseListQuery<
...
           (obj) => this.#getObjectCacheKey(obj),
+          (key) => {
+            const value = batch.read(key)?.value;
+            return value != null && typeof value === "object";
+          },
         );
 
-        const existingTotalCount = batch.read(this.cacheKey)?.value?.totalCount;
+        const status =
+          (this.pendingFetch != null && existingEntry?.status === "loading") ||
```
KuberSethi: there's a lot happening here, could we pull this out into some boolean vars to make this more readable

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
```diff
@@ -543,18 +550,30 @@ function reconcileListChanges(
...
   getObjectCacheKey: (obj: ObjectHolder | InterfaceHolder) => ObjectCacheKey,
+  hasCachedObject: (key: ObjectCacheKey) => boolean,
 ): { newList: ObjectCacheKey[]; needsRevalidation: boolean } {
-  const objectsToInsert = new Set<ObjectHolder | InterfaceHolder>(
-    addedDefiniteMatches,
-  );
+  const objectsToInsert = new Set<ObjectHolder | InterfaceHolder>();
   const keysToRemove = new Set<CacheKey>(deleted);
 
   let needsRevalidation = false;
```
KuberSethi, line 611: not super opinionated either way but thoughts on pulling out all of the below revalidation handling (esp now wrt optimistic updates) into a helper function so it's a bit more clear to readers? Generally very pro anything we can do to keep complexity down in complex files / places like this

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
KuberSethi, line 458: I think this is fine for now but I'm wondering if we should make changes even more upstream to distinguish requests between revalidations vs fetching more data and then pushing all of this logic there (like into `BaseListQuery` and `Query`). This way the other Query types can benefit and we don't need to add this complexity / level of concern to each Query type. Maybe something to do in the future as y'all ramp up!

KuberSethi (commented):
Generally looks good, thank you for doing this! Just a few small questions.

packages/client/src/observable/internal/object/ObjectsHelper.test.ts
nihalbhatnagar: I just validated using test failure/pass before and after the fix. I think relying on manual validation isn't something we should require, and instead make sure our tests are reliable, which it seems to be correct in this case. That being said, I'll test it to get a hang of how to do it

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
nihalbhatnagar: Yup makes this more clear

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
nihalbhatnagar: Yeah done

packages/client/src/observable/internal/objectset/ObjectSetQuery.ts
nihalbhatnagar: Claude generated issue for this https://github.com/palantir/osdk-ts/issues/3996

PR #3951, opened 2026-08-31
Add empty backing media set support

colin4554:
### Changes:
- Add includeEmptyBackingMediaSet flag so media ref properties can include a backing media set as an output

### DIFF
--- a/packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ObjectTypeShapeExtractor.ts
+++ b/packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ObjectTypeShapeExtractor.ts
@@ -204,18 +204,6 @@ export class ObjectTypeShapeExtractor {
       );
       for (const [id, shape] of Array.from(dsShapes.entries())) {
         blockShapes.inputShapes.set(id, shape);
-        if (
-          shape.type === "filesDatasource" &&
-          shape.filesDatasource.supportedTypes?.some(
-            ({ type }) => type === "mediaSet",
-          )
-        ) {
-          blockShapes.inputShapeMetadata.set(id, {
-            isOptional: false,
-            isAccessedInReconcile: true,
-            reconcileAccessRequirements: "RESOURCE_EXISTENCE_REQUIRED",
-          });
-        }
       }
     }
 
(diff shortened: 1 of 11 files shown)


packages/maker-experimental/src/util/generateRid.ts
```diff
@@ -282,6 +282,10 @@ export class ReadableIdGenerator {
     return `media-set-view-${mediaSetViewName}` as ReadableId;
   }
 
+  static getForMediaSetOutput(mediaSetViewName: string): ReadableId {
```
ethanai22: nit: `getForMediaSetViewOutput` and
return media-set-view-output-${mediaSetViewName} as ReadableId;

packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ObjectTypeShapeExtractor.ts
```diff
@@ -204,6 +204,18 @@ export class ObjectTypeShapeExtractor {
       );
       for (const [id, shape] of Array.from(dsShapes.entries())) {
         blockShapes.inputShapes.set(id, shape);
+        if (
+          shape.type === "filesDatasource" &&
+          shape.filesDatasource.supportedTypes?.some(
+            ({ type }) => type === "mediaSet",
+          )
+        ) {
+          blockShapes.inputShapeMetadata.set(id, {
```
ethanai22: Do you need this? this changes the behavior of existing media set datasources, and it looks like you already do this here: https://github.com/palantir/osdk-ts/pull/3951/changes#diff-795b961c7aa4803259f292fd14137283d7f27c77cf4e3fd6b647b9794bc58ac5R87

packages/maker-experimental/src/util/generateRid.ts
colin4554: Updated

packages/maker-experimental/src/conversion/toMarketplace/shapeExtractors/ObjectTypeShapeExtractor.ts
colin4554: Good call

ethanai22 (approved):

PR #3961, opened 2026-09-01
Fix behavior when multiple useOsdkObjects hooks were sharing max pageSize

williamhe723:
Previously, if I had two `useOsdkObjects` hooks making the same query, with one of them setting a `pageSize` of 20, and the other without `pageSize`, both would only fetch 20.

With my changes, the query without a `pageSize` will now default to 100, which was the expected behavior.

### DIFF
--- a/.changeset/stale-bees-wonder.md
+++ b/.changeset/stale-bees-wonder.md
@@ -1,0 +1,5 @@
+---
+"@osdk/client": patch
+---
+
+Fix bug where empty page size parameter was behaving incorrectly
--- a/.changeset/stale-bees-wonder.md
+++ b/.changeset/stale-bees-wonder.md
@@ -2,4 +2,4 @@
 "@osdk/client": patch
 ---
 
-Fix bug where empty page size parameter was behaving incorrectly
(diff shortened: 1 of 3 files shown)


.changeset/stale-bees-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@osdk/client": patch
+---
+
+Fix bug where empty page size parameter was behaving incorrectly
```
KuberSethi: hypernit: this could be a little more specific - what was the bug / what did we change?

KuberSethi (commented):
Looks good, thank you for fixing this!! Just curious - how did you go about testing this? Will also leave final +1 to Bryant as I want someone from the team to review and stamp.

williamhe723:
> Looks good, thank you for fixing this!! Just curious - how did you go about testing this? Will also leave final +1 to Bryant as I want someone from the team to review and stamp.

I created a local react OSDK app that made 2 requests, one with `pageSize`=20 and one without. Before my changes, both of them loaded 20 entries, but after my changes, the first one loaded 20 while the second request loaded 100. Both still had the same cache entry.

bryantpark04:
:+1:

nihalbhatnagar (approved):

PR #3962, opened 2026-09-02
Materialize array-valued struct interface implementations

nihalbhatnagar:
## Summary

Interface materialization treated every struct implementation as a non-array value, leaving array-valued struct properties unresolved. If we saw a struct array implementation backed by a struct array, we need to map the array correctly.

This PR:

- materialize array-valued struct implementations element-by-element
- validate the expected single array-valued backing property shape
- add regression coverage for valid arrays and malformed scalar values

### DIFF
--- a/packages/e2e.generated.catchall/ontology.json
+++ b/packages/e2e.generated.catchall/ontology.json
@@ -1116,25 +1116,25 @@
           "rid": "ri.ontology.main.interface.ee112779-ed04-44c6-bf84-9543513eb80b",
           "properties": {},
           "propertiesV2": {
-            "stringFromNonMainValueOfStruct": {
+            "stringFromStruct": {
               "type": "structFieldImplementation",
               "structFieldOfProperty": {
                 "propertyApiName": "struct",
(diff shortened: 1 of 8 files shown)


packages/e2e.generated.catchall/ontology.json
```diff
@@ -1116,25 +1116,25 @@
           "rid": "ri.ontology.main.interface.ee112779-ed04-44c6-bf84-9543513eb80b",
           "properties": {},
           "propertiesV2": {
-            "stringFromNonMainValueOfStruct": {
+            "stringFromStruct": {
```
nihalbhatnagar, line 1119: These mappings were out of date but not relevant to this PR specifically

jameszhang244 (dismissed):

jameszhang244:
I think build is failing because the `ontology.json` changes require regenerating the catchall SDK files. Can you run `pnpm turbo codegen --filter=@osdk/e2e.generated.catchall`?

jameszhang244 (approved):

PR #3964, opened 2026-09-02
Deprecate initial PDF viewer props

ziyunp:
## Summary

Align PDF viewer uncontrolled state props with the react-components contribution guidelines while preserving backwards compatibility.

## Changes

- Add defaultPage, defaultScale, defaultAutoSize, and defaultSidebarOpen across PDF viewer components and hooks.
- Keep the corresponding initial* props as deprecated aliases, with default* taking precedence.
- Update DocumentViewer examples, PDF viewer stories, documentation, tests, and the package changeset.


packages/react-components/src/pdf-viewer/BasePdfViewer.tsx
```diff
@@ -48,10 +48,14 @@ export const BasePdfViewer: React.ForwardRefExoticComponent<
     formData,
     onFormSubmit,
     onFormChange,
-    initialPage = 1,
```
ziyunp, line 51: Default set in the hooks, this is duplicated

nfdiop (dismissed):

nfdiop (approved):

PR #3965, opened 2026-09-02
Test Blueprint next tokens in React components

ziyunp:
## Summary

Tests replacing the React Components base-token vocabulary with Blueprint's proposed next tokens from palantir/blueprint#8250. The goal is to evaluate whether product themes can override only Blueprint's public `--bp-*` contract while existing OSDK component-token names continue to work.

## Changes

- Import `next-tokens.css` instead of the legacy Blueprint token copy.
- Map retained OSDK base aliases to tokens from the new Blueprint vocabulary.
- Preserve token categories: retained `--osdk-intent-*` aliases map to `--bp-intent-*`, typography to typography, palette to palette, and emphasis to emphasis.
- Map the legacy `default` intent to Blueprint's `neutral` intent, using `500 / 600 / 700` for rest, hover, and active.
- Use base surface color for resting ground and neutral layer colors for interactive default hover and active backgrounds.
- Use default and strong divider tokens for the legacy neutral separator aliases.
- Use component-level `--bp-emphasis-disabled-opacity` instead of disabled color aliases.
- Replace remaining component-token references to removed legacy Blueprint names with valid next tokens.
- Remove 18 OSDK aliases with no runtime `var(...)` consumers and remove their documentation entries.
- Apply Storybook and People App themes as later CSS layers containing only next-token `--bp-*` overrides; remove Storybook's runtime one-to-many token expansion.
- Keep the legacy `blueprint-design-tokens.css` and `dark.css` files unchanged and unimported.

## Mapping model

The RFC defines `palette → intent → applied token → component`.

- Palette tokens are used directly only for values without semantic meaning, such as shadow construction and the dialog scrim.
- Intent aliases remain intent-to-intent. Foregrounds use the matching intent foreground.
- Opaque surfaces use `--bp-surface-background-color-*`.
- Translucent `background-color` consumers use `--bp-surface-layer-color-*`; stackable background-image consumers use `--bp-surface-layer-*`.
- Standard and semantic text use `--bp-typography-color-base` and `--bp-typography-color-intent-*`.
- Focus, disabled state, motion, spacing, radius, elevation, and stacking use their corresponding emphasis or surface tokens.

## Concerns and token gaps

- **Intent aliases have mixed consumers.** Existing `--osdk-intent-*` aliases are used for filled backgrounds as well as text, icons, borders, and focus treatments. Keeping them in the intent category satisfies the mapping constraint, but those text and structural consumers still bypass the RFC's applied typography and surface roles.
- **Dark interaction ramps differ.** Category-preserving aliases use `500 / 600 / 700`. Blueprint's applied dark filled surfaces use `500 / 400 / 600`, so existing components that treat an OSDK intent alias as a fill will not follow the new dark surface interaction ramp.
- **Generic border aliases mix roles.** OSDK's default border token is used for both component outlines and separators. Blueprint distinguishes intent-bound component borders from neutral dividers, so no single mapping is semantically correct for every current consumer.
- **No strong neutral component-border role.** The new set has a strong divider but only one border color per intent. A component that needs a stronger neutral outline has no direct applied token.
- **No disabled intent or surface color states.** The RFC uses opacity on the entire component. This works for current disabled controls, but there is no token for a design that needs a distinct disabled color without dimming the whole component.
- **No stateful typography intent colors.** Semantic text has one color per intent, with no hover or active variants.
- **No dedicated placeholder/secondary-foreground token.** Placeholder and muted content currently share neutral intent typography, although these roles may require different contrast.
- **No dedicated backdrop/scrim token.** The dialog backdrop maps to mode-independent `--bp-palette-black-500` because it has no semantic intent.
- **Line height exposes only a factor.** Preserving the previous OSDK line height requires a consumer multiplier (`1.105 × 1.164 ≈ 1.286`) rather than a direct semantic line-height token.
- **Base surfaces deliberately do not react.** Existing interactive default backgrounds therefore use translucent neutral layer colors. This preserves state feedback, but their result depends on the surface underneath.
- **Skeleton contrast changes.** The old animation used approximately 5% → 40%; the new neutral layers use 3% → 12% in light mode, producing a much subtler glow.
- **Typography changes are visible.** Body sizes move from `10/12/14/16px` to `9/11/13/15px`, strong weight moves from 600 to 500, and disabled opacity moves from 0.5 to 0.4.
- **Intent foregrounds require theme validation.** The new warning foreground is white instead of the previous dark foreground. Any custom palette, intent, foreground, or applied filled-surface override must rerun the RFC's 4.5:1 contrast checks.

## Assumptions

- This is intentionally a breaking-change experiment; possible external consumers of removed OSDK aliases are not preserved.
- A runtime consumer means a `var(--osdk-...)` reference in this repository. Documentation and the unimported legacy `dark.css` do not count as consumers.
- Component-token names remain the compatibility boundary even where their current usage does not cleanly fit the new architecture.

### DIFF
--- a/packages/react-components/src/tokens/base-tokens/base.css
+++ b/packages/react-components/src/tokens/base-tokens/base.css
@@ -1,25 +1,23 @@
-@import "./blueprint-design-tokens.css";
+@import "./next-tokens.css";
 
-/* Level 1 - Blueprint Tokens */
 :root {
-  /* Blueprint token mappings - Surface tokens */
   --osdk-surface-z-index-1: var(--bp-surface-z-index-1);
   --osdk-surface-z-index-2: var(--bp-surface-z-index-2);
   --osdk-surface-z-index-3: var(--bp-surface-z-index-3);
   --osdk-surface-z-index-4: var(--bp-surface-z-index-4);
   --osdk-surface-border-radius: var(--bp-surface-border-radius);
-  --osdk-surface-layer-primary: var(--bp-surface-layer-primary);
+  --osdk-surface-layer-primary: var(--bp-surface-layer-primary-rest);
   --osdk-surface-border-width: var(--bp-surface-border-width);
-  --osdk-surface-border-color-default: var(--bp-surface-border-color-default);
-  --osdk-surface-border-color-strong: var(--bp-surface-border-color-strong);
+  --osdk-surface-border-color-default: var(--bp-surface-divider-color-default);
+  --osdk-surface-border-color-strong: var(--bp-surface-divider-color-strong);
   --osdk-surface-background-color-default-rest: var(
-    --bp-surface-background-color-default-rest
+    --bp-surface-background-color-base-rest
   );
   --osdk-surface-background-color-default-hover: var(
-    --bp-surface-background-color-default-hover
+    --bp-surface-background-color-base-hover
   );
   --osdk-surface-background-color-default-active: var(
-    --bp-surface-background-color-default-active
+    --bp-surface-background-color-base-active
   );
   --osdk-surface-background-color-danger-rest: var(
     --bp-surface-background-color-danger-rest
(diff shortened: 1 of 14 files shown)


packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -1,25 +1,23 @@
...
   --osdk-surface-z-index-3: var(--bp-surface-z-index-3);
   --osdk-surface-z-index-4: var(--bp-surface-z-index-4);
   --osdk-surface-border-radius: var(--bp-surface-border-radius);
-  --osdk-surface-layer-primary: var(--bp-surface-layer-primary);
+  --osdk-surface-layer-primary: var(--bp-surface-layer-primary-rest);
   --osdk-surface-border-width: var(--bp-surface-border-width);
-  --osdk-surface-border-color-default: var(--bp-surface-border-color-default);
-  --osdk-surface-border-color-strong: var(--bp-surface-border-color-strong);
+  --osdk-surface-border-color-default: var(--bp-surface-border-color-neutral);
+  --osdk-surface-border-color-strong: var(--bp-surface-border-color-neutral);
```
ziyunp: Check these, shouldn't use the same value

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -1,25 +1,23 @@
...
   --osdk-surface-background-color-default-rest: var(
-    --bp-surface-background-color-default-rest
+    --bp-surface-background-color-base-rest
```
ziyunp, line 14: Should this be using base?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,122 @@
...
-  );
-  --osdk-typography-color-default-active: var(
-    --bp-typography-color-default-active
-  );
+  --osdk-typography-family-default: var(--bp-typography-family-body);
+  --osdk-typography-color-muted: var(--bp-typography-color-intent-neutral);
+  --osdk-typography-color-default-rest: var(--bp-typography-color-base);
+  --osdk-typography-color-default-hover: var(--bp-typography-color-base);
+  --osdk-typography-color-default-active: var(--bp-typography-color-base);
   --osdk-typography-color-default-disabled: var(
```
ziyunp: Disabled state - should apply disabled opacity with intent color?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,122 @@
...
+  );
+  --osdk-typography-color-primary-rest: var(
+    --bp-typography-color-intent-primary
   );
-  --osdk-typography-color-primary-rest: var(--bp-typography-color-primary-rest);
-  --osdk-typography-color-danger-rest: var(--bp-typography-color-danger-rest);
+  --osdk-typography-color-danger-rest: var(--bp-typography-color-intent-danger);
   --osdk-typography-color-danger-active: var(
-    --bp-typography-color-danger-active
+    --bp-typography-color-intent-danger
```
ziyunp: How do we use active state for danger font color?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,122 @@
...
-  --osdk-typography-size-body-medium: var(--bp-typography-size-body-medium);
-  --osdk-typography-size-body-large: var(--bp-typography-size-body-large);
-  --osdk-typography-line-height-default: var(
-    --bp-typography-line-height-default
+  --osdk-typography-size-body-x-small: var(--bp-typography-size-2xs);
+  --osdk-typography-size-body-small: var(--bp-typography-size-sm);
+  --osdk-typography-size-body-medium: var(--bp-typography-size-md);
+  --osdk-typography-size-body-large: var(--bp-typography-size-xl);
+  --osdk-typography-line-height-default: calc(
+    var(--bp-typography-line-height-running-factor) * 1.164
```
ziyunp, line 42: Why the multiple?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,122 @@
...
-  --osdk-intent-default-hover: var(--bp-intent-default-hover);
-  --osdk-intent-default-active: var(--bp-intent-default-active);
-  --osdk-intent-default-disabled: var(--bp-intent-default-disabled);
-  --osdk-intent-default-foreground: var(--bp-intent-default-foreground);
-  --osdk-intent-primary-rest: var(--bp-intent-primary-rest);
-  --osdk-intent-primary-hover: var(--bp-intent-primary-hover);
-  --osdk-intent-primary-active: var(--bp-intent-primary-active);
-  --osdk-intent-primary-disabled: var(--bp-intent-primary-disabled);
+  --osdk-typography-weight-bold: var(--bp-typography-weight-strong);
+  --osdk-intent-default-rest: var(--bp-intent-neutral-500);
```
ziyunp: How are we using intent rest, hover, active states?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,122 @@
...
+  --osdk-palette-dark-gray-2: var(--bp-palette-grey-800);
+  --osdk-palette-dark-gray-3: var(--bp-palette-grey-700);
+  --osdk-palette-light-gray-1: var(--bp-palette-grey-200);
+  --osdk-palette-light-gray-3: var(--bp-palette-grey-100);
+  --osdk-palette-light-gray-4: var(--bp-palette-grey-100);
+  --osdk-palette-light-gray-5: var(--bp-palette-grey-100);
+  --osdk-palette-white: var(--bp-palette-white-1000);
+  --osdk-palette-black: var(--bp-palette-black-1000);
+  --osdk-typography-color-error: var(--bp-typography-color-intent-danger);
+  --osdk-typography-color-placeholder: var(
```
ziyunp, line 78: What color should we use for placeholder / muted color? base / neutral?

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,122 @@
...
+    --bp-typography-color-intent-neutral
+  );
+  --osdk-background-primary: var(--bp-surface-background-color-base-rest);
+  --osdk-background-secondary: var(--bp-surface-layer-color-neutral-rest);
+  --osdk-background-tertiary: var(--bp-surface-layer-color-neutral-hover);
+  --osdk-background-skeleton-from: var(--bp-surface-layer-color-neutral-rest);
+  --osdk-background-skeleton-to: var(--bp-surface-layer-color-neutral-active);
+  --osdk-background-backdrop: oklch(
+    from var(--bp-surface-background-color-neutral-active) l c h / 0.5
```
ziyunp: How is this backdrop color used?

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: Agreed — these should not use the same value. The legacy default and strong borders represented roughly 12% and 25% opacity, and --osdk-surface-border-color-strong is used for higher-emphasis boundaries such as frozen table columns. next-tokens.css only exposes one neutral border color, so there is no faithful strong-border mapping; this is a gap in the new token set.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: Base is appropriate for the resting background, but not sufficient for the interactive states. OSDK uses the hover and active aliases on buttons, menus, table controls, checkboxes, and dialogs, while the new base rest/hover/active values are identical. The intended model appears to be a base surface plus neutral hover/active layers, but the existing aliases are consumed as standalone background-color values, so that composition cannot be represented faithfully by mapping all three to base.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: --osdk-typography-color-default-disabled has no consumers. Disabled controls instead apply --osdk-disabled-opacity to the whole control in Switch, Select, Combobox, DatePicker, inputs, and other components. Applying opacity again in the text color could double-dim it. For this codebase, we should remove this unused alias and apply --bp-emphasis-disabled-opacity once at the control level.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: --osdk-typography-color-danger-active is currently unused. Error and validation text generally use the danger rest intent, while callouts use the hover intent. We should remove this unused alias. If interactive danger-colored text is required later, the new vocabulary would need stateful typography-intent tokens; a filled danger surface token would not be the right semantic substitute.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: The multiplier preserves the previous line height: 1.105 × 1.164 ≈ 1.286, versus the old 1.28581. It is only a compatibility conversion from the new running factor. A direct semantic line-height token in the new set would be preferable, since consumer-specific multipliers make consistent typography harder.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: These states directly style button backgrounds, checked checkbox/radio backgrounds, checked and unchecked switch tracks, selected calendar dates, and interactive table/header controls. Rest intent colors are also currently used for text, icons, annotations, validation, and focus borders. Those text/icon usages should generally move to --bp-typography-color-intent-{intent}; the numbered intent states are most appropriate for filled intent surfaces.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: Neutral is the better available choice. --bp-typography-color-base is the primary readable foreground, while placeholders and muted content are intentionally de-emphasized. However, OSDK uses the muted token for many roles besides placeholders, including descriptions, icons, table headers, and empty states. If placeholder contrast needs to differ from general secondary text, the new set needs a dedicated placeholder or secondary-foreground token.

packages/react-components/src/tokens/base-tokens/base.css
ziyunp: This is only used as the full-screen scrim behind Dialog: --osdk-background-backdrop → --osdk-dialog-backdrop-bg → .backdrop. The old value was black at 50% opacity. Deriving it from the neutral active filled surface produces a theme-dependent grey scrim and is not a strong semantic match. The new token set appears to need a dedicated backdrop/scrim token.

packages/react-components/src/tokens/base-tokens/base.css
```diff
@@ -32,200 +30,136 @@
...
-    --bp-typography-color-default-hover
-  );
-  --osdk-typography-color-default-active: var(
-    --bp-typography-color-default-active
-  );
+  --osdk-typography-family-default: var(--bp-typography-family-body);
+  --osdk-typography-color-muted: var(--bp-typography-color-intent-neutral);
+  --osdk-typography-color-default-rest: var(--bp-typography-color-base);
+  --osdk-typography-color-default-hover: var(--bp-typography-color-base);
+  --osdk-typography-color-default-active: var(--bp-typography-color-base);
```
ziyunp: How are these tokens used in components? What is the styling regression to map them to the same token?

PR #3969, opened 2026-09-02
[3/5] ITSM OAC DSL: Add ITSM authoring DSL

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/3968

Adds the interface type schema migration (ITSM) DSL and static validation, but does not yet emit the ITSMs in block-data.

### DIFF
--- a/packages/maker-experimental/src/conversion/toMarketplace/convertInterface.ts
+++ b/packages/maker-experimental/src/conversion/toMarketplace/convertInterface.ts
@@ -32,6 +32,9 @@ export function convertInterface(
     __type,
     status,
     linkedInterfaces: _linkedInterfaces,
+    // schema migrations travel in their own block data section rather than on the interface type
+    // directly, so we explicitly exclude it from "other"
+    schemaMigrations: _schemaMigrations,
     ...other
   } = interfaceType;
   // Normalize deprecated deadline format to match Java (strip .000 milliseconds)
--- a/packages/maker/src/api/interface/InterfaceSchemaMigrations.ts
+++ b/packages/maker/src/api/interface/InterfaceSchemaMigrations.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2026 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Interface Schema Migrations declare impending, but not yet enforced, schema changes
+ * for an interface. These give implementing object types a grace period to comply with the
+ * schema change before it becomes enforced, enabling rollout of otherwise breaking schema changes.
+ *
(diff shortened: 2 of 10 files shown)


packages/maker-experimental/src/conversion/toMarketplace/convertInterface.ts
```diff
@@ -32,6 +32,9 @@ export function convertInterface(
     __type,
     status,
     linkedInterfaces: _linkedInterfaces,
+    // schema migrations travel in their own block data section rather than on the interface type
+    // directly, so we explicitly exclude it from "other"
+    schemaMigrations: _schemaMigrations,
```
MatthewMeacham, line 37: It'll get manually included in the next PR (at a separate layer from the IT)

packages/maker/src/api/interface/InterfaceSchemaMigrations.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Interface Schema Migrations declare impending, but not yet enforced, schema changes
+ * for an interface. These give implementing object types a grace period to comply with the
+ * schema change before it becomes enforced, enabling rollout of otherwise breaking schema changes.
+ *
+ * Presence of this field on an interface type definition opts the interface type into
+ * stricter backwards-compatibility checks (e.g. you may not add a new required-to-be-implemented
+ * property except via an interface schema migration).
+ */
+export interface InterfaceSchemaMigrations {
```
MatthewMeacham, line 26: We don't rely on the generated ontology-metadata ITSM bindings for a few reasons:
* the bindings use conjure-typescript unions which aren't ergonomic typescript; instead we use discriminated typescript union 
* we rename a few properties for more human-readable DSL
* we make title required 

We should maybe upstream the last two, but in any case, the best we'd be able to re-use is this top-level object and `InterfaceSchemaTransition` which won't see a ton of changes anyhow

packages/maker/src/api/interface/validateInterfaceSchemaMigrations.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  switch (instruction.type) {
+    case "addRequiredProperty":
+      invariant(
+        !isInterfacePropertyRequired(propertiesV3[propertyApiName]),
+        `Schema migration transition "${transitionId}" on interface ${apiName} targets property "${propertyApiName}" via ${instruction.type}, but that property is required. Only properties declared required: false may be targeted.`,
+      );
+      return;
+    default:
+      // TODO: add a never exhaustiveness check once there's more than one instruction type
```
MatthewMeacham, line 177: The compiler refuses a `never` exhaustiveness check here -- something about a discriminated union with only one type/variant actually get compiled to just an object under-the-hood, so there's nothing to discriminate against

packages/maker/src/api/test/links.test.ts
```diff
@@ -2122,6 +2122,7 @@ describe("Link Types", () => {
                     "permission": undefined,
                     "propertiesV2": {},
                     "propertiesV3": {},
+                    "schemaMigrations": undefined,
```
MatthewMeacham, line 2140: Unrelated to this PR, but this is a bit strange? Linked interface types are embedded wholesale into an IT? I would've expected just a reference or something

MatthewMeacham: Ah it seems these linkedInterfaces are included for X-OAC imports, but gets stripped before emission: https://github.com/palantir/osdk-ts/blob/eb0c920910f024ebff428f50c9481b32b2f74a5f/packages/maker/src/cli/main.ts#L164-L168

So ig it's fine to include this verbatim here since it never ends up in the IR

SpecificGenerality (dismissed):

SpecificGenerality (dismissed):

davidpaquin (dismissed):

davidpaquin (approved):

PR #3976, opened 2026-09-03
Support version ranges for functions

aqding:
Adds support for generating an sdk using version ranges, which will:
- Generate bindings based on the function spec returned from gateway, when requesting with a veresion range
- Execute the version range at runtime rather than a concrete version

### DIFF
--- a/packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
+++ b/packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
@@ -62,7 +62,7 @@ export interface OntologyInfo {
   requestedMetadata: OntologyFullMetadata;
   externalInterfaces: Map<string, string>;
   externalObjects: Map<string, string>;
-  fixedVersionQueryTypes: string[];
+  queryVersionReferences: ReadonlyMap<string, string>;
 }
 
 export class OntologyMetadataResolver {
@@ -333,7 +333,7 @@ export class OntologyMetadataResolver {
         requestedMetadata: filteredFullMetadata,
         externalInterfaces,
         externalObjects,
-        fixedVersionQueryTypes: [],
+        queryVersionReferences: new Map(),
       });
     } else {
       const objectTypes = new Set(entities.objectTypesApiNamesToLoad);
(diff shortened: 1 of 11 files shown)


palantirtech:
Thanks for your interest in [palantir/osdk-ts](https://github.com/palantir/osdk-ts), @aqding! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

packages/generator/src/v2.0/generatePerQueryDataFiles.ts
```diff
@@ -127,9 +127,11 @@ async function generateV2QueryFile(
...
   );
 
-  const isUsingFixedVersion = fixedVersionQueryTypes.includes(
-    query.fullApiName,
-  );
+  const requestedVersion = queryVersionReferences.get(query.fullApiName);
+  const isUsingFixedVersion = requestedVersion !== undefined;
+  const runtimeProps = requestedVersion === undefined
+    ? baseProps
+    : { ...baseProps, version: requestedVersion };
```
aqding, line 134: use the requested version (could be a version range at request time) rather than the version returned on the query adapter (a concrete version)

packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -353,19 +353,27 @@ export class OntologyMetadataResolver {
...
-        if (queryTypes.has(queryType)) {
+        const lastColonIndex = queryType.lastIndexOf(":");
+        const queryTypeApiName = lastColonIndex === -1
+          ? queryType
+          : queryType.substring(0, lastColonIndex);
+
+        if (queryTypeApiNames.has(queryTypeApiName)) {
           return Result.err([
-            `Query type ${queryType} was specified multiple times.`,
+            `Query type ${queryTypeApiName} was specified multiple times.`,
```
aqding, line 367: This is "technically" a break, but if the same function api name was requested multiple times, the behavior was nondeterministic/undefined. If you requested the same function at different versions, you would generate code in parallel for them, but they would all write to the same `apiName.ts` file. Given that this behavior is clearly not intended/was incredibly unstable even if it was being used, it seems ok to just throw up front now

PR #3977, opened 2026-09-03
Support version ranges for functions

aqding:
Adds support for generating an sdk using version ranges, which will:
- Generate bindings based on the function spec returned from gateway, when requesting with a veresion range
- Execute the version range at runtime rather than a concrete version


packages/foundry-sdk-generator/src/ontologyMetadata/ontologyMetadataResolver.ts
```diff
@@ -353,19 +353,27 @@ export class OntologyMetadataResolver {
...
 
       for (const queryType of entities.queryTypesApiNamesToLoad ?? []) {
-        if (queryTypes.has(queryType)) {
+        const lastColonIndex = queryType.lastIndexOf(":");
+        const queryTypeApiName = lastColonIndex === -1
+          ? queryType
+          : queryType.substring(0, lastColonIndex);
+
+        if (queryTypeApiNames.has(queryTypeApiName)) {
           return Result.err([
```
aqding, line 377: This is "technically" a break, but if the same function api name was requested multiple times, the behavior was nondeterministic/undefined. If you requested the same function at different versions, you would generate code in parallel for them, but they would all write to the same `apiName.ts` file. Given that this behavior is clearly not intended/was incredibly unstable even if it was being used, it seems ok to just throw up front now

bryantpark04: Is it possible for this to happen if you define your OSDK in Functions/Dev Console UI? I think we should avoid starting to throw from in platform if possible but through CLI only seems acceptable to me

aqding: As far as I can tell, no, you can't do this in the UI

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
```diff
@@ -39,7 +39,7 @@ export async function generateClientSdkVersionTwoPointZero(
   externalInterfaces: Map<string, string> = new Map<string, string>(),
   externalSpts: Map<string, string> = new Map<string, string>(),
   forInternalUse: boolean = false,
-  fixedVersionQueryTypes: string[] = [],
```
ssanjay1, line 42: We can't really remove this since its a break on our public API surface, probably need to keep both for now

packages/client/src/ontology/loadQueryMetadata.ts
```diff
@@ -23,10 +23,26 @@ export async function loadQueryMetadata(
   client: MinimalClient,
   queryTypeApiNameAndVersion: string,
 ): Promise<QueryMetadata> {
-  const [apiName, version] = queryTypeApiNameAndVersion.split(":");
-  const r = await QueryTypes.get(client, await client.ontologyRid, apiName, {
+  const separatorIndex = queryTypeApiNameAndVersion.lastIndexOf(":");
```
ssanjay1: I might be missing something, but when would we have more than one colon here?

packages/generator/src/v2.0/generateClientSdkVersionTwoPointZero.ts
aqding: ah ok rip, I've kept the old one for back compat

packages/client/src/ontology/loadQueryMetadata.ts
aqding: We don't, I've reverted this change

packages/e2e.sandbox.todoapp/src/CreateTodoForm.tsx
```diff
@@ -39,7 +39,7 @@ export default function CreateTodoForm() {
       // reset the form if we succeeded
       formRef.current.reset();
     },
-    [inputRef.current, applyAction],
+    [applyAction],
```
ssanjay1: Why this change?

aqding: Skill issue from linter, reverted

ssanjay1 (approved):

PR #3979, opened 2026-09-03
Add Chromatic visual regression testing

ziyunp:
## Summary

Add Chromatic visual regression testing for the React components Storybook so UI changes can be reviewed against captured baselines.

## Changes

- add a scoped GitHub Actions workflow for component and Storybook changes
- extract shared build-storybook github action for consistency across deploy-storybook and chromatic workflows
- publish the root-based Storybook build with the repository Chromatic project secret
- report visual changes without blocking CI while baselines are established
- updated a few storybook interaction tests with longer timeout as they failed to find the first row of data in chromatic run.

**Successful run on this PR** https://github.com/palantir/osdk-ts/actions/runs/34097907261/job/101665520740
Published chromatic build: https://www.chromatic.com/build?appId=6a1de290682467a65a801062&number=6


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Build dependencies for Storybook
+        run: pnpm exec turbo transpileBrowser --filter='@osdk/react-components-storybook^...'
+
+      - name: Build Storybook
+        env:
+          STORYBOOK_BASE_PATH: /
+        run: pnpm --filter @osdk/react-components-storybook build
+
+      - name: Publish to Chromatic
+        uses: chromaui/action@c93e0bc3a63aa176e14a75b61a31847cbfdd341c # v1
```
ziyunp: blocking: this workflow has never successfully started. All three `Chromatic` runs on this branch are `startup_failure` at 0s ([latest](https://github.com/palantir/osdk-ts/actions/runs/33776014491)) with "This run likely failed because of a workflow file issue", so nothing here has been exercised end to end.

The YAML parses and validates against the GitHub workflow schema, so my best guess is the org's allowed-actions policy: every other `uses:` in `.github/workflows/` is `actions/*` or `pnpm/action-setup`, and `chromaui/action` would be the first third-party action in the repo. Worth checking with whoever owns the Actions policy before this can land.

Separately, the pin comment is off — `c93e0bc` is `v11.27.0`, not `v1` (current is `v18.7.x`). Given the CLI is pinned to `~18.7.1` below, mind bumping the action to a matching v18 SHA and fixing the comment?

packages/react-components-storybook/package.json
```diff
@@ -41,6 +42,7 @@
     "@types/react-dom": "^18.3.7",
     "@vitejs/plugin-react": "^4.3.4",
     "@vitest/browser": "^3.2.5",
+    "chromatic": "~18.7.1",
```
ziyunp: blocking: `Check for changesets` is failing — `Some packages have been changed but no changesets were found`. Since this package is `private`, `pnpm changeset add --empty` should be enough.

packages/react-components-storybook/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "build": "storybook build",
+    "chromatic": "STORYBOOK_BASE_PATH=/ chromatic --build-script-name build",
```
ziyunp: issue (non-blocking): this script and the workflow describe two different ways to run Chromatic — the script lets Chromatic drive the build via `--build-script-name build`, while the workflow pre-builds and hands over `storybookBuildDir: storybook-static`. Two paths to the same outcome tends to drift.

A couple of things also make the local script hard to use as-is:
- it doesn't build the workspace deps first, so `pnpm chromatic` picks up whatever `@osdk/react-components` build output happens to be on disk (the `dev` script runs `turbo transpileBrowser --filter='@osdk/react-components-storybook^...'` for exactly this reason)
- `chromatic@18.7.1` declares `engines: node >=22`, but `.nvmrc` is `18.20.4`

Would it make sense to either drop the script and keep Chromatic CI-only, or have the workflow just call `pnpm --filter @osdk/react-components-storybook chromatic` so there's one definition?

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 24
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build dependencies for Storybook
+        run: pnpm exec turbo transpileBrowser --filter='@osdk/react-components-storybook^...'
```
ziyunp: thought: this is the third workflow that builds this Storybook, and now the third different way of building its deps — `deploy-storybook.yml` and `storybook-preview.yml` both use `pnpm --filter "@osdk/*" transpileBrowser`, this one uses the narrower turbo filter. The narrow one is nicer, but with three copies a future change to the build has to land in three files. Worth consolidating into a composite action or a root script as a FLUP?

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Build Storybook
+        env:
+          STORYBOOK_BASE_PATH: /
```
ziyunp: question: what does `STORYBOOK_BASE_PATH: /` buy us here? `deploy-storybook.yml` doesn't set it at all and `.storybook/main.ts` treats it as optional, so I'd have expected the default to be fine for a Chromatic upload. If it is required, a one-line comment saying why would help.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: pnpm --filter @osdk/react-components-storybook build
+
+      - name: Publish to Chromatic
+        uses: chromaui/action@c93e0bc3a63aa176e14a75b61a31847cbfdd341c # v1
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          workingDir: packages/react-components-storybook
+          storybookBuildDir: storybook-static
+          exitZeroOnChanges: true
```
ziyunp: suggestion (non-blocking): the PR description explains that `exitZeroOnChanges` is temporary while baselines settle, but the workflow doesn't — mind adding a short comment with the reason and what has to be true before flipping it?

Also worth considering `autoAcceptChanges: main` for push builds: without it, main-branch builds land as unreviewed changes and later PRs keep diffing against the last accepted baseline.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
+name: Chromatic
+
+env:
+  DO_NOT_TRACK: 1
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/react-components/**'
```
ziyunp: thought: `packages/react` isn't in the path filter, but it's a peer dep of `@osdk/react-components` and the Storybook renders through it, so a change to hook loading/error behavior could shift rendered output without triggering Chromatic. Not sure it's worth widening the filter versus just accepting the gap, but flagging in case you'd rather include it.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
...
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'packages/react-components/**'
+      - 'packages/react-components-storybook/**'
+      - '.github/workflows/chromatic.yml'
+
+permissions: {}
+
+concurrency:
+  group: chromatic-${{ github.workflow }}-${{ github.ref }}
```
ziyunp: nit: `github.workflow` is already `Chromatic`, so the group renders as `chromatic-Chromatic-refs/...`. `${{ github.workflow }}-${{ github.ref }}` is enough.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,69 @@
...
+  cancel-in-progress: ${{ github.event_name != 'push' }}
+
+jobs:
+  chromatic:
+    name: Run Chromatic
+    # Fork PRs cannot read repository secrets.
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
+    timeout-minutes: 20
```
ziyunp: praise: nice — `timeout-minutes` plus the fork guard and the `fetch-depth: 0` rationale comment are all things the sibling storybook workflows are missing.

packages/react-components-storybook/src/stories/ObjectTable/objectTableStoryHelpers.tsx
```diff
@@ -445,6 +445,7 @@ export const columnDefinitions: ColumnDefinition<
 // ===========================================================================
 
 export const TARGET_DATA = "Ahmed Williams";
+const TABLE_DATA_TIMEOUT = 5_000;
```
nfdiop, line 450: Questions:
- Could you share how we came up with 5_000? 
- If we need a timeout here, do we also need it in other components?
nit: could you add the time unit?

packages/react-components-storybook/package.json
```diff
@@ -41,6 +43,7 @@
     "@types/react-dom": "^18.3.7",
     "@vitejs/plugin-react": "^4.3.4",
     "@vitest/browser": "^3.2.5",
+    "chromatic": "18.7.2",
```
nfdiop: curious why pin the version?

packages/react-components-storybook/README.md
```diff
@@ -61,6 +61,38 @@ Notes:
...
+
+# upload it (get the token from the project's Chromatic settings)
+CHROMATIC_PROJECT_TOKEN=… pnpm --filter @osdk/react-components-storybook chromatic
+```
+
+Notes:
+
+- The CLI needs Node >= 22; this repo's `.nvmrc` is older, so switch with
+  `nvm use 24` (the version CI uses) first.
+- CI passes `--exit-zero-on-changes` while the baselines settle, so visual
```
nfdiop, line 92: Curious how we report visual diffs? Is it by navigating to it?

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,67 @@
...
+
+      # --exit-zero-on-changes is temporary: visual diffs are reported but do
+      # not fail the build while the initial baselines settle. Drop it once
+      # main has an accepted baseline for every story and no story diffs run
+      # to run.
+      #
+      # --auto-accept-changes main accepts main's builds as the baseline
+      - name: Publish to Chromatic
+        env:
+          CHROMATIC_PROJECT_TOKEN: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
```
nfdiop, line 66: question: how do we keep the token safe since it is long-lived? Could someone access it by overriding `chromatic` in a PR then extracting the token?

packages/react-components-storybook/src/stories/ObjectTable/objectTableStoryHelpers.tsx
ziyunp: I just tried with 5_000 to see if the failed tests pass. Only ObjectTable needs this because it asserts that the first row data loaded

packages/react-components-storybook/README.md
ziyunp: yes, for now, we don't want to fail the build when visual changes yet. This flag is also set in blueprint's CI run

.github/workflows/chromatic.yml
ziyunp: It's added as github secret of this repo

packages/react-components-storybook/package.json
ziyunp: It was left from previous setup, updated to "^18.7.2"

PR #3981, opened 2026-09-03
Make Foundry CLI installation nonfatal

nihalbhatnagar:
## Summary
- catch automatic Foundry CLI installation failures during postinstall and warn instead of failing package installation
- require the Foundry CLI when creating an integration client directly
- probe the CLI executable directly and add regression coverage

## Testing
- `pnpm --filter @osdk/integration-testing test`
- `pnpm --filter @osdk/integration-testing typecheck`
- `pnpm --filter @osdk/integration-testing lint`
- `pnpm exec turbo run check-api --filter @osdk/integration-testing`
- `git diff --check`

### DIFF
--- a/packages/integration-testing/src/createIntegrationClient.ts
+++ b/packages/integration-testing/src/createIntegrationClient.ts
@@ -25,9 +25,11 @@ import type {
 import type { Experiment } from "@osdk/api/unstable";
 import { createClient, type Client } from "@osdk/client";
 import { createMockClient } from "@osdk/unit-testing";
+import invariant from "tiny-invariant";
 import { fetch as undiciFetch, Agent } from "undici";
 
 import type { IntegrationClient, IntegrationClientConfig } from "./types.js";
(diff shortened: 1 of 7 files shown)


bryantpark04 (approved):

packages/integration-testing/src/createIntegrationClient.ts
```diff
@@ -40,6 +42,17 @@ type ClientArg =
 
 export async function createIntegrationClient(
   config: IntegrationClientConfig,
+): Promise<IntegrationClient> {
+  const foundryCli = await checkFoundryCliVersion();
+  invariant(
+    foundryCli.type !== "not-found",
+    "Foundry CLI must be installed before creating an integration client.",
+  );
+  return createIntegrationClientForRunningServer(config);
```
bryantpark04, line 51: Feels a bit weird but I suppose there's no better way to do this

PR #3986, opened 2026-09-07
[ScenariosV2] Add optional expireAfter when creating scenarios

lczm:
# Description

Adds a `expireAfter` optional property when creating scenarios. Feature has been recently established in the backend, wiring it through here for end-to-end support.

# Tests

- Added a test for creating a scenario with `expireAfter`.
- Added a test confirming `expireAfter` works alongside a branch.

### DIFF
--- a/packages/client/src/scenarios/createScenario.ts
+++ b/packages/client/src/scenarios/createScenario.ts
@@ -42,7 +42,7 @@ import {
  * ```ts
  * import { createScenario } from "@osdk/client/unstable-do-not-use";
  *
- * const expireAfter = new Date(Date.now() + 60 * 60 * 1000).toISOString();
+ * const expireAfter = new Date(Date.now() + 60 * 60 * 1000);
  * const scenario = await createScenario(client, { expireAfter });
  * const scenarioRid = scenario.getScenarioReference();
  * ```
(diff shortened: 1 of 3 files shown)


packages/client/src/scenarios/createScenario.ts
```diff
@@ -38,12 +42,16 @@ import {
  * ```ts
  * import { createScenario } from "@osdk/client/unstable-do-not-use";
  *
- * const scenario = await createScenario(client);
+ * const expireAfter = new Date(Date.now() + 60 * 60 * 1000).toISOString();
```
bryantpark04: This is pretty clunky

packages/client/src/scenarios/createScenario.ts
```diff
@@ -38,12 +42,16 @@ import {
...
+ * const scenario = await createScenario(client, { expireAfter });
  * const scenarioRid = scenario.getScenarioReference();
  * ```
  */
 export async function createScenario(
   client: Client,
+  options?: {
+    expireAfter?: string;
```
bryantpark04: Let's have a proper type rather than a `string`

bryantpark04 (changes requested):

packages/client/src/scenarios/createScenario.ts
lczm: updated to take in a Date!

bryantpark04 (approved):

PR #3987, opened 2026-09-07
Pass itemToStringLabel to ObjectSelectField

nfdiop:


packages/react-components/src/action-form/FormFieldApi.ts
```diff
@@ -450,8 +450,14 @@ export type ObjectSelectFieldProps<
   | "onQueryChange"
   | "disableClientSideFiltering"
   | "renderItemList"
-> &
-  ObjectSelectDataSource<Q>;
+> & {
+  /**
+   * Converts an object to its visual and accessible label.
+   *
+   * @default The object's title, falling back to its primary key.
```
ziyunp, line 457: Why not just remove itemToStringLabel from the omit list?

PR #3984, opened 2026-09-04
Version Packages 2.65.0

nihalbhatnagar:
This PR was opened by automation. When you're ready to do a release, you can merge this and publish to npm yourself.
     If you're not ready to do a release yet, that's fine, whenever you re-run the release script in main, this PR will be updated.

# Releases
### Minor Changes

-   e3a381d: Bump the `@osdk/foundry.*` and `@osdk/internal.foundry.*` catalog entries to `2.77.0`. Generated experimental ontology metadata now includes full action type metadata and logic rules.

### Minor Changes

-   e3a381d: Bump the `@osdk/foundry.*` and `@osdk/internal.foundry.*` catalog entries to `2.77.0`. Generated experimental ontology metadata now includes full action type metadata and logic rules.
-   b1bb84a: Fix multiple useOsdkObjects hooks that share the same query incorrectly sharing a page size. Each hook now respects its configured pageSize, or the default of 100, regardless of subscription order.
-   aacb86b: Honor expected value of $isNull in client-side where matching

### Patch Changes

-   Updated dependencies [e3a381d]
    -   @osdk/api@2.65.0
    -   @osdk/generator-converters@2.65.0
    -   @osdk/client.unstable@2.65.0

### Minor Changes

-   e3a381d: Bump the `@osdk/foundry.*` and `@osdk/internal.foundry.*` catalog entries to `2.77.0`. Generated experimental ontology metadata now includes full action type metadata and logic rules.

### Patch Changes

-   Updated dependencies [e3a381d]
    -   @osdk/api@2.65.0
    -   @osdk/generator-converters@2.65.0

### Minor Changes

-   e3a381d: Bump the `@osdk/foundry.*` and `@osdk/internal.foundry.*` catalog entries to `2.77.0`. Generated experimental ontology metadata now includes full action type metadata and logic rules.

### Patch Changes

-   Updated dependencies [e3a381d]
-   Updated dependencies [b1bb84a]
-   Updated dependencies [aacb86b]
    -   @osdk/api@2.65.0
    -   @osdk/client@2.65.0
    -   @osdk/generator@2.65.0
    -   @osdk/generator-converters@2.65.0
    -   @osdk/client.unstable@2.65.0
    -   @osdk/generator-utils@2.65.0

### Minor Changes

-   e3a381d: Bump the `@osdk/foundry.*` and `@osdk/internal.foundry.*` catalog entries to `2.77.0`. Generated experimental ontology metadata now includes full action type metadata and logic rules.

### Patch Changes

[...]

PR #3992, opened 2026-09-08
[ScenariosV2] Add support for `applyBatch` in scenarios

kurzdev:
## Change
Support for sending a `scenarioRid` on calling `applyBatch` to use batch action application within scenarios was added to the platform SDK in https://github.com/palantir/foundry-platform-typescript/pull/383. This new SDK was brought into OSDK in https://github.com/palantir/osdk-ts/pull/3984.
We can therefore just remove the guard against this behaviour and wire through the `scenarioRid`.


pnpm-lock.yaml
kurzdev, line 1: This fixes the semantic merge conflict introduced through https://github.com/palantir/osdk-ts/pull/3432 that currently is breaking main.

Looks like the bundle size task is still using a cached version of the lockfile, which is why it's failing

bryantpark04 (approved):

PR #3995, opened 2026-09-08
Sorted objects no longer flip-flop in rendering order

williamhe723:
Previously, when using useOsdkObjects in a query like this:
```
const { data: tasks } = useOsdkObjects(Task, {
orderBy: { dueDate: "desc" },
});
```
calling `observableClient.invalidateAll()` would cause the order to change. This is because there was client side sorting happening after the server had already sorted, and client side ordering breaks ties by PK, while server side is non-deterministic.

With these changes, `observableClient.invalidateAll()` now keeps the sorted order server side. However there is a race condition that could still happen that causes a brief flash of order change, but the final state is consistent. This race condition happens because object fetches and list fetches happen concurrently, so depending on which finishes first, there could be a client side sort order for the list that is briefly displayed.

### DIFF
--- a/packages/client/src/observable/internal/list/ListQuery.test.ts
+++ b/packages/client/src/observable/internal/list/ListQuery.test.ts
@@ -760,6 +760,60 @@ describe("ListQuery sort stability across pages", () => {
       resolved.map((o) => (o as { derivedLong: string }).derivedLong),
     ).toEqual(["1", "42", "100", "9007199254740993"]);
   });
+
+  it("invalidateAll preserves server order for tied sort keys", async () => {
+    // Regression test: a revalidation that returns byte-identical data used to
+    // fall through the list's self-recursion guard (the deep-equal write never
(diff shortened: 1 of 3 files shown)


packages/client/src/observable/internal/list/ListQuery.test.ts
```diff
@@ -760,6 +760,60 @@ describe("ListQuery sort stability across pages", () => {
       resolved.map((o) => (o as { derivedLong: string }).derivedLong),
     ).toEqual(["1", "42", "100", "9007199254740993"]);
   });
+
+  it("invalidateAll preserves server order for tied sort keys", async () => {
+    // Regression test: a revalidation that returns byte-identical data used to
```
KuberSethi: nit: can we make this comment a bit more concise and readable?

KuberSethi (commented):
LGTM just one nit!

bryantpark04 (approved):

PR #3997, opened 2026-09-08
Have `branchPlugin` and `createClient` utilize the window for branch env var

aqding:
We had some issues with Vite chunk caching that would make it so stale branches would be read. Specifically, the issue was that:
- With the current Vite setup, Vite would resolve the global branch on dev server startup, write that to a chunk on the local machine, and then serve that chunk when the browser requests it.
- However, the browser is optimized to not refetch chunks if the hash of the chunk has not changed. Changing the Git branch (and subsequently, what the exported meta.env global branch variable would resolve to) does not cause the chunk to be recomputed, nor does Vite recompute the cached chunk when a new dev server instance starts
- The result is that you would have a stale global branch value that was being served via Vite, that could diverge with what the environment variable actually is.

We would instead like to ensure that the branch that is passed to the client is not subject to such caching concerns. We also wanted this to not live in the user code. The new approach is:
- Utilize Vite's `transformIndexHtml` to inject a meta tag with a `osdk-foundry-branch-rid` attribute into the HTML. The branch is resolved by first reading the Node env's `FOUNDRY_BRANCH_RID`, which is set by in-platform Jemma CI/CD to support PR previews. If that Node envvar is not set, then we read the Git branch.
- Vite by default tells the browser to not cache HTML, meaning that the browser will always fetch the HTML from the Vite server, and Vite always injects the branch on request to fetch the HTML. Note that this is just a nicety but we only request HTML on a page refresh, meaning that a user that does not refresh the page after switch branches may still technically have a page with a "stale branch". We've separately enforced restarting the dev server whenever the branch changes, which then trivially solves this problem.
- The client then reads the `osdk-foundry-branch-rid` from the meta tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,6 @@ on:
       - main
       - release/*
       - next
-      - ading/window
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
@@ -77,7 +76,7 @@ jobs:
         run: pnpm exec turbo run build
 
       - name: Update versions for snapshots
-        run: pnpm exec changeset version --snapshot ${GITHUB_REF_NAME//\//-}
+        run: pnpm exec changeset version --snapshot ${GITHUB_REF_NAME//\//__}
 
       - name: Make sure code is up to date
         run: pnpm exec turbo run postVersioning
@@ -86,4 +85,4 @@ jobs:
         run: pnpm exec turbo run build
 
       - name: Publish results
-        run: pnpm exec changeset publish --no-git-tag --tag next-${GITHUB_REF_NAME//\//-}
+        run: pnpm exec changeset publish --no-git-tag --tag next-${GITHUB_REF_NAME//\//__}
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -9,6 +9,7 @@ on:
       - main
       - release/*
       - next
+      - ading/window
```
amishshah: remove changes to this file

packages/vite-plugin-branch/src/branchPlugin.ts
```diff
@@ -52,40 +53,50 @@ export interface BranchPluginOptions {
...
-          : path.resolve(root, config.envDir ?? ".");
+    configResolved(config) {
+      root = config.root;
+      mode = config.mode;
+      envDir = config.envDir;
+      logger = config.logger;
+    },
 
-      const configuredBranch = loadEnv(mode, envDir, "VITE_")[
+    async transformIndexHtml() {
```
amishshah, line 72: I wonder if we should inject this meta tag for all html files?

amishshah (commented):
Looks generally good just two comments

amishshah (approved):

PR #3972, opened 2026-09-02
[4/5] ITSM OAC DSL: Include ITSMs in ontology block data

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/3969

ITSMs authored in the OAC DSL are now included in the emitted ontology-ir and ontology block data.

PR #4002, opened 2026-09-09
[5/5] ITSM OAC DSL: Adopt latest bindings which include RID

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/3972

Bumps to the latest ITSM bindings which include the RID.

### DIFF
--- a/packages/maker-experimental/src/util/generateRid.ts
+++ b/packages/maker-experimental/src/util/generateRid.ts
@@ -389,7 +389,7 @@ export class ReadableIdGenerator {
     interfaceApiName: string,
     transitionId: string,
   ): ReadableId {
-    return `interface-type-schema-transition-${interfaceApiName}-${transitionId}` as ReadableId;
+    return `interface-schema-transition-${interfaceApiName}-${transitionId}` as ReadableId;
   }
 
   static getForInterfaceParameterConstraint(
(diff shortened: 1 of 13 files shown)


packages/maker-experimental/src/util/generateRid.ts
```diff
@@ -382,7 +389,7 @@ export class ReadableIdGenerator {
     interfaceApiName: string,
     transitionId: string,
   ): ReadableId {
-    return `interface-schema-transition-${interfaceApiName}-${transitionId}` as ReadableId;
```
SpecificGenerality, line 385: Are we sure we want this? The method below just uses `interface-*` terminology instead of `interface-type-*`

MatthewMeacham: I _think_ this has to match OMS's RID format, right? Or can we just make up the format...?

SpecificGenerality: Oh okay great, yeah no reason to deviate then imo.

SpecificGenerality (dismissed):

davidpaquin (dismissed):

packages/maker-experimental/src/util/generateRid.ts
MatthewMeacham: Well, we could change OMS's generation.. easiest to do now if possible

MatthewMeacham: Going to change this internally and here

davidpaquin (approved):

PR #4004, opened 2026-09-09
[1/x] ITSM lockfile: adopt latest bindings and refactor in prep

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/3972

Mostly drive-by changes/simplifications as I was prototyping that aren't directly related to lockfile functionality, but should be made.

PR #4007, opened 2026-09-09
[2/x] ITSM lockfile: Add the interface schema lockfile data model

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4004

Defines the `ontology-schema-lock.json` model for tracking the last-used schema of every IT enrolled in schema migrations. It's not _perfect_ since we cannot load the truly last-published block-data/ontology-ir from Artifactory or some store, but it'll be good enough to help capture/prevent authors from publishing something that we know would break at installation-time.

Generation of this lockfile is a pure function of source and never consults a previously-persisted lockfile; this way, generation is simple to test and reason about. _Validations_ will consult both the previously-persisted and incoming lockfile since those checks depend on diffing history against now/


packages/maker/src/lockfile/applyInterfaceSchemaTransition.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+function editsByProperty(
+  instruction: InterfaceSchemaMigrationInstruction,
+  mode: ApplyMode,
+): ReadonlyMap<string, PropertyEdit> {
+  switch (instruction.type) {
+    case "addRequiredProperty":
+      return new Map([[instruction.property, { required: mode === "strict" }]]);
+    default:
+      // TODO: add a never exhaustiveness check once there's more than one instruction type
```
MatthewMeacham, line 67: Same reasoning as the previous chain -- TS doesn't allow never on discriminated union types w/ only one variant

packages/maker/src/lockfile/OntologySchemaLockfile.ts
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * The interface's locally-declared properties (i.e. non-inherited), keyed by the api name they are published under.
+   */
+  properties: Record<string, LockedProperty>;
+}
+
+export interface LockedProperty {
+  type: LockedPropertyType;
+  required: boolean;
+}
```
MatthewMeacham, line 67: Have to handle array-ness in a FLUP (trying to minimize this diff)

packages/maker/src/lockfile/test/serializeLockfile.test.ts
```diff
@@ -0,0 +1,149 @@
...
+};
+
+describe("serializeLockfile", () => {
+  it("writes the persisted format this version of maker promises", () => {
+    expect(serializeLockfile(golden)).toBe(
+      `{
+  "//": [
+    "This is a generated file, do not modify manually.",
+    "Run \`maker --write-locks\` to regenerate this file."
+  ],
```
MatthewMeacham, line 76: We could use a different file extension, but I liked keeping it JSON since there's so many utilities already applicable to JSON -- and since it's a machine-generated file

SpecificGenerality (dismissed):

SpecificGenerality (dismissed):

SpecificGenerality (approved):

PR #4011, opened 2026-09-09
[5/x] ITSM lockfile: Diff a incoming lockfile against the persisted lockfile to detect breaks/finalizations

MatthewMeacham:
## Targeting: https://github.com/palantir/osdk-ts/pull/4024

Add support for diffing an incoming lockfile against the previously-persisted one to detect breaks and finalizations. Breaks are surfaced as things that will be rejected at installation-time, and carry remediation/alternative instructions. Finalizations/deletions will eventually require confirmation from the user when running `maker --write-locks`.

Note that the break detection is nowhere near exhaustive: to keep the diff small, I've done the basic ones for the existing `addRequiredProperty` ITSM. FLUPs can introduce other break detections.


packages/maker/src/lockfile/test/generateOntologySchemaLockfile.test.ts
```diff
@@ -80,3 +83,37 @@ describe("generateOntologySchemaLockfile", () => {
...
+    expect([...censusOfSource(ontology).interfaces.keys()].sort()).toEqual([
+      "com.palantir.Enrolled",
+      "com.palantir.Unenrolled",
+    ]);
+    expect(
+      Object.keys(generateOntologySchemaLockfile(ontology).interfaces),
+    ).toEqual(["com.palantir.Enrolled"]);
+  });
+
+  it("records the schema of an interface that opted back out", () => {
```
SpecificGenerality, line 105: Hm, did the interface ever opt in?

MatthewMeacham: hah thanks

SpecificGenerality (approved):

PR #4012, opened 2026-09-09
[6/x] ITSM lockfile: Render lockfile findings and diffs

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4024

Lockfile changes and diff findings are now rendered in a pretty-ish format for user consumption.


packages/maker/src/lockfile/LockedPropertyType.ts
```diff
@@ -49,6 +49,11 @@ export function normalizePropertyType(
   return strip(type) as LockedPropertyType;
 }
 
+/** How a property's type is named in error messages and in the rendered lockfile diff. */
+export function describeType(type: LockedPropertyType): string {
+  return JSON.stringify(type);
+}
```
SpecificGenerality, line 55: AI complaining about the return type here: `string | undefined`.

MatthewMeacham: Can it return `undefined`? The `lib.es5.d.ts` bindings don't have undefined:
image

It might return undefined for an undefined input, but our usages will never use undefined as an input

SpecificGenerality (dismissed):

SpecificGenerality (approved):

PR #4013, opened 2026-09-09
[7/x] ITSM lockfile: Reconcile the interface schema lockfile on disk

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4012

Add function for the command-line to invoke (in the next PR) for determining whether the lockfile is stale, printing the differences, and writing the new lockfile to disk.


packages/maker/src/lockfile/test/reconcileOntologySchemaLockfile.test.ts
```diff
@@ -0,0 +1,370 @@
...
+    it("accepts deleting an interface that opted out in an earlier release", async () => {
+      await published(lastNameFinalized);
+      await maker(
+        person({ firstName: REQUIRED_STRING, lastName: REQUIRED_STRING }),
+        { writeLocks: true },
+      );
+      await maker(() => {}, { writeLocks: true });
+      expect(await readLockfile()).toStrictEqual({
+        version: 1,
+        interfaces: {},
```
SpecificGenerality, line 124: Should we record the interface as deleted? Or perhaps not if we don't track historical transitions otherwise.

SpecificGenerality (dismissed):

packages/maker/src/lockfile/test/reconcileOntologySchemaLockfile.test.ts
MatthewMeacham: I don't think there's a reason to -- deleting an IT just orphans that IT on existing installs, so the change is trivially valid

SpecificGenerality (dismissed):

SpecificGenerality (approved):

PR #4014, opened 2026-09-10
[8/x] ITSM lockfile: Integrate the lockfile with the maker CLI

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4013

Ads `--lockfile`, `--write-locks`, and `--yes` to the maker CLI for writing the ontology schema lockfile. 

The `lockfile` path defaults to sitting besides `--input` since it is checked-in source rather than a build artifact.


packages/maker/src/cli/main.ts
```diff
@@ -44,6 +48,10 @@ export default async function main(
     codeSnippetPackageName: string;
     codeSnippetDir: string;
     randomnessKey?: string;
+    lockfile?: string;
+    writeLocks?: boolean;
+    yes?: boolean;
```
SpecificGenerality, line 53: Nit: Should we change this to like `acceptBreaks` even if the alias is still `y`?

SpecificGenerality (dismissed):

packages/maker/src/cli/main.ts
MatthewMeacham: Well, it's not exactly accepting breaks -- it's just ACK'ing that the inferred FINALIZE/DELETEs are indeed what the author intended. Breaks always fail (there's currently no escape hatch for that short of hand-modifying or outright deleting the lockfile itself) during reconcile

In any case, I think we should keep `yes/y` since `@osdk/cli` already uses it for automatic confirmations, so idk if I wanna introduce another concept: https://github.com/palantir/osdk-ts/blob/5c8bdb05994f93f5517efc5c4b6ddaef19cc46ab/packages/cli/src/commands/site/version/unset/index.ts#L26-L32

SpecificGenerality (dismissed):

SpecificGenerality (approved):

PR #4015, opened 2026-09-10
[9/x] ITSM lockfile: ETE test

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4014

ETE runs an ITSM lifecycle: addition, finalization, supersession, and mutation through the whole generate/validate/write pipeline.


packages/maker/src/lockfile/test/ontologySchemaLockfile.test.ts
```diff
@@ -0,0 +1,314 @@
...
+          ),
+          { writeLocks: true },
+        ),
+      ).rejects.toThrowError(
+        /neither a finalization nor a deletion[\s\S]*"lastName" changed type from "string" to "integer"/u,
+      );
+    });
+  });
+
+  describe("a second migration and a superseding checkpoint", () => {
```
SpecificGenerality, line 94: Can we also test finalizing/checkpointing the second migration before the first?

packages/maker/src/lockfile/test/ontologySchemaLockfile.test.ts
```diff
@@ -0,0 +1,314 @@
...
+        email: { type: "string", required: false },
+        firstName: { type: "string", required: true },
+        lastName: { type: "string", required: true },
+      });
+      expect(
+        lockfile.interfaces.Person.transitions.map(({ id }) => id),
+      ).toStrictEqual(["requireEmail"]);
+    });
+
+    it("finalizes the second migration without disturbing the first", async () => {
```
SpecificGenerality, line 109: Is last `lastName` already finalized at this point though? Do we want to test when it's not finalized? Same as above comment I believe.

packages/maker/src/lockfile/test/ontologySchemaLockfile.test.ts
MatthewMeacham: wow great call -- the RFC was written before independent checkpointing, so the story there didn't convey that case... Expanding this suite

SpecificGenerality (dismissed):

SpecificGenerality (approved):

PR #4019, opened 2026-09-10
Add OTDP Flag to React hooks

williamhe723:
## Summary

Adds `$UNSTABLE_loadOntologyDefinedDerivedProperties` support to the React OSDK hooks and corresponding observable-client APIs:

- `useOsdkObject`
- `useOsdkObjects`
- `useObjectSet`
- `useLinks`

The option is forwarded through object, list, object-set, link, pagination, batching, and interface-to-object reload paths.

## Cache behavior

The cache treates each setting as a distinct variant:

- `undefined`: use the server default
- `false`: explicitly disable ontology-defined derived properties
- `true`: explicitly enable ontology-defined derived properties

Values fetched under one setting are not propagated into cache entries for another setting. This prevents derived properties from leaking into variants where they were not requested.

## Optimistic updates

Optimistic value updates are intentionally not propagated between different ontology-defined-derived-property settings.

The existing optimistic-update path writes through the server-default cache variant. Consequently, hooks using an explicit `true` or `false` setting will display the updated server value after the action completes and the query is refreshed, rather than displaying the optimistic value. 

This could lead to some times when the UI is out of sync, but it will always later reconsolidate. An alternative to this approach is allow optimistic writes to update all three variants, but this will cause the OTDP values to be temporarily inconsistent during an optimistic update.

## Testing

Added coverage for:

- Forwarding explicit `false` through each React hook
- Separating bulk requests by OTDP setting
- Partitioning object, list, object-set, and link cache entries
- Preventing fetched and optimistic values from crossing OTDP cache variants

Also tested e2e locally

### DIFF
--- a/packages/client/src/observable/internal/list/ListQuery.ts
+++ b/packages/client/src/observable/internal/list/ListQuery.ts
@@ -59,6 +59,7 @@ import { EMPTY_RDP_SET } from "../utils/rdpFieldOperations.js";
 import {
   INCLUDE_ALL_BASE_PROPERTIES_IDX,
   INTERSECT_IDX,
+  LOAD_ONTOLOGY_DEFINED_DERIVED_PROPERTIES_IDX,
   type ListCacheKey,
   ORDER_BY_IDX,
   PIVOT_IDX,
@@ -205,6 +206,14 @@ export abstract class ListQuery extends BaseListQuery<
     return this.cacheKey.otherKeys[INCLUDE_ALL_BASE_PROPERTIES_IDX] === true;
   }
 
+  public override get loadOntologyDefinedDerivedProperties():
+    | boolean
+    | undefined {
+    return this.cacheKey.otherKeys[
+      LOAD_ONTOLOGY_DEFINED_DERIVED_PROPERTIES_IDX
+    ];
+  }
+
   get objectTypes(): ReadonlySet<string> {
     return this.#objectTypesCache ?? new Set([this.apiName]);
   }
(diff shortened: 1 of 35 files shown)


williamhe723:
The main thing I'm concerned about is the optimistic writes. We could go through and refactor optimistic updates to not have temporary inconsistencies (either UI or in the cache), but that would be a more involved change. I'm not sure how many use cases of the React OSDK load the same object twice, one with OTDPs and one without, so this may not even matter much.

williamhe723:
> The main thing I'm concerned about is the optimistic writes. We could go through and refactor optimistic updates to not have temporary inconsistencies (either UI or in the cache), but that would be a more involved change. I'm not sure how many use cases of the React OSDK load the same object twice, one with OTDPs and one without, so this may not even matter much.

If it's not used much, may be better to just have a cache inconsistency than a UI inconsistency because the UI one will show up if they use the flag at all, but the cache one only shows up if you load an object both with and without OTDPS

packages/client/src/observable/internal/list/ListQuery.ts
```diff
@@ -708,6 +724,10 @@ export abstract class ListQuery extends BaseListQuery<
       obj.$objectType,
       pk,
       this.rdpConfig ?? undefined,
+      undefined,
+      undefined,
+      undefined,
+      this.loadOntologyDefinedDerivedProperties,
```
KuberSethi, line 730: I don't think we always want to do this as if it's nonexistent things might blow up. Can we be defensive and follow what @nihalbhatnagar did in #3945 with `getCachedObjectKey` usage where we only insert the key if its value is cached, otherwise keep the existing rows and revalidate? For here and for `ObjectSetQuery`

williamhe723: Added the defensive checks in `ListQuery`. It seems that #3945 has not been merged yet, but it would add the logic needed in `ObjectSetQuery` and I would just have to pipe the flag in this PR through

PR #4020, opened 2026-09-10
[vite-plugin-branch] auto-refresh on branch change

amishshah:
Auto-refresh the dev server for OSDK apps using the Vite plugin when the Git branch changes


.github/workflows/release.yml
```diff
@@ -9,6 +9,7 @@ on:
       - main
       - release/*
       - next
+      - amishshah/cw/vite-plugin-changes
```
olliejenningsPT, line 12: is this so you can test releasing from this branch?

packages/vite-plugin-branch/package.json
```diff
@@ -44,7 +44,7 @@
     "execa": "^9.6.0"
   },
   "peerDependencies": {
-    "vite": "*"
+    "vite": "^6.0.0 || ^7.0.0 || ^8.0.0"
```
olliejenningsPT, line 47: maybe just worth flagging in a comment that anything below v6 will get a peer dep resolution / install failure

.github/workflows/release.yml
amishshah: Yep!

packages/vite-plugin-branch/package.json
amishshah: in the README or something?

olliejenningsPT: Yeah something like that, or could just make sure int he changelog entry thats detailed

olliejenningsPT (approved):

PR #4021, opened 2026-09-10
Fix test flakes

MatthewMeacham:
Kept seeing test flakes due to concurrent execution of defineOntoloy against the same write dirs; this meant multiple tests may put stuff into the dir, and then when they go to delete the dir after the test, they'd fail due to the dir not being totally empty.

These tests don't actually assert on the written output (and the tests that _do_ already use unique directories); so let's just stop providing a write directory and avoid this problem.

### DIFF
--- a/.changeset/maker-tests-no-shared-tmp-codegen-dir.md
+++ b/.changeset/maker-tests-no-shared-tmp-codegen-dir.md
@@ -1,6 +1,0 @@
----
-"@osdk/maker": patch
-"@osdk/maker-experimental": patch
----
-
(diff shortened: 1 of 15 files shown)


.changeset/maker-tests-no-shared-tmp-codegen-dir.md
```diff
@@ -0,0 +1,6 @@
+---
```
MatthewMeacham, line 1: CI complained if I didn't have a changeset..

packages/maker-experimental/src/api/overall.test.ts
```diff
@@ -65,7 +65,7 @@ function apiNamePreset(apiName: string) {
 
 describe("Experimental Test Suite", () => {
   beforeEach(async () => {
-    await defineOntology("com.palantir.", () => {}, "/tmp/");
+    await defineOntology("com.palantir.", () => {}, undefined);
```
MatthewMeacham, line 68: None of these tests actually need the output written (there are a few tests that _do_, and those use randomly-generated dirs)

So I figured not writing anything was the cheapest way to avoid the flakes -- `defineOntology` just writes nothing when the input is `undefined`: https://github.com/palantir/osdk-ts/blob/aec9d54682aef9497dbbc593ddec50d345567702/packages/maker-experimental/src/api/defineOntologyV2.ts#L162-L164

SpecificGenerality (approved):

PR #4026, opened 2026-09-11
Extract the Functions alias runtime into @osdk/aliases

jameszhang244:
Extracts the existing Functions alias runtime into `@osdk/aliases/node` so it can be reused without depending on `@osdk/functions`. The implementation, fixtures, and existing tests move unchanged.

`@osdk/functions` continues to export the same six alias readers and types through `Aliases`. Calls remain synchronous, including thrown errors, with no deprecation warnings. Added compatibility tests cover synchronous return values, missing-alias errors, and the exported readers.

This is the prerequisite for #3841, which adds the experimental async API and browser alias loading. Merge this refactor first, then rebase #3841 onto `main`.

Validation: all 50 alias tests and 3 Functions compatibility tests pass. Browser, ESM, CommonJS, and declaration builds pass, along with lint, type checks, API extraction, package export validation, and monorepo lint. All GitHub Actions checks pass, including tests on Node 18, 20, 22, and 24, E2E tests, and type/API checks. CircleCI reports a missing configuration.

End-to-end validation: installed the exact published Functions, Client, and Aliases PR snapshots with npm in a TypeScript v2 function repository. In both live preview and published execution, two configured custom aliases returned the expected strings synchronously and repeated reads matched. All six readers synchronously rejected missing aliases, and an uncaught missing custom alias surfaced the expected runtime error. Successful resource resolution was tested for custom aliases.


packages/aliases/package.json
```diff
@@ -0,0 +1,73 @@
...
+    ".": {
+      "browser": "./build/browser/index.js",
+      "import": {
+        "types": "./build/types/index.d.ts",
+        "default": "./build/esm/index.js"
+      },
+      "require": "./build/cjs/index.cjs",
+      "default": "./build/browser/index.js"
+    },
+    "./node": {
```
ssanjay1: We should probably name this in such a way that its like internal or something, maybe even just calling it `./internal`? Like we don't want people importing from here in the osdk/aliases package in the future right, like want to move people to the async version. If we name this node, this becomes a part of the public API surface we have to keep supporting

packages/functions/package.json
```diff
@@ -70,6 +70,7 @@
     "typecheck": "tsc --noEmit --emitDeclarationOnly false"
   },
   "dependencies": {
+    "@osdk/aliases": "workspace:~",
```
ssanjay1: Since aliases is technically still in beta (0.x version) we should probably keep this as a `*` dep for now

tests/verify-cjs-node16/src/test.ts
```diff
@@ -22,3 +23,15 @@ Unstable.augment({ type: "object", apiName: "foo" } as any);
 if (sdk.$Objects.Employee.apiName !== "Employee") {
   throw new Error("Expected Employee");
 }
+
+try {
```
ssanjay1: Why did we need this?

packages/aliases/package.json
jameszhang244: done! https://github.com/palantir/osdk-ts/pull/4026/changes/5a73bceb41c0ebe13228c00186e917065b74b698

packages/functions/package.json
jameszhang244: done! https://github.com/palantir/osdk-ts/pull/4026/changes/7f33926537b2917d3fcb4b2b1d9d4dbb14079dd7

tests/verify-cjs-node16/src/test.ts
jameszhang244: I was trying to test that Node could import `@osdk/aliases/internal` and find `custom`. Calling `InternalAliases.custom(...)` without a Functions environment throws, so the try/catch handled that expected error. I changed the test to check that the function is exported, since the unit tests already cover alias resolution and error behavior

tests/verify-cjs-node16/package.json
```diff
@@ -13,6 +13,7 @@
     "typecheck": "tsc"
   },
   "dependencies": {
+    "@osdk/aliases": "workspace:~",
```
ssanjay1: Should be * deps too

ssanjay1 (dismissed):

packages/functions/src/aliases/index.ts
```diff
@@ -14,9 +14,23 @@
  * limitations under the License.
  */
 
-export * from "./custom.js";
-export * from "./dataset.js";
-export * from "./mediaset.js";
-export * from "./model.js";
-export * from "./source.js";
-export * from "./stream.js";
+// API Extractor cannot represent a namespace that re-exports `export *` from
```
ssanjay1: can we remove this comment in FLUP?

jameszhang244: done! https://github.com/palantir/osdk-ts/pull/4026/changes/02a8c74a03a6152bc880178416815307094c6f68

ssanjay1 (approved):

PR #4033, opened 2026-09-14
Use service discovery for Foundry CLI installation

nihalbhatnagar:
We want to use the postInstall script in Foundry CI. To do so, we need to find the correct URIs to download the foundry cli installer from. Foundry CLI exposes service discovery files we can use to route to the correct URLs as regular egress to the head of the stack is not allowed


packages/integration-testing/src/test/installFoundryCli.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  expect(fetchMock).toHaveBeenCalledWith(
+    new URL(
+      `${discovery ? codeUrl : "https://local.example.com/code/api"}/extension/install-script`,
+    ),
+    { headers: { authorization: "Bearer test-token" } },
+  );
+  expect(spawn).toHaveBeenCalledWith("bash", ["-s", "--"], {
+    env: expect.objectContaining({
+      FOUNDRY_URL: discovery ? artifactsBaseUrl : "https://local.example.com",
```
nihalbhatnagar, line 97: The installer script itself is what calls this

packages/integration-testing/src/scripts/foundry-services.ts
```diff
@@ -0,0 +1,91 @@
...
+  );
+  const url = new URL(uris[0]);
+  invariant(
+    url.protocol === "https:" || url.protocol === "http:",
+    `Invalid URL for ${serviceName}`,
+  );
+  return trimTrailingSlashes(uris[0]);
+};
+
+export const resolveFoundryServices = async (
```
nihalbhatnagar, line 54: The structure of what we're reading here is a bit "magic" to the repo as theres no generated types or structure to base it off of, but as long as we stick to the contract defined in our tests we can assume it will continue to work

packages/integration-testing/src/scripts/download.ts
```diff
@@ -106,10 +106,11 @@ export const installFoundryCli = async (
   const env = options.env ?? process.env;
   const isWindows = process.platform === "win32";
 
-  invariant(
-    !isWindows || !env.FOUNDRY_SERVICE_DISCOVERY_V2,
-    "Foundry CLI installation using FOUNDRY_SERVICE_DISCOVERY_V2 is not supported on Windows.",
-  );
+  if (isWindows && env.FOUNDRY_SERVICE_DISCOVERY_V2) {
```
nihalbhatnagar, line 109: Invariant read as "!isWindows || !env.FOUNDRY_SERVICE_DISCOVERY_V2" so I made this more clear

ssanjay1 (approved):

PR #4045, opened 2026-09-16
[12/x] ITSM lockfile: track IDP vs SPT

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4043

An SPT and an IDP are different bindings even when they resolve to the same underlying type -- since an implementing OT would have to remap their IT<>OT property mappings if the an SPT is swapped in place of an IDP, or vice versa. 

We now track SPT vs IDP in the IT lockfile so we can diff them.


packages/maker/src/lockfile/validateOntologySchemaLockfile.ts
```diff
@@ -321,6 +329,21 @@ function validateSchemaDiff(
       continue;
     }
 
+    // Ahead of the type check: when the binding itself was swapped, the types are incidental, and
+    // reporting them would point the author at the wrong thing to restore.
+    const previousDeclaration = declarationOf(previousProperty);
+    const nextDeclaration = declarationOf(nextProperty);
+    if (previousDeclaration !== nextDeclaration) {
```
SpecificGenerality: Are we concerned about skew between the IPT and SPT API names that would cause the change to be detected as a removal? i.e. are the `authoredApiName` and `apiName` need to be consistent for this to work? https://github.com/palantir/osdk-ts/blob/fc508663855e4eb2633b364598da7cb22bb2d224/packages/maker/src/api/interface/InterfacePropertyType.ts#L96

MatthewMeacham: Ope great catch -- SPTs are always namespaced

SpecificGenerality (approved):

PR #4051, opened 2026-09-16
[16/x] ITSM lockfile: track value types used by ITs

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4050

VTs encode constraints similar to how properties define their type, nullability, etc. Adding a VT is a break for existing OTs, while removing is a warning .

The lockfile now tracks VTs used by ITs and can report breaks over them.


packages/maker/src/lockfile/generateOntologySchemaLockfile.ts
```diff
@@ -129,6 +134,23 @@ function lockProperty(property: InterfacePropertyType): LockedProperty {
...
+ * The identity of the value type a property references, rebuilt field by field: the wire reference
+ * also carries `displayMetadata` and a `version`, neither of which `LockedValueType` records.
+ */
+function lockValueType(
+  valueType: OntologyIrValueTypeReferenceWithMetadata | undefined,
+): LockedValueType | undefined {
+  if (valueType === undefined) {
+    return undefined;
+  }
+  return {
```
MatthewMeacham:
```suggestion
function lockValueType(
  valueType: OntologyIrValueTypeReferenceWithMetadata | undefined,
): LockedValueType | undefined {
  if (valueType === undefined) {
    return undefined;
  }

  // Re-built field-by-field so key order is stable
  return {
```

packages/maker/src/lockfile/generateOntologySchemaLockfile.ts
```diff
@@ -129,6 +134,21 @@ function lockProperty(property: InterfacePropertyType): LockedProperty {
...
+}
+
+function lockValueType(
+  valueType: OntologyIrValueTypeReferenceWithMetadata | undefined,
+): LockedValueType | undefined {
+  if (valueType === undefined) {
+    return undefined;
+  }
+
+  // Re-built field-by-field so key order is stable
```
SpecificGenerality, line 148: Should we add a comment about why the value type version isn't considered for back-compat diffing?

SpecificGenerality: Oh perhaps this is handled in the next PR...

MatthewMeacham: Hm, should we consider version..? I figured we couldn't actually know what changed b/w versions to know if it's back-compat or not

SpecificGenerality (approved):

PR #4052, opened 2026-09-16
[17/x] ITSM lockfile: track IT extensions

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4051

Adding a new IT extension is a break since implementing OTs need to map the new properties; removing one is a warning (the mapping just disappears, but the OT keeps its physical properties -- removing only impacts readers).

The lockfile now tracks extensions and can report breaks/warnings.


packages/maker/src/lockfile/test/generateOntologySchemaLockfile.test.ts
```diff
@@ -476,6 +477,99 @@ describe("generateOntologySchemaLockfile", () => {
       expect(ascending.typeClasses).toEqual([GEO, SELECTABLE, SORTABLE]);
     });
   });
+
+  describe("extended interfaces", () => {
```
SpecificGenerality, line 481: Should we test the case where the parent is not enrolled in schema migrations but authors a break?

MatthewMeacham: The next PR should make that case impossible to avoid that exact situation (by requiring the whole family opts in): https://github.com/palantir/osdk-ts/pull/4053

SpecificGenerality (approved):

PR #4053, opened 2026-09-16
[18/x] ITSM lockfile: Require fully family ITSM opt-in

MatthewMeacham:
## Targeting https://github.com/palantir/osdk-ts/pull/4052

If any IT in a family hierarchy is opted into ITSMs, the whole family must be opted in. We do this since an OT implementing any of the ITs will want the same back-compat guarantees regardless of where a property us coming from in the hierarchy.


packages/maker/src/api/test/interfaceSchemaMigrations.test.ts
```diff
@@ -182,6 +185,140 @@ describe("Interface schema migrations", () => {
...
+      expect(() =>
+        defineInterface({ apiName: "Child", extends: parent }),
+      ).toThrow();
+
+      expect(
+        Object.keys(dumpOntologyFullMetadata().ontology.interfaceTypes),
+      ).toEqual(["com.palantir.Parent"]);
+    });
+
+    it("says nothing about a parent imported from another ontology", () => {
```
SpecificGenerality, line 298: Should supporting this be a flup?

MatthewMeacham: Do we actually have the schema for the remote/linked ontology beyond the API name? I'm not familiar with the ontology importing workflow here

SpecificGenerality: It seems like we import more than the just the API name [here](https://github.com/palantir/osdk-ts/blob/ef25a0229e71a181ef90badbba8b9aaf15429ef2/packages/maker-import/src/generate/convertInterfaceType.ts#L38), but we lose the `schemaMigrationsEnabled` marker that you might need to determine the break. I wonder if we need to consider these sorts of interfaces for the v0 though...

MatthewMeacham: Ok I see, it literally loads the definition into the in-memory registry so that you can reference it w/o copying it

It's probably worth a FLUP to include these in locking and these checks

SpecificGenerality (approved):
Opting to add support for imported inherited interfaces in a FLUP.

